PR #1, opened 2016-12-16
Layout for the new Files app

GoOz:
Add default layout for Files app.
It still needs to be cut into modules but here is a start to work on.


config/webpack.cozyui.config.js
```diff
@@ -13,9 +13,9 @@ module.exports = {
       {
         test: /\.styl$/,
         loader: ExtractTextPlugin.extract([
-          'css-loader?importLoaders=1',
+          'css-loader?importLoaders=1&modules',
           'postcss-loader',
-          'stylus-loader'
+          'stylus-loader?paths=node_modules/cozy-ui/stylus/'
```
CPatchane: For my information, is this loader still available for other stylus files (out of cozy-ui)?

GoOz: @CPatchane This line is from @m4dz so i'm not really sure what it does, but i think it just gives a styl folder to look up to for Stylus, on top of the folder Stylus is already watching (.styl files from the app in this case).

m4dz:
Woohoo 🎉 

Nice job! I just made a little fix on buttons styles and rmove useless classnames calls. We should probably splt the table into subcomponents like table-row, but let's see this later we dealing with real data. It's now ALIVE!

Thanks!

PR #3, opened 2016-12-23
Add navigation + breadcrumb enhancements

CPatchane:
This PR:

* [x] Add the navigation feature using react-router
* [x] Rename `location` component to `breadcrumb` capitalize all `.jsx` files name to match their component name
* [x] Add breadcrumb root name detection using router path
* [x] Add EN locales for breadcrumb and sidebar items


src/components/TopBar.jsx
```diff
@@ -2,12 +2,12 @@ import styles from '../styles/topbar'
 
 import React from 'react'
 
-import Location from './location'
-import Toolbar from './toolbar'
+import Toolbar from './Toolbar'
+import Breadcrump from './Breadcrumb'
```
m4dz: Var name should be `Breadcrumb`, not `Breadcrump`

src/main.jsx
```diff
@@ -4,17 +4,55 @@ import './styles/main'
...
-      <App />
+      <Router history={hashHistory}>
+        <Route component={(props) =>
+          <App {...props} />}
+        >
+          <Redirect from='/' to='files' />
+          <Route
+            path='files'
+            component={(props) =>
+              <Table {...props} />
```
m4dz: As long as the `App` children is always a `Table` component, is it really relevant to externalize it in the _main.jsx_ file instead of letting it live in the _app.jsx_ file directly?

src/components/TopBar.jsx
CPatchane: Oops, type. Thanks!

src/main.jsx
CPatchane: Yes, because I think that we will have to pass the files/folders list to the Table as property here (not the `App` component). And that will be different according to the route. What do you think about?

m4dz: Hmm… I understand but I'm a little bit confuse about having the file list loaded in the _main.jsx_ file. Is it because I think it has more it's place in a store that should be called from the `Table` component itself? I can't really what anoy me in the way… @goldoraf, If you have some suggestion about the best way to do this 😃

CPatchane: @m4dz That's it, but the store will be provided to the Router element in the main.jsx. It should be provided to the most top level component as possible I think.

CPatchane:
@m4dz About naming, I think it's a sort of "standard/convention/practice" in (p)React to keep the file name exactly the same as the component name. We can use camelCase in this case.

m4dz:
If it's considered as a good practice, let's do this, it's ok for me 👍

goldoraf:
@CPatchane is right: it is a common practice in the React world.

src/main.jsx
goldoraf: @CPatchane used here the same routing style we used for MyAccounts, with data passed as props in the router file. I think we'll probably use Redux for Files v3, and the code organization will evolve consequently.

goldoraf:
:+1:

PR #5, opened 2016-12-23
Enable transifex

m4dz:
Thi PR enables Transifex on the project. It uses a nested dictionnaries, relevant to each component (a dict name is the related JSX component).


CPatchane:
Great this idea to use a dictionary to separate components translations 👍

PR #9, opened 2016-12-27
Fix Travis config

m4dz:
This PR fix travis config and enables auto-build to the `build` branch.

It still has a bug on Transifex locales pulling that we need to address in another PR (poke @jsilvestre)


src/components/App.jsx
```diff
@@ -5,7 +5,7 @@ import { translate } from '../plugins/preact-polyglot'
 import classNames from 'classnames'
 
 import Sidebar from './Sidebar'
-import Topbar from './Topbar'
+import Topbar from './TopBar'
```
CPatchane: Oops, that seems to be a typo from me since I renamed the variable name before. Could you rename the TopBar.jsx to Topbar.jsx instead please? Thanks.

m4dz: Al'right :)

m4dz: Done 🌮

CPatchane:
Thanks @m4dz!

PR #13, opened 2016-12-30
Enable Jest tests

m4dz:
This PR enables Jest tests and add a basic example on the Toolbar component

### DIFF
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -4,9 +4,9 @@ import React from 'react'
 import { translate } from '../plugins/preact-polyglot'
 
 const Toolbar = ({ t }) => (
-  <div class={styles['fil-content-toolbar']} role='toolbar'>
-    <button role='button' class='coz-btn coz-btn--regular coz-btn--upload'>{ t('toolbar.item_upload') }</button>
-    <button role='button' class='coz-btn coz-btn--more'><span class='coz-hidden'>{ t('toolbar.item_more') }</span></button>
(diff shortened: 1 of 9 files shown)


src/components/Toolbar.jsx
```diff
@@ -4,9 +4,9 @@ import React from 'react'
 import { translate } from '../plugins/preact-polyglot'
 
 const Toolbar = ({ t }) => (
-  <div class={styles['fil-content-toolbar']} role='toolbar'>
-    <button role='button' class='coz-btn coz-btn--regular coz-btn--upload'>{ t('toolbar.item_upload') }</button>
-    <button role='button' class='coz-btn coz-btn--more'><span class='coz-hidden'>{ t('toolbar.item_more') }</span></button>
+  <div className={styles['fil-content-toolbar']} role='toolbar'>
+    <button role='button' className='coz-btn coz-btn--regular coz-btn--upload'>{ t('toolbar.item_upload') }</button>
+    <button role='button' className='coz-btn coz-btn--more'><span class='coz-hidden'>{ t('toolbar.item_more') }</span></button>
```
CPatchane: Do we have to use className or class attribute here? I thought pReact used class but maybe it's className using preact-compat. But I think we should not keep a mix of the two ways here. What do you think?

goldoraf: That's a good question, @CPatchane. I love using class, but perhaps we should stay with className in order to keep the compatibility with React and/or inferno.

PR #14, opened 2017-01-02
Add ‘Coming soon’ temporary text to all views except the first one

CPatchane:
It was a requirement for the navigation feature card currently in validation.


GoOz:
👍

PR #15, opened 2017-01-06
Basic navigation

goldoraf:
Hi! This PR adds the ability to display the actual contents of folders. The app state is managed by a redux store, which code I tried to keep as idiomatic as possible. 

Bonus point: you can upload files and create new folders too ;)


src/components/File.jsx
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from 'react'
+import classNames from 'classnames'
+import { Link } from 'react-router'
+
+import styles from '../styles/table'
+import { translate } from '../lib/I18n'
+import FilenameInput from '../components/FilenameInput'
+
+const splitFilename = filename => {
+  let dotIdx = filename.lastIndexOf('.') - 1 >>> 0
```
gregorylegarec: Pure detail but I am not a big fan of this trick, I find it not enough self-explicit. What @CPatchane thinks about it ?

src/lib/I18n.js
```diff
@@ -35,20 +36,39 @@ const init = function (context, lang) {
...
 
+const initFormat = lang => {
+  const locales = {
+    en: require('date-fns/locale/en')
+  }
+  if (lang && lang !== 'en') {
+    try {
+      locales[lang] = require(`date-fns/locale/${lang}`)
+    } catch (e) {
+      console.warning(`The "${lang}" locale isn't supported by date-fns`)
```
goldoraf: Right. We should extract this lib into cozy-ui anyway, because it's now used in 2 apps.

src/containers/UploadButton.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { translate } from '../lib/I18n'
+
+import { uploadFile } from '../actions'
+
+const styles = {
```
goldoraf: well, the styles in question are needed for functional reasons, not for aesthetics. Without these styles, the component doesn't work. That's why I think these styles belong to the component.

src/components/File.jsx
goldoraf: I don't like that kind of trick either, but the alternative is rather verbose and miss some edge cases, so...

kosssi: maybe add a comment ?

gregorylegarec:
Thanks @goldoraf 👍

src/components/File.jsx
gregorylegarec: @goldoraf linked us to http://www.jstips.co/en/get-file-extension/

PR #17, opened 2017-01-09
Fix cozy-client-js :bug:

kosssi:


enguerran:
I think we must publish a new version of https://github.com/cozy/cozy-client-js on npm.
Another workaround, that do not change the actual project, is to use [`yarn link`](https://yarnpkg.com/en/docs/cli/link):

```
cd $SOMEWHERE/cozy-client-js
yarn && yarn link
cd $SOMEWHERE/cozy-files-v3
yarn && yarn link cozy-client-js
```

But changing the dependency in the project is not sustainable.

kosssi:
with cozy-client-js v0.0.2 it's ok

PR #21, opened 2017-01-11
Fix enter key on safari :bug:

kosssi:
KeyboardEvent.key is not supported on safari: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key


goldoraf:
Bad browser, good contrib :+1:  Thanks!

PR #27, opened 2017-01-12
[feat] Add error alerts for file fetching

gregorylegarec:
This PR :

Needs this branch of Cozy UI : https://github.com/gregorylegarec/cozy-ui/tree/feat/improve-alerts

* Connects App to redux
* Add Alerter component, which show alerts. Based on `Notifier.jsx` in `MyAccounts` app, but the behaviour is a little bit different (Overlay, reloading, button, etc.)
* At this time, the alerter mechanism only show critical errors.
* To trigger CSS transition, I used a `setTimeout` in `componentDidMount` method. I clearly not a big fan and I'd love to replace it by a better solution.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,14 +1,26 @@
 import styles from '../styles/app'
 
 import React from 'react'
+import { connect } from 'react-redux'
 import { translate } from '../lib/I18n'
 import classNames from 'classnames'
 
+import Alerter from './Alerter'
 import Sidebar from './Sidebar'
 import Topbar from './Topbar'
 
-const App = ({ t, children }) => (
+const reload = () => {
+  window.location.reload()
+}
+
+const App = ({ t, error, children }) => (
   <div class={classNames(styles['fil-wrapper'], styles['coz-sticky'])}>
+    { error &&
+    <Alerter
+      error={error}
+      reload={reload}
+      />
+    }
 
     <Sidebar />
 
@@ -19,4 +31,13 @@ const App = ({ t, children }) => (
   </div>
 )
 
-export default translate()(App)
+const mapStateToProps = (state) => ({
+  error: state.ui.error
+})
+
+const mapDispatchToProps = (dispatch) => ({})
+
+export default translate()(connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(App))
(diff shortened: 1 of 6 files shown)


src/components/App.jsx
```diff
@@ -1,14 +1,26 @@
...
 import Topbar from './Topbar'
 
-const App = ({ t, children }) => (
+const reload = () => {
+  window.location.reload()
+}
+
+const App = ({ t, error, children }) => (
   <div class={classNames(styles['fil-wrapper'], styles['coz-sticky'])}>
+    { error &&
```
goldoraf: I'd prefer a one-liner here but it's only a matter of personal preference :)

src/reducers/ui.js
```diff
@@ -2,17 +2,21 @@ import { combineReducers } from 'redux'
...
 import {
   FETCH_FILES,
+  FETCH_FILES_FAILURE,
   RECEIVE_FILES,
   CREATE_FOLDER,
   CREATE_FOLDER_SUCCESS,
   UPLOAD_FILE,
-  UPLOAD_FILE_SUCCESS
+  UPLOAD_FILE_SUCCESS,
+  RELOAD_PAGE
```
goldoraf: This const is now useless, right?

src/reducers/ui.js
```diff
@@ -2,17 +2,21 @@ import { combineReducers } from 'redux'
...
 } from '../actions'
 
 const isFetching = (state = false, action) => {
   switch (action.type) {
     case FETCH_FILES:
       return action.folderId
+    case FETCH_FILES_FAILURE:
+      return false
```
goldoraf: This line isn't necessary: you can pile up 'case' statements when the instructions to be executed are the same.

src/reducers/ui.js
```diff
@@ -51,8 +55,23 @@ const updating = (state = [], action) => {
   }
 }
 
+const error = (state = null, action) => {
+  switch (action.type) {
+    case FETCH_FILES_FAILURE:
+      console.debug('FETCH_FILES_FAILURE')
```
goldoraf: Forgotten console.debug?

gregorylegarec:
@goldoraf Sorry about the evident lint issue. I updated the two lines of code you mentionned, let's try to unify the code as most as possible. :)

PR #28, opened 2017-01-16
Go back to previous hash in history before reloading the page after a critical error

gregorylegarec:
@goldoraf, this one is for you, I just want to be sure that my code is OK.

As requested by @jsilvestre, we need to go back in hash history before reloading the page, otherwise we will try to reload the folder that we are not able to load, it is not logical in an UI point a view and it could make the whole app crashing infinetly.

Thanks :)

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -2,6 +2,7 @@ import styles from '../styles/app'
 
 import React from 'react'
 import { connect } from 'react-redux'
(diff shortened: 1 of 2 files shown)


gregorylegarec:
We will change the way the error is shown to user in a near future. For now, we are keeping it as it. The main concern of this request is now out of scope.

PR #29, opened 2017-01-17
Make the route change only after the folder's content has been successfully fetched

goldoraf:


gregorylegarec:
Well done @goldoraf, it seems ok to me. Just a little remarks, when the error is displayed, the background displays "Loading" with no UI. Could it be a quick win to keep the view where the folder is loading with the spinner ?

goldoraf:
I think I fixed the issue: could you have a look?

gregorylegarec:
The behavior is ok for me. Just one regression about the timer : it does not reload the page anymore after the 10 seconds. I think the solution should be in component Alerter to move the code form function `componentWillReceiveProps` to `constructor` (or maybe `componentDidMount` ?). I tested and it worked for me.

```javascript
  constructor (props) {
    super(props)
    this.state = {
      hidden: true,
      reloadTimer: 10
    }

    const { error, reload } = props
    const { reloadTimer } = this.state

    if (error && error.critical) {
      setTimeout(() => {
        console.debug('reload')
        if (typeof reload === 'function') reload()
      }, reloadTimer * 1000)
    }
  }
```

goldoraf:
I fixed the problem by moving the related code into componentDidMount, that seems to me a better place than the constructor.

gregorylegarec:
Thanks @goldoraf 👍

PR #34, opened 2017-01-24
use the navigator language rather than the doc lang attribute

enguerran:
I found that `index.ejs` set the lang attribute (`<html lang="en">`) and the application use this lang attribute to translate the application.

In cordova, we could get the navigator language that uses the up-to-date user's language preference.
The format is `fr-FR` or `en-US`, so we could `slice(0,2)` to get `fr` or `en`.

I don't know if the web application is well translated but with this PR, the mobile application will be.

### DIFF
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -23,10 +23,7 @@ cozy.init({
 })
 
 const context = window.context
-let lang = document.documentElement.getAttribute('lang') || 'en'
-if (navigator && navigator.language) {
-  lang = navigator.language.slice(0, 2)
-}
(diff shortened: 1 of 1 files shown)


mobile/src/main.jsx
```diff
@@ -23,7 +23,10 @@ cozy.init({
 })
 
 const context = window.context
-const lang = document.documentElement.getAttribute('lang') || 'en'
+let lang = document.documentElement.getAttribute('lang') || 'en'
```
kosssi: Can you remove document.documentElement.getAttribute('lang')?

PR #35, opened 2017-01-25
Feature: download a file selection (or just one file for now)

goldoraf:


goldoraf:
Well, I'm a bit disappointed 'cause I tried really hard to respect the mockup's design ;) 

Concerning the selection checkboxes, 63a5edc8e75a8c549cc3c65af26b84bb355d8e5c was supposed to fix the alignment (at least it works for me on Firefox and Chrome), did you fetch all the commits before testing?

We should talk about this with the whole team of course, but I think that a peer programming session between a CSS specialist like @GoOz and our favourite UX designer in order to fix these styling issues would be a better solution than trying to fix it now.

src/components/File.jsx
```diff
@@ -68,21 +85,28 @@ class File extends Component {
...
-            {attributes.isOpening === true && <div class={styles['fil-loading']} />}
+            {attributes.isOpening === true && <div className={styles['fil-loading']} />}
           </a>
         </td>
       )
-      : <td class={classes}><a target='_blank' href={`http://cozy.local:8080/files/download/${attributes.id}`}>{filename}<span class={styles['fil-content-ext']}>{extension}</span></a></td>
+    }
+    return (
+      <td className={classes}>
+        <a target='_blank' href={`http://cozy.local:8080/files/download/${attributes.id}`}>
```
enguerran: The url should be given by the cozy-stack.

On mobile, we let the user register an url, then it is stored in indexedDB.

src/components/FileList.jsx
```diff
@@ -1,16 +1,24 @@
...
 import File from '../components/File'
 
-const FileList = ({ t, f, files, onFileEdit, onFolderOpen }) => (
+import SelectionBar from '../containers/SelectionBar'
+
+const FileList = ({ t, f, files, showSelection, onFileEdit, onFolderOpen, onFileToggle }) => (
   <div role='contentinfo'>
-    <table class={styles['fil-content-table']}>
+    { showSelection && <SelectionBar /> }
+    <table class={classNames(
```
enguerran: `<table class={classNames(` → `<table className={classNames(`

src/components/FileList.jsx
```diff
@@ -1,16 +1,24 @@
...
+      styles['fil-content-table'],
+      { [styles['fil-content-table-selection']]: showSelection }
+    )}>
       <thead>
         <tr>
+          <th />
           <th class={styles['fil-content-file']}>{ t('table.head_name') }</th>
```
enguerran: `<th class={styles['fil-content-file']}>{ t('table.head_name') }</th>` → `<th className={styles['fil-content-file']}>{ t('table.head_name') }</th>`

src/containers/Toolbar.jsx
```diff
@@ -5,19 +5,39 @@ import { connect } from 'react-redux'
...
-      onClick={() => addFolder()}
-    >
-      <span class='coz-hidden'>{ t('toolbar.item_more') }</span>
-    </button>
+    <MenuButton>
+      <button
+        role='button'
+        className='coz-btn coz-btn--more'
+      >
+        <span class='coz-hidden'>{ t('toolbar.item_more') }</span>
```
enguerran: `<span class='coz-hidden'>{ t('toolbar.item_more') }</span>` → `<span className='coz-hidden'>{ t('toolbar.item_more') }</span>`

gregorylegarec:
@goldoraf 
> We should talk about this with the whole team of course, but I think that a peer programming session between a CSS specialist like @GoOz and our favourite UX designer in order to fix these styling issues would be a better solution than trying to fix it now.

Ok for me 👍

src/styles/selectionbar.styl
```diff
@@ -0,0 +1,65 @@
...
+  z-index 1000 // because... you know...
+
+  .fil-selectionbar-separator
+    background-color black
+    border-radius 50%
+    width 4px
+    height 4px
+    margin 0 0.5em 0 2em
+
+  button
```
CPatchane: Maybe we should use a class .cozy-bar-button from cozy-ui here? Because we will need these cozy-bar components in other apps. /poke @GoOz

src/components/File.jsx
CPatchane: I agree with @enguerran here, but how to do that? I had the same case for cozy-photos-v3, should we store it in the store in a specific reducer at the app rendering?

GoOz:
@goldoraf @gregorylegarec 

    We should talk about this with the whole team of course, but I think that a peer programming session between a CSS specialist like @GoOz and our favourite UX designer in order to fix these styling issues would be a better solution than trying to fix it now.

Yup! I agree, it's not that far from the mockup (in Zeplin), I don't know why but the mockup in Marvel is too different on many points so I guess this is definitely something we should see directly with our favourite UX designer :)

src/styles/selectionbar.styl
```diff
@@ -0,0 +1,65 @@
...
+  justify-content center
+  align-items center
+  width 100%
+  top 0
+  left 0
+  color white
+  background-color slate-grey
+  height em(56px)
+  font-weight bold
+  z-index 1000 // because... you know...
```
GoOz: Duuuude! 1000! really? I get that the bar should be on top of everything but… I seriously doubt that there will 1000 layers of z-index. z-index 50 should be enough, by far.

src/components/File.jsx
enguerran: Yes, we should have a specific reducer for that url that we (mobile developers) could use as well.
This value should be set at the beginning of the app, somewhere in `main.jsx` around the lines where `cozy-client-js` is initialized.

src/styles/selectionbar.styl
GoOz: Definitely. It should, at some point, be a shared component with shared style, directly from cozy-ui. Not sure it should delay the PR to be merged though, as we didn't engaged on a solution for shared components yet.

src/components/File.jsx
goldoraf: To the best of my knowledge, there is still work to do so that cozy-stack exposes that URL to the client. Additionally, this "download" feature is a temporary one, we have time to correct this.

src/styles/selectionbar.styl
goldoraf: Yes, the selection bar should become a cozy-ui thing, but I think it's better to wait the second time we need it, so that we can choose how to componentize it.

src/styles/selectionbar.styl
goldoraf: Yeah, that was more a joke :)

PR #36, opened 2017-01-26
Stop webpack when failed fix travis green build :bug:

kosssi:
You can see https://github.com/webpack/webpack/issues/708


enguerran:
TIL (*Today I Learned*) `--bail` (cf. http://webpack.github.io/docs/configuration.html#bail)

PR #42, opened 2017-01-26
breadcrumb amendments

ysa000:


src/styles/breadcrumb.styl
```diff
@@ -7,6 +7,10 @@
 .fil-inside-path
     color #95999d
 
+.fil-inside-path
```
gregorylegarec: We don't need to repeat `.fil-inside-path` here. The following code uses nesting and is equivalent :
```stylus
.fil-inside-path
  color #95999d

  a
    cursor pointer
```
Also, we use the compact form, without `:` character.

gregorylegarec:
Thanks @ysa000 !

PR #51, opened 2017-01-30
Fix breadcrump router behaviour

CPatchane:
In the current breadcrumb, when you click on the root Files, the path doesn't change to the correct path means '/files'.
This PR fixes it.


gregorylegarec:
Thanks @CPatchane !

PR #55, opened 2017-01-31
add mobile settings :fork_and_knife:

enguerran:
I am adding a menu in the Nav component and a route in the AppRoute component. Both to have a settings page in our mobile application.
With this settings component, users will be able to:
- see the cozy server's url
- see the application's version number

Mockup: https://marvelapp.com/3e1j0ig/screen/24244093

- [x] settings is added to nav
- [x] only for mobile application
- [x] redirects to a settings page
- [x] only accessible for mobile application
- [x] displaying app's about (with version number)
- [x] displaying cozy server's url
- [x] the whole UI could be translated in french

tests:
- [ ] ~nav: it should display settings page when clicking on the settings button~ (don't know how to test that behavior)
- [x] settings: it should contains version number
- [x] settings: it should contains cozy server's url

PR #56, opened 2017-01-31
Clean and improve build runtimes

m4dz:
This PR is an improvement for our build system. It:

- remove unwanted deps
- improve linting
- clean webpack configs
- add webpack target contexts
- fix webpack warnings
- improve index.html build process
- add support for compil variables
- ~~make the best ☕️ you've ever drunk~~ (not for this one)


package.json
```diff
@@ -18,12 +18,14 @@
     "pretest": "npm run lint",
     "prewatch:browser": "npm run clean:browser",
     "prewatch:mobile": "npm run clean:mobile",
-    "lint": "standard 'src/**/*.js' 'src/**/*.jsx' 'mobile/src/**/*.js' 'mobile/src/**/*.jsx'",
+    "lint": "npm-run-all --parallel 'lint:*'",
+    "lint:styles": "stylint src/styles --config ./.stylintrc",
+    "lint:js": "standard 'src/**/*.js' 'src/**/*.jsx' 'test/**/*.js' 'mobile/src/**/*.js' 'mobile/src/**/*.jsx' 'mobile/test/**/*.js'",
     "test": "node --harmony_proxies node_modules/.bin/jest",
     "watch": "webpack --config webpack.config.js --watch --display-modules --display-chunks",
```
jinroh: `npm run watch` does not work anymore since it does not define a `NODE_ENV`:

```
/Users/pierre/code/cozy-files-v3/webpack.config.js:6
const target = process.env.NODE_ENV.match(/^(\w+):/)[1]
                                   ^

TypeError: Cannot read property 'match' of undefined
```

config/webpack.mobile.target.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  module: {
+    noParse: [
+      /localforage\/dist/
+    ]
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      __ALLOW_HTTP__: !production,
+      __TARGET__: JSON.stringify('mobile')
```
enguerran: What's the aim? As `JSON.stringify('mobile') === '"mobile"'` :neckbeard: 

In `src/` we would be able to write conditions like:
```javascript
if (__TARGET__ === 'mobile') {
  // do something awesome for mobile application
}
```

mobile/src/main.jsx
```diff
@@ -33,6 +35,8 @@ const store = createStore(
 )
 
 document.addEventListener('DOMContentLoaded', () => {
+  console.debug(__ALLOW_HTTP__)
```
enguerran: No need to add this usage example, imho.

config/webpack.base.config.js
```diff
@@ -2,24 +2,18 @@
...
   output: {
-    path: path.resolve(__dirname, '..', outputFolder),
-    filename: build ? 'app.[hash].js' : 'app.js'
+    filename: 'app.js'
   },
   resolve: {
     extensions: ['', '.js', '.jsx', '.json']
   },
-  devtool: build ? '#cheap-module-source-map' : 'eval',
+  devtool: 'cheap-source-map',
```
enguerran: Is this enough to have real source attached to the debugger?
I may test the visual studio code's remote debugger to have real source attached, but it is comfortable to have real source in the browser's dev tools.

package.json
kosssi: @jinroh You could use `yarn watch:browser` or `yarn watch:mobile`.

But you are right, add default NODE_ENV to 'browser:development' it's better.

enguerran: I faced the same issue

kosssi: It's the same for `yarn build`;)

config/webpack.base.config.js
enguerran: I just try `devtool: 'source-map',` and [it displays](http://webpack.github.io/docs/configuration.html#devtool) the real source code in browser's devtool, very convenient:

package.json
enguerran: In fact `yarn watch` is a factorization for `yarn watch:browser` and `yarn watch:mobile`, I guess it is not intended to be used.

webpack.config.js
```diff
@@ -2,15 +2,17 @@
 
 const merge = require('webpack-merge')
 
-const build = /production$/.test(process.env.NODE_ENV)
+const build = /:production$/.test(process.env.NODE_ENV)
```
kosssi: const build = /:production$/.test(process.env.NODE_ENV || 'development') ?

package.json
m4dz: So what do you prefer? Do we need to have a shadowed task like `sub:watch` and alias `watch` to `watch:browser`? Or do we keep it like this and never use the `watch` or `build` tasks alone?

enguerran:
You have to know I am very impatient about this PR because of #55. Could we have an update?

config/webpack.mobile.target.js
m4dz: The `stringify` is mandatory in `webpack.definePlugin` config (https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin). So the condition to test is just `if (__TARGET__ === 'mobile') { }`

mobile/src/main.jsx
m4dz: Sure? Obviously having a compil variable dedicated to it is probably a better practice than relying on the global `__TARGET__` one to do some feature-flip (see the @gregorylegarec point about it)

config/webpack.base.config.js
m4dz: Agree, the problem the `source-map` is that it's particurlarly slow on rebuild tasks (https://github.com/webpack/docs/wiki/configuration#devtool), when `cheap-source-map`is faster and doesn't rely on the unwanted `eval` we need to avoid because of CSP. But if the rebuild time isn't an issue, let's go for `source-map`

webpack.config.js
m4dz: I'm not sure… what's the goal of this change?

m4dz:
@enguerran yep, just waiting for responses to all discussions and let close it after 👍

webpack.config.js
kosssi: @m4dz it's for resolve when you launch `yarn build` or `yarn watch` without :browser or :mobile

kosssi: Not only here ;)

config/webpack.mobile.target.js
enguerran: It's weird, but ok, I like weird things.

config/webpack.base.config.js
enguerran: Yes, I know, but debugging at dev stage may be easier than fast, imo.

enguerran:
I guess we can merge it as this.

package.json
enguerran: I suggest to let as this and describe the modus operandi in `README.md` like

> if you want to watch files' changes and rebuild, use `yarn watch:browser` or `yarn watch:mobile`. (you don't have to run `yarn watch`. Because this project contains two different workflows, you muse choose one of them.)

m4dz:
@gregorylegarec ok, ready to merge as it, and we'll improve it later. Thanks to all 🍻

PR #57, opened 2017-02-01
Add mobile action menu

goldoraf:


src/containers/FileActionMenu.jsx
```diff
@@ -0,0 +1,73 @@
...
+import withGestures from 'react-bosonic/lib/withGestures'
+
+import { splitFilename, getClassFromMime } from '../components/File'
+import { getActionableFile } from '../reducers'
+import { downloadFile, hideFileActionMenu } from '../actions'
+
+const Menu = ({ t, file, onDownload }) => {
+  const { filename, extension } = splitFilename(file.name)
+  return (
+    <div className={styles['fil-actionmenu']}>
```
CPatchane: I would use `action-menu` instead of `actionmenu`, it sounds weird. What do you think?

src/containers/FileActionMenu.jsx
```diff
@@ -0,0 +1,73 @@
...
+      <Item>
+        <div className={classNames(styles['fil-actionmenu-file'], getClassFromMime(file))}>
+          {filename}
+          <span className={styles['fil-actionmenu-file-ext']}>{extension}</span>
+        </div>
+      </Item>
+      <hr />
+      <Item>
+        <a className={styles['fil-action-openwith']}>
+          {t('mobile.actionmenu.open_with')}
```
CPatchane: Here too, `mobile.action_menu.open_with`

src/containers/FileActionMenu.jsx
```diff
@@ -0,0 +1,73 @@
...
+    <div className={styles['fil-actionmenu']}>
+      <Item>
+        <div className={classNames(styles['fil-actionmenu-file'], getClassFromMime(file))}>
+          {filename}
+          <span className={styles['fil-actionmenu-file-ext']}>{extension}</span>
+        </div>
+      </Item>
+      <hr />
+      <Item>
+        <a className={styles['fil-action-openwith']}>
```
CPatchane: Maybe the same thing for `openwith`

src/styles/actionmenu.styl
```diff
@@ -0,0 +1,37 @@
...
+    bottom         em(9px)
+    width          "calc(100% - %s)" % em(16px)
+    margin         0 em(8px)
+    padding-bottom em(5px)
+
+    hr
+        margin-top 0
+
+    a
+        &.fil-action-openwith
```
CPatchane: I think we don't need to target the html tag `a` here imho, just `fil-action-openwith` and `fil-action-download` would be enough.

src/containers/FileActionMenu.jsx
goldoraf: I did this because `actionmenu` is one element, I tend to use carets to separate elements, aka: `<app>-<element>-<sub-element>`. But maybe I'm too influenced by BEM here...

src/containers/FileActionMenu.jsx
goldoraf: Same remark

src/containers/FileActionMenu.jsx
goldoraf: I'm OK here with your suggestion.

src/styles/actionmenu.styl
goldoraf: Good catch!

src/containers/FileActionMenu.jsx
GoOz: @goldoraf no you're right. It's not quite clear yet, but the future guidelines will tend to use BEM-like naming conventions. If the base element's name is actionmenu then actionmenu it is. Using dash between the two words would be confusing, I think.

src/containers/FileActionMenu.jsx
GoOz: Same answer 😛

CPatchane:
Also, I just test it and it seems that the cross is not correctly vertically centered:

screen shot 2017-02-01 at 14 32 36

CPatchane:
Thanks @goldoraf 🎉

enguerran:
By the way, @goldoraf I got an error on mobile.

PR #58, opened 2017-02-01
Handle permissions

jinroh:
Based on #56.

This PR aims at adding support for permissions in cozy-files-v3.

  - cozy-client-js in now an external served by the stack
  - cozy-bar is initialized using the data-* attributes of the application div
  - the manifest.webapp is copied to build/ during the build. the slug of the manifest is changed to "app" for dev builds

This is still a *wip*, it also should be tested with future releases of the cozy-stack.

cc @m4dz


src/index.ejs
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{.Locale}}">
```
kosssi: @m4dz it's a problem with mobile version I'm not sure it's a good idea to have the same index.ejs ;)

enguerran: @kosssi this is not the PR you are looking for --> https://github.com/cozy/cozy-files-v3/pull/56

enguerran: Nevertheless, within the mobile application version we use the `navigator` object, see https://github.com/cozy/cozy-files-v3/blob/master/mobile/src/main.jsx#L25

kosssi: Yop @enguerran it's this PR which update index.ejs 😜  but you are right it's finally not a problem, sorry for this alert.

nono:
-> #81

PR #59, opened 2017-02-01
Changed behaviour when a folder cannot be opened

goldoraf:
… because of a network failure : we now display a error page similar to the empty folder one


src/components/Oops.jsx
```diff
@@ -0,0 +1,25 @@
+import styles from '../styles/oops'
+
+import React from 'react'
+import { translate } from '../lib/I18n'
+
+const reload = () => {
+  window.location.reload()
+}
+
+const Oops = ({ t }) => (
```
CPatchane: Maybe 'errorView' could be more convenient here?

goldoraf: It's less fun :)

src/components/UploadButton.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+
+const styles = {
```
GoOz: I don't understand why this style is not handle by external CSS. Not sure it's you're doing though but I'm still wondering, in case you know why…

goldoraf: We already debated on this in a previous PR, basically the reason is the component cannot possibly work witout these styles: these are "functional" styles, without them, the component is broken. That's why I think they belong to the component and not to an external stylesheet.

PR #64, opened 2017-02-02
Get rid of node-uuid dep and win ~100ko minified

jinroh:


goldoraf:
:tada: thanks a lot! You probably saved a few polar bears ;)

PR #66, opened 2017-02-02
Fil 16 - Trash a selection

y-lohse:
Early PR for this as I'm guessing there will be things to review!

The css for the modal depends on a new component in cozy-ui, PR for that will follow soon.

Things left to do:

- [x] Enable folder trashing
- [x] Error handling
- [x] Minor style updates


src/containers/DeleteConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+    <div className={styles['coz-overlay']}>
+      <div className={styles['coz-modal']}>
+        <h2 className={styles['coz-modal-title']}>
+          {t('deleteconfirmation.title')}
+        </h2>
+        <button
+          className={classNames('coz-btn', 'coz-btn--close', styles['coz-modal-close'])}
+          onClick={() => onDismiss(selected)}
+          >
+          <span className='coz-hidden'>{t('deleteconfirmation.close')}</span>
```
y-lohse: I guess this could be the same as `deletionconfirmation.cancel`?

y-lohse:
⚠️⚠️⚠️
This PR is now going to conflict with #59 ; #59 removed the alerter behavior in the App component, while this PR still needs it.

y-lohse:
Integration tests are currently failing because this PR needs the `modals` components from cozy-ui, currently in [this PR](https://github.com/cozy/cozy-ui/pull/40).

src/locales/en.json
```diff
@@ -49,6 +49,15 @@
     "rename": "Rename",
     "close": "Close"
   },
+  "deleteconfirmation": {
```
enguerran: I don't know if it should be `"deletion_confirmation"` or `"deletionconfirmation"`.
I let others decide.

src/components/File.jsx
```diff
@@ -67,12 +67,9 @@ class File extends Component {
           <span data-input='checkbox'>
             <input
               type='checkbox'
-              disabled={isDir(attributes)}
               checked={attributes.selected}
              />
-            {isDir(attributes)
-              ? <label />
-              : <label onClick={e => this.toggle(e)} />}
+            <label onClick={e => this.toggle(e)} />
```
enguerran: You can write
```jsx
<label onClick={this.toggle} />
```
as `this.toggle` takes the same argument as `onClick`

src/components/FileList.jsx
```diff
@@ -10,12 +10,14 @@ import File from '../components/File'
...
 import FileActionMenu from '../containers/FileActionMenu'
+import DeleteConfirmation from '../containers/DeleteConfirmation'
 
 const FileList = ({
-  t, f, files, error, showSelection, showActionMenu, onFileEdit, onFolderOpen, onFileToggle, onShowActionMenu
+  t, f, files, error, showSelection, showActionMenu, showDeleteConfirmation, onFileEdit, onFolderOpen, onFileToggle, onShowActionMenu
 }) => (
   <div role='contentinfo'>
     {showSelection && <SelectionBar />}
+    { showDeleteConfirmation && <DeleteConfirmation /> }
```
enguerran: I don't know if there should be 1 or zero space inside braces on L19 and L20. But it is not homogeneous.

src/containers/DeleteConfirmation.jsx
```diff
@@ -0,0 +1,66 @@
...
+        </button>
+        <p className={classNames(styles['fil-deleteconfirmation-text'], styles['icon-trash'])}>
+          {t('deleteconfirmation.trash')}
+        </p>
+        <p className={classNames(styles['fil-deleteconfirmation-text'], styles['icon-restore'])}>
+          {t('deleteconfirmation.restore')}
+        </p>
+        <p className={classNames(styles['fil-deleteconfirmation-text'], styles['icon-shared'])}>
+          {t('deleteconfirmation.shared')}
+        </p>
```
enguerran: Maybe you can factorize those `<p />` like:
```jsx
const deleteconfirmationTexts = ['trash','restore','shared'].map(type => (
  <p className={classNames(styles['fil-deleteconfirmation-text'], styles[`icone-${type}`])}>
    {t(`deleteconfirmation.${type}`)}
  </p>
)
```
and in `render` function
```jsx
<button><span></span></button>
{deleteconfirmationTexts}
<div><button></button><button></button></div>
// etc.
```

src/reducers/folder.js
```diff
@@ -20,6 +20,8 @@ export const files = (state = [], action) => {
         ...state,
         action.file
       ]
+    case TRASH_FILE_SUCCESS:
+      return state.map(file => file.id === action.file.id ? action.file : file)
```
enguerran: If `id` are the same we return `action.file` but if not we return `file`. Why don't we always return `file`?
Is it possible that `action.file !== file` even when `action.file.id === file.id`?

src/components/File.jsx
goldoraf: @enguerran no, because `toggle` is not bound to this explicitely in the constructor.

enguerran: oh yes,

```jsx
<label onClick={this.toggle.bind(this)} />
```

should do the trick.

But it may be less explicit than

```jsx
<label onClick={e => this.toggle(e)} />
```

src/containers/DeleteConfirmation.jsx
y-lohse: Is this how you would normally do it? I tend to find the longer, more explicit version easier to read and maintain, but I'm happy to conform to the rest of the code base!

src/reducers/folder.js
y-lohse: Yes. After a file has been trashed server side, `action.file.dir_id`is going to be different from `file.dir_id`. I suppose the same logic could be used for file moving, renaming, etc.
Does that make sense?

src/containers/DeleteConfirmation.jsx
enguerran: If `className` changes, you change it once. If you had to change something, you change it once. DRY

y-lohse: Cool, makes sense. Change incoming. Thanks for the review btw!

enguerran: np

PR #67, opened 2017-02-02
[mobile] Fix some bugs

kosssi:
- [x] Add offline even if onboarding is set
- [x] Fix onboarding on ios

### DIFF
--- a/mobile/src/actions/index.js
+++ b/mobile/src/actions/index.js
@@ -54,7 +54,7 @@ export const registerDevice = (router, location) => {
             const inAppBrowser = InAppBrowser.open(url, target, options)
 
             return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


mobile/src/actions/index.js
```diff
@@ -24,14 +24,17 @@ export const registerDevice = (router, location) => {
...
                 const accessCode = /\?access_code=(.+)$/.test(url)
                 const state = /\?state=(.+)$/.test(url)
 
                 if (accessCode || state) {
                   resolve(url)
                 }
-              })
+              }
+
+              inAppBrowser.addEventListener('loadstop', validate)
```
kosssi: Maybe I can use only 'loadstart', I will test it.

PR #74, opened 2017-02-06
feature: open with in cordova app :busstop:

enguerran:
On mobile device, you can *"Open With..."* a file stored in your cozy.

It downloads the file to a temporary folder then let user opens the file in another application.

open with...

- [x] "Open with..." disney scenario
- [ ] Works on both Android and iOS
- [x] "Open with..." if no application can open this type
- [x] "Open with..." if there is no network
- [x] T9n and i18n
- [x] Add a loader in the menu
- [ ] Publish the android-FIL-7.apk

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -231,3 +231,13 @@ export const renameFolder = (newName, id) => {
 export const hideToast = () => ({
   type: HIDE_TOAST
 })
+
+export const actionMenuLoading = (menu) => ({
+  type: 'SHOW_SPINNER',
+  menu
+})
+
+export const actionMenuLoaded = (menu) => ({
+  type: 'HIDE_SPINNER',
+  menu
+})
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -2,6 +2,7 @@ import cozy from 'cozy-client-js'
 
 import { ROOT_DIR_ID } from '../constants/config'
 import { saveFileWithCordova, openFileWithCordova } from '../../mobile/src/lib/filesystem'
+import { offlineError, noAppError } from '../../mobile/src/actions'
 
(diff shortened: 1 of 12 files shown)


enguerran:
Here is how an error is displayed

ezgif com-optimize 1

enguerran:
And now, ladies and gentlemen, please welcome the spinner:

ezgif com-optimize 2

src/actions/index.js
```diff
@@ -160,6 +164,26 @@ export const downloadFile = id => {
   }
 }
 
+export const openFileWith = (id, filename) => {
+  return async (dispatch, getState) => {
+    if (window.cordova && window.cordova.plugins.fileOpener2) {
+      dispatch({ type: OPEN_FILE_WITH, id })
+      const response = await cozy.files.downloadById(id).catch((error) => {
```
kosssi: I didn't know It's possible to use await with .catch 👍

enguerran:
I rebase this branch to allow anyone to merge it.
Are we cool like Fonzy?

src/components/App.jsx
```diff
@@ -38,6 +39,7 @@ class App extends Component {
         <main class={styles['fil-content']}>
           <Topbar />
           { children }
+          { toastMessage && <Toast />}
```
kosssi: It's normal to call an container in component?

enguerran: Why not?
Containers and Components are both Components as known as Dumb Components and Smart Components.

What differentiate smart and dumb? Within `redux` and `react-redux`, a Smart Component is a component that fetches data from state or store's dispatch (*actions*), so you could have a Component tree with different layers including SC or DC in any order.

Do you agree @goldoraf?

PR #75, opened 2017-02-06
Unlink a device from Cozy

kosssi:
- [x] Refactoring Settings container
- [x] Add translation
- [x] Add unlink section on Settings
- [x] Display unlink popup
- [x] Remove device on Cozy
- [x] Update old translation on transifex
- [ ] Add new translation on transifex


mobile/src/containers/Settings.jsx
```diff
@@ -1,30 +1,19 @@
...
-)
+import Settings from '../Components/Settings'
 
 const mapStateToProps = (state, ownProps) => ({
   version: window.navigator.appInfo ? window.navigator.appInfo.version : 'dev',
   serverUrl: state.mobile.serverUrl
 })
 
-export default translate()(connect(mapStateToProps)(Settings))
+const mapDispatchToProps = (dispatch, ownProps) => ({
```
enguerran: No need to have `ownProps` in the function signature if we don't use it. Same for my code [L5](https://github.com/cozy/cozy-files-v3/pull/75/files#diff-c870ef95a44c7511a2259f2dfd57c015R5) ;)

mobile/src/reducers/mobile.js
```diff
@@ -1,25 +1,35 @@
...
+    backupImages: false,
+    displayUnlinkConfirmation: false
   }
 }
 
 export const mobile = (state = initialState, action) => {
   switch (action.type) {
     case SET_STATE:
-      return action.state
+      return Object.assign({}, initialState, action.state)
```
enguerran: if default value of `state` is `initialState`, we may stay pure and use `state` instead of `initialState`:

```javascript
return Object.assign({}, state, action.state)
```

but as soon as we write this, we understand that the action is not correct:

- SET_STATE should set **a** state
- UPDATE_STATE should update **the** state
- INITIALIZE_STATE should initialize **the** state
- and so on...

We may create another case to match the wanted behavior.

mobile/src/reducers/mobile.js
```diff
@@ -1,25 +1,35 @@
...
+    case UNLINK:
+      return Object.assign({}, initialState)
```
enguerran: I don't think the name `UNLINK` describe the result of this reducer.

We may call it `INIT_STATE` instead.

enguerran: and as `state` default value is `initialState`, we may use `state` instead, to stay pure.

mobile/src/reducers/mobile.js
```diff
@@ -1,25 +1,35 @@
...
+    displayUnlinkConfirmation: false
   }
 }
 
 export const mobile = (state = initialState, action) => {
   switch (action.type) {
     case SET_STATE:
       return action.state
+    case UNLINK:
+      return Object.assign({}, initialState)
```
enguerran: I don't think the word `UNLINK` descibes the behavior of the result.

To return `initialState` (reset current `state` to `initialState`, we may:
- `reduce(state, { type: SET_STATE, state: initialState })` or
- `reduce(null, { type: NOOP })` (but I don't beleive it available in redux

enguerran:
@kosssi I may have found what's wrong with `jest`.

Here is what the travis' console displays:
```
  ● Test suite failed to run
    /home/enguerran/Development/Cozy/cozy-files-v3/node_modules/cozy-ui/react/Modal/index.jsx:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import React from 'react'
                                                                                             ^^^^^^
    SyntaxError: Unexpected token import
      at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
      at Object.<anonymous> (mobile/src/components/Settings.jsx:2:40)
      at Object.<anonymous> (mobile/test/settings.spec.js:3:43)
```

Jest assumes that `node_modules` packages' code are full JavaScript. To [configure `jest` (to not ignore) to transform the wanted JavaScript package's code you have to configure `jest` a bit more](http://facebook.github.io/jest/docs/tutorial-react-native.html#transformignorepatterns-customization) with something like that (in `package.json`):

```json
"transformIgnorePatterns": [
    "node_modules/(?!cozy-ui)"
]
```

As soon as you had it, there is still an error but on `.styl` file. It may be the same and it means we have to add some configuration for `.styl` file. This configuration is out of my capabilities for the evening.

So I let you find what more is wrong.

```
  ● Test suite failed to run
    /home/enguerran/Development/Cozy/cozy-files-v3/node_modules/cozy-ui/react/Modal/index.styl:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){@import '../stylus/ui-base/fonts'
                                                                                             ^
    SyntaxError: Invalid or unexpected token
      at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
      at Object.<anonymous> (node_modules/cozy-ui/react/Modal/index.jsx:4:40)
      at Object.<anonymous> (mobile/src/components/Settings.jsx:2:40)
```

enguerran:
One more thing: you refactored Settings and move the file from `containers` to `components`. You miss this path https://github.com/cozy/cozy-files-v3/blob/master/mobile/test/settings.spec.js#L3 (I know, it comes from a bad habit first 🐈)

enguerran:
This PR do not pass all tests because of a Jest configuration. But ... we will fix this later (note for the future: is "will fix later"'s still not fixed in the near future?)

PR #77, opened 2017-02-07
fix: set a unique name to the client name

enguerran:
With OAuth, a device must have a unique clientName to be identified by the server.

### DIFF
--- a/mobile/src/actions/index.js
+++ b/mobile/src/actions/index.js
@@ -11,6 +11,7 @@ export const ERROR = 'ERROR'
 const WRONG_ADDRESS = 'mobile.onboarding.server_selection.wrong_address'
 
 const error = () => ({ type: ERROR, error: WRONG_ADDRESS })
+const genId = () => Math.random().toString(36).slice(2)
 
 export class OnBoardingError extends Error {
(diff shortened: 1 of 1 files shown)


mobile/src/actions/index.js
```diff
@@ -49,7 +50,7 @@ export const registerDevice = (router, location) => {
         clientParams: {
           redirectURI: 'http://localhost',
           softwareID: 'io.cozy.mobile.files',
-          clientName: 'Mobile Cozy Files',
+          clientName: genId(),
```
aenario: > Hi, @enguerran would you like to let 3M (http://localhost)` access to your files ?

May be something along the lines of `CozyFiles Application on Device' + cordova.device.model + genID()` would be clearer.

enguerran: I do not want to add [](https://github.com/apache/cordova-plugin-device) to get `cordova.device.model`.

But without any plugin we can have something like `${cordova.platformId} (${genId()})` (in a template string) to have something like: **android (5ogdneqea9e)**. What do you think about, @aenario?

enguerran:
Replaced by #83

PR #81, opened 2017-02-08
Jinroh handle permissions

nono:
I've rebased #58 and updated yarn.lock

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{.Locale}}">
 <head>
   <meta charset="utf-8">
   <title><%= htmlWebpackPlugin.options.title %></title>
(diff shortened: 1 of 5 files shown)


src/index.ejs
```diff
@@ -12,6 +12,16 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <% if(__TARGET__ === 'mobile') { %>
   <script src="cordova.js" defer></script>
+  <% } else { %>
+  {{.CozyClientJS}}
```
m4dz: @kosssi do you need it in mobile flavor or do we just inject it in web version?

src/main.jsx
```diff
@@ -10,24 +10,11 @@ import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
 import { Router, hashHistory } from 'react-router'
 import cozy from 'cozy-client-js'
```
m4dz: We probably need to remove this dep as the `cozy-client-js` asset is now injected by the template to stuck the last compatible version with the stack, nope?

nono: Done with 35eb281

src/index.ejs
enguerran: That line removes the need to `import cozy from 'cozy-client-js' and may inject cozy in the global scope, does not it?

But it will be replaced by a script with a local file, we must ship our mobile app with every local files needed.

We, mobile app developers, want to be the closer we can to front developers stack.

m4dz: @enguerran if I understand well, you package a standalone version of `cozy-client-js` in your mobile bundle, so it's ok to have the dependency injected in web version only, true?

enguerran: no, sorry if I did not explain very well: `cozy-client-js` is bundled in an unique `app.js` file with the other JavaScript code.

m4dz:
@nono could you rebase it so we can merge it? Thanks!

nono:
@m4dz done

m4dz:
Great, thanks!

PR #82, opened 2017-02-08
Display a toast error on missing download file & offline download :fire:

kosssi:


src/styles/toast.styl
```diff
@@ -2,6 +2,11 @@
 
 .toast
     padding 0 5px
+    position fixed
+    bottom 50px
+    left 0
+    right 0
+    padding 0 10px
```
enguerran: There is 2 padding rules.

PR #83, opened 2017-02-08
Refactoring cozy init :dizzy:

enguerran:
I don't know how to update #79.

And I fix as well #77 in this PR.

So @aenario and @kosssi, are we cool?

### DIFF
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -10,7 +10,7 @@ import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
 import { Router, Route, hashHistory } from 'react-router'
 import localforage from 'localforage'
-import cozy from 'cozy-client-js'
+import { init } from './lib/cozy_init'
 
 import { I18n } from '../../src/lib/I18n'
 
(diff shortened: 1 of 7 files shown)


mobile/src/main.jsx
```diff
@@ -39,10 +39,7 @@ document.addEventListener('DOMContentLoaded', () => {
   localforage.getItem('state').then(state => {
     if (state) {
       store.dispatch({ type: SET_STATE, state })
-      cozy.init({
-        cozyURL: state.serverUrl,
-        offline: {doctypes: ['io.cozy.files'], timer: 15}
-      })
+      init(state.serverUrl)
```
kosssi: I have always this error: https://github.com/cozy/cozy-client-js/blob/master/src/index.js#L180

kosssi: This problem is only on ios. I change it on another PR

enguerran: I will check that with `cozy-client-js-0.0.9` and will push a fix.

enguerran: I checked with @aenario:

For third party application, cozy-client must be init with a oauth block. So I add this oauth block in the cozy-helper `init` function. No need for `onRegister` callback as I explain in https://github.com/cozy/cozy-files-v3/pull/83/commits/b90e252120d709fd4da48075c6182461e94116c6.

kosssi:
\o/

PR #84, opened 2017-02-08
[fix] Fil 16 fixes

y-lohse:
Fixes some minor issues:

- modal close button and mobile view
- a visual regression affecting checkboxes
- simplified translations
- better notifications

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "classnames": "^2.2.0",
     "cozy-bar": "3.0.0-beta3",
     "cozy-client-js": "0.0.7",
-    "cozy-ui": "3.0.0-beta6",
+    "cozy-ui": "3.0.0-beta7",
     "date-fns": "^1.22.0",
     "filesize": "^3.3.0",
     "localforage": "^1.4.3",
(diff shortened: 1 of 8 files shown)


y-lohse:
I agree that the names should be the same. However, I was more inclined to eventually  rename "Alerters" as "Notifications" as I feel it's a more accurate description of what it does. But you confirm it's a conscious choice that has already been discussed, I'm happy to rename the few instances of "notifications".

enguerran:
Did you see @kosssi add a modal component in cozy-ui? (see https://github.com/cozy/cozy-ui/pull/53)

y-lohse:
Yes, I think it will make sense to use that modal eventually. But that implies another cozy-ui update first, so I'd like to land FIL-16 before. It's been in the works for way too long already.

gregorylegarec:
@y-lohse Yep we already discussed about the naming of notifications/alerts, we concluded that notification was a more complex model, (they can be for example stacked and "waiting" for consultation), as the alert was more the UI aspect of it. By the way, the opinions in the team were divided, so if this naming confuse too much developers and contributors, we may discuss to change it again in the future.

gregorylegarec:
Great job @y-lohse !

PR #87, opened 2017-02-09
[feat] Use error type on alerts related to failures

gregorylegarec:
@y-lohse Related to our previous work on alerts, what do you think about setting type to error on alerts related to failures ?


y-lohse:
Actually, I had a chat with @jsilvestre and all notifications are meant to be grey at the moment, if I remember correctly. I left the success and error styles just in case, but I'm not sure they are going to be used.

gregorylegarec:
@y-lohse OK thanks, let's close this for the moment.

PR #88, opened 2017-02-09
[fix] added specific styles for the input

y-lohse:
This input needs to be smaller than the usual ones defined in cozy-ui.

### DIFF
--- a/src/styles/filenameinput.styl
+++ b/src/styles/filenameinput.styl
@@ -4,5 +4,5 @@
     @extends $form-text
 
     input[type=text]
-        padded(6px)
-        width 100%
+        padded 6px
+        width  100%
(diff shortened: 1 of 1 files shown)


y-lohse:
Ow, thanks? And here I was, thinking my IDE was being stupid.

A lot of other stylus files have inconsistent indentation, I'll create a different PR to fix those.

src/styles/filenameinput.styl
```diff
@@ -1,5 +1,8 @@
 @require 'cozy-ui'
 
 .fil-file-name-input
-    @extends $form
     @extends $form-text
+
+    input[type=text]
+        padded(6px)
```
GoOz: Could you please write (I know i'm a little monomaniac on this…):
`padded 6px` instead of `padded(6px)`
It will do exactly the same thing but values would be aligned. 😄

PR #89, opened 2017-02-09
Archive download & other things

goldoraf:
This PR:

- adds the download of a selection of files as an archive
- fixes the missing manifest.webapp requested by the stack
- fixes the error related to bosonic's menu

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -53,28 +53,21 @@ export const openFolder = (folderId = ROOT_DIR_ID, isInitialFetch = false, route
       dispatch({ type: FETCH_FILES, folderId })
     }
     dispatch({ type: OPEN_FOLDER, folderId })
-    let folder
+    let folder, parent
     try {
       folder = await cozy.files.statById(folderId)
-    } catch (err) {
-      return dispatch({type: OPEN_FOLDER_FAILURE, error: err})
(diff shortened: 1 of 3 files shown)


src/actions/index.js
```diff
@@ -212,8 +206,13 @@ export const downloadSelection = () => {
...
     dispatch({ type: DOWNLOAD_SELECTION, selected })
-    selected.forEach(id => dispatch(toggleFileSelection(id, true)))
-    return downloadFile(selected[0])
+    if (selected.length === 1 && getFileById(getState().files, selected[0]).type !== 'directory') {
+      return dispatch(downloadFile(selected[0]))
+    }
+    const paths = getFilePaths(getState(), selected)
+    const href = await cozy.files.getArchiveLink(paths)
+    const fullpath = await cozy.fullpath(href)
+    forceFileDownload(fullpath, 'files.zip')
```
y-lohse: Should this be localized?

y-lohse:
Just the one tidbit if you think it should be localized, otherwise it's all good 👍

src/actions/index.js
goldoraf: I don't think so... Maybe later, let's see what our POs think about that ;)

PR #90, opened 2017-02-09
[fix] FIL-16 : Prevent notifications when the aborting folder creation on purpose

y-lohse:
When a user aborts a folder creation with escape, no notification should be shown.


gregorylegarec:
Thanks @y-lohse 👍

PR #97, opened 2017-02-10
Improve webpack config

m4dz:


enguerran:
It may be a good time to migrate to webpack2, just sayin' 😜

m4dz:
@enguerran I know, and don't forget to do it, but not for this one, and not until I have enough time :wink:

PR #100, opened 2017-02-13
[refactor] Fil 18

y-lohse:
~~Before being complete, this PR needs a new version of cozy-ui, pending [here ](https://github.com/cozy/cozy-ui/pull/59).~~ All good now.

But the code is ready for review.

This PR replaces the Toaster component with the Alerter component from cozy-ui, since they both look the same on mobile now.

The other notable change is the way alerts are generated. Until now, we did it in 2 different ways:

- a generic action type (`DISPLAY_TOAST`) with additional info in the action object
- several specific action types (`TRASH_FAILURE`, `DUPLICATE_FOLDER_NAME`, ...), with the additional info being computed by the reducers

The problem with the first approach was that these actions were unusable by any other reducer as they were too generic. The problem with the *second* approach was that the reducers were growing in complexity with each new alert, when they really shouldn't be doing a lot of work at all.

The new approach is -- I hope -- a happy compromise between both: action types are specific, which makes them verbose and re-usable, but an alert is generated for [every action that has an `alert` property](https://github.com/cozy/cozy-files-v3/compare/master...y-lohse:FIL-18?expand=1#diff-56772dffec7d4d45eb4ac3f8d9709e77R237), regardless of the type. So the reducer can stay as it is from now on, and any extra logic is done in the action creators.


y-lohse:
Heads up for @enguerran : The file-opening errors are affected by this change. I tested them in the browser and they appeared to work, but I haven't been able to test the cordova specific parts.

src/actions/index.js
```diff
@@ -33,19 +33,28 @@ export const DOWNLOAD_SELECTION = 'DOWNLOAD_SELECTION'
...
+export const DOWNLOAD_FILE_E_MISSING = 'DOWNLOAD_FILE_E_MISSING'
+export const DOWNLOAD_FILE_E_OFFLINE = 'DOWNLOAD_FILE_E_OFFLINE'
 export const OPEN_FILE_WITH = 'OPEN_FILE_WITH'
-export const DISPLAY_TOAST = 'DISPLAY_TOAST'
-export const HIDE_TOAST = 'HIDE_TOAST'
+export const OPEN_FILE_E_OFFLINE = 'OPEN_FILE_E_OFFLINE'
+export const OPEN_FILE_E_NO_APP = 'OPEN_FILE_E_NO_APP'
 export const ALERT_CLOSED = 'ALERT_CLOSED'
 
+const T_DOWNLOAD_FILE_MISSING = 'error.download_file.missing'
```
goldoraf: Not sure these constants are necessary (there are already a lot of action types consts here...). What do you think about using string keys in the action creators instead of these consts?

y-lohse: I'd be in favor, if only because the translation keys and action types have very similar names.

kosssi:
@y-lohse can you naming your PR without trello reference 😜

y-lohse:
Uuh ok, sure.

PR #102, opened 2017-02-13
refactor: mobile settings are persisted :vhs:

enguerran:
We use localforage to store settings in IndexedDB within redux thanks to really good redux API.

See https://egghead.io/lessons/javascript-redux-persisting-the-state-to-the-local-storage

### DIFF
--- a/mobile/src/actions/index.js
+++ b/mobile/src/actions/index.js
@@ -19,6 +19,8 @@ const WRONG_ADDRESS_ERROR = 'mobile.onboarding.server_selection.wrong_address'
 const OPEN_WITH_OFFLINE_ERROR = 'mobile.error.open_with.offline'
 const OPEN_WITH_NO_APP_ERROR = 'mobile.error.open_with.noapp'
 
+export const setUrl = url => ({ type: SET_URL, url })
+
 export const wrongAddressError = () => ({ type: ERROR, error: WRONG_ADDRESS_ERROR })
 export const openWithOfflineError = () => ({ type: DISPLAY_TOAST, message: OPEN_WITH_OFFLINE_ERROR })
 export const openWithNoAppError = () => ({ type: DISPLAY_TOAST, message: OPEN_WITH_NO_APP_ERROR })
@@ -60,7 +62,7 @@ export const checkURL = url => async dispatch => {
   if (!url.startsWith(scheme)) {
     url = `${scheme}${url}`
   }
-  return dispatch({ type: SET_URL, url: url })
+  return dispatch(setUrl(url))
 }
 
 const openRegistrationWith = inAppBrowser => new Promise((resolve) => {
--- a/mobile/src/actions/index.js
+++ b/mobile/src/actions/index.js
@@ -95,7 +95,7 @@ const onRegistered = dispatch => (client, url) => {
   }
 }
 
-export const registerDevice = (router, location) => async (dispatch, getState) => {
+export const registerDevice = () => async (dispatch, getState) => {
   await dispatch(checkURL(getState().mobile.settings.serverUrl))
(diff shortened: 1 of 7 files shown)


mobile/src/reducers/mobile.js
```diff
@@ -16,9 +16,7 @@ export const mobile = (state = initialState, action) => {
     case SET_STATE:
       return action.state
     case SET_URL:
-      return Object.assign({}, state, { serverUrl: action.url, error: null })
-    case SETUP:
-      return Object.assign({}, state, { isSetup: true, error: null })
+      return Object.assign({}, state, { settings: { serverUrl: action.url }, error: null })
```
kosssi: now it's possible to use spread operator ;)

mobile/src/actions/index.js
```diff
@@ -95,7 +95,7 @@ const onRegistered = dispatch => (client, url) => {
   }
 }
 
-export const registerDevice = (router, location) => async (dispatch, getState) => {
+export const registerDevice = () => async (dispatch, getState) => {
```
kosssi: 👍

mobile/src/lib/localStorage.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const loadState = async () => {
+  try {
+    const persistedState = await localforage.getItem('state')
+    if (persistedState === null) {
+      return undefined
+    }
+    return persistedState
+  } catch (err) {
+    return undefined
```
kosssi: may be you can console.warn(err)?

mobile/src/reducers/mobile.js
```diff
@@ -20,7 +20,7 @@ export const mobile = (state = initialState, action) => {
     case ERROR:
       return Object.assign({}, state, { error: action.error })
     case UPDATE_SETTINGS:
-      return Object.assign({}, state, Object.assign({}, state.settings, action.newSettings))
+      return Object.assign({}, state, { settings: Object.assign({}, state.settings, action.newSettings) })
```
kosssi: same with spread operator ;)

mobile/src/actions/index.js
```diff
@@ -92,6 +92,15 @@ const onRegistered = dispatch => (client, url) => {
         throw err
       }
     )
+  } else {
+    // TODO open a new tab ?
```
kosssi: Could you explain this code is only for development?

mobile/src/reducers/mobile.js
enguerran: Now we add spread operator for ojects, we must refactor our reducer with it. Agreed!

mobile/src/actions/index.js
enguerran: As there is no inapp browser, we need to open the `oauth/` page in a new tab (or maybe an iframe) then get the response's url to give it back to `cozy-client-js`. For now, a prompt ask for this url (the one with `access_code` query parameter) and it initialize `cozy-client-js` letting it to add the token in its persistence layer.

mobile/src/lib/localStorage.js
enguerran: I don't know what error could append here, so I mute it and just let the redux store be initialized without any previous persisted state by returning `undefined`.
Nevertheless I guess it could help if we add a `console.warn` in the catch block.

mobile/src/actions/index.js
kosssi: I would like to say you can explain it on code comment ;)

enguerran: done: https://github.com/cozy/cozy-files-v3/pull/102/commits/09bac036149f51deea79e7652143591b84a2d14b#diff-d531d6391203bbbe5792216cf13ff430R96

PR #103, opened 2017-02-13
[mobile] Upload local picture :camera:

kosssi:


mobile/src/actions/media_backup.js
```diff
@@ -0,0 +1,20 @@
...
+  let photos = await getPhotos()
+  for (let photo of photos) {
+    const blob = await getBlob(photo)
+    const options = {
+      dirID: 'io.cozy.files.root-dir',
+      name: photo.filename
+    }
+    const created = await cozy.files.create(blob, options).catch(err => {
+      console.log(err)
+    })
```
enguerran: I do like this kind of syntax:

```javascript
const created = await cozy.files.create(blob, options).catch(console.log.bind(console))
```

Don't you?

mobile/src/actions/media_backup.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const MEDIA_UPLOAD_START = 'MEDIA_UPLOAD_START'
+export const MEDIA_UPLOAD_END = 'MEDIA_UPLOAD_END'
+
+export const mediaBackup = () => async dispatch => {
+  dispatch({ type: MEDIA_UPLOAD_START })
+  let photos = await getPhotos()
+  for (let photo of photos) {
+    const blob = await getBlob(photo)
+    const options = {
```
enguerran: I do prefer this:

```javascript
photos.forEach(photo => {
  const blob = await getBlob(photo)
  // etc.
})
```

mobile/src/components/Settings.jsx
```diff
@@ -13,7 +13,7 @@ const SubCategory = ({ id, label, value, title }) => (
   </div>
 )
 
-export const Settings = ({ t, version, serverUrl, backupImages, setBackupImages, showUnlinkConfirmation, displayUnlinkConfirmation, hideUnlinkConfirmation, unlink }) => (
+export const Settings = ({ t, version, serverUrl, backupImages, setBackupImages, showUnlinkConfirmation, displayUnlinkConfirmation, hideUnlinkConfirmation, unlink, mediaUploading, launchBackup }) => (
```
enguerran: Reading this, I start to consider it will be a good thing to connect SubSettings component rather than the whole Settings component. Do you agree?

Note: it is not necessary to include this re-factorization in this exact PR.

mobile/src/lib/media.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const requestAuthorization = async () => {
+  if (hasCordovaPlugin()) {
+    return new Promise((resolve, reject) => {
+      window.cordova.plugins.photoLibrary.requestAuthorization(
+        () => resolve(true),
+        (error) => resolve(false)
+      )
+    })
```
enguerran: I believe you can write something like:

```javascript
return async () => {
  try {
    await window.cordova.plugins.photoLibrary.requestAuthorization()
    return true
  } catch (error) {
    return false
  }
}
```

mobile/src/lib/media.js
```diff
@@ -0,0 +1,59 @@
...
+                resolve(defaultReturn)
+              }
+            })
+          } else {
+            console.warn(err)
+            resolve(defaultReturn)
+          }
+        }
+      )
+    })
```
enguerran: It may also be rewritten with `async/await` syntax, doesn't it?

mobile/src/lib/media.js
```diff
@@ -0,0 +1,59 @@
...
+        (error) => resolve(false)
+      )
+    })
+  }
+  return false
+}
+
+export const getBlob = async (libraryItem) => {
+  if (hasCordovaPlugin()) {
+    return new Promise((resolve, reject) => {
```
enguerran: same as other comments about `async/await`.

mobile/src/actions/media_backup.js
kosssi: forEach not support async function ;)

mobile/src/lib/media.js
kosssi: It's not possible because requestAuthorization is not a Promise ;)

mobile/src/lib/media.js
kosssi: I make promise because photoLibrary don't return Promise ;)

mobile/src/actions/media_backup.js
kosssi: I don't like `.bind` and it's not really use on cozy-files-v3 ;)

PR #107, opened 2017-02-14
Don't upload image which is already uploaded ☔️

kosssi:


mobile/src/actions/media_backup.js
```diff
@@ -1,21 +1,31 @@
 import cozy from 'cozy-client-js'
 import { getPhotos, getBlob } from '../lib/media'
+import { HTTP_CODE_CONFLICT } from '../../../src/actions'
 
 export const MEDIA_UPLOAD_START = 'MEDIA_UPLOAD_START'
 export const MEDIA_UPLOAD_END = 'MEDIA_UPLOAD_END'
+export const IMAGE_UPLOAD_SUCCESS = 'IMAGE_UPLOAD_SUCCESS'
 
-export const mediaBackup = () => async dispatch => {
+export const mediaBackup = () => async (dispatch, getState) => {
```
enguerran: Be ware that using `getState` in a redux action creator is an anti-pattern. In [this SO answer](http://stackoverflow.com/questions/35411423/how-to-dispatch-a-redux-action-with-a-timeout/35415559#35415559), it said:

> Don’t abuse this pattern. It is good for bailing out of API calls when there is cached data available, but it is not a very good foundation to build your business logic upon. If you use getState() only to conditionally dispatch different actions, consider putting the business logic into the reducers instead.

PR #110, opened 2017-02-15
Simplify dev process on stack assets

m4dz:
:warning: This PR relies on https://github.com/cozy/cozy-bar/pull/12 and https://github.com/cozy/cozy-client-js/pull/74, merge them before pliz

This PR allow devs to use local version of stack assets rather than the ones delivered by the stack in development condition.

# How it works?

Stack assets like `cozy-client-js` and `cozy-bar` are injected into the DOM as legacy-scripts by the stack when it serves the app to the browser. Sometimes, we need to overrides them with local versions when a developper needs to works on them and tests them in a real app environment.

When lanching the app in _watch mode_ (e.g. `yarn run watch`), then local versions (downloaded from the npm registry into the `node_modules` directory) are used by the app instead of the stack's ones (they are implicitely detected and bundled by webpack). 

If a developper need to work on those assets, he simply have to link them in the local app:

```sh
$ cd ~/cozy-client-js
$ yarn link
$ cd ~/cozy-files-v3
$ yarn run link cozy-client-js
$ yarn run watch
```

This way, the local version of `cozy-client-js` is bundled in the dev version of `cozy-files-v3`. Launching a watcher on the `cozy-client-js` directory in parallel of a watcher in the app directory results of a rebuild of the app automatically when saving some changes in `cozy-client-js` :raised_hands:

# About the build

Local versions of assets are only used in _watch mode_. The build does *not* bundle them in the `app.js` file, and stack assets still continue to be used in _production mode_.

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -1,0 +1,24 @@
+'use strict'
+
+const webpack = require('webpack')
+
+module.exports = {
+  devtool: '#source-map',
+  externals: ['cozy'],
+  module: {
+    loaders: [{
+      test: require.resolve('cozy-bar/dist/cozy-bar.js'),
+      loader: 'imports?css=./cozy-bar.css'
+    }]
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      __SERVER__: JSON.stringify('http://app.cozy.local'),
+      __STACK_ASSETS__: false
+    }),
+    new webpack.ProvidePlugin({
+      'cozy.client': 'cozy-client-js/dist/cozy-client.js',
+      'cozy.bar': 'cozy-bar/dist/cozy-bar.js'
+    })
+  ]
+}
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -20,7 +20,7 @@ module.exports = {
     loaders: [
       {
         test: /\.js$/,
-        exclude: /node_modules/,
+        exclude: /(node_modules|cozy-(bar|client-js))/,
         loader: 'babel-loader'
       },
       {
@@ -34,6 +34,9 @@ module.exports = {
           'postcss-loader'
         ])
       }
+    ],
+    noParse: [
+      /localforage\/dist/
     ]
   },
   plugins: [
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -5,7 +5,7 @@ const HtmlWebpackPlugin = require('html-webpack-plugin')
 const ScriptExtHtmlWebpackPlugin = require('script-ext-html-webpack-plugin')
 const PostCSSAssetsPlugin = require('postcss-assets-webpack-plugin')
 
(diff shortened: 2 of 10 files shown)


enguerran:
Once again, there is some failing tests. Could you fix them?

nono:
@enguerran as it is said above, this PR depends on other PRs that have not yet been merged. I think the failure of the build is due to that

enguerran:
Oops, ok. My bad. I'll wait.

config/webpack.config.dev.js
```diff
@@ -0,0 +1,24 @@
...
+  externals: ['cozy'],
+  module: {
+    loaders: [{
+      test: require.resolve('cozy-bar/dist/cozy-bar.js'),
+      loader: 'imports?css=./cozy-bar.css'
+    }]
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      __SERVER__: JSON.stringify('http://app.cozy.local'),
```
gregorylegarec: What the role of `JSON.stringify` here ?

Should we provide the port ? (8080)

config/webpack.config.dev.js
```diff
@@ -0,0 +1,24 @@
...
+      __STACK_ASSETS__: false
+    }),
+    new webpack.ProvidePlugin({
+      'cozy.client': 'cozy-client-js/dist/cozy-client.js',
+      'cozy.bar': 'cozy-bar/dist/cozy-bar.js'
```
gregorylegarec: Just for my own culture : I need explanation here, do `cozy-client-js/` and `cozy-bar/` have to be subfolders of our current app ?

config/webpack.config.dev.js
gregorylegarec: Other though here : will it work with the new mechanism of `cozy-stack serve --appdirs` https://github.com/cozy/cozy-stack/blob/master/cmd/serve.go#L35, like on a locally served `http://files.cozy.local` for example ?
ping @nono

m4dz: Pretty good question, I thinks it's a perfectly valid way to use local assets in an app served locally. I never tested it, but it's the same process like the `volume` flag on the docker image :smiley:

config/webpack.config.dev.js
m4dz: Nope, they're node modules, so in the `node_modules` folder. They're installed from the NPM registry as a _peerDependency_, but can be overriden with `cozy link` (or `yarn link`) command with local versions.

Webpack can properly handle them, as their subdir contents. The `ProvidePlugin` here acts like if you do `const cozy.client = require('cozy-client-js/dist/cozy-client.js')` at the top of each file which contains the variable `cozy.client`. And so on, Webpack access subdir contents of the module when you require a pattern like `<moduule_slug>/path/to/file`.

m4dz:
Tests are now :green_apple: ! @nono does it sounds good to you?

PR #111, opened 2017-02-16
[feat] FIL 20 -- The Trash Edition

y-lohse:
First, I apologize to the person(s) who will review this PR, considering the number and nature of the changes. I'm going to annote and highlight the trickier changes, or the ones that may be improved.

This PR adds the Trash view. It's a whole new view, but it's also very similar to the existing Files view: you mainly browse files and folder. The differences are in the interactions, as you can only choose to restore some files, or delete them permanently (the latter is not part of this PR, only the UI bits).


src/actions/index.js
```diff
@@ -51,6 +54,11 @@ export const downloadFileOffline = () => ({ type: DOWNLOAD_FILE_E_OFFLINE, alert
 
 export const openFolder = (folderId = ROOT_DIR_ID, isInitialFetch = false, router = null) => {
   return async dispatch => {
+    let routePrefix = '/files'
```
y-lohse: `openFolder` is now used to navigate in `/files`and `/trash`, so having a hard-coded route prefix wasn't an option any more. The prefix could be supplied as an argument, but since we already have the router at hand, I think it's safe to extract the prefix from there.

src/actions/index.js
```diff
@@ -193,8 +201,8 @@ export const deleteFileOrFolder = (id, isNew = false) => {
   }
 }
 
-export const trashFile = (id) => {
```
y-lohse: Just some unrelated syntax changes here.

src/assets/icons/icon-trash-dark.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
+  <g fill="#32363f" fill-rule="evenodd" transform="translate(-96 -32)">
```
y-lohse: @GoOz next on my wishlist: figuring out a way to change svg colors on the fly 😁

src/components/App.jsx
```diff
@@ -1,30 +1,17 @@
...
 
-import { openFolder, alertClosed } from '../actions'
+import { alertClosed } from '../actions'
 
 import Alerter from 'cozy-ui/react/Alerter'
 import Sidebar from './Sidebar'
 import Topbar from './Topbar'
 
 class App extends Component {
-  componentWillMount () {
```
y-lohse: One of the major changes: the App component is no longer the one who initializes data load. Since it's used across the whole app and the data that needs to be loaded changes on each view, I feel like this is the wrong place to do it.
Now the data is loaded by the appropriate containers, eg. `Folder` and `Trash`. This is also consistent with what we do in cozy-photos AFAIK.

src/components/Nav.jsx
```diff
@@ -9,6 +9,7 @@ import { Link, withRouter } from 'react-router'
 
 const Nav = ({ t, router }) => {
   let isBrowsingFiles = router.location.pathname.match(/^\/files/) !== null
+  let isBrowsingTrash = router.location.pathname.match(/^\/trash/) !== null
```
y-lohse: I feel like there should be a better, more global way to do this. As you can see later in this PR, I've even added a reducer that does it and used it in a couple of places. However, we haven't one for `isBrowsingFiles `, so maybe we should add that too? Then again, the router is a relatively good way of finding out what the user is viewing.

src/containers/Folder.jsx
```diff
@@ -10,6 +10,18 @@ import FileList from '../components/FileList'
 const isDir = attrs => attrs.type === 'directory'
 
 class Folder extends Component {
+  componentWillMount () {
```
y-lohse: So here we find the routing code that I removed from `App.jsx`.

src/reducers/index.js
```diff
@@ -54,4 +54,6 @@ export const getActionableFile = ({ files, ui }) => getFileById(files, ui.action
 
 export const mustShowSelectionBar = state => state.ui.showSelectionBar || state.ui.selected.length !== 0
 
+export const isBrowsingTrash = state => state.folder && state.folder.path && state.folder.path.indexOf(TRASH_DIR_PATH) === 0
```
y-lohse: I'm not overly fond of th new `TRASH_DIR_PATH` constant, but I really can't see a better way to do this. The main problem is, we only ever know:

- what the current folder's id is
- what the parent's folder id is
- the id of that parent's dir

But if the current folder is 4 levels deep inside the trash, there's no way to find that information. Except in stringified path.

src/styles/breadcrumb.styl
```diff
@@ -8,6 +8,8 @@
     color #95999d
     a
         cursor pointer
+    .separator
```
y-lohse: Sort of unrelated to the PR, but the spacing in the breadcrumb was relying on literal spaces and was really inconsistent. Now it's handled with css!

src/assets/icons/icon-trash-dark.svg
GoOz: Something that worked so far but I didn't get the chance to make it a rule is to fill the SVG with the `currentColor` keyword and then then change the `color` property of the element wrapping the SVG **BUT ONLY** when we're talking about inline SVGs, not in CSS background :P

src/styles/toolbar.styl
```diff
@@ -6,6 +6,18 @@
     .coz-btn:last-child
         margin-right 0
 
+.fil-toolbar-trash
+    @extends $buttons
```
y-lohse: @GoOz  and @m4dz : this one was an interesting problem. The related DOM is [here](https://github.com/y-lohse/cozy-files-v3/blob/FIL-20/src/containers/TrashToolbar.jsx#L15).

The problem I encountered was, the first button (`coz-btn--delete`) really needs the `danger-outline` class that is defined [here](https://github.com/cozy/cozy-ui/blob/v3/stylus/ui-components/button.styl#L118). But when I do that, I import the selector `[role=button], button`, and now [the other button](https://github.com/y-lohse/cozy-files-v3/blob/FIL-20/src/containers/TrashToolbar.jsx#L21) *also* inherits these styles -- because this selector is stronger than `coz-btn coz-btn--more`.

Is there a way to @extend only some parts of the placeholder? That would be quite handy in cases like this.

src/assets/icons/icon-trash-dark.svg
y-lohse: Yes, this is what I used to do as well and it worked great. But as you said, it only works for inline SVG so I was only using that, whereas the current approach in cozy is to use background everywhere. Unless there's a reason to prefer backgrounds, I'd be in favor of switching to inline svg tbh.

src/styles/toolbar.styl
GoOz: Should be fixed with the Cozy-UI refactoring as $buttons doesn't exist anymore and the danger-outline is included in the $button placeholder… but until then I don't really have a clean solution :/

src/assets/icons/icon-trash-dark.svg
GoOz: I'm too questioning that choice more and more lately. It's semantically more correct to use the background technique but not really the more pragmatic way. Any thought @m4dz?

src/styles/toolbar.styl
y-lohse: Alright, thanks. Good to know a fix is on the way :-)

src/actions/index.js
goldoraf: I understand, but I'd suggest to add a comment here, because the code is not very explicit.

src/assets/icons/icon-trash-dark.svg
goldoraf: https://medium.com/@david.gilbertson/icons-as-react-components-de3e33cb8792#.xzmogaxo8
What do you think of this solution?

src/components/Nav.jsx
goldoraf: Yeah, and the v4 of react-router should make this more easy to do.

src/containers/Trash.jsx
```diff
@@ -0,0 +1,63 @@
...
+import { connect } from 'react-redux'
+import { withRouter } from 'react-router'
+
+import { openFolder, toggleFileSelection, showFileActionMenu } from '../actions'
+import { getVisibleFiles, mustShowSelectionBar, isBrowsingTrash } from '../reducers'
+import { TRASH_DIR_ID } from '../constants/config'
+
+import FileList from '../components/FileList'
+
+class Folder extends Component {
```
goldoraf: It should be `class Trash` here ;)

y-lohse:
Changes made, thanks for the review! Should be ready to merge now.

PR #114, opened 2017-02-16
Add File action menu on selection bar

GoOz:
⚠️ To be merged after Cozy-ui's PR https://github.com/cozy/cozy-ui/pull/60

- [x] Updated selection bar on mobile
- [x] Added FileActionMenu on selection bar
- [x] Added actions in FileActionMenu such as download a file or files and delete files

Some improvements can certainly be made, please share your wisdom with me. :)


src/containers/DeleteConfirmation.jsx
```diff
@@ -19,30 +20,32 @@ const DeleteConfirmation = ({ t, selected, onConfirm, onDismiss }) => {
...
-  onConfirm: (selected) => {
+  onConfirm: (selected, dropSelection) => {
     selected.forEach(item => {
-      dispatch(trashFile(item))
-      dispatch(toggleFileSelection(item, true))
+      dispatch(trashFile(item.id))
+      if (dropSelection) dispatch(toggleFileSelection(item.id, true))
     })
     dispatch(hideDeleteConfirmation())
+    dispatch(hideFileActionMenu())
```
y-lohse: IMO, this can stay as it is for this PR, but we should probably change this later. Seeing it now, I think we are doing too many things un this component that aren't it's job (dropping the selection, hiding stuff). I think those things could be handled by reducers that react to a trash success / failure rather than explicits dispatch. Any thoughts, @goldoraf ?

src/containers/FileActionMenu.jsx
```diff
@@ -8,36 +8,81 @@ import { Item } from 'react-bosonic/lib/Menu'
...
-        </a>
-      </Item>
-      <Item>
-        <a className={styles['fil-action-download']} onClick={() => onDownload(file.id)}>
-          {t('mobile.action_menu.download')}
-          {actionMenu.download && <div className={styles['fil-loading']} />}
-        </a>
-      </Item>
+      {files.length === 1 && <ItemOpenWith {...props} />}
+      {files.length === 1 ? <DownloadFile file={files[0]} {...props} /> : <DownloadSelection {...props} />}
```
y-lohse: Nice, I like these changes!

src/containers/FileActionMenu.jsx
```diff
@@ -58,19 +103,28 @@ const FileActionMenu = props => (
...
+  onDownloadFile: id => {
     const clearUI = () => {
       dispatch(actionMenuLoaded('download'))
       dispatch(hideFileActionMenu())
     }
     dispatch(actionMenuLoading('download'))
     dispatch(downloadFile(id)).then(clearUI).catch(clearUI)
   },
+  onDownloadSelection: () => {
+    const clearUI = () => {
```
y-lohse: Have you considered declaring this function once and for all, outside of the body of `onDownloadSelection ()`?

src/containers/DeleteConfirmation.jsx
```diff
@@ -19,30 +20,32 @@ const DeleteConfirmation = ({ t, selected, onConfirm, onDismiss }) => {
...
     validateText={t('deleteconfirmation.delete')}
-    validateAction={() => onConfirm(selected)}
+    validateAction={() => onConfirm(selected, dropSelection)}
    />)
 }
 
 const mapStateToProps = (state, ownProps) => {
   return {
-    selected: state.ui.selected
+    selected: getActionableFiles(state),
```
GoOz: true :)

src/containers/DeleteConfirmation.jsx
GoOz: True. As discussed with @goldoraf while coding the downloading part, we came to a conclusion that it's starting to get a little messy, although I don't think I'm the right person to clean it up 😛 
But it surely should be refactored.

src/containers/FileActionMenu.jsx
GoOz: Indeed. I think I can work something out. :)

GoOz: Actually… I can't find a suitable and flexible solution.

y-lohse: Ok, no problem :-)

y-lohse:
After trying it out, everything seems to work quite well -- of course there are some UI problems, but I assume they will be fixed by the PR in cozy-ui. The only thing I noticed that might be a bug is that when you download a file from the selection bar, the selection is dropped. But if you download it via the FAM from the selection bar, the selection stays. Not sure which one is right, but I assume they should be the same?

GoOz:
Oh good catch. You're right! 😄

enguerran:
When the original comment say:
> ⚠️ To be merged after Cozy-ui's PR cozy/cozy-ui#60

Don't we need to wait for https://github.com/cozy/cozy-ui/pull/60 to be merged?

PR #118, opened 2017-02-20
feature: add sentry to watch errors

enguerran:
This new feature gives access to sentry online logger.
JavaScript uses Raven to connect to sentry.cozycloud.cc.

Sentry's Dashboard looks like that:

screen shot 2017-02-20 at 18 07 16-fullpage

In another development, we will add an option (activable via settings) to trigger or not the error's capture.
As Sentry takes data from customer (private data such as IP address, and less sensible data), we have to check twice if the user is ok to send data, and what data the application sends. And we need to make a declaration for the purpose of our collect.

### DIFF
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -9,6 +9,7 @@ import { createStore, applyMiddleware } from 'redux'
 import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
 import { Router, Route, hashHistory } from 'react-router'
+import Raven from 'raven-js'
 
 import { I18n } from '../../src/lib/I18n'
 
@@ -21,6 +22,8 @@ import Settings from './containers/Settings'
 
 import { loadState, saveState } from './lib/localStorage'
 import { init } from './lib/cozy-helper'
+
+Raven.config('https://29bd1255b6d544a1b65435a634c9ff67@sentry.cozycloud.cc/2').install()
 
 const context = window.context
 const lang = (navigator && navigator.language) ? navigator.language.slice(0, 2) : 'en'
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -1,3 +1,5 @@
+/* global __SENTRY_TOKEN__ */
+
 import 'babel-polyfill'
 
 import '../../src/styles/main'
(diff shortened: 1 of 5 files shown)


mobile/src/main.jsx
```diff
@@ -22,6 +23,8 @@ import Settings from './containers/Settings'
 import { loadState, saveState } from './lib/localStorage'
 import { init } from './lib/cozy-helper'
 
+Raven.config('https://29bd1255b6d544a1b65435a634c9ff67@sentry.cozycloud.cc/2').install()
```
nono: I don't like hard-coded URLs like this. Can we move this URL to the webpack config?

enguerran: Yes, we can. I may have @m4dz advice about how webpack can handle this.

enguerran:
@nono We will add it on the next sprint as explained in the trello card:

screenshot at 09-15-27

As I say in the first comment, fetching user's data is a big deal and I had a small presentation at Apero Web Toulouse about this. The speaker told us about https://www.cnil.fr/fr/comprendre-vos-obligations with a lot a useful resources.

mobile/src/main.jsx
m4dz: @enguerran does the token in the URL always the same, or may it change during runtime?

enguerran: It is build as: `https://<key>@sentry.cozycloud.cc/<project>` where `key` is generated from [the settings page](https://sentry.cozycloud.cc/sentry/cozy-files-v3/settings/keys/) and `project` is the ID of the project.

So, it will be always the same until we revoke it and generate a new one.

m4dz: Cool! I suggest we add in the `ProvidePlugin` (https://github.com/cozy/cozy-files-v3/blob/master/config/webpack.target.mobile.js#L19-L22) a compil variable named `__SENTRY_TOKEN__` and we inject it like:

```js
Raven.config(`https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`).install()
```

it'll be injected properly on build.

enguerran: Ok, I'll do it. Thanks @m4dz

PR #120, opened 2017-02-21
[Chore] Update Cozy-UI dep

GoOz:


y-lohse:
Shouldn't there be a change in the yarn file as well?

GoOz:
@y-lohse You're damn right. Fixed 😄

y-lohse:
No worries, I noticed because the same happened to me last time :-).

PR #123, opened 2017-02-21
update sentry integration

enguerran:
It appears to be a good tips to add the current state if a dispatch throws an exception.

Cf. https://blog.sentry.io/2016/08/24/redux-middleware-error-logging.html

So these commits introduce two concepts:
- an automatic redux dispatch exception logger
- a helper `logException` to let developer add exception in Sentry

### DIFF
--- a/mobile/src/lib/crash-reporter.js
+++ b/mobile/src/lib/crash-reporter.js
@@ -1,0 +1,15 @@
+/* global __SENTRY_TOKEN__ */
+import Raven from 'raven-js'
+
+export function logException (err, context) {
+  if (!Raven.isSetup()) {
+    Raven.config(`https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`).install()
+  }
+  Raven.captureException(err, {
+    extra: context
+  })
+  console.groupCollapsed('Raven is recording exception')
+  console.error(err)
+  console.info(context)
+  console.groupEnd()
+}
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -11,7 +11,7 @@ import { createStore, applyMiddleware } from 'redux'
 import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
 import { Router, Route, hashHistory } from 'react-router'
-import Raven from 'raven-js'
+import RavenMiddleWare from 'redux-raven-middleware'
 
 import { I18n } from '../../src/lib/I18n'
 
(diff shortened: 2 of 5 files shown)


enguerran:
As I move Raven in a helper file, I am convinced @kosssi will like that.

enguerran:
For information, Sentry recommends to wrap `React.createElement` calls in a big `try {} catch` as describe here https://docs.sentry.io/clients/javascript/integrations/react/

mobile/src/lib/crash-reporter.js
```diff
@@ -0,0 +1,15 @@
+/* global __SENTRY_TOKEN__ */
+import Raven from 'raven-js'
+
+export function logException (err, context) {
+  if (!Raven.isSetup()) {
+    Raven.config(`https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`).install()
```
kosssi: it's better to have `__SENTRY_URL__` ?

mobile/src/main.jsx
```diff
@@ -38,6 +36,7 @@ document.addEventListener('DOMContentLoaded', () => {
       filesApp,
       persistedState,
       applyMiddleware(
+        RavenMiddleWare(`https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`),
```
kosssi: like other comment, it's normal you have `/2`.

kosssi: Sorry I haven't see `/2` on the first url.

mobile/src/lib/crash-reporter.js
enguerran: This were discussed in #118 . We decide to put the variable part of the sentry url in webpack, and to let the url in the code.

mobile/src/main.jsx
enguerran: `/2` is the `key` of the application as explained in #118

mobile/src/lib/crash-reporter.js
kosssi: ok, I think it's better to add parameter for project id.

mobile/src/main.jsx
kosssi: I know ;)

mobile/src/lib/crash-reporter.js
enguerran: You do approve this PR, but you suggest to add a new webpack parameter for `project ID`. What should I do?

kosssi: Like you want ;)

PR #125, opened 2017-02-22
Add HammerJS for swiping and taping on mobile

GoOz:
⚠️ **I wasn't able to test the swipe (I don't know, it seems like my fingers just don't work) so I'm not 100% sure it's all working.**

For testing, just open the file's action menu by taping on the tree dots button and then:
- Tap on the transparent background to close the menu;
- Swipe down on the menu to close it;

Although, if anyone sees some kind of optimisation to be made, just tell me.

### DIFF
--- a/src/lib/withGestures.jsx
+++ b/src/lib/withGestures.jsx
@@ -1,0 +1,46 @@
+import React, { Component } from 'react'
+import ReactDOM from 'react-dom'
+import Hammer from 'hammerjs'
+
+const shouldListenToSwipe = handlers => Object.keys(handlers).find(h => /^swipe/.test(h)) !== undefined
+const shouldListenToTap = handlers => Object.keys(handlers).find(h => /^tap/.test(h)) !== undefined
+
+const withGestures = (eventHandlers) => {
+  return WrappedComponent => {
+    return class WithGesturesComponent extends Component {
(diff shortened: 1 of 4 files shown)


src/lib/withGestures.jsx
```diff
@@ -0,0 +1,46 @@
...
+import Hammer from 'hammerjs'
+
+const shouldListenToSwipe = handlers => Object.keys(handlers).find(h => /^swipe/.test(h)) !== undefined
+const shouldListenToTap = handlers => Object.keys(handlers).find(h => /^tap/.test(h)) !== undefined
+
+const withGestures = (eventHandlers) => {
+  return WrappedComponent => {
+    return class WithGesturesComponent extends Component {
+      componentDidMount () {
+        this.hammer = new Hammer(ReactDOM.findDOMNode(this))
```
y-lohse: Swipe up and down don't work at the moment because Hammer [disables them by default](http://hammerjs.github.io/getting-started/).
You can add a line such as `this.hammer.get('swipe').set({ direction: Hammer.DIRECTION_VERTICAL })` and it will work. I also suggest we disable all recognizers except tap and swipe, since we don't use them at the moment.

y-lohse:
Just waiting on #127 and we're good to merge this.

PR #126, opened 2017-02-22
Add unregisterClient when unlink :punch:

kosssi:
- [x] Add unregisterClient

### DIFF
--- a/mobile/src/containers/Settings.jsx
+++ b/mobile/src/containers/Settings.jsx
@@ -9,15 +9,16 @@ const mapStateToProps = (state, ownProps) => ({
   version: window.navigator.appInfo ? window.navigator.appInfo.version : 'dev',
   serverUrl: state.mobile.settings.serverUrl,
   backupImages: state.mobile.settings.backupImages,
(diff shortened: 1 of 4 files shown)


mobile/src/containers/Settings.jsx
```diff
@@ -9,15 +9,16 @@ const mapStateToProps = (state, ownProps) => ({
...
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   launchBackup: () => dispatch(mediaBackup()),
   showUnlinkConfirmation: () => dispatch(showUnlinkConfirmation()),
   hideUnlinkConfirmation: () => dispatch(hideUnlinkConfirmation()),
-  unlink: () => {
-    dispatch(unlink())
+  unlink: (client) => {
+    dispatch(unlink(client))
     ownProps.router.replace('/onboarding')
```
enguerran: I don't know if replace is the correct function to use (cf. https://github.com/ReactTraining/react-router/blob/master/docs/API.md#replacepathorloc) or maybe we should replace after clearing the whole navigation history.

enguerran: It appears `replace` is the exact function to use here as it clears the history, instead of `push`.

PR #128, opened 2017-02-23
[WIP] Remove false url :shit:

kosssi:


src/components/File.jsx
```diff
@@ -69,7 +67,7 @@ class File extends Component {
       this.props.onOpen(id)
     } else {
       // TODO Handle files opening throught the app instead of doing it throught the browser
-      window.open(`${STACK_URL}/files/download/${id}`, '_blank')
+      window.open(`${cozy._url}/files/download/${id}`, '_blank')
```
m4dz: Are we exposing a `_url` property in the global `cozy` object?

enguerran: Yes we are!

And in the *mobile* app, we store that url in the global state, because we need it to init the `cozy-client-js` instance.

m4dz: Can you show me where? I can't figure how this can be defined, and more surely how it can work in both `dev` and `production` mode…

enguerran: In cozy-client-js, here it is: https://github.com/cozy/cozy-client-js/search?utf8=%E2%9C%93&q=_url

In mobile cozy-files-v3: https://github.com/cozy/cozy-files-v3/search?utf8=%E2%9C%93&q=seturl&type=Code

kosssi:
It will be resolved by https://github.com/cozy/cozy-files-v3/pull/136

PR #130, opened 2017-02-23
Refactored Folder & FileList

goldoraf:


src/containers/Folder.jsx
```diff
@@ -24,14 +37,37 @@ class Folder extends Component {
...
     if (props.isFetching === true) {
-      return <p>Loading</p>
+      return (
+        <div role='contentinfo'>
+          <p>Loading</p>
+        </div>
+      )
     }
-    return <FileList {...props} {...state} />
+    const isTrashContext = props.context === TRASH_CONTEXT
```
y-lohse: For clarity, I would suggest to also have a `isFilesContext` variable here. It makes things more readable and paves the way for future additional screens.

goldoraf: The reason I didn't is that the linter complained that the variable was unused ;)

goldoraf: Regarding your other comments, this refactoring is a WIP, there's still a lot to do in my mind to make things nicer. But let's merge this one so that we can move the card ;)

y-lohse: I think [here](https://github.com/cozy/cozy-files-v3/pull/130/files#diff-96149b386f16fb111f96b0c5a54afe8bR56) and [here](https://github.com/cozy/cozy-files-v3/pull/130/files#diff-96149b386f16fb111f96b0c5a54afe8bR61) would be good candidates (and yes, I'm more or less the one who wrote them in the first place ¯\_(ツ)_/¯

PR #131, opened 2017-02-24
Use promises instead of reducers for opening spinners

goldoraf:


enguerran:
I like this, it makes a lot *moar* sense to use *"internal state"* with a good promise management to handle this kind of transition.

Just a quick question: what is the purpose of `withState` instead of using classical `mapStateToProps` and `mapDispatchToProps` or downright using a stateful component object instead of a functional component.

y-lohse:
Same as @enguerran, I like this new approach very much, but I'm a bit confused as to what the advantage of `withState` is here.

enguerran:
@goldoraf do you have something to add to this PR and could you explain why you use `withState` to us? Maybe it can be merged.

goldoraf:
@enguerran @y-lohse the only advantage of using withState is you don't have to use a stateful component for a single toggle boolean. I like this, but it may look overkill.

src/components/File.jsx
```diff
@@ -66,14 +67,15 @@ class File extends Component {
...
+      this.setState({ opening: true })
+      this.props.onOpen(id).then(() => this.setState({ opening: false }))
     } else {
       // TODO Handle files opening throught the app instead of doing it throught the browser
       window.open(`${STACK_URL}/files/download/${id}`, '_blank')
     }
   }
 
-  render ({ t, f, attributes, showSelection, onShowActionMenu }, { editing }) {
+  render ({ t, f, attributes, showSelection, onShowActionMenu }, { editing, opening }) {
```
gregorylegarec: Second parameter here is state ? Nice to know, I'll use it in cozy-photos.

gregorylegarec:
Nice, I love this.

PR #133, opened 2017-02-27
Publish & update information by fastlane

kosssi:
This PR allow to update play store information with fastlane.

All description is not up to date @jsilvestre can update after ;)


enguerran:
It looks great. I see nothing to avoid the merge.

PR #134, opened 2017-02-28
add fr locale

ysa000:


nono:
Do we want to keep the translated locales in the repository, or only in transifex?

y-lohse:
No definitely only in Transifex. Sorry, we forgot to close this PR.

PR #136, opened 2017-03-01
Refactored folder creation and loading behaviors

goldoraf:
My goal was to avoid using the previous hackish way of adding a new folder (ie: we added a temp entry into the files reducer...) by introducing a new `AddFolder` container, thereby eliminating some UI reducers.


goldoraf:
I just added some commits that solve our loading behavior problems in a more-or-less correct way in terms of code (not perfect, but still better).

y-lohse:
I'm about to look at the actual code, but just a quick head's up: when the input loses the focus while empty, there is supposed to be a notification ;)

src/containers/TrashToolbar.jsx
```diff
@@ -10,13 +10,11 @@ import Menu, { MenuButton, Item } from 'react-bosonic/lib/Menu'
...
+const TrashToolbar = ({ t, error, isSelectionBarVisible, showSelectionBar }) => (
   <div className={styles['fil-toolbar']} role='toolbar'>
-    <div className={styles['fil-toolbar-trash']}>
-      <button className={classNames(styles['danger-outline'], styles['coz-btn--delete'])}>
-        {t('toolbar.delete_all')}
-      </button>
-    </div>
+    <button className={classNames(styles['fil-toolbar-trash'], 'coz-btn', 'coz-btn--danger-outline')}>
+      {t('toolbar.delete_all')}
+    </button>
```
y-lohse: I noticed this and the associated css changes, but the button doesn't have the correct appearance after these changes.

src/actions/index.js
```diff
@@ -136,24 +113,24 @@ export const renameFolder = (newName, id) => ({
...
-      return dispatch({
+      dispatch({
         type: CREATE_FOLDER_FAILURE_DUPLICATE,
-        id: tempId,
         alert: {
           message: 'alert.folder_name',
           messageData: { folderName: name }
         }
       })
+      throw new Error('alert.folder_name')
```
goldoraf: Throwing here allows to catch() the error instead of having a single generic then()

src/components/FileList.jsx
```diff
@@ -21,10 +22,11 @@ const FileList = ({
       <div className={classNames(styles['fil-content-header'], styles['fil-content-status'])}>{ t('table.head_status') }</div>
     </div>
     <div className={styles['fil-content-body']}>
+      <AddFolder />
       {!error && files.map((file, idx) => (
         <File
+          context={context}
           onEdit={onFileEdit}
-          onEditAbort={onFileEditAbort}
```
goldoraf: I hesitated to remove it completely: like you said, today it's impossible to edit a filename because of the click/double-click behavior on the row. I opted to not remove everything because the golbal refactoring is still in progress, and I hope to find a functional solution when our PO comes back next week.

src/components/Sidebar.jsx
```diff
@@ -2,7 +2,7 @@ import styles from '../styles/sidebar'
 
 import React from 'react'
 
-import Nav from './Nav'
+import Nav from '../containers/Nav'
 
 const Sidebar = () => (
```
goldoraf: You're right, and `Topbar` has the same problem. But I don't know yet what's the best way to refactor the sidebar and the topbar, and, again, this PR is a part of a bigger refactoring ;)

src/actions/index.js
y-lohse: But it isn't caught anywhere, is it?

src/containers/AddFolder.jsx
```diff
@@ -0,0 +1,46 @@
...
+  visible: mustShowAddFolder(state)
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  create: name => {
+    return dispatch(createFolder(name))
+  },
+  hide: () => {
+    return dispatch(abortAddFolder(name))
+  }
```
enguerran: We could simplify the syntax by writing:

```javascript
const mapDispatchToProps = dispatch => ({
  create: name => dispatch(createFolder(name)),
  hide: () => dispatch(abortAddFolder(name))
})
```

src/containers/Folder.jsx
```diff
@@ -92,7 +92,10 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     dispatch(openFolder(folderId, true))
   },
   onFolderOpen: (folderId) => {
-    return dispatch(openFolder(folderId, false, ownProps.router))
+    return dispatch(openFolder(folderId, false))
+      .then(() => {
+        ownProps.router.push(`/${ownProps.context}/${folderId}`)
+      })
```
enguerran: I think that arrow function returns the result of the first instruction.
I don't know what should return `onFolderOpen` and I don't know if this code has to be changed.

*I may be Jon Snow as I know nothing.*

src/components/File.jsx
```diff
@@ -128,4 +130,4 @@ class File extends Component {
   }
 }
 
-export default translate()(File)
+export default withRouter(translate()(File))
```
enguerran: I already ask this: but are you sure `withRouter` is mandatory to have router's functions injected in the component's props? (for `this.props.router.push`)

src/components/Loading.jsx
```diff
@@ -0,0 +1,13 @@
...
+
+const Loading = ({ message }) => {
+  return (
+    <div className={styles['fil-loading']}>
+      <p>{message}</p>
+    </div>
+  )
+}
+
+export default Loading
```
enguerran: Does it should go to [cozy-ui](https://github.com/cozy/cozy-ui) ?

enguerran: It may use a translation's key instead of a string according to the other components, doesn't it?

src/components/Topbar.jsx
```diff
@@ -1,31 +1,32 @@
...
   )
 }
 
-export default withRouter(Topbar)
+const mapStateToProps = (state, ownProps) => ({
+  context: state.context
+})
+const mapDispatchToProps = (dispatch, ownProps) => ({})
+
+export default connect(mapStateToProps, mapDispatchToProps)(Topbar)
```
enguerran: If there is no `mapDispatchToProps`, we could write:

```javascript
const mapStateToProps = state => ({
  context: state.context
})

export default connect(mapStateTopProps)(Topbar)
```

src/containers/Breadcrumb.jsx
```diff
@@ -95,4 +108,4 @@ export default translate()(withRouter(connect(
   toggle: () => {
     setState(state => ({ opening: !state.opening }))
   }
-}))(Breadcrumb))))
+}))(Breadcrumb)))
```
enguerran: I though we use `translate()` for dumb component and not for smart component.

src/containers/Folder.jsx
```diff
@@ -20,27 +21,28 @@ import DeleteConfirmation from '../containers/DeleteConfirmation'
...
   }
 
   componentWillReceiveProps (newProps) {
-    if (this.props.params.file !== undefined && // we're not in the root dir
-      newProps.params.file !== this.props.params.file && // the route has changed
-      newProps.params.file !== newProps.folderId) { // but the folder has not been fetched
-      this.props.onRouteChange(newProps.params.file)
+    // if we're already fetching, that's because of a 'openFolder' call
+    // no need to check if the URL has changed
+    if (this.props.isFetching || newProps.isFetching) return
```
enguerran: We may return something more explicit because we don't use semicolons: `return null` or `return undefined`.

src/containers/Folder.jsx
```diff
@@ -67,35 +69,24 @@ class Folder extends Component {
...
-    if (ownProps.params.file !== undefined) {
-      folderId = ownProps.params.file
-    } else {
-      folderId = ownProps.context === TRASH_CONTEXT
-        ? TRASH_DIR_ID
-        : ROOT_DIR_ID
-    }
-    dispatch(openFolder(folderId, true))
+    dispatch(openFolder(ownProps.params.folderId, ownProps.context))
   },
```
enguerran: We may write:

```javascript
onMount: () => dispatch(openFolder(ownProps.params.folderId, ownProps.context))
```

src/containers/Folder.jsx
```diff
@@ -67,35 +69,24 @@ class Folder extends Component {
...
-  onRouteChange: (folderId) => {
-    dispatch(openFolder(folderId, true))
+  onRouteChange: (folderId, context) => {
+    dispatch(openFolder(folderId, context))
   },
```
enguerran: Same as above `onRouteChange: (folderId, context) => dispatch(openFolder(folderId, context))`

src/containers/Folder.jsx
```diff
@@ -67,35 +69,24 @@ class Folder extends Component {
...
   onFolderOpen: (folderId) => {
-    return dispatch(openFolder(folderId, false))
-      .then(() => {
-        ownProps.router.push(`/${ownProps.context}/${folderId}`)
-      })
+    return dispatch(openFolder(folderId, ownProps.context))
   },
```
enguerran: And same here, and in addition arrow function returns the first instruction's result:

```javascript
onFolderOpen: folderId => dispatch(openFolder(folderId, ownProps.context))
```

src/containers/Nav.jsx
```diff
@@ -0,0 +1,87 @@
...
+    </nav>
+  )
+}
+
+const mapStateToProps = (state, ownProps) => ({
+  context: state.context
+})
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  openFiles: () => dispatch(openFolder(null, FILES_CONTEXT)),
+  openTrash: () => dispatch(openFolder(null, TRASH_CONTEXT))
```
enguerran: 💓

package.json
```diff
@@ -113,7 +113,7 @@
     "preact": "^7.1.0",
     "preact-compat": "^3.9.0",
     "raven-js": "^3.10.0",
-    "react-bosonic": "^1.0.0-beta5",
+    "react-bosonic": "beta",
```
kosssi: can you rebase your PR and remove this change ;) https://github.com/cozy/cozy-files-v3/pull/141

src/components/AppRoute.jsx
```diff
@@ -3,16 +3,12 @@ import { Route, Redirect } from 'react-router'
...
 
 const AppRoute = (
   <Route component={App}>
     <Redirect from='/' to='files' />
-    <Route path='files(/:file)' component={props => <Folder context='files' {...props} />} />
-    <Route path='recent' component={ComingSoon} />
-    <Route path='shared' component={ComingSoon} />
-    <Route path='activity' component={ComingSoon} />
-    <Route path='trash(/:file)' component={props => <Folder context='trash' {...props} />} />
+    <Route path='files(/:folderId)' component={props => <Folder context='files' {...props} />} />
```
CPatchane: Could we use another word that context here? `folderType` maybe?

src/components/Loading.jsx
CPatchane: I think maybe it's better to put the translated message as argument instead of include the translation library here in this little component.

y-lohse: It should, as it's used in all our apps now. But I also think this PR is big enough as it is, so maybe that can be done separately.

enguerran: I know, right? Sometime earlier, I asked to keep *dumb component* out of third libraries like `i18n.js`. We told me it's better to have a dumb component requesting keys instead of fully translated sentences. I think it's wrong, we may have *dumb component* requesting strings instead of i18n keys:

```jsx
const myDumbComponent = ({ title, subtitle, paragraph }) => (
  <section>
    <h1>{title}</h1>
    <h2>{subtitle}</h2>
    <p>{paragraph}</p>
  </section>
)
```

instead of a strongly coupled component:

```jsx
const myNotReallyDumbComponent = ({ t, titleKey, subtitleKey, paragraphKey }) => (
  <section>
    <h1>{t(titleKey)}</h1>
    <h2>{t(subtitleKey)}</h2>
    <p>{t(paragraphKey)}</p>
  </section>
)
```

src/actions/index.js
goldoraf: It's caught in the `FilenameInput`: https://github.com/cozy/cozy-files-v3/pull/136/files#diff-ba80c61f3bfd7509f2878d491217a8b3R54

src/actions/index.js
```diff
@@ -175,32 +151,20 @@ export const createFolder = (name, tempId) => {
...
-      folder: extractFileAttributes(folder),
-      tempId
+      folder: extractFileAttributes(folder)
     })
   }
 }
 
-export const deleteFileOrFolder = (id, isNew = false) => {
-  return async (dispatch, getState) => {
-    dispatch({ type: DELETE_FILE, id: id })
```
goldoraf: Done

package.json
goldoraf: OK

src/components/File.jsx
goldoraf: Yes

src/components/Loading.jsx
goldoraf: @enguerran like a lot of things in life, it depends: with some simple components like `Loading`, it makes totally sense to me to pass a translated string as arg. For more complex (but still dumb) components, it'll sometimes (often) be necessary to translate things inside the component. I'm against any hard rules related to this concern, let's simply use our best judgement when writing components.

src/components/AppRoute.jsx
goldoraf: I agree `context` is not ideal, but honestly renaming it now would require a bit more work than I'd be inclined to ;) Also, I find `folderType`worse.

src/containers/AddFolder.jsx
goldoraf: Yes, and I prefer this shorter syntax, but I was afraid to hide the importance of `return`, that allows us to get a promise when calling the action. What do you think?

src/containers/Folder.jsx
goldoraf: Well, I just want to exit the function here, not returning anything. I don't think it's necessary in this context to return null.

src/containers/AddFolder.jsx
enguerran: Even if nowadays, small part of developers knows that arrow function returns the result of the first instruction. I think it's gonna be wide shared and it helps developers to take the plunge.

> You must be the change you want to see in the world.
— Ghandi

src/containers/Folder.jsx
enguerran: Let's say we have some code like this:

```javascript
function noSemicolon () {
  if (true) return
  doSomething()
}
```

could become

```javascript
function noSemicolon () {
  if (true) return doSomething()
}
```

That's just my point to add `null` or `undefined` (or a semicolon).

Exactly the same idea as
> Time to eat
> children

could become 

> Time to eat, children

or

> Time to eat children

src/components/Loading.jsx
enguerran: 👍 alright

> any decent answer to an interesting question begins, "it depends..."
— Kent Beck, https://twitter.com/KentBeck/status/596007846887628801

src/components/Topbar.jsx
goldoraf: Done

src/actions/index.js
```diff
@@ -42,45 +39,36 @@ export const OPEN_FILE_E_OFFLINE = 'OPEN_FILE_E_OFFLINE'
 export const OPEN_FILE_E_NO_APP = 'OPEN_FILE_E_NO_APP'
 export const ALERT_CLOSED = 'ALERT_CLOSED'
 
+// Temporary
+const STACK_URL = 'http://cozy.local:8080'
```
y-lohse: Should we still be using this? We know the actual stack url from the inject data, and I don't think this workaround will work on the test servers?

goldoraf: Fixed!

src/components/File.jsx
```diff
@@ -1,14 +1,12 @@
...
 import classNames from 'classnames'
 import filesize from 'filesize'
+import { withRouter } from 'react-router'
 
 import styles from '../styles/table'
 import { translate } from '../lib/I18n'
 import FilenameInput from '../components/FilenameInput'
 
-// Temporary
-const STACK_URL = 'http://cozy.local:8080'
```
kosssi: 👍

src/actions/index.js
```diff
@@ -90,6 +75,19 @@ export const openFolder = (folderId = ROOT_DIR_ID, isInitialFetch = false, route
...
 
+export const openFileInNewTab = (file) => {
+  return async (dispatch, getState) => {
+    const folder = getState().folder
+    // TODO: replace this with cozy.client.getFilePath(file, folder)
+    const folderPath = folder.path.endsWith('/')
+      ? folder.path
+      : `${folder.path}/`
+    const filePath = `${folderPath}${file.name}`
+    const href = await cozy.client.files.getDowloadLink(filePath)
```
enguerran: 🎃 there is a hidden typo here, be ware or the pumpkin will get angry.

```javascript
 const href = await cozy.client.files.getDownloadLink(filePath)
```

PR #137, opened 2017-03-01
Fix/cosmetic

GoOz:
⚠️ **Not to be merge before https://github.com/cozy/cozy-ui/pull/66**

- Updated with Cozy-UI@beta13 (pending…) 
- Various rendering and cleaning fix


GoOz:
@y-lohse thanks for your thorough review.

- the cozy bar is over the selection bars --> **Yarn is supposed to load the last version of Cozy-bar which is beta9 which fix this bug… but I fixed the version to beta9 so we're clear**
- the "more" button in the selection bar is also visible on desktop --> **Good catch :)**
- on the delete confirmation button, the delete button is not aligned with the closing cross (there's some right margin) --> **Damn right**
- the alerter goes to 100% height when exiting --> **Indeed**
- successive alerts are not displayed, only the first one (probably unrelated to this PR though) --> **Yeah I think there's a queue feature missing here, so i won't fix this myself with this PR**
- the FAM is below the grey overlay --> **Finally caught the problem, thanks for pointing that out**
- on mobile, some icons in the selection menus (both files and trash) have the wrong color --> **Actually, everything should be white on desktop but on mobile only the number of selected items and the closing button should be white according to the last mockup I had, but I fixed the missing one**

Everything fixable should be fixed now.

y-lohse:
> Actually, everything should be white on desktop but on mobile only the number of selected items and the closing button should be white according to the last mockup I had, but I fixed the missing one

Oh, my bad! 

Everything looks ok now, thanks for the fixes. Feel free to merge it whenever cozy/cozy-ui#66 is ready!

enguerran:
I recomand to list needed PR in another comment thant the first one (PR's description) as it let github to add a reference to the commit with a label referencing the state of the remote PR.

Demo: https://github.com/cozy/cozy-ui/pull/66

GoOz:
@enguerran oh nice, that explains a lot. Thanks for the tips. 😃

PR #138, opened 2017-03-01
feat: manage device revocation

enguerran:
To unregister a device, you need to retrieve the `device id` then `DELETE` this device:

```
OAUTH_TOKEN=$(./cozy-stack instances token-oauth cozy.local:8080 \
$(./cozy-stack instances client-oauth cozy.local:8080 \
http://localhost curl curl) 'io.cozy.oauth.clients')

http http://cozy.local:8080/settings/clients \
--auth user:$OAUTH_TOKEN

http DELETE http://cozy.local:8080/settings/clients/17d49c2fef1f1b951635881e3bc0d1b12 \
--auth user:$OAUTH_TOKEN
```

Here is the scenario:

1. open the android app
1. go through the onboarding
1. your device is registered
1. make above commands to unregister the client
1. restart the android app
1. you should see the onboarding screen

I don't know yet how to manage revocation during app usage.

*hints:*

- `cozy-client-js` throws an error with a specific message on `4xx` HTTP error saying the device is revoked. See https://github.com/cozy/cozy-client-js/pull/96
- but when do we get this error if we use pouchDB data? On the next sync? So it has to be added in the repeated replication flow ;)

### DIFF
--- a/mobile/src/actions/settings.js
+++ b/mobile/src/actions/settings.js
@@ -10,6 +10,8 @@ export const BACKUP_IMAGES_ENABLE = 'BACKUP_IMAGES_ENABLE'
 export const ERROR = 'ERROR'
 export const SET_CLIENT = 'SET_CLIENT'
 
+import { revokeClient, unrevokeClient } from './authorization'
+
 // url
 
 export const setUrl = url => ({ type: SET_URL, url })
@@ -71,9 +73,17 @@ export const registerDevice = () => async (dispatch, getState) => {
   dispatch(checkURL(getState().mobile.settings.serverUrl))
   initClient(getState().mobile.settings.serverUrl, onRegister(dispatch), device)
   await cozy.client.authorize().then(async ({ client }) => {
+    dispatch(unrevokeClient())
     dispatch(setClient(client))
     await cozy.client.offline.replicateFromCozy('io.cozy.files')
-    cozy.client.offline.startRepeatedReplication('io.cozy.files', 15)
+    const options = {
+      onError: (err) => {
+        console.log('on error fron the client', err)
+        console.warn(`Your device is no more connected to your server: ${getState().mobile.settings.serverUrl}`)
+        dispatch(revokeClient())
+      }
+    }
+    cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
   }).catch(err => {
     dispatch(wrongAddressError())
     logException(err)
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -34,3 +34,14 @@ export const initBar = () => {
     lang: 'en'
   })
 }
+
+export const isClientRegistered = async (client) => {
+  return await cozy.client.auth.getClient(client).then(client => true).catch(() => false)
+}
+
+export function resetClient () {
+  // reset pouchDB
+  cozy.client.offline.destroyAllDatabase()
+  // reset cozy-client-js
+  cozy.client._storage.clear()
+}
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -36,7 +36,16 @@ export const initBar = () => {
 }
 
 export const isClientRegistered = async (client) => {
-  return await cozy.client.auth.getClient(client).then(client => true).catch(() => false)
+  return await cozy.client.auth.getClient(client).then(client => true).catch(err => {
+    if (err.message === 'Client has been revoked') {
+      return false
+    }
+    // this is the error sent if we are offline
+    if (err.message === 'Failed to fetch') {
+      return true
+    }
+    throw err
+  })
 }
 
 export function resetClient () {
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -40,6 +40,8 @@ export const isClientRegistered = async (client) => {
 }
 
 export function resetClient () {
+  // reset cozy-bar
+  // TODO
   // reset pouchDB
   cozy.client.offline.destroyAllDatabase()
   // reset cozy-client-js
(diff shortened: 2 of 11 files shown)


enguerran:
I change the feature: now you see a modal saying your device is revoked.

See below both scenarios:

**Login again**

videotogif_2017 03 02_16 13 26

**Log out**

videotogif_2017 03 02_16 16 01

What you don't see is that local data is erased in the logout scenario.

enguerran:
I had to rebase due to #140 by @CPatchane but it's great now!

enguerran:
@goldoraf do you think it's a good idea to wrap the whole application to manage the Modal regardless of the application state.

enguerran:
Now we register the `repeatedReplication` with an `onError` handler. If something is wrong, we display the *revoked Modal* as shown below:

videotogif_2017 03 06_13 37 44

But I need https://github.com/cozy/cozy-client-js/pull/102 to be merged.

nono:
Maybe I'm reading correctly the code, but I have the impression that you check that the revocation only at boot, not when a call to cozy-client-js fails. Am I right?

mobile/src/lib/cozy-helper.js
```diff
@@ -34,3 +34,16 @@ export const initBar = () => {
     lang: 'en'
   })
 }
+
+export const isClientRegistered = async (client) => {
+  return await cozy.client.auth.getClient(client).then(client => true).catch(() => false)
```
nono: When the device is offline, `isClientRegistered` will return false. I'd have made the opposite supposition: that the client is still registered.

mobile/src/actions/settings.js
```diff
@@ -71,9 +73,17 @@ export const registerDevice = () => async (dispatch, getState) => {
   dispatch(checkURL(getState().mobile.settings.serverUrl))
   initClient(getState().mobile.settings.serverUrl, onRegister(dispatch), device)
   await cozy.client.authorize().then(async ({ client }) => {
+    dispatch(unrevokeClient())
     dispatch(setClient(client))
     await cozy.client.offline.replicateFromCozy('io.cozy.files')
-    cozy.client.offline.startRepeatedReplication('io.cozy.files', 15)
+    const options = {
+      onError: (err) => {
```
kosssi: where do you see this parameter?

kosssi: Replication is an event emitter maybe it ok with onError : https://github.com/pouchdb/pouchdb/blob/master/docs/_includes/api/replication.html#L9

kosssi: Ok I'm not see https://github.com/cozy/cozy-client-js/pull/102. So it's ok ;)

mobile/src/lib/cozy-helper.js
enguerran: You are absolutely right, and forgot the *offline* mode.
How do you do to make the opposite supposition? I am really interested.

mobile/src/main.jsx
```diff
@@ -56,28 +58,46 @@ document.addEventListener('DOMContentLoaded', () => {
...
+        isClientRegistered(client).then(clientIsRegistered => {
+          if (clientIsRegistered) {
+            const options = {
+              onError: (err) => {
+                console.log('on error fron the client', err)
+                console.warn(`Your device is no more connected to your server: ${store.getState().mobile.settings.serverUrl}`)
+                store.dispatch(revokeClient())
+              }
+            }
+            cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
```
enguerran: @nono I added this callback (`options.onError`) to `cozy.client.offline` to be warned when the scheduled replication fails. In this PR https://github.com/cozy/cozy-client-js/pull/102 I added the requisite change to handle this case.

mobile/src/lib/cozy-helper.js
nono: In the catch, you can use the error. I think the message should be `Client has been revoked` if the client has been revoked. If it's not the case, you can check the status code. If you have a status code (probably 401), you are online. If the status code is missing, you can suppose you are offline.

enguerran: In fact, here is the failing request on `getClient`:

```
curl 'http://cozy.local:8080/auth/register/977aa3b0d9e520699f9f9650e000d64a' -H 'authorization: Bearer eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJyZWdpc3RyYXRpb24iLCJpYXQiOjE0ODg4MTE5NzIsImlzcyI6ImNvenkubG9jYWw6ODA4MCIsInN1YiI6Ijk3N2FhM2IwZDllNTIwNjk5ZjlmOTY1MGUwMDBkNjRhIn0.Mtk5qDQlDrrdbyawE9LYvn3UOOIVuJEgvh1k9EZSsHZhpcETRYxNIEBDcEYPeJm6rXx0ZOnptSoGY7LsY6yqWA' -H 'accept: application/json'

HTTP/1.1 404 Not Found
Content-Security-Policy: default-src 'self';img-src *;
Content-Type: application/json; charset=UTF-8
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
Date: Mon, 06 Mar 2017 15:15:43 GMT
Content-Length: 28
{"error":"Client not found"}
```

And the error in the catch block is:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
```

nono: I think that the argument in the catch callback is a `FetchError` and has a `status` field. Can you try this?

```js
.catch((err) => !(err.status && err.status >= 400) )
```

enguerran: I added some code in cozy-client-js (see below https://github.com/cozy/cozy-client-js/issues/105) to handle this edge case.

enguerran:
I need to change something in https://github.com/cozy/cozy-client-js/issues/105 before this PR to be merged.

Stay tuned! 🌷

enguerran:
@kosssi you get the final word. If it looks good to you, please tell me and I'll merge it.

enguerran:
Rebase + update `cozy-client-js` version. Be my guest.

PR #139, opened 2017-03-02
Mobile onboarding UI improvements

CPatchane:
This PR improve the mobile onboarding mobile to v3 design.

* Welcome screen
* Server selection screen
* Backup photos & videos screen


mobile/src/containers/onboarding/SelectServer.jsx
```diff
@@ -1,30 +1,73 @@
...
-      ownProps.nextStep()
-    })
+  goBack: () => {
+    ownProps.previousStep()
+  },
+  selectServer: (serverUrl, authorized) => {
+    if (!serverUrl) return
+    dispatch(registerDevice())
+      .then(() => {
+        if (authorized) ownProps.nextStep()
```
enguerran: This is not correct:

`authorized` is a redux state's object. It changes value through a dispatch value. But here, `authorized` is a parameter, and in this scope, its value doesn't change. So when `selectServer(serverUrl, authorized)` is called by the `onClick` handler, the value is `false` and it doesn't change even if you update the redux global state.

Anyway, there is no need for such a flag here as `registerDevice()` returns a success promise's result if we enter the `.then()`. So we are `authorized`. So I'll remove the `authorized` variable from the whole file as it is useless.

If you want to keep the `authorized` variable (to rely on the redux global state), you need to write your component in another way:

1. when the device is registered, it updates the global state
1. a component is subscribed to that state changes
1. if the flag is true the nextComponent is displayed
1. if the flag is false we stay on this component and display an error message

It is exactly how the error is handled: no action in the catch() but a subscription to the `error` object from the redux global state and a conditional render depending on it.

CPatchane: Agree with you, that probably was a previous fix. I merged the PR you did to fix that. Thanks 👍

PR #141, opened 2017-03-03
Fix gesture problems

y-lohse:
Post-mortem, for anyone curious:
When viewing the file action menu, you're supposed to be able to tap the background or do a swipe down gesture to dismiss the menu. This worked fine during our tests, but not in production, not on the iphone simulator, etc.

One of our dependencies was injecting a polyfill for PointerEvents. This caused `hammer` to rely on them to detect gestures. However, the polyfill in question requires that elements for which you want PointerEvents be marked with a special attribute, which we didn't use, and so no PointerEvents were fired, and Hammer wasn't doing anything.

We didn't see the problem in dev because we were either using browsers that natively support PointerEvents, or because the polyfill was being blocked for CSP reasons.

Anyway, tested in Firefox, Chrome and Safari iOS, so I'm pretty confident it's fixed.

I also took the opportunity to disable all the Hammer recognizers we don't use at the moment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "preact": "^7.1.0",
     "preact-compat": "^3.9.0",
     "raven-js": "^3.10.0",
-    "react-bosonic": "beta",
+    "react-bosonic": "1.0.0-beta6",
     "react-redux": "^5.0.1",
     "react-router": "^3.0.0",
     "redux": "^3.6.0",
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "preact": "^7.1.0",
     "preact-compat": "^3.9.0",
     "raven-js": "^3.10.0",
-    "react-bosonic": "^1.0.0-beta5",
+    "react-bosonic": "beta",
     "react-redux": "^5.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "preact": "^7.1.0",
     "preact-compat": "^3.9.0",
     "raven-js": "^3.10.0",
-    "react-bosonic": "^1.0.0-beta5",
+    "react-bosonic": "beta",
```
kosssi: beta is not the good version.

    "react-bosonic": "1.0.0-beta6",

if you use yarn you do this:

    yarn upgrade react-bosonic

this change package.json and yarn.lock so it's better to add the correct version ;)

GoOz: truedat. I noticed also that when you just write 'beta', on yarn install, it takes the newer local version and doesn't bother checking on NPM if there's a even newer version…

y-lohse:
If you have the time and given that this bug has been hard to reproduce reliably so far, I wouldn't mind a second look at this :-)

GoOz:
Yup.  🎉 

It's fixed

PR #142, opened 2017-03-03
Upgrade cozy-client-js :rocket:

kosssi:


mobile/src/actions/settings.js
```diff
@@ -73,6 +73,7 @@ export const registerDevice = () => async (dispatch, getState) => {
   await cozy.client.authorize().then(async ({ client }) => {
     dispatch(setClient(client))
     await cozy.client.offline.replicateFromCozy('io.cozy.files')
+    cozy.client.offline.startRepeatedReplication('io.cozy.files', 15)
```
enguerran: Do we have to `replicateFromCozy` and `startRepeatedReplication` sequencially. Don't we have to `replicateFromCozy.then(startRepeatedReplication)`?

mobile/src/main.jsx
```diff
@@ -66,6 +66,7 @@ document.addEventListener('DOMContentLoaded', () => {
         })
       } else {
         initClient(url)
+        cozy.client.offline.startRepeatedReplication('io.cozy.files', 15)
```
enguerran: If we set up the `replication` workflow in a way (cf. https://github.com/cozy/cozy-files-v3/pull/142/files#r104361067), don't we have to start it here the same way: first `replicateFromCozy` then `startRepeatedReplication`?

mobile/src/actions/settings.js
kosssi: `replicateFromCozy have an await, so the replication is finish and we launch the repeated replication.
When I remove `replicationFromCozy` this app failed (OPEN_FOLDER_FAILURE) on first connection.

mobile/src/main.jsx
kosssi: the first replication is only here to initialize `io.cozy.files.root-dir` ;)

enguerran:
Now that it is clearer, I approve and merge this PR.

kosssi:
👍

PR #147, opened 2017-03-06
Upgrade cozy-client-js :rocket:

kosssi:


kosssi:
It is necessary to https://github.com/cozy/cozy-files-v3/pull/138

PR #152, opened 2017-03-08
Fix alerts & alerts' redux stuff move to cozy-ui

goldoraf:
This PR is paired with https://github.com/cozy/cozy-ui/pull/73


goldoraf:
Yes, the tests were failing because of the non-update of cozy-ui, I was waiting its new release. Should be fixed now.

PR #154, opened 2017-03-08
[mobile] refactor main.jsx

enguerran:
I found that `main.jsx` was too heavy.

So.

- I create `MobileAppRoute.jsx` in regard to the original `AppRoute.jsx`.
- I move some code inside the `main.jsx` file

### DIFF
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -10,102 +10,88 @@ import { Provider } from 'react-redux'
 import { createStore, applyMiddleware } from 'redux'
 import thunkMiddleware from 'redux-thunk'
 import createLogger from 'redux-logger'
-import { Router, Route, hashHistory } from 'react-router'
+import { Router, hashHistory } from 'react-router'
 import RavenMiddleWare from 'redux-raven-middleware'
 
 import { I18n } from '../../src/lib/I18n'
 
 import filesApp from './reducers'
-import AppRoute from '../../src/components/AppRoute'
-import App from '../../src/components/App'
-
-import OnBoarding from './containers/OnBoarding'
-import Settings from './containers/Settings'
+import MobileAppRoute from './components/MobileAppRoute'
 
 import { loadState, saveState } from './lib/localStorage'
 import { initClient, initBar, isClientRegistered, resetClient } from './lib/cozy-helper'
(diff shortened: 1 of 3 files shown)


mobile/src/main.jsx
```diff
@@ -10,102 +10,88 @@ import { Provider } from 'react-redux'
...
-      const stateToBeSaved = {
-        mobile: {
-          settings: store.getState().mobile.settings,
-          mediaBackup: {
-            uploaded: store.getState().mobile.mediaBackup.uploaded
-          }
-        }
+  store.subscribe(() => saveState({
+    mobile: {
+      settings: console.log(store.getState().mobile.settings) || store.getState().mobile.settings,
```
kosssi: I'don't understand why you add console.log?

mobile/src/main.jsx
```diff
@@ -10,102 +10,88 @@ import { Provider } from 'react-redux'
...
-            store.dispatch(revokeClient())
-            callback()
+  function requireSetup (nextState, replace, callback) {
+    const client = store.getState().mobile.settings.client
+    const isSetup = store.getState().mobile.settings.authorized
+    if (isSetup) {
+      isClientRegistered(client).then(clientIsRegistered => {
+        if (clientIsRegistered) {
+          const options = {
+            onError: onError(store, callback)
```
kosssi: Are you sure it's good here? onError didn't have `err` parameter.

mobile/src/main.jsx
enguerran: I forget to remove it... really sorry about that. Thanks your sharped eyes!

mobile/src/main.jsx
enguerran: We could add `err` parameter in the onError HOF here https://github.com/cozy/cozy-files-v3/pull/154/files/566f31ecc6325cd4dd01b1d32d9322beb1b67a71#diff-c609607c962d635b37d6c6f38848804fR27

But with standardjs, we don't want to have an error not handled: `cozy-files-v3/mobile/src/main.jsx:27:38: Expected error to be handled.`. And because we have nothing to handle... I do not add the err parameter.

enguerran:
@kosssi Do you have other requests?

PR #155, opened 2017-03-08
[fix] Fix thead on Safari (requires updated UI & Bar)

GoOz:
That's only a part of the fix.

### DIFF
--- a/src/styles/table.styl
+++ b/src/styles/table.styl
@@ -1,23 +1,22 @@
 @import 'cozy-ui'
 
 .fil-content-table
-    position    relative
-    height      100%
-    text-align  left
-    color       grey-11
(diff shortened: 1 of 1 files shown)


GoOz:
https://github.com/cozy/cozy-ui/pull/74

m4dz:
I assume we don't hav eto do many tweaks in other apps to support cozy/cozy-ui#74 / cozy/cozy-bar#24, right?

GoOz:
@m4dz hopefuly, none. Didn't test for real but it mostly fix the layout which shouldn't impact any other app. And for the thead part, if other apps don't like it, we'll only have to separate what I just did from what existed before and make a variant class.

PR #156, opened 2017-03-08
Breadcrumb for mobile

y-lohse:
👓 Please wear your safety goggles while reviewing this PR.

It's hard to explain what's going on, but here is the gist: the topbar, which contains the breadcrumbs and the "more" button, is removed from the document flow and gets absolutely positioned *above* the cozy bar -- effectively on the same `z-index`, but because it appears later in the DOM it gets rendered above. However, the breadcrumb is moved to the right so that the hamburger menu button of the bar can still be clicked.

For the more button, that's all there is to know.

For the breadcrumb, everything except the current folder is hidden. When we're in a folder that is *not* the root folder,  2 things happen:

- A back arrow appears. It is absolutely positioned above the hamburger menu and gets a white background to hide the hamburger below (You still wearing them' goggles? Good.)
- A down arrow appears next to the folder's name, and the whole thing becomes interactive. When you click it, you get the deployed view.

When the deployed view is on, the Breadcrumb Component creates a backdrop, and the hidden sub folders are revealed. They are all set to `display block` and their order is reversed using flexbox.

Finally, when the hamburger is clicked and cozy-bar's lateral menu opens, we have a problem, because at this point cozy-bar is still rendered below the topbar. So when it's open, we bump it's z-index up a notch. This is by far the thing that bothers me the most in this PR, but I can't see any other workaround.

That's about it. There are a few hacks here and there and if someone has ideas on how to do this differently, I'm all ears.

The PR also contains a change that makes the upload button inside the "more" menu work, but that's small beans compared to the rest.

Edit: I strongly recommend we do a squash and merge this one, it has gone through a few hard rebases.


src/styles/topbar.styl
```diff
@@ -7,6 +7,7 @@
     right-spaced(32px)
     left-spaced(32px)
 
-    @media (max-width: (768/basefont)rem)
-        right-spaced(16px)
-        left-spaced(16px)
+@media (max-width: (768/basefont)rem)
+    .fil-content-header
```
GoOz: I wonder if this is necessary now as the `.fil-content-head`, parent of every `.fil-content-header`, is hidden below 768px

src/styles/toolbar.styl
```diff
@@ -34,7 +39,7 @@ b-item
     margin-top 1px
     min-width 220px
 
-    a
+    a, label
```
GoOz: Oh some rogue double quotes! :p

src/styles/toolbar.styl
```diff
@@ -7,10 +7,15 @@
     .coz-btn:last-child
         margin-right 0
 
+    .fil-toolbar-more-btn
+        @extends $button--secondary
```
GoOz: Can you use `@extend` instead of `@extends`, please? I tend to unify things through every app

src/styles/breadcrumb.styl
```diff
@@ -1,18 +1,96 @@
...
-    @media (max-width: (768/basefont)rem)
-        font-size(20px)
-        .fil-path-hidden
+.fil-path-previous
+    position   absolute
+    right      100%
+    top        0
+    display    none
+    width      $coz-bar-size
+    height     $coz-bar-size - 1px//don't cover the cozy bar bottom border
```
GoOz: Thanks for the comment 👍

y-lohse:
@GoOz : All patched up!

PR #157, opened 2017-03-08
feat: enhance offline behavior

enguerran:
When a replication is complete AND when there is some new document(s), then
we refresh the current folder's file list by dispatching an action creator.

And on startup, we don't get localfile first as in that case we have to need the end of the replication to display files.
Instead we load remote files and start a repeated replication. When the first replication is going to start, it creates a local database.

It is related to https://github.com/cozy/cozy-client-js/pull/109

I think it's better, what do **you** think?

### DIFF
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -2,6 +2,8 @@
 
 import { LocalStorage as Storage } from 'cozy-client-js'
 
+import { openFolder } from '../../../src/actions'
+
 const getStorage = () => new Storage()
 const getClientName = device => `Cozy Files Application on ${device} (${Math.random().toString(36).slice(2)})`
 
@@ -23,7 +25,6 @@ export const initClient = (url, onRegister = null, device = 'Device') => {
     console.log(`Cozy Client initializes a connection with ${url}`)
     cozy.client.init({
       cozyURL: url,
-      offline: {doctypes: ['io.cozy.files']},
       oauth: getAuth(onRegister, device)
     })
   }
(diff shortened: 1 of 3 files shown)


mobile/src/lib/cozy-helper.js
```diff
@@ -22,7 +24,7 @@ export const initClient = (url, onRegister = null, device = 'Device') => {
   console.log(`Cozy Client initializes a connection with ${url}`)
   cozy.client.init({
     cozyURL: url,
-    offline: {doctypes: ['io.cozy.files']},
+    // offline: {doctypes: ['io.cozy.files']},
```
kosssi: can you remove it?

mobile/src/lib/cozy-helper.js
```diff
@@ -62,3 +64,12 @@ export function resetClient () {
     cozy.client._storage.clear()
   }
 }
+
+export function refreshFolder (dispatch, getState) {
+  return result => {
+    if (result.docs_written !== 0) {
+      dispatch(openFolder(getState().folder.id))
+    }
+    console.log('replication complete', result)
```
kosssi: It's normal to have left the log?

enguerran: Oh no, I missed it... again! ;)

mobile/src/lib/cozy-helper.js
enguerran: Oh, yes, yet another too early push.

kosssi:
@enguerran can you rebase & upgrade cozy-client-js?

enguerran:
This branch is up-to-date, I need to double-check the behavior before a potential merge.

enguerran:
According to me, it works like a charm.

enguerran:
As soon as @kosssi validates the changes, I merge it.

PR #158, opened 2017-03-08
Fix/empty

GoOz:
- Fix Empty folder (by moving it from Folder.jsx to FileList.jsx)
- Fix spinner position
- Fix FAM button's position

### DIFF
--- a/src/components/FileList.jsx
+++ b/src/components/FileList.jsx
@@ -6,9 +6,10 @@ import { translate } from '../lib/I18n'
 
 import File from '../components/File'
 import AddFolder from '../containers/AddFolder'
+import Empty from '../components/Empty'
 
 const FileList = ({
-  t, f, context, files, error, showSelection, onFileEdit, onFolderOpen, onFileOpen, onFileToggle, onShowActionMenu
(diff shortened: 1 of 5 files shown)


GoOz:
@y-lohse I don't know if I passed the ` isTrashContext` correctly, please tell me if there's a better way to do it.

y-lohse:
@GoOz I think there is, actually :).
Right now, you're [computing](https://github.com/cozy/cozy-files-v3/blob/master/src/containers/Folder.jsx#L48) `isTrashContext` in the Folder component, only to pass it down to the Empty component.
Instead, I suggest to compute it directly in the Empty component (you can because you have the `context` prop, you just need to import [the constant](https://github.com/cozy/cozy-files-v3/blob/master/src/containers/Folder.jsx#L8). WIth this approach you don't have to pass an extra prop.

GoOz:
@y-lohse mmmh `isTrashContext` is also used for [selectionbar](https://github.com/cozy/cozy-files-v3/blob/master/src/containers/Folder.jsx#L53), so I still need it computed in the Folder.jsx file, isn't it?

PR #159, opened 2017-03-09
Refacto Folder/FileList

goldoraf:


src/components/FileList.jsx
```diff
@@ -1,43 +1,28 @@
...
-          onFileOpen={onFileOpen}
-          onToggle={onFileToggle}
-          onShowActionMenu={onShowActionMenu}
-          attributes={file}
-          showSelection={showSelection}
-        />
-      ))}
-      {!error && files.length === 0 && <Empty canUpload={!isTrashContext} />}
-    </div>
+  <div className={styles['fil-content-body']}>
```
GoOz: I'm not sure this will work, I think the Empty component should be inside `.fil-content-body`. That's the way I fixed the bug anyway.

y-lohse: I confirm, the empty element appears too low on the page :/

goldoraf: Right, should be fixed now

src/components/FileList.jsx
```diff
@@ -9,7 +7,7 @@ import AddFolder from '../containers/AddFolder'
 const FileList = ({
   t, f, context, files, error, showSelection, onFileEdit, onFolderOpen, onFileOpen, onFileToggle, onShowActionMenu
 }) => (
-  <div className={styles['fil-content-body']}>
+  <div>
```
GoOz: I know you have to wrap the render in a single element, but given the hard ass CSS architecture, I wonder if everything still works fine. I gotta test this to be sure.

GoOz: It still works fine 🎉

PR #165, opened 2017-03-09
fix: highlight the correct nav item after content's loading

enguerran:
As explained in IRC, sometimes we have a context, sometimes we don't have a context.
This PR reflects this.


enguerran:
It seems that the @goldoraf's PR (https://github.com/cozy/cozy-files-v3/pull/169) will replace this one. So I may close this one.

PR #166, opened 2017-03-09
[fix] Prevent ghostclicks after taps

y-lohse:
After a tap, a click event is *also* fired not long after -- the infamous GhostClick. See the last tip here: http://hammerjs.github.io/tips/

We could prevent the click as wildly suggested elsewhere, but that requires a bit of code and doesn't integrate very well with our current set-up.
Instead, I'm slighty delaying the hiding of the backdrop, so the ghost click happens on the backdrop and has no side effects.


gregorylegarec:
Seems ok to me. I think we should use some kind of `debounce` function to be more robust but it will be ok for now. 👍

PR #168, opened 2017-03-10
feat: add option to allow backup media only on wifi

enguerran:
We need to:
- [x] add a setting in settings (`wifiOnly`)
- [x] add the connection state in the global state
- [x] check `wifiOnly && !onWifi` to disable backup
- [x] update connection state when connection state is updated ([cordova-plugin-network-information](https://github.com/apache/cordova-plugin-network-information#sample-upload-a-file-depending-on-your-network-state-))

That is more or less what this PR aims to do.

### DIFF
--- a/mobile/config.xml
+++ b/mobile/config.xml
@@ -58,4 +58,5 @@
     <plugin name="cordova-plugin-photo-library" spec="2.0.3">
         <variable name="PHOTO_LIBRARY_USAGE_DESCRIPTION" value="This app requires access to the photo library." />
     </plugin>
+    <plugin name="cordova-plugin-network-information" spec="~1.3.2" />
 </widget>
--- a/mobile/config.xml
+++ b/mobile/config.xml
@@ -58,5 +58,5 @@
     <plugin name="cordova-plugin-photo-library" spec="2.0.3">
         <variable name="PHOTO_LIBRARY_USAGE_DESCRIPTION" value="This app requires access to the photo library." />
     </plugin>
-    <plugin name="cordova-plugin-network-information" spec="~1.3.2" />
+    <plugin name="cordova-plugin-network-information" spec="1.3.2" />
 </widget>
--- a/mobile/src/actions/mediaBackup.js
+++ b/mobile/src/actions/mediaBackup.js
@@ -1,6 +1,7 @@
 /* global cozy */
 
 import { getFilteredPhotos, getBlob } from '../lib/media'
+import { onWifi } from '../lib/network'
 import { HTTP_CODE_CONFLICT } from '../../../src/actions'
 
 export const MEDIA_UPLOAD_START = 'MEDIA_UPLOAD_START'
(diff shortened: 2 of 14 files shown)


kosssi:
Can you update config.xml to add new cordova plugin 🙃

enguerran:
@kosssi If you could approve or reject this PR (and add a comment to explain your choice), I'll be glad.

Note: there is no way within the cordova plugin to add an event if the network type change. So we need to check every time needed what is the network type. And in this code, I'll do it in `mediaBackup`:
1. check the networkType
1. dispatch the networkType
1. if backup is allowed, launch backup

enguerran:
In a further PR, we'll add the `on wifi only` option in the onboarding process.

mobile/config.xml
```diff
@@ -58,4 +58,5 @@
     <plugin name="cordova-plugin-photo-library" spec="2.0.3">
         <variable name="PHOTO_LIBRARY_USAGE_DESCRIPTION" value="This app requires access to the photo library." />
     </plugin>
+    <plugin name="cordova-plugin-network-information" spec="~1.3.2" />
```
kosssi: can you remove the ~?

mobile/src/actions/mediaBackup.js
```diff
@@ -25,24 +27,28 @@ async function getDirID (dir) {
...
-        name: photo.fileName
-      }
-      await cozy.client.files.create(blob, options).then(() => {
-        dispatch(successImageUpload(photo))
-      }).catch(err => {
-        if (err.status === HTTP_CODE_CONFLICT) {
-          dispatch(successImageUpload(photo))
+  const connectionType = getConnectionType()
+  dispatch(setConnectionState(connectionType))
+  if (backupAllowed(getConnectionType(), getState().mobile.settings.wifiOnly)) {
```
kosssi: may be you can use the constant `connectionType`?

mobile/src/lib/network.js
```diff
@@ -0,0 +1,12 @@
+export const watchNetworkState = (onConnectionChange) => {
+  document.addEventListener('offline', onConnectionChange, false)
+  document.addEventListener('online', onConnectionChange, false)
+}
+
+export const hasCordovaPlugin = window.navigator.connection !== undefined && window.Connection !== undefined
+
+export const getConnectionType = () => hasCordovaPlugin ? navigator.connection.type : 'wifi'
```
kosssi: Can you replcate 'wifi' to `Connection.WIFI`?

Why you use wifi by default?

mobile/src/lib/network.js
```diff
@@ -0,0 +1,12 @@
...
+
+export const onWifi = connection => connection === 'wifi'
```
kosssi: Same `Connection.WIFI`

enguerran: `Connection.WIFI` exists only if you have cordova plugins loaded. So, in a standalone web version (as I use in development), we don't have the constant `Connection.WIFI`.

But it may be a good think to change this string into a built-in constant.

mobile/src/lib/network.js
enguerran: If I check `hasCordovaPlugin` to get either the `navigator.connection.type` value or a constant value, it is because we don't have `Connection.WIFI` outside of cordova. And I want to build a lib that can work outside cordova environment (first for development time, but also for tomorrow when mobile navigator will have support for that kind of things... Firefox OS we miss you).

mobile/src/actions/mediaBackup.js
enguerran: Oh you are right. My bad.

enguerran:
@kosssi I add a commit to include your changes request. Please review and approve if everything seems ok to you.

PR #169, opened 2017-03-10
Fix nav links active styling

goldoraf:


goldoraf:
@y-lohse you're right, with the pending refactoring, this PR can be closed.

PR #170, opened 2017-03-10
Fix/spinner

GoOz:
Micro fixes
- Fix page spinner position
- Fix spinner height in folder list when browsing


CPatchane:
Thanks @GoOz 👍

PR #171, opened 2017-03-10
[fix] Prevent ghost clicks, attempt #2

y-lohse:
Following up on #166, a hopefully more robust fix.

### DIFF
--- a/src/lib/withGestures.jsx
+++ b/src/lib/withGestures.jsx
@@ -11,7 +11,7 @@ const withGestures = (eventHandlers) => {
   return WrappedComponent => {
     return class WithGesturesComponent extends Component {
       componentDidMount () {
-        let node = ReactDOM.findDOMNode(this);
+        const node = ReactDOM.findDOMNode(this)
         this.hammer = new Hammer.Manager(node, {
           recognizers: [[Hammer.Tap], [Hammer.Swipe, { direction: Hammer.DIRECTION_ALL }]]
         })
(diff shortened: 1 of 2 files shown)


src/lib/withGestures.jsx
```diff
@@ -5,19 +5,44 @@ import Hammer from 'hammerjs'
...
+            lastTouchEvent = e
+            // if nothing happens in the next couple of ms, we fire the tap event
+            ghostClickTimeout = setTimeout(() => {
+              this.onTap(e)
+              ghostClickTimeout = lastTouchEvent = null
+            }, GHOST_CLICK_DELAY)
+          })
+          node.addEventListener('click', e => {
+            // when a click event happens, we check to see if it's related to the last touch event we received
+            if (ghostClickTimeout !== null &&
```
gregorylegarec: Pure personal opinion, when a condition becomes pretty complex I like to store it in an explicit boolean variable like : 
```javascript
const eventAreRelated = ghostClickTimeout !== null && lastTouchEvent !==
  Math.round(lastTouchEvent.srcEvent.pageX) === Math.round(e.pageX) &&
  Math.round(lastTouchEvent.srcEvent.pageY) === Math.round(e.pageY)

if (eventAreRelated) {
  [...]
}
```
It allows us to not comment the code, and make it easier to read (I generally try to read only the left part of an affectation)

gregorylegarec:
Looks good to me 👍

PR #172, opened 2017-03-10
feat: revocation Modal has no cross

enguerran:
In https://github.com/cozy/cozy-ui/pull/78 we had a new props to control the cross in the modal.
This PR aims to use it to not display cross in our revocation modal.

screenshot at 18-29-19


enguerran:
As cozy-ui is up-to-date, why not merge this one?

PR #173, opened 2017-03-13
Fix typo 😞

kosssi:


kosssi:
Frankly it depresses this kind of return, I did it a week ago but we did not want to make the change requested, I am obliged to do it myself ... @goldoraf

PR #174, opened 2017-03-13
Mobile selection great again

GoOz:
- [x] Prevent the selection bar to hide the last item of a long list when it is placed to the bottom (mobile and tablet)
- [x] Add disabled button feature when not any item is selected.
- [x] Bonus: some indentation cleaning

### DIFF
--- a/src/styles/table.styl
+++ b/src/styles/table.styl
@@ -45,12 +45,12 @@
     color grey-11
 
 .fil-content-header
-  padding         em(8px, 12px) em(28px, 12px)
-  font-size       12px
-  text-transform  uppercase
+    padding         em(8px, 12px) em(28px, 12px)
+    font-size       12px
+    text-transform  uppercase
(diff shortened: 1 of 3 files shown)


src/styles/table.styl
```diff
@@ -104,49 +104,53 @@
     padding em(15px) em(8px)
 
 .fil-file-audio
-  background embedurl("../assets/icons/icon-type-audio.svg") 0 center no-repeat
+    background embedurl("../assets/icons/icon-type-audio.svg") 0 center no-repeat
```
y-lohse: My turn:  double quotes! Lots of double quotes!

(it makes me stupidly happy to catch that 😛 )

GoOz: HAHAHA nice catch 😛

GoOz: Fixed :)

PR #176, opened 2017-03-13
Remove inappbrowser cache :punch:

kosssi:
I remove default options:
- location=yes
- hidden=no

I add clear cache parameter.

You can see the documentation [here](https://github.com/apache/cordova-plugin-inappbrowser).

### DIFF
--- a/mobile/src/lib/registration.js
+++ b/mobile/src/lib/registration.js
@@ -18,7 +18,7 @@ export const onRegistered = (client, url) => {
   if (hasCordovaPlugin()) {
     return new Promise((resolve, reject) => {
       const target = '_blank'
-      const options = 'location=yes,hidden=no'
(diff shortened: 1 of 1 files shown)


mobile/src/lib/registration.js
```diff
@@ -18,7 +18,7 @@ export const onRegistered = (client, url) => {
   if (hasCordovaPlugin()) {
     return new Promise((resolve, reject) => {
       const target = '_blank'
-      const options = 'location=yes,hidden=no'
+      const options = 'clearcache=yes'
```
enguerran: I let this default option because I would like to change it later.
I don't believe `location=yes` is a good UI for end-user.
I do believe we can load the content before displaying the page `hidden=yes` and add a instruction to display the inapp-browser `loadstop`.

PR #179, opened 2017-03-15
Fix :bug:: After a timeout does not revoke the client

kosssi:


mobile/src/main.jsx
```diff
@@ -19,17 +19,10 @@ import filesApp from './reducers'
...
-import { initClient, initBar, isClientRegistered, resetClient, refreshFolder } from './lib/cozy-helper'
-import { revokeClient } from './actions/authorization'
+import { initClient, initBar, isClientRegistered, resetClient, refreshFolder, onError } from './lib/cozy-helper'
 
 const loggerMiddleware = createLogger()
 
-const onError = (store, callback) => () => {
-  console.warn(`Your device is no more connected to your server: ${store.getState().mobile.settings.serverUrl}`)
-  store.dispatch(revokeClient())
-  callback()
```
enguerran: I did a mistake by adding `callback()` in this function. So we should remove it from here and call it in the `react-router` function.

mobile/src/main.jsx
```diff
@@ -59,14 +52,14 @@ const renderAppWithPersistedState = persistedState => {
...
-            onError: onError(store, callback),
+            onError: onError(store.dispatch, store.getState, callback),
             onComplete: refreshFolder(store.dispatch, store.getState)
           }
           cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
           initBar()
           callback()
         } else {
-          onError(store, callback)()
+          onError(store.dispatch, store.getState, callback)()
```
enguerran: We must extract `callback` from the parameter and call it directly here (cf. https://github.com/cozy/cozy-files-v3/pull/179/files#r106179418)

mobile/src/main.jsx
```diff
@@ -59,14 +52,14 @@ const renderAppWithPersistedState = persistedState => {
       isClientRegistered(client).then(clientIsRegistered => {
         if (clientIsRegistered) {
           const options = {
-            onError: onError(store, callback),
+            onError: onError(store.dispatch, store.getState, callback),
```
enguerran: There is no need to pass `callback` here as `callback` is used by `react-router` to validate an asyncrhonous check on routes. Cf. https://github.com/cozy/cozy-files-v3/pull/179/files#r106179418

mobile/src/lib/cozy-helper.js
```diff
@@ -73,3 +75,17 @@ export function refreshFolder (dispatch, getState) {
...
+export const onError = (dispatch, getState, callback) => (err) => {
+  if (err.message === clientRevokedMsg) {
+    console.warn(`Your device is no more connected to your server: ${getState().mobile.settings.serverUrl}`)
+    dispatch(revokeClient())
+  } else if (err.message === 'ETIMEDOUT' || err.message === 'getCheckpoint rejected with ') {
+    console.log('timeout')
+  } else {
+    console.warn(err)
+  }
+  if (callback) {
```
enguerran: If you accept all the others comments, then you should remove `callback` from this function. Cf. https://github.com/cozy/cozy-files-v3/pull/179/files#r106179418

mobile/src/lib/cozy-helper.js
```diff
@@ -73,3 +75,17 @@ export function refreshFolder (dispatch, getState) {
     }
   }
 }
+
+export const onError = (dispatch, getState, callback) => (err) => {
```
enguerran: No need for `callback`. Cf. https://github.com/cozy/cozy-files-v3/pull/179/files#r106179418

PR #185, opened 2017-03-17
Swipe to dismiss the FAM

y-lohse:
The whole FileActionMenu can now be swipped, and the menu follows the gesture.

swipe

### DIFF
--- a/src/lib/withGestures.jsx
+++ b/src/lib/withGestures.jsx
@@ -3,7 +3,6 @@ import ReactDOM from 'react-dom'
 import Hammer from 'hammerjs'
 
 const shouldListenToSwipe = handlers => Object.keys(handlers).find(h => /^swipe/.test(h)) !== undefined
-const shouldListenToPan = handlers => Object.keys(handlers).find(h => /^pan/.test(h)) !== undefined
 const shouldListenToTap = handlers => Object.keys(handlers).find(h => /^tap/.test(h)) !== undefined
 
 const GHOST_CLICK_DELAY = 350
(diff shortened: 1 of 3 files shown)


src/lib/withGestures.jsx
```diff
@@ -13,12 +14,17 @@ const withGestures = (eventHandlers) => {
       componentDidMount () {
         const node = ReactDOM.findDOMNode(this)
         this.hammer = new Hammer.Manager(node, {
-          recognizers: [[Hammer.Tap], [Hammer.Swipe, { direction: Hammer.DIRECTION_ALL }]]
+          recognizers: [[Hammer.Tap], [Hammer.Swipe, { direction: Hammer.DIRECTION_ALL }], [Hammer.Pan, { direction: Hammer.DIRECTION_ALL }]]
         })
-        this.handlers = eventHandlers(this.props)
+        this.handlers = eventHandlers.call(this.wrappedComponentInstance, this.props)
```
y-lohse: @goldoraf : I'm especially curious about your opinion on this bit. When calling the function passed to `withGesture`, it's quite useful to have access to the Component you are extending, typically because you might want to work with the DOM.

I decided to make it accessible as the `this` context, because I think it makes sense that when you're writing code like [this](https://github.com/cozy/cozy-files-v3/pull/185/files#diff-04eb4bee8114ff9645da8e6f5fd59270R107), `this` would refer to the component you are enriching. But it could also be a second parameter, or maybe there are even other approaches I didn't think of.

PR #190, opened 2017-03-20
[WIP] feat: Add background service for media backup :sparkles:

kosssi:
- [x] cordova: Add background plugin :heavy_plus_sign:
- [ ] Launch background service


mobile/src/lib/background.js
```diff
@@ -28,18 +27,22 @@ const launchIosBackground = () => {
 
       logException('C\'est moi le Service Background!!!')
 
-      if (canStartBackup(store.getState)) {
+      if (store.getState().mobile.settings.backupImages && backupAllowed(getState().mobile.settings.wifiOnly)) {
```
enguerran: There is something weird in the call of `store.getState()` and the `getState()` in the two conditions.

mobile/config.xml
```diff
@@ -27,8 +27,8 @@
         <preference name="StatusBarOverlaysWebView" value="false" />
         <allow-intent href="itms:*" />
         <allow-intent href="itms-apps:*" />
-        <config-file target="*-Info.plist" parent="ITSAppUsesNonExemptEncryption" mode="add">
-            <false/>
+        <config-file mode="add" parent="ITSAppUsesNonExemptEncryption" target="*-Info.plist">
+            <false />
```
enguerran: if you say so... I hope the attributes' order will not change anymore.

mobile/src/lib/background.js
```diff
@@ -0,0 +1,59 @@
...
+  const fetchCallback = () => {
+    console.log('[js] BackgroundFetch initiated')
+
+    loadState().then(persistedState => {
+      let store = getStore(persistedState)
+      initService(store)
+
+      logException('C\'est moi le Service Background!!!')
+
+      if (store.getState().mobile.settings.backupImages && backupAllowed(getState().mobile.settings.wifiOnly)) {
```
enguerran: There is something weird here as you call `store.getState()` in one hand and keep `getState()` on the other hand.

mobile/src/lib/background.js
```diff
@@ -0,0 +1,60 @@
...
```
enguerran: There is something weird here as you call `store.getState()` in one hand and keep `getState()` on the other hand.

mobile/src/actions/settings.js
```diff
@@ -38,7 +39,10 @@ export const checkURL = url => dispatch => {
 // settings
 
 export const setAnalytics = analytics => ({ type: SET_ANALYTICS, analytics })
-export const setBackupImages = backupImages => ({type: BACKUP_IMAGES, backupImages})
+export const setBackupImages = backupImages => {
+  backupImages ? startBackgroundService() : stopBackgroundService()
+  return {type: BACKUP_IMAGES, backupImages}
```
enguerran: miss spaces between brackets `{ type: BACKUP_IMAGES, backupImages }`

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+import { initPolyglot } from '../../../src/lib/I18n'
+
+const hasIosCordovaPlugin = () => {
+  return window.cordova !== undefined && window.cordova.platformId === 'ios' && window.BackgroundFetch !== undefined
+}
+
+export const startBackgroundService = () => {
+  if (hasIosCordovaPlugin()) {
+    startIosBackgroundService()
+  }
```
enguerran: We shall add and `else` with a `console.warn` instruction to prevent developers why the code does not seem to do anything.

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export const stopBackgroundService = () => {
+  if (hasIosCordovaPlugin()) {
+    stopIosBackgroundService()
+  }
+}
+
+const startIosBackgroundService = () => {
+  let Fetcher = window.BackgroundFetch
```
enguerran: Why do you declare `Fetcher` with a First-Upper-Case?

enguerran: Why do you declare `Fetcher` with a `let` and not a `const`?

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+
+  const fetchCallback = () => {
+    console.log('[js] BackgroundFetch initiated')
+
+    loadState().then(persistedState => {
+      let store = getStore(persistedState)
```
enguerran: Why do you declare `store` with a `let` and not a `const`?

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+      initService(store)
+
+      logException('C\'est moi le Service Background!!!')
+
+      if (store.getState().mobile.settings.backupImages && backupAllowed(store.getState().mobile.settings.wifiOnly)) {
```
enguerran: If you request twice `store.getState()` don't it be reasonable to use a variable?
```js
const state = store.getState()
if (state.mobile.settings.backupImages && backupAllowed(state.mobile.settings.wifiOnly)) { 
// etc.
```

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+  let Fetcher = window.BackgroundFetch
+
+  const fetchCallback = () => {
+    console.log('[js] BackgroundFetch initiated')
+
+    loadState().then(persistedState => {
+      let store = getStore(persistedState)
+      initService(store)
+
+      logException('C\'est moi le Service Background!!!')
```
nono: I think you let some debug logging

enguerran: You forgot to remove or translate this exception. What is its purpose?

mobile/src/lib/background.js
```diff
@@ -0,0 +1,72 @@
...
+
+      if (store.getState().mobile.settings.backupImages && backupAllowed(store.getState().mobile.settings.wifiOnly)) {
+        logException('start backup on Background')
```
enguerran: Not sure it worth to be sent to sentry?

enguerran:
I cannot test the code as I do not have a iOS device.

Do you have a modus operandi to test this?

mobile/src/lib/background.js
kosssi: Thank you for your comment @nono, I forgot to write in English.

Since the background service of iOS is rather obscure, we want to leave the log to enable us to validate more easily and allow us to see the frequency of emission.

mobile/src/lib/background.js
kosssi: I have copy paste the code from the sample...

kosssi: It's a mistake.

mobile/src/lib/background.js
kosssi: It's a mistake.

mobile/src/lib/background.js
kosssi: I forgot to write in English.

Since the background service of iOS is rather obscure, we want to leave the log to enable us to validate more easily and allow us to see the frequency of emission.

mobile/src/lib/background.js
kosssi: you are right ;)

mobile/src/lib/background.js
kosssi: I will remove it

enguerran:
LGTM

kosssi:
@enguerran I updated the test process in the description

enguerran:
Nice, but I still do not have an iOS to test it out.

PR #193, opened 2017-03-21
[fix] Dismiss menu when already down

y-lohse:
Handles the case where the gesture ends when the menu is already past the bottom of the screen, and no `transitionend` event is fired.


CPatchane:
Thanks @y-lohse 👍

PR #195, opened 2017-03-22
[WIP] First refactoring pass

goldoraf:


src/components/AppRoute.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react'
 import { Route, Redirect } from 'react-router'
 
 import App from './App'
-import Folder from '../containers/Folder'
+import FileExplorer from '../containers/FileExplorer'
```
enguerran: ❤️ this new component's name

src/components/File.jsx
```diff
@@ -5,7 +5,8 @@ import { withRouter } from 'react-router'
 
 import styles from '../styles/table'
 import { translate } from '../lib/I18n'
-import FilenameInput from '../components/FilenameInput'
+
+import { getUrlFromParams } from '../reducers'
```
enguerran: Why do a reducer do in a component file?
Is it a selector?

src/components/FileList.jsx
```diff
@@ -5,13 +5,15 @@ import File from '../components/File'
 import AddFolder from '../containers/AddFolder'
 
 const FileList = ({
-  t, f, context, files, error, showSelection, onFileEdit, onFolderOpen, onFileOpen, onFileToggle, onShowActionMenu
+  t, f, virtualRoot, displayedFolder, files, selected, error, showSelection, onFileEdit, onFolderOpen, onFileOpen, onFileToggle, onShowActionMenu
```
enguerran: 👃 two many parameters: is it a smell?

src/containers/Breadcrumb.jsx
```diff
@@ -118,12 +126,12 @@ const Breadcrumb = ({ t, context, folder, opening, deployed, toggleOpening, togg
...
 const mapStateToProps = (state) => ({
-  folder: state.folder,
-  context: state.context
+  folder: state.view.displayedFolder,
+  virtualRoot: state.view.virtualRoot
 })
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  goToFolder: (parentId, context) => dispatch(openFolder(parentId, context))
+  goToFolder: (virtualRoot, parentId) => dispatch(openFolder(virtualRoot, parentId))
```
enguerran: We change from one parameters order to another. Is it really on purpose?

src/containers/FileExplorer.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { withRouter } from 'react-router'
+
+import { locationChange } from '../actions'
+import { getVirtualRootFromUrl, getUrlFromParams } from '../reducers'
```
enguerran: Is it selectors or reducers?

src/containers/Folder.jsx
```diff
@@ -23,84 +23,64 @@ import TrashSelectionBar from '../containers/TrashSelectionBar'
...
+      {!isTrashContext && showSelection && <FilesSelectionBar />}
+      {isTrashContext && showSelection && <TrashSelectionBar />}
+      {showDeleteConfirmation && <DeleteConfirmation />}
+      <div className={classNames(
+        styles['fil-content-table'],
+        { [styles['fil-content-table-selection']]: showSelection }
+      )}>
+        <FileListHeader />
+        <div className={styles['fil-content-body']}>
+          <FolderContent {...props} />
```
enguerran: :heart: very cool

src/containers/Nav.jsx
```diff
@@ -5,14 +5,15 @@ import styles from '../styles/nav'
...
 import { connect } from 'react-redux'
+import { withRouter } from 'react-router'
 
 import Spinner from '../components/Spinner'
-import { FILES_CONTEXT, TRASH_CONTEXT } from '../constants/config'
+import { ROOT_DIR_ID, TRASH_DIR_ID } from '../constants/config'
 import { openFolder } from '../actions'
 
-class ActiveLink extends Component {
+class Link extends Component {
```
enguerran: Isn't it dangerous to use a `<Link />` component that differs to the `react-router` `<Link />` component?

src/reducers/view.js
```diff
@@ -0,0 +1,126 @@
...
+  virtualRoot,
+  displayedFolder,
+  files,
+  fetchStatus,
+  lastFetch
+})
+
+const sortFiles = files => files.sort((a, b) => a.name.localeCompare(b.name))
+
+const getSortedFiles = allFiles => {
```
enguerran: Do the selectors go to another file than the reducers' file?

src/actions/index.js
```diff
@@ -1,15 +1,14 @@
...
+export const LOCATION_CHANGE = 'LOCATION_CHANGE'
 export const OPEN_FOLDER = 'OPEN_FOLDER'
 export const OPEN_FOLDER_SUCCESS = 'OPEN_FOLDER_SUCCESS'
 export const OPEN_FOLDER_FAILURE = 'OPEN_FOLDER_FAILURE'
 export const ADD_FOLDER = 'ADD_FOLDER'
 export const ABORT_ADD_FOLDER = 'ABORT_ADD_FOLDER'
-export const RENAME_FOLDER = 'RENAME_FOLDER'
 export const CREATE_FOLDER = 'CREATE_FOLDER'
 export const CREATE_FOLDER_FAILURE_GENERIC = 'CREATE_FOLDER_FAILURE_GENERIC'
 export const CREATE_FOLDER_FAILURE_DUPLICATE = 'CREATE_FOLDER_FAILURE_DUPLICATE'
```
gregorylegarec: @enguerran 
IMHO :
Action -> verb
Action creator -> I love the idea of an action being a noun. What told us redux guidelines about this point ?

goldoraf: Well, actions and action creators should be verbs, because they're actions, not events. But this causes one major problem: what about *_SUCCESS and *_FAILURE actions? We use actions like events here, and that's not ideal, even if it's commonly done in the redux world.

src/actions/index.js
```diff
@@ -46,25 +45,28 @@ const ALERT_LEVEL_ERROR = 'error'
 export const downloadFileMissing = () => ({ type: DOWNLOAD_FILE_E_MISSING, alert: { message: 'error.download_file.missing', level: ALERT_LEVEL_ERROR } })
 export const downloadFileOffline = () => ({ type: DOWNLOAD_FILE_E_OFFLINE, alert: { message: 'error.download_file.offline', level: ALERT_LEVEL_ERROR } })
 
-export const openFolder = (folderId, context = FILES_CONTEXT) => {
+export const locationChange = (virtualRoot, folderId) => {
```
goldoraf: Regarding precisely `LOCATION_CHANGE`, it's clearly ill-named, but your suggestion is worse because the location has already changed. The action's type must reflect the consequence of the location change (a view refresh in a sense), but I can't find a better name. Any suggestions people?

src/actions/index.js
enguerran: Let me be clearer:

```js
const actionType = 'ACTION_TYPE' // maybe a noun
const action = { type: actionType } // nothing special
const actionCreator = (parameter) => { // maybe a verb
  type: actionType,
  payload: parameter
}
```

src/actions/index.js
enguerran: Yeah, good remark: I dispatch something to alert redux global state to be updated. What is this *something*?
If I refer to the *flux* architecture, I can see that an action is something that dispatch an event to trigger a state change.

src/actions/index.js
```diff
@@ -46,25 +45,28 @@ const ALERT_LEVEL_ERROR = 'error'
...
-    dispatch({ type: OPEN_FOLDER, folderId, context })
+    dispatch({ type: LOCATION_CHANGE, virtualRoot, folderId })
+    return dispatch(openFolder(virtualRoot, folderId))
+  }
+}
+
+export const openFolder = (virtualRoot, folderId) => {
+  return async dispatch => {
+    if (!folderId) folderId = virtualRoot
+    dispatch({ type: OPEN_FOLDER, virtualRoot, folderId })
```
goldoraf: `virtualRoot` is the new `context`... The problem is that we need to keep the view's "type" (normal or trash) or "root" (/files or /trash) in the store... I can't find a better way to implement the specified nav behavior.

src/components/AppRoute.jsx
```diff
@@ -1,13 +1,13 @@
...
+import Layout from './Layout'
 import FileExplorer from '../containers/FileExplorer'
-import Files from '../containers/Files'
-import Trash from '../containers/Trash'
+import { Container as Files } from '../ducks/files'
+import { Container as Trash } from '../ducks/trash'
 
 const AppRoute = (
-  <Route component={App}>
+  <Route component={Layout}>
```
enguerran: We have to be careful when refactoring the front codebase as the mobile's one depends on it, cf. https://github.com/cozy/cozy-files-v3/blob/master/mobile/src/components/MobileAppRoute.jsx#L15

src/ducks/files/index.js
```diff
@@ -0,0 +1,3 @@
+import Container from './Container'
+
+export { Container }
```
enguerran: What's `ducks`?

goldoraf:
Closing this for a new one in order to reduce noise on this huge PR

PR #198, opened 2017-03-23
[fix] Handle undefined contents relation

y-lohse:
The result of `folder.relations('contents')` is not always an array, it [can be `null`, `undefined`, and more](https://github.com/cozy/cozy-client-js/blob/e6657ce1c5b472757c5e822700b3b5cbb1e2a9f5/src/jsonapi.js#L16).

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -67,7 +67,7 @@ export const openFolder = (folderId, context = FILES_CONTEXT) => {
       context,
       folder: Object.assign(extractFileAttributes(folder), {
         parent: extractFileAttributes(parent)}),
-      files: folder.relations('contents').map(
+      files: (folder.relations('contents') || []).map(
         c => extractFileAttributes(c)
       )
     })


src/actions/index.js
```diff
@@ -67,7 +67,7 @@ export const openFolder = (folderId, context = FILES_CONTEXT) => {
       context,
       folder: Object.assign(extractFileAttributes(folder), {
         parent: extractFileAttributes(parent)}),
-      files: folder.relations('contents').map(
+      files: (folder.relations('contents') || []).map(
```
enguerran: Is it the stack which defines what is `folder.relations('contents')` value? May we change the stack to send a empty array rather than `null` or `undefined`?

y-lohse: Yes, see [this](https://github.com/cozy/cozy-stack/blob/master/docs/files.md#response-1). I think `relationships.contents.data` used to be an empty array, and now it's null (when a folder has nothing in it, for example).

Not sure where this should be fixed, here or in the stack. I can live with both. Ping @aenario ?

enguerran: I think a Rest API must be consistent in the type of data it sends. If it should be an array, the array should be empty if there is no data, and a null value must declare something else like an error or an inconsistent data.

enguerran:
Any update on this? @aenario @y-lohse

aenario:
From what I have seen in the stack, I see no reason we ever send a null or undefined there. It might be worth investigating when it happens. 

If we cant find it, I'd say cozy-client-js is a better place for this kind of defensive programming.

enguerran:
@y-lohse I did not see that error since a while. I am quite agree with @aenario . Do you mind we close this PR?

y-lohse:
Nope, no problem at all. Let's see if it pops up again.

PR #203, opened 2017-03-28
feat: send usage information to sentry

enguerran:
1. [x] when user change *backup images* settings
1. [x] when user opens the app once a day

some refactors

#### Checklist

- ~[ ] Faithful integration of the mockups at all screen sizes~
- ~[ ] Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)~
- ~[ ] Localized in english and french~
- ~[ ] All changes have test coverage~
- ~[ ] Updated README, if necessary~


enguerran:
@goldoraf I added you as a reviewer because I would like you to see the redux code in `./mobile/src` to anticipate for a potential refactorization. So do the tests in `./mobile/src/tests`.

mobile/src/lib/background.js
```diff
@@ -32,7 +32,8 @@ const startIosBackgroundService = () => {
     console.log('BackgroundFetch initiated')
 
     loadState().then(persistedState => {
-      const store = getStore(persistedState)
+      initStore(persistedState)
```
kosssi: why initStore can't return the store directly?

enguerran: Because of the naming. But since this commit, I saw a pattern `configureStore`, and I can admit that `const store = configureStore()` is a way better than `const store = initStore()`!

mobile/src/lib/reporter.js
```diff
@@ -7,7 +7,8 @@ export const ANALYTICS_URL = `https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`
 
 export function getAnalyticsConfiguration () {
   return {
-    shouldSendCallback: () => getState().mobile.settings.analytics
+    shouldSendCallback: () => getState().mobile.settings.analytics,
+    environment: __DEVMODE__ ? 'development' : 'production'
```
kosssi: Why in Sentry you are 2 keys? (development and production) https://sentry.cozycloud.cc/sentry/cozy-files-v3/settings/keys/

Is the same?

mobile/src/lib/reporter.js
```diff
@@ -27,3 +28,21 @@ export function logException (err, context) {
   console.info(context)
   console.groupEnd()
 }
+
+export function logInfo (message, context) {
```
kosssi: context is not used.

mobile/src/lib/reporter.js
enguerran: I forgot to explain: within the same project, you can create multiple DSN (access keys). I though it permits to discriminate sentry's message, but it doesn't. So I can remove one of the DSN as it is not useful, nor used.

mobile/src/lib/reporter.js
enguerran: No, you are right.

mobile/src/main.jsx
```diff
@@ -50,6 +52,10 @@ const renderAppWithPersistedState = persistedState => {
     }
   }
 
+  document.addEventListener('resume', () => {
```
kosssi: what is resume event?

enguerran: It comes from cordova ⇒ https://cordova.apache.org/docs/en/latest/cordova/events/events.html#resume

kosssi: Are you sure that the event is called when launching the application?

mobile/src/lib/reporter.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+export function configureReporter (value) {
+  getState = value
+}
+
+export function logException (err) {
+  if (!Raven.isSetup()) {
+    Raven.config(ANALYTICS_URL, getAnalyticsConfiguration()).install()
```
kosssi: maybe you can refactor this :p

PR #207, opened 2017-03-30
fix: revoke not handled with replication :beetle:

enguerran:
The message between the automatic pouchDB replication
was not handled well by the application.

### DIFF
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -83,10 +83,10 @@ export function refreshFolder (dispatch, getState) {
 }
 
 export const onError = (dispatch, getState) => (err) => {
-  if (err.message === clientRevokedMsg) {
+  if (err.message === clientRevokedMsg || err.error === 'code=400, message=Invalid JWT token') {
     console.warn(`Your device is no more connected to your server: ${getState().mobile.settings.serverUrl}`)
     dispatch(revokeClient())
-  } else if (err.message === 'ETIMEDOUT' || err.message === 'getCheckpoint rejected with ') {
+  } else if (err.message === 'ETIMEDOUT') {
     console.log('timeout')
   } else {
     console.warn(err)
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -37,7 +37,7 @@ const renderAppWithPersistedState = persistedState => {
           cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
           initBar()
         } else {
-          onError(store.dispatch, store.getState)()
+          onError(store.dispatch, store.getState)({ message: 'Client has been revoked' })
         }
         callback()
       })


mobile/src/main.jsx
```diff
@@ -37,7 +37,7 @@ const renderAppWithPersistedState = persistedState => {
           cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
           initBar()
         } else {
-          onError(store.dispatch, store.getState)()
+          onError(store.dispatch, store.getState)({ message: 'Client has been revoked' })
```
kosssi: maybe you can use https://github.com/cozy/cozy-files-v3/pull/207/files#diff-f96699990a60f265f8d3b7bd25d3e9a6R8 ?

mobile/src/lib/cozy-helper.js
```diff
@@ -83,10 +83,10 @@ export function refreshFolder (dispatch, getState) {
 }
 
 export const onError = (dispatch, getState) => (err) => {
-  if (err.message === clientRevokedMsg) {
+  if (err.message === clientRevokedMsg || err.error === 'code=400, message=Invalid JWT token') {
```
kosssi: it's not the same here https://github.com/cozy/cozy-files-v3/pull/207/files#diff-f96699990a60f265f8d3b7bd25d3e9a6R51 ?

mobile/src/lib/cozy-helper.js
```diff
@@ -83,10 +83,10 @@ export function refreshFolder (dispatch, getState) {
...
     console.warn(`Your device is no more connected to your server: ${getState().mobile.settings.serverUrl}`)
     dispatch(revokeClient())
-  } else if (err.message === 'ETIMEDOUT' || err.message === 'getCheckpoint rejected with ') {
+  } else if (err.message === 'ETIMEDOUT') {
```
kosssi: why you remove this?

enguerran: It appears that pouchDB replication encapsulates network error in an error with message `getCheckpoint rejected with`. For instance when your token is not correct, cozy-stack responds with a `Invalid JWT Token` translated by cozy-client-js to `Client has been revoked`. On the other hand, the pouchDB replication encapsulates the `Invalid JWT Token` in an error with `getCheckpoint rejected with` as the message.
So it is not very accurate as it catches others errors than `timeout` errors and I prefer to remove this and use `err.error` to see what's the real error message is.

mobile/src/lib/cozy-helper.js
enguerran: PouchDB encapsulates network error message in another error with a different generic message. The real error is available in `err.error`.

mobile/src/main.jsx
enguerran: I write that code this way as it brings to light the need of a refactoring. It is a big smell and I would like to fix the bug, not to refactor the whole error management in pouchDB replication.

This is not an easy move as it involves a lot of different actors:
- the different cozy-stack's error ;
- the cozy-client-js layer ;
- the pouchDB internal layer ;
- the cozy-files-v3 handling those all together.

It worth to be refactored once and for all now we have all the actors in the place, imho.

PR #210, opened 2017-04-03
refactor: const is better :hammer:

enguerran:
A tiny refactoring pass to let `let` be `const` when variable are not redefined.
All changes do not have test coverage... We need to cover code as soon as possible.

#### Checklist

- [ ] ~Faithful integration of the mockups at all screen sizes~
- [ ] ~Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)~
- [ ] ~Localized in english and french~
- [ ] All changes have test coverage
- [ ] ~Updated README, if necessary~

### DIFF
--- a/mobile/src/actions/mediaBackup.js
+++ b/mobile/src/actions/mediaBackup.js
@@ -45,7 +45,7 @@ export const mediaBackup = (dir) => async (dispatch, getState) => {
     const dirID = await getDirID(dir)
     const totalUpload = photosToUpload.length
     let uploadCounter = 0
-    for (let photo of photosToUpload) {
+    for (const photo of photosToUpload) {
       if (backupAllowed(getState().mobile.settings.wifiOnly)) {
         dispatch(currentUploading(photo, uploadCounter++, totalUpload))
         const blob = await getBlob(photo)
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -95,7 +95,7 @@ export const uploadFile = (file, folder) => {
 }
 
 export const abortAddFolder = (accidental) => {
-  let action = {
+  const action = {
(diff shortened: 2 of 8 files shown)


mobile/src/actions/mediaBackup.js
```diff
@@ -45,7 +45,7 @@ export const mediaBackup = (dir) => async (dispatch, getState) => {
     const dirID = await getDirID(dir)
     const totalUpload = photosToUpload.length
     let uploadCounter = 0
-    for (let photo of photosToUpload) {
+    for (const photo of photosToUpload) {
```
kosssi: are you sure here?

src/actions/index.js
```diff
@@ -127,12 +127,15 @@ export const createFolder = name => {
...
     })
 
-    let folder
     try {
-      folder = await cozy.client.files.createDirectory({
+      const folder = await cozy.client.files.createDirectory({
         name: name,
         dirID: getState().view.displayedFolder.id
       })
+      dispatch({
```
kosssi: why you add this dispatch on this commit?

mobile/src/actions/mediaBackup.js
enguerran: If you don't reassign the variable inside the loop, you can use `const` for the iterable variable. Cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Examples

src/actions/index.js
enguerran: It is not an addition, it is a move. As a variable was declared outside the `try...catch` block because this `dispatch` was outside the `try...catch` block for no reason.

PR #214, opened 2017-04-04
feat: Add online view after subscribe 🏃

kosssi:
Now after an user is registered, he can browse directly in its folders

This feature adds 2 flags:
- offline : on browser is false, on mobile is true after client is defined
- firstReplication : after client is registered, a replication is launched. At the end of this replication `firstReplication` is set to true. Now we can display offline folder.

### DIFF
--- a/mobile/src/actions/settings.js
+++ b/mobile/src/actions/settings.js
@@ -1,6 +1,6 @@
 /* global cozy, __ALLOW_HTTP__ */
 
-import { initClient, startReplication } from '../lib/cozy-helper'
+import { initClient, startReplication as startPouchReplication } from '../lib/cozy-helper'
 import { setClient, setFirstReplication } from '../../../src/actions/settings'
 import { openFolder } from '../../../src/actions'
 import { onRegistered } from '../lib/registration'
@@ -87,15 +87,19 @@ export const registerDevice = () => async (dispatch, getState) => {
   initClient(getState().mobile.settings.serverUrl, onRegister(dispatch), device)
   await cozy.client.authorize().then(({ client }) => {
     dispatch(setClient(client))
-
-    const firstReplication = getState().settings.firstReplication
-    const refreshFolder = () => { dispatch(openFolder(getState().folder.id)) }
-    const revokeClient = () => { dispatch(reduxRevokeClient()) }
-    const firstReplicationFinished = () => { dispatch(setFirstReplication(true)) }
-    startReplication(firstReplication, firstReplicationFinished, refreshFolder, revokeClient, dispatch, getState)
+    startReplication(dispatch, getState)
   }).catch(err => {
     dispatch(wrongAddressError())
     logException(err)
     throw err
   })
+}
+
+export const startReplication = (dispatch, getState) => {
+  const firstReplication = getState().settings.firstReplication
+  const refreshFolder = () => { dispatch(openFolder(getState().folder.id)) }
+  const revokeClient = () => { dispatch(reduxRevokeClient()) }
+  const firstReplicationFinished = () => { dispatch(setFirstReplication(true)) }
+
+  startPouchReplication(firstReplication, firstReplicationFinished, refreshFolder, revokeClient)
 }
--- a/mobile/src/actions/settings.js
+++ b/mobile/src/actions/settings.js
@@ -1,6 +1,6 @@
 /* global cozy, __ALLOW_HTTP__ */
 
-import { initClient, startFirstReplication } from '../lib/cozy-helper'
+import { initClient, startReplication } from '../lib/cozy-helper'
 import { setClient } from '../../../src/actions/settings'
 import { onRegistered } from '../lib/registration'
 import { logException, logInfo } from '../lib/reporter'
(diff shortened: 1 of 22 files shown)


mobile/src/actions/settings.js
```diff
@@ -80,14 +81,11 @@ export const registerDevice = () => async (dispatch, getState) => {
...
-  await cozy.client.authorize().then(async ({ client }) => {
+  await cozy.client.authorize().then(({ client }) => {
     dispatch(unrevokeClient())
     dispatch(setClient(client))
-    const options = {
-      onError: onError(dispatch, getState),
-      onComplete: refreshFolder(dispatch, getState)
-    }
-    cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
+    dispatch(setOffline(true))
```
goldoraf: Dispatching a bunch of actions sequentially for a given action creator is an antipattern: these dispatches look a lot like setter methods. It'd be better to dispatch a single action with a richer payload that will modify the state in multiple reducers.

mobile/src/lib/cozy-helper.js
```diff
@@ -82,6 +82,27 @@ export function refreshFolder (dispatch, getState) {
...
+    onComplete: refreshFolder(dispatch, getState)
+  }
+  cozy.client.offline.startRepeatedReplication('io.cozy.files', 15, options)
+}
+
+export const startFirstReplication = (dispatch, getState) => {
+  const options = {
+    onError: onError(dispatch, getState),
+    onComplete: () => {
+      dispatch(setFirstReplication(true))
```
enguerran: I don't know how, but the lib should not depend on redux.
Maybe you can pass `onError` and `onCompleete` as parameters instead of redux variable.

mobile/src/lib/cozy-helper.js
```diff
@@ -82,6 +82,27 @@ export function refreshFolder (dispatch, getState) {
...
+      startRepeatedReplication(dispatch, getState)
+    }
+  }
+  cozy.client.offline.replicateFromCozy('io.cozy.files', options).then(() => {
+    console.log('End of Replication')
```
enguerran: You forget a `console.log` here.

mobile/src/main.jsx
```diff
@@ -17,24 +15,26 @@ import { loadState } from './lib/localStorage'
...
+    const client = state.settings.client
+    const isSetup = state.mobile.settings.authorized
+    const isFirstReplicationFinished = state.settings.firstReplication
     if (isSetup) {
       isClientRegistered(client).then(clientIsRegistered => {
         if (clientIsRegistered) {
-          const options = {
-            onError: onError(store.dispatch, store.getState),
-            onComplete: refreshFolder(store.dispatch, store.getState)
+          if (isFirstReplicationFinished) {
```
enguerran: If we read this instruction out loud, we may say *"if first replication is finished, then"*.
It is just a matter of wording, but I'd prefer a `firstReplicationIsFinished` name.

enguerran: Maybe it won't be this scope's responsibility to know if it starts *the first* replication or *a repeated* replication.
Maybe it may delegate that decision to the lib with a `function startReplication (onError, onComplete)` and implements the decision to be taken in the lib file.

mobile/src/reducers/settings.js
```diff
@@ -22,7 +22,7 @@ export const settings = (state = initialState, action) => {
     case INIT_STATE:
       return initialState
     case SET_CLIENT:
-      return { ...state, client: action.client, authorized: true }
+      return { ...state, authorized: true }
```
enguerran: This seems to be a refactoring's debt as `SET_CLIENT` is not accurate anymore for this action's reducer.

src/reducers/settings.js
```diff
@@ -0,0 +1,23 @@
...
+export const settings = (state = initialState, action) => {
+  switch (action.type) {
+    case SET_CLIENT:
+      return { ...state, offline: true, client: action.client }
+    case SET_OFFLINE:
+      return { ...state, offline: action.offline }
+    case SET_FIRST_REPLICATION:
+      return { ...state, firstReplication: action.firstReplication }
+    case INIT_STATE:
+      return initialState
```
enguerran: I'm not a big fan of `INIT_STATE` action.
The reducer will take a default value from `initialState` if the state is not yet defined (`undefined` or `null`).

If we want to load a full (persisted) state, we must call a `SET_STATE` action.

But I don't know how we need to call `INIT_STATE` rather than real state changing's actions like `DO_SOMETHING`, `CHANGE_THIS`, `UPDATE_VALUE`, ...

It seems like a smell to me.

mobile/src/reducers/settings.js
kosssi: I'm not sure because when I remove it, all application freeze... I can't debug. So I prefer no touch it ;)

kosssi:
@enguerran if you have anothers comments do not hesitate. I create another PR ;)

mobile/src/lib/cozy-helper.js
kosssi: oups I don't remove this log after my refactoring. But It highlights a bug [I specified on Trello](https://trello.com/c/DDUppsPT/222-requiresetup-est-appele-plusieurs-fois).

mobile/src/reducers/settings.js
enguerran: I did not say to remove it. I said that the name was not reflecting the behavior anymore... Now it's in the code base, good luck for the 6 months old ourselves if they want to understand what does the code do.

PR #215, opened 2017-04-05
Kill two :bird: with one stone

enguerran:
- logInfo is launched when user change the backup image settings.
Now we know if he set the value in the onboarding or later in
the settings page.
- pingOnceADay was not triggered on fist launch as we initialized
the timestamp state with the current date before checking if
the timestamp is today or not.

### DIFF
--- a/mobile/src/actions/timestamp.js
+++ b/mobile/src/actions/timestamp.js
@@ -2,15 +2,13 @@ import { pingOnceADay as pingOnceADayReporter } from '../lib/reporter'
 
 export const SET_TIMESTAMP = 'SET_TIMESTAMP'
 
-export const pingOnceADay = (timestamp = new Date().getTime()) => (dispatch, getState) => {
-  if (getState().mobile.settings.analytics) {
-    const previousDate = new Date(timestamp)
-    if (!isToday(previousDate)) {
+export const pingOnceADay = (timestamp, analytics = false) => dispatch => {
+  if (analytics) {
+    if (!timestamp || !isToday(new Date(timestamp))) {
       pingOnceADayReporter()
       return dispatch({ type: SET_TIMESTAMP, timestamp: new Date().getTime() })
     }
   }
-  return dispatch({ type: SET_TIMESTAMP, timestamp })
 }
 
 function isToday (date) {
--- a/mobile/src/main.jsx
+++ b/mobile/src/main.jsx
@@ -51,12 +51,21 @@ const renderAppWithPersistedState = persistedState => {
     }
   }
 
+  function pingOnceADayWithState () {
+    const state = store.getState()
+    if (state.mobile) {
+      const timestamp = state.mobile.timestamp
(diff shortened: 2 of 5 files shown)


mobile/src/main.jsx
```diff
@@ -51,12 +51,20 @@ const renderAppWithPersistedState = persistedState => {
     }
   }
 
+  function pingOnceADayWithState () {
```
kosssi: This feature should not be in main.jsx

mobile/src/actions/timestamp.js
```diff
@@ -2,15 +2,14 @@ import { pingOnceADay as pingOnceADayReporter } from '../lib/reporter'
 
 export const SET_TIMESTAMP = 'SET_TIMESTAMP'
 
-export const pingOnceADay = (timestamp = new Date().getTime()) => (dispatch, getState) => {
-  if (getState().mobile.settings.analytics) {
-    const previousDate = new Date(timestamp)
-    if (!isToday(previousDate)) {
+export const pingOnceADay = (timestamp, analytics = false) => dispatch => {
+  console.log('DISPATCH THE TIMESTAMP')
```
kosssi: This log is necessary?

enguerran: rah, once again... Crap!

mobile/src/main.jsx
enguerran: why not? Where should it be?

mobile/src/main.jsx
```diff
@@ -51,12 +51,20 @@ const renderAppWithPersistedState = persistedState => {
...
+    if (store.getState().mobile) {
+      const timestamp = store.getState().mobile.timestamp
```
kosssi: can you refactor you call 3 times 'store.getState().mobile'?

mobile/src/main.jsx
```diff
@@ -51,12 +51,21 @@ const renderAppWithPersistedState = persistedState => {
     }
   }
 
+  function pingOnceADayWithState () {
+    const state = store.getState()
+    if (state.mobile) {
```
kosssi: when state.mobile is not true?

mobile/src/main.jsx
kosssi: 👀 it's the morning ^^ I did not see it's a refactoring of 'resume' & 'deviceready' event.

mobile/src/main.jsx
enguerran: I don't know, but I know that if `state.mobile` is `undefined`, then `state.mobile.timestamp` will break the code execution. And I don't want that. 

I can't remember when I saw that, but there is a standard proposal to accept `state.mobile?.timestamp` as a shortcut for `state.mobile ? state.mobile.timestamp : undefined`.

mobile/src/main.jsx
enguerran: cool ;)

PR #216, opened 2017-04-05
Rename Cozy Files to Cozy Drive

kosssi:
On this repository:
- [x] Rename cozy-files-v3 to cozy-drive
- [x] Rename Cozy Files to Cozy Drive
- [x] Update mobile id
- [x] Update translation on project

Update service:
- [ ] Update android app id
- [ ] Update iOS app id
- [ ] Rename Github repository
- [ ] Rename Transifex project
- [ ] Update translation on Transifex
- [ ] Update Travis

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -1,9 +1,9 @@
 {
   "name": "Files",
-  "slug": "cozy-files-v3",
+  "slug": "cozy-drive",
   "icon": "app-icon.svg",
   "description": "File manager for Cozy v3",
-  "source": "https://github.com/cozy/cozy-files-v3.git@build",
+  "source": "https://github.com/cozy/cozy-drive.git@build",
   "developer": {
     "name": "Cozy",
     "url": "https://cozy.io"
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -1,5 +1,5 @@
 {
-  "name": "Cozy Drive",
+  "name": "Drive",
   "slug": "cozy-drive",
   "icon": "app-icon.svg",
   "description": "File manager for Cozy v3",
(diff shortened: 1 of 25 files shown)


manifest.webapp
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Files",
+  "name": "Cozy Drive",
```
enguerran: I would do s/Files/Drive/ as this manifest is used by the cozy-stack (in which the word Cozy is obvious)

src/locales/en.json
```diff
@@ -1,14 +1,14 @@
 {
   "nav": {
-    "item_files": "Files",
+    "item_files": "Drive",
```
enguerran: I would keep *Files* in that context.
But if you change the value, consider changing the key as well.

src/locales/en.json
```diff
@@ -1,14 +1,14 @@
...
     "item_recent": "Recent",
     "item_shared": "Shared by me",
     "item_activity": "Activity",
     "item_trash": "Trash",
     "item_settings": "Settings"
   },
   "breadcrumb": {
-    "title_files": "Files",
+    "title_files": "Drive",
```
enguerran: I would keep *Files* in that context.
But if you change the value, consider changing the key as well.
And to depend less on the translation, maybe `breadcrumb.title` is explicit enough.

jsilvestre: We chose to go for "Drive" here to match what is done by Google, but @pierrevdk will have the last word anyway

src/locales/en.json
jsilvestre: Same comment than for the breadcrumb, @pierrevdk will decide what they want, but as for now we copy what Google does (= using Drive everywhere).

PR #220, opened 2017-04-10
fix: handle setup :fries:

enguerran:
[react-router](https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#onenternextstate-replace-callback) let us register a function called each time a route is going to enter.

That is the way we drive user to the onboarding or to the application main page.

This is not a way to *initialize* the application.

So I moved the initialization into a function called once the JS file is loaded and evaluated.

And I renamed the `requireSetup` function with a more explicit name.

I would like to initialized the [cozy-bar](https://github.com/cozy/cozy-bar) at two places:
- when the application is opened in classical condition (user already onboarded)
- when the onboarding is finished

Because we don't want to initialized the cozy-bar each time the *normal route* is called (`#/files`).

*Note: I fixed a bug on initialization when user is not registered (because the device was revoked) [introduced earlier](https://github.com/cozy/cozy-drive/pull/214/commits/63a5670815f298ad076c47fad56541e8541a2eef#diff-a8d74123e70b3c36093d784c4d702199R26)*

### DIFF
--- a/mobile/src/components/MobileAppRoute.jsx
+++ b/mobile/src/components/MobileAppRoute.jsx
@@ -4,6 +4,8 @@ import { Route } from 'react-router'
 import AppRoute from '../../../src/components/AppRoute'
 import App from '../../../src/components/App'
 
+import { initBar } from '../lib/cozy-helper'
+
 import OnBoarding from '../containers/OnBoarding'
 import Settings from '../containers/Settings'
 import RevokableWrapper from '../containers/RevokableWrapper'
@@ -16,7 +18,7 @@ const MobileAppRoute = requireSetup => (
         <Route path='settings' name='mobile.settings' component={Settings} />}
       </Route>
     </Route>
-    <Route path='onboarding' component={OnBoarding} />
+    <Route path='onboarding' component={OnBoarding} onLeave={() => initBar()} />
   </Route>
 )
 
--- a/mobile/src/lib/init.js
+++ b/mobile/src/lib/init.js
@@ -1,7 +1,23 @@
 import { configureReporter } from './reporter'
-import { initClient } from './cozy-helper'
+import { initClient, initBar, isClientRegistered } from './cozy-helper'
+import { revokeClient } from '../actions/authorization'
+import { startReplication } from '../actions/settings'
 
-export const initService = (store) => {
+export const initServices = (store) => {
   configureReporter(store.getState)
   initClient(store.getState().mobile.settings.serverUrl)
+
+  const client = store.getState().settings.client
+  if (client) {
+    isClientRegistered(client)
+      .then((clientIsRegistered) => {
+        if (clientIsRegistered) {
+          startReplication(store.dispatch, store.getState) // don't like to pass `store.dispatch` and `store.getState` as parameters, big coupling
+          initBar()
+        } else {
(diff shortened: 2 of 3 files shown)


mobile/src/components/MobileAppRoute.jsx
```diff
@@ -16,7 +18,7 @@ const MobileAppRoute = requireSetup => (
         <Route path='settings' name='mobile.settings' component={Settings} />}
       </Route>
     </Route>
-    <Route path='onboarding' component={OnBoarding} />
+    <Route path='onboarding' component={OnBoarding} onLeave={() => initBar()} />
```
enguerran: @kosssi or @goldoraf I am not sure it is a good way to init the bar after the onboarding steps. What do you think?

mobile/src/lib/init.js
```diff
@@ -1,7 +1,23 @@
...
+export const initServices = (store) => {
   configureReporter(store.getState)
   initClient(store.getState().mobile.settings.serverUrl)
+
+  const client = store.getState().settings.client
+  if (client) {
+    isClientRegistered(client)
+      .then((clientIsRegistered) => {
+        if (clientIsRegistered) {
+          startReplication(store.dispatch, store.getState) // don't like to pass `store.dispatch` and `store.getState` as parameters, big coupling
```
enguerran: What do you think about this comment? Do you think it's a thing or may I remove it?

mobile/src/lib/init.js
```diff
@@ -1,7 +1,23 @@
...
+          initBar()
+        } else {
+          console.warn('Your device is no more connected to your server')
```
enguerran: Do we really want this kind of `console` message in the code? I can't know.

mobile/src/components/MobileAppRoute.jsx
kosssi: it's a good idea.

mobile/src/lib/init.js
kosssi: I thinks it's normal can you remove it?

mobile/src/lib/init.js
kosssi: I will leave the message to be able to debug

mobile/src/lib/init.js
enguerran: `startReplication` is related to pouchDB, and `store` is related to redux.

How does `startReplication` become related to redux? What are the architectural needs to link redux to pouchDB? How can they be related?

What do you mean by:
> I think it's normal

mobile/src/lib/init.js
enguerran: If it's a debug message, don't you think the message content is out of scope: what is *"Your device"* when you're a developer? May it contain something more introspective?

mobile/src/components/MobileAppRoute.jsx
enguerran: thanks for your answer, wait for another advice and I'll merge this PR if it's ok.

mobile/src/lib/init.js
kosssi: Maybe, but will be changed when needed ;)

enguerran:
I don't wait any longer for other reviews

PR #222, opened 2017-04-11
feat: breadcrumb animation :moneybag:

enguerran:
We have two behaviors within the breadcrumb:
1. on breadcrumb deploy: animate the dropdown menu
1. on back button: no animation

*Animation includes a deployed menu, an overlay and a collapsed menu.*


enguerran:
I don't wait longer for other reviews.

PR #224, opened 2017-04-12
fix: unable to start on plane mode

enguerran:
Even if it looked a good idea to remove https://github.com/cozy/cozy-drive/pull/157/files#diff-f96699990a60f265f8d3b7bd25d3e9a6L26 earlier.

It is mandatory while https://github.com/cozy/cozy-client-js/blob/master/src/index.js#L177 is not refactored.

Without that, the mobile application does not start when the phone is on plane mode.


kosssi:
Sorry but you were right 😘

kosssi:
[(2) [FIL-33] Faire en sorte que l'initialisation de cozy client js soit offline-friendly](https://trello.com/c/RWOFSwgQ/231-2-fil-33-faire-en-sorte-que-l-initialisation-de-cozy-client-js-soit-offline-friendly)

PR #225, opened 2017-04-13
[feat] add developer tools

ptbrowne:
Enable React Developer tools and Redux developer tools in Chrome.

[Redux dev tools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd)

[React dev tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi)

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -14,7 +14,8 @@ module.exports = {
   plugins: [
     new webpack.DefinePlugin({
       __SERVER__: JSON.stringify('http://app.cozy.local'),
-      __STACK_ASSETS__: false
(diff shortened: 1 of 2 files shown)


goldoraf:
Wow, thanks for this :+1: ! There is just a small fix to do if you don't mind ; currently the build is failing because of linting: __DEVELOPMENT__ is not explicitely defined or imported in main.jsx, and the linter doesn't like that. Adding `/* global __DEVELOPMENT__ */` to the header of main.jsx will fix that problem.

goldoraf:
Thanks a lot :smile:

PR #226, opened 2017-04-13
refactor: Add a specific library to the device :phone:

kosssi:


mobile/src/lib/device.js
```diff
@@ -0,0 +1,11 @@
...
+export const isIos = () => (getDeviceName() === 'ios')
+export const isAndroid = () => (getDeviceName() === 'android')
+
+export const getDeviceName = () => {
+  if (isCordova()) {
+    return window.cordova.platformId
+  }
+
+  return undefined
+}
```
enguerran: functionnal versions (just for fun, or not):

```js
const platformId = window.cordova ? window.cordova.platformId : undefined
export const getDeviceName = () => platformId

const isPlatform = platform => () => platformId === platform
export const isIos = isPlatform('ios')
export const isAndroid = isPlatform('android')
export const isCordova = !isPlatform(undefined)
```

mobile/src/lib/device.js
```diff
@@ -0,0 +1,11 @@
+export const isCordova = () => (window.cordova !== undefined)
+export const isIos = () => (getDeviceName() === 'ios')
```
enguerran: I think the `const fun = () => {}` version is not equivalent as `fun() {}`: the first must be declared before it is used, not the second. Am I right?

kosssi: 👍 https://scotch.io/tutorials/understanding-hoisting-in-javascript

PR #228, opened 2017-04-14
fix: Selection Bar UI

enguerran:
As asked in [this card](https://trello.com/c/oppcRVha/217-3-fil-21-cosmetique-selection-bar), here are some cosmetic changes:

- [x] use cozy-ui for trash icon
- [x] add `open with` in the selection bar

See the bad screenshot below

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "classnames": "^2.2.0",
     "cozy-bar": "3.0.0-beta15",
     "cozy-client-js": "0.2.0",
-    "cozy-ui": "3.0.0-beta18",
+    "cozy-ui": "3.0.0-beta20",
     "date-fns": "^1.22.0",
     "filesize": "^3.3.0",
     "hammerjs": "^2.0.8",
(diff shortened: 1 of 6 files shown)


enguerran:
IDK why we want the `open with` action in the selection bar as it can only be triggered if the selection is one and only one file. I guess the selection bar is good to perform bulk operation on several files.

The fact remains that it is now added to the code base with this PR.

package.json
```diff
@@ -105,7 +105,7 @@
     "classnames": "^2.2.0",
     "cozy-bar": "3.0.0-beta15",
     "cozy-client-js": "0.2.0",
-    "cozy-ui": "3.0.0-beta18",
+    "cozy-ui": "3.0.0-beta20",
```
GoOz: This means you'll need an new release of Cozy-UI, preferably before merging this PR, right?

GoOz: Oh I just saw that the release has already been done. Nevermind my comment. :)

PR #230, opened 2017-04-14
feat: Add Support for editor's name

GoOz:
Adding Editor's name to the manifest so the stack can use it and make sure this data is shared so the cozy-bar can use it.

tl;dr: It displays the "Cozy" before the "Drive" in the Cozy-bar

Related PR: https://github.com/cozy/cozy-bar/pull/37

### DIFF
--- a/mobile/src/lib/cozy-helper.js
+++ b/mobile/src/lib/cozy-helper.js
@@ -35,14 +35,11 @@ export const initClient = (url, onRegister = null, device = 'Device') => {
 }
 
 export const initBar = () => {
-  const root = document.querySelector('[role=application]')
-  const data = root.dataset
-
   cozy.bar.init({
-    appName: data.cozyAppName,
-    appEditor: data.cozyAppEditor,
-    iconPath: data.cozyIconPath,
-    lang: data.cozyLocale,
+    appName: 'Cozy Drive',
+    appEditor: 'Cozy',
+    iconPath: require('../../../vendor/assets/app-icon.svg'),
+    lang: 'en',
     replaceTitleOnMobile: true
   })
 }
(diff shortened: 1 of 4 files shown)


GoOz:
[(1) [BAR-11 / E-36] ETDeveloppeur d'application tiers, je n'ai pas 'Cozy' devant le nom de mon app](https://trello.com/c/zsq1dQq4/298-1-bar-11-e-36-etdeveloppeur-d-application-tiers-je-n-ai-pas-cozy-devant-le-nom-de-mon-app)

kosssi:
Can you update [here](https://github.com/cozy/cozy-drive/blob/master/mobile/src/lib/cozy-helper.js#L39)?

GoOz:
@kosssi I just updated and tested with a watch:mobile, but i'm not 100% sure this is the way to test it. Anyway I have the appEditor string in the DOM but hidden because of… well because we're on mobile and we hide this like before.

mobile/src/lib/cozy-helper.js
```diff
@@ -35,10 +35,14 @@ export const initClient = (url, onRegister = null, device = 'Device') => {
 }
 
 export const initBar = () => {
+  const root = document.querySelector('[role=application]')
+  const data = root.dataset
```
kosssi: Sorry but you can't do it like that. This method is when app is render by server which replace data.

You can write directly data:

    cozy.bar.init({
      appName: 'Drive',
      appEditor: 'Cozy',
      iconPath: require('../../../vendor/assets/app-icon.svg'),
      lang: getLang()

getLang is on this PR https://github.com/cozy/cozy-drive/pull/231

GoOz: AAAAAH ok I absolutely didn't get it at first, my bad. I'll correct it.

PR #236, opened 2017-04-21
fix: cordova displays app in portrait orientation

enguerran:
I don't think landscape mode is very accurate for this kind of application.
I rather to forbid it by adding this configuration preference in cordova's `config.xml`.


kosssi:
I'm not really in favor of fixing orientation, it's an unnecessary limitation and I've already had cases of people with broken phones who was causing problem to fix orientation (on cozy-mobile v2).

enguerran:
If we address broken phone users... :trollface:

PR #239, opened 2017-04-26
feat: Verify authorization when backup parameter change :lock:

kosssi:


kosssi:
[(2) [FIL-36] ET utilisateur Android et iOS la popup d'autorisation d'accès aux fichiers doit apparaitre pendant l'onboarding](https://trello.com/c/NKzqehc3/248-2-fil-36-et-utilisateur-android-et-ios-la-popup-d-autorisation-d-acces-aux-fichiers-doit-apparaitre-pendant-l-onboarding)

mobile/src/lib/media.js
```diff
@@ -49,21 +64,7 @@ export const getPhotos = async () => {
         (response) => resolve(response.library),
         (err) => {
           console.warn(err)
-          if (err.startsWith('Permission')) {
-            requestAuthorization().then(authorization => {
```
kosssi: We don't want to ask permission when the backup starts but only when the user makes changes in the interface.
PS: we are on getPhotos function

mobile/src/actions/mediaBackup.js
```diff
@@ -40,35 +43,82 @@ async function getDirID (dir) {
...
+        }
       }
+      dispatch(endMediaUpload())
+    }
+  })
+}
+
+// backupImages
+
+export const backupImages = backupImages => async (dispatch, getState) => {
```
kosssi: It's main function to pilot backupImages

mobile/src/actions/mediaBackup.js
```diff
@@ -39,36 +42,96 @@ async function getDirID (dir) {
...
+  if (canStart(force, getState)) {
     const photosOnDevice = await getFilteredPhotos()
     const alreadyUploaded = getState().mobile.mediaBackup.uploaded
     const photosToUpload = photosOnDevice.filter(photo => !alreadyUploaded.includes(photo.id))
     const dirID = await getDirID(dir)
     const totalUpload = photosToUpload.length
     let uploadCounter = 0
     for (const photo of photosToUpload) {
-      if (getState().mobile.mediaBackup.cancelMediaBackup) {
+      if (getState().mobile.mediaBackup.cancelMediaBackup || !canStart(force, getState)) {
```
enguerran: `canStart` is used to control `startMediaBackup` but it also used in the loop. But in the loop, it cannot be named `canStart`. You may name it `canBackup`, imho.

mobile/src/actions/mediaBackup.js
```diff
@@ -39,36 +42,96 @@ async function getDirID (dir) {
...
+  if (backupImages) {
+    dispatch(startMediaBackup(getMediaFolderName()))
+  }
+
+  return backupImages
+}
+
+const updateValueAfterRequestAuthorization = async (value) => {
+  if (value) {
+    value = await requestAuthorization()
```
enguerran: You should pass `requestAuthorization` function as a parameter if this function depends on a specific plugin.

mobile/src/containers/Settings.jsx
```diff
@@ -31,8 +31,17 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     dispatch(unlink(client))
     ownProps.router.replace('/onboarding')
   },
-  setBackupImages: (e) => dispatch(setBackupImages(e.target.checked)),
-  setWifiOnly: (e) => dispatch(setWifiOnly(e.target.checked)),
+  setBackupImages: async (e) => {
+    const value = e.target.checked
+    const action = await dispatch(backupImages(value))
+    if (value === true && value !== action) {
+      e.target.checked = false
```
enguerran: The side effect is not compatible with the controlled input component that uses this `setBackupImages` function.

mobile/src/actions/mediaBackup.js
kosssi: the dependency is manage on `requestAuthorization` 😀

mobile/src/containers/Settings.jsx
kosssi: when we call `backupImages` may be user don't authorized files access. So we ask him. If he refuses, the state don't change. Checkbox and state are different so we uncheck it ;)

goldoraf: @kosssi not good, if the checkbox state is controlled by a reducer, it must stay that way: manipulating DOM directly like this is an antipattern. Your `backupImages` action creator should dispatch whatever actions necessary to update the reducer for the checkbox state.

PR #242, opened 2017-05-02
Add react-virtualized for large lists of files

goldoraf:
#### Checklist

- [x] Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)
- [x] The package.json and yarn.lock files include the correct dependency version, in those versions have been released
- [x] [cozy-client-js](https://github.com/cozy/cozy-client-js) documentation and tests are up to date


src/reducers/view.js
```diff
@@ -97,26 +135,16 @@ const lastFetch = (state = null, action) => {
...
-
-export const getVisibleFiles = ({ view }) => {
-  const { files } = view
-  return getSortedFiles(files)
-}
+export const getVisibleFiles = ({ view }) => view.files
 
 export const getFileById = ({ view }, id) => {
-  const file = view.files.find(f => f.id === id)
+  const file = view.files.find(f => f && f.id && f.id === id)
```
kosssi: Are you sure is nessacerry `&& f.id &&`:

    const file = view.files.find(f => f && f.id === id)

src/components/FileList.jsx
```diff
@@ -1,14 +1,78 @@
...
+              />
+            )}
+          </AutoSizer>
+        )}
+      </InfiniteLoader>
+    )
+  }
+
+  isRowLoaded = ({ index }) => {
+    return !!this.requestedFiles[index]
```
kosssi: Can you explain `!!`?

src/components/FileList.jsx
```diff
@@ -1,14 +1,78 @@
...
+const STATUS_LOADING = 1
+const STATUS_LOADED = 2
+
+const FETCH_LIMIT = 50
+
+class FileList extends PureComponent {
+  constructor (props) {
+    super(props)
+    this.requestedFiles = {}
+    for (var i = 0; i < props.files.length; i++) this.requestedFiles[i] = STATUS_LOADED
```
kosssi: can you replace all `for (var` by `for (const` please?

src/components/FileList.jsx
```diff
@@ -1,14 +1,78 @@
...
+      >
+        {({ onRowsRendered, registerChild }) => (
+          <AutoSizer>
+            {({ width, height }) => (
+              <List
+                ref={registerChild}
+                height={height}
+                rowCount={fileCount}
+                onRowsRendered={onRowsRendered}
+                rowHeight={50}
```
kosssi: maybe `50` value can be on a variable.

src/components/FileList.jsx
```diff
@@ -1,14 +1,78 @@
...
+    )
+  }
+
+  isRowLoaded = ({ index }) => {
+    return !!this.requestedFiles[index]
+  }
+
+  loadMoreRows = ({ startIndex, stopIndex }) => {
+    const limit = stopIndex - startIndex + 1
+    for (var i = startIndex; i <= stopIndex; i++) this.requestedFiles[i] = STATUS_LOADING
```
kosssi: `var` to `const` ;)

src/components/FileList.jsx
```diff
@@ -1,14 +1,78 @@
...
+    return this.props.fetchMoreFiles(this.props.displayedFolder.id, startIndex, limit)
+      .then(() => {
+        for (var i = startIndex; i <= stopIndex; i++) this.requestedFiles[i] = STATUS_LOADED
```
kosssi: `var` to `const` ;)

src/components/UploadButton.jsx
```diff
@@ -26,11 +26,12 @@ const UploadButton = ({ label, disabled, onUpload, className }) => (
     {label}
     <input
       type='file'
+      multiple
```
kosssi: 👍

src/reducers/view.js
goldoraf: It's necessary for now because of a bug on the stack side. I'll add a comment.

src/components/FileList.jsx
goldoraf: No entry in this map signifies that the row has never been loaded before. An entry (either LOADING or LOADED) can be treated as loaded as far as InfiniteLoader is concerned.

src/components/FileList.jsx
```diff
@@ -1,14 +1,81 @@
...
-}) => (
-  <div>
-    {files.map((file, idx) => (
+const STATUS_LOADING = 1
+const STATUS_LOADED = 2
+
+const FETCH_LIMIT = 50
+const ROW_HEIGHT = 50
+
+class FileList extends PureComponent {
```
enguerran: What's the point to use `PureComponent` instead of `Component`?

src/components/FileList.jsx
enguerran: Don't you think the declarative way is clearer?

```js
this.requestedFiles = props.files.map(file => STATUS_LOADED)
```

src/components/FileList.jsx
```diff
@@ -1,14 +1,81 @@
...
+  constructor (props) {
+    super(props)
+    this.requestedFiles = {}
+    for (var i = 0; i < props.files.length; i++) this.requestedFiles[i] = STATUS_LOADED
```
enguerran: What's the point to declare `this.requestedFiles` as an object indexed by properties representing a number instead of an array:

```js
const obj = { 0: STATUS_LOADED, 1: STATUS_LOADED, 2: STATUS_LOADED }
// vs
const arr = [STATUS_LOADED, STATUS_LOADED, STATUS_LOADED]

// with access
obj[0] // STATUS_LOADED
// vs
arr[0] // STATUS_LOADED
```

src/components/FileList.jsx
enguerran: And with a declarative way:

```js
this.requestedFiles = [
  ...this.requestedFiles.slice(0, startIndex),
  ...this.requestedFiles.slice(startIndex, stopIndex).map(file => STATUS_LOADING),
  ...this.requestedFiles.slice(stopIndex)
]
```

src/ducks/files/Toolbar.jsx
```diff
@@ -56,8 +56,8 @@ const mapStateToProps = (state, ownProps) => ({
 })
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  uploadFile: (file, displayedFolder) => {
-    dispatch(uploadFile(file, displayedFolder))
+  uploadFiles: (files, displayedFolder) => {
+    dispatch(uploadFiles(files, displayedFolder))
```
enguerran: It will be clearer to rename that 2 variables as we can miss than the second comes from an import.

src/reducers/view.js
```diff
@@ -32,21 +32,60 @@ const displayedFolder = (state = null, action) => {
   }
 }
 
+const fileCount = (state = null, action) => {
```
enguerran: To avoid runtime exception, we should declare `action` with the default value `{}`, giving:

```js
const fileCount = (state = null, action = {}) => {
```

src/components/FileList.jsx
goldoraf: Because there could be gaps in the requested files : if the user is scrolling very fast, the lib will not always fetch the next batch of files, but may skip some. That's why using an object is simpler.

src/components/FileList.jsx
goldoraf: `PureComponent` is used in react-virtualized samples, so I presumed it could have a performance impact.

PR #243, opened 2017-05-02
Some fixes for the README

nono:


.transifexrc.tpl
```diff
@@ -1,4 +0,0 @@
-[https://www.transifex.com]
-hostname = https://www.transifex.com
-username = cozy
-token =
```
enguerran: Removing this file does not please travis: https://travis-ci.org/cozy/cozy-drive/builds/227905003#L405

nono: OK, let's keep it

PR #254, opened 2017-05-12
chore: add piwik tracking

enguerran:
I don't know how to configure webpack the best way to:

- [x] add `piwik.js` in index.html
- [x] configure `piwik.js`
- [x] enable the `piwik-react-router` history tracker

Please help! 🆘

**[EDIT]**
I find something better, see below.

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -18,10 +18,11 @@ module.exports = {
     }),
     new webpack.DefinePlugin({
       'process.env.NODE_ENV': JSON.stringify('production'), // to compile on production mode (redux)
-      __SERVER__: false,
       __DEVELOPMENT__: false,
       __DEVTOOLS__: false,
-      __STACK_ASSETS__: true
+      __STACK_ASSETS__: true,
+      __PIWIK_SITEID__: 8,
+      __PIWIK_TRACKER_URL__: JSON.stringify('https://piwik.cozycloud.cc')
     })
   ]
 }
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -10,6 +10,7 @@
   <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
   <meta name="theme-color" content="#ffffff">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <script src="//{{.Domain}}/assets/js/piwik.js" defer></script>
   <% if (__TARGET__ === 'mobile') { %>
   <meta name="format-detection" content="telephone=no">
   <script src="cordova.js" defer></script>
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -13,9 +13,10 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      __SERVER__: JSON.stringify('http://app.cozy.local'),
       __STACK_ASSETS__: false,
-      __DEVELOPMENT__: true
+      __DEVELOPMENT__: true,
+      __PIWIK_SITEID__: 8,
+      __PIWIK_TRACKER_URL__: JSON.stringify('https://piwik.cozycloud.cc')
     }),
     new webpack.ProvidePlugin({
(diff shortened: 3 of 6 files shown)


enguerran:
For now, I assume following:

- cozy-stack have a `{{.CozyPiwik}}` replaced by `assets/js/piwik.js`
- copy `src/tracking.js` with hard-coded configuration (`siteId` and `url`)
- use a third-party plugin to wrap `react-router`

enguerran:
[(2) [GLOB-7] Traces d'usage](https://trello.com/c/zUe4D0KA/269-2-glob-7-traces-d-usage)

config/webpack.config.prod.js
```diff
@@ -17,10 +17,11 @@ module.exports = {
       }
     }),
     new webpack.DefinePlugin({
-      __SERVER__: false,
       __DEVELOPMENT__: false,
       __DEVTOOLS__: false,
-      __STACK_ASSETS__: true
+      __STACK_ASSETS__: true,
+      __PIWIK_SITEID__: 8,
+      __PIWIK_TRACKER_URL__: 'https://piwik.cozycloud.cc/index.php'
```
nono: Why is there a `JSON.stringify` for dev, but not for prod?

src/tracking.js
```diff
@@ -0,0 +1,8 @@
+var idSite = 8
+var piwikTrackingApiUrl = 'https://piwik.cozycloud.cc/piwik.php'
```
nono: Is it useful to have the piwik URL both here and in the init of react router?

src/index.ejs
```diff
@@ -10,6 +10,8 @@
   <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
   <meta name="theme-color" content="#ffffff">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  {{.CozyPiwik}}
```
nono: Oh, you can just use:

```html
<script src="{{.Domain}}/js/piwik.js"></script>
```

It will let you add `defer` if you prefer

src/tracking.js
enguerran: Oh? No! That's a mistake.

My wish was to let the configuration under webpack's responsibility, but I could not configure webpack (such a noob I am!) to create the correct `tracking.js`.

As I said in [my previous comment](https://github.com/cozy/cozy-drive/pull/254#issuecomment-301110069), I hard-coded `siteId` and `URL` in `tracking.js`.

src/index.ejs
enguerran: Nicer! 🎉

src/tracking.js
m4dz: You can't replace data with webpack compilation variables (as defined in the `Define` plugin) when you _copy_ the file. You must `require` it to take care of the defined compilation vars.

The point is that we can here not copying the `tracking.js` file but require it in `main.jsx|router` to enable the tracking. As the tracking file rely on the global `_paq` variable, we can use the [`external` config](https://webpack.github.io/docs/library-and-externals.html) key to declares

```js
externals: {
  "_paq": "_paq"
}
```

and do a `require('_paq') in `tracking.js` to use the `_paq` var injected by the `piwiki.js` external script.

enguerran:
@m4dz @nono Do you agree these changes?

src/index.ejs
```diff
@@ -10,6 +10,7 @@
   <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
   <meta name="theme-color" content="#ffffff">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <script src="//{{.Domain}}/assets/js/piwik.js" async defer></script>
```
m4dz: Hmm, a `script` can't be `async` and `defer` at the same time. We choose to rely on `defer` as it's sufficient for webapps perfs

enguerran: However I *cargo cult* this code from Piwik's website: https://piwik.org/faq/general/faq_20904/

PR #262, opened 2017-05-22
Intents to open a file and download a file

ptbrowne:
Downloading a file is useful for example for the EDF app when you want to download your bills.
It also can be useful to view your files in an iframe.

### DIFF
--- a/src/containers/IntentService.jsx
+++ b/src/containers/IntentService.jsx
@@ -54,9 +54,9 @@ export default class IntentService extends Component {
       })
   }
 
-  terminate (account) {
+  terminate (msg) {
     const { service } = this.state
-    service.terminate(account)
+    service.terminate(msg)
   }
 
   cancel () {
--- a/src/containers/IntentService.jsx
+++ b/src/containers/IntentService.jsx
@@ -1,3 +1,5 @@
+/* global cozy */
+
 import styles from '../styles/services'
 
 import React, { Component } from 'react'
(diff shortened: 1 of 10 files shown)


src/services.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+  cozy.client.init({
+    cozyURL: '//' + data.cozyDomain,
+    token: data.cozyToken
+  })
+
+  render((
+    <I18n context={context} lang={data.cozyLocale}>
+      <Provider store={store}>
+        <IntentService window={window} />
```
enguerran: I don't think that it is useful to pass a global object into another scope.

src/services.jsx
```diff
@@ -0,0 +1,58 @@
...
+  require('preact/devtools')
+
+  // Export React to window for the devtools
+  window.React = React
+}
+
+// Enable Redux dev tools
+const composeEnhancers = (__DEVELOPMENT__ && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) || compose
+
+const store = createStore(
```
enguerran: There is no need of a redux store in this sub-application.

src/containers/IntentService.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+import styles from '../styles/services'
+
+import React, { Component } from 'react'
+
+import Loading from '../components/Loading'
+import ServiceBar from '../components/services/ServiceBar'
+import ViewerService from '../components/services/ViewerService'
+
+export default class IntentService extends Component {
```
enguerran: I guess we could write something simpler than this very big and not designed component.

ptbrowne:
Better PR here : https://github.com/cozy/cozy-drive/pull/269

PR #265, opened 2017-05-23
feat: Add a user id and an app variable

y-lohse:


src/main.jsx
```diff
@@ -66,6 +66,12 @@ document.addEventListener('DOMContentLoaded', () => {
       injectScript: false
     }))
     piwikTracker.push(['enableHeartBeatTimer'])
+    let userId = data.cozyDomain
+    let indexOfPort = userId.indexOf(':')
+    if (indexOfPort >= 0) userId = userId.substring(0, indexOfPort)
```
enguerran: I think if `userId` is the user's domain, it may contain the port since en user could have several cozy's installation on his same domain.

y-lohse: That makes sense. Let's keep the domain intact then.

PR #267, opened 2017-05-24
feat: Add piwik events

y-lohse:
I've created a new redux middleware that can be used as a hook to track certain actions in piwik. The upside is that we'll be able to track most actions quite easily; the downside is that sometimes (as is the case here), we're going to need new actions / modify existing one solely for the middleware, not for the store.

Tell me what you think about this approach, and once we've found a good one I'll deploy the same thing to photos.

### DIFF
--- a/src/middlewares/piwik.js
+++ b/src/middlewares/piwik.js
@@ -22,8 +22,11 @@ const piwik = store => next => action => {
   }
 
   if (event && event.category && event.action) {
-    const tracker = Piwik.getTracker()
-    tracker.trackEvent(event.category, event.action, event.name, event.value)
+    try {
+      const tracker = Piwik.getTracker()
+      tracker.trackEvent(event.category, event.action, event.name, event.value)
+    }
+    catch (err) { }
   }
 
   return next(action)
--- a/src/middlewares/piwik.js
+++ b/src/middlewares/piwik.js
@@ -14,7 +14,7 @@ const ACTIONS = {
 const piwik = store => next => action => {
   let event = null
 
-  switch (action.type){
+  switch (action.type) {
     case PRE_UPLOAD_FILE:
       event = {
         category: CATEGORY.INTERACTION,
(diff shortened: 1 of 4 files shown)


src/middlewares/piwik.js
```diff
@@ -0,0 +1,40 @@
+/* global Piwik */
+import {
+  PRE_UPLOAD_FILE
+} from '../actions'
```
enguerran: Not sure this middleware should depend on another file. It prevents to be easily extracted and reused elsewhere.

src/middlewares/piwik.js
```diff
@@ -0,0 +1,40 @@
...
+    try {
+      const tracker = Piwik.getTracker()
+      tracker.trackEvent(event.category, event.action, event.name, event.value)
+    } catch (err) { }
+  }
+
+  return next(action)
+}
+
+export default piwik
```
enguerran: The name should reflect the responsibility: s/piwik/piwikMiddleware/

src/middlewares/piwik.js
y-lohse: I agree, although I don't see a better way to do this. Keeping the actions somewhere else would mean the middleware needs an API to be configured, and would be practically empty. I don't mind the "extract and modify" approach in this case, but I'm open to other ways to structure this :)

src/middlewares/piwik.js
y-lohse: The file being in a folder called `middlewares` makes this pretty clear. It also seems pretty standard ([refux-logger](https://github.com/evgenyrodionov/redux-logger/blob/master/src/index.js#L126) and [redux-thunk](https://github.com/gaearon/redux-thunk/blob/master/src/index.js#L14) typically do the same). And at [import time](https://github.com/cozy/cozy-drive/blob/master/src/main.jsx#L16), the middleware-thing is kept explicit.

PR #268, opened 2017-05-26
feat: OPEN intent service to embed pdf

enguerran:
Create a "sub-app" for intents: OPEN:id.

It renders a html page with the id's content in a `embed` HTML element.
Need to be stylized a little bit more.

It is used in cozy-bank to display a bill's pdf file. (cf. https://gitlab.cozycloud.cc/enguerran/cozy-bank/tree/intents)


ptbrowne:
I think we can close this in favor of : https://github.com/cozy/cozy-drive/pull/269 ?

PR #269, opened 2017-05-29
feat: GET_URL and OPEN intents

ptbrowne:
Enables applications to open a file via an intent. Here we display a PDF in cozy-bank via `<embed />`.

Also add the GET_URL intent that is useful for the cozy-edf application


src/services.ejs
```diff
@@ -0,0 +1,28 @@
...
+  <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
+  <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32">
+  <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16">
+  <link rel="manifest" href="/manifest.json">
+  <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
+  <meta name="theme-color" content="#ffffff">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <% if (__TARGET__ === 'mobile') { %>
+  <meta name="format-detection" content="telephone=no">
+  <script src="cordova.js" defer></script>
```
enguerran: Don't think we need this for `services.ejs`

src/services.ejs
```diff
@@ -0,0 +1,28 @@
...
+  <meta charset="utf-8">
+  <title><%= htmlWebpackPlugin.options.title %></title>
+  <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
+  <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32">
+  <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16">
+  <link rel="manifest" href="/manifest.json">
+  <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
+  <meta name="theme-color" content="#ffffff">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <% if (__TARGET__ === 'mobile') { %>
```
enguerran: We could remove this line.

config/webpack.config.base.js
```diff
@@ -48,6 +48,17 @@ module.exports = {
...
       title: pkg.name,
+      excludeChunks: ['services'],
+      inject: 'head',
+      minify: {
+        collapseWhitespace: true
+      }
+    }),
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, '../src/services.ejs'),
+      title: pkg.name,
```
enguerran: The title for this page should render the fact it is a service.

src/components/FileViewer.jsx
```diff
@@ -0,0 +1,38 @@
+/* global cozy */
+
+import utilStyles from '../styles/utils'
+import React from 'react'
+import Loading from './Loading'
+
+class FileViewer extends React.Component {
```
enguerran: The fact this file and object are called `FileViewer` and the fact it requests an `intent` as a props seem inconsistent.
As cozy-drive is a file explorer, I guess we will have a file viewer feature.
In the meantime the file and object's name do not reflect this very feature: service OPEN intent_id

config/webpack.config.base.js
ptbrowne: True. Is the title of an iFrame ever displayed ?

src/services.ejs
ptbrowne: Why ? Cozy-drive will never be used on mobile ?

src/components/FileViewer.jsx
ptbrowne: You're right, I am thinking of `IntentHandler` since it is also used to handle the `GET_URL` intent. What do you think ?

config/webpack.config.base.js
ptbrowne: I think we should just remove the title for the moment, as having it means that we need to care about i18n. Since the iframe is embedded in another page, I do not think this title will be used.

src/services.ejs
ptbrowne: Ah sorry I understood, it is only for Cordova apps. I'm removing it.

src/components/FileViewer.jsx
```diff
@@ -12,17 +12,32 @@ class FileViewer extends React.Component {
...
   componentDidMount () {
-    const intent = this.props.intent
-    cozy.client.intents.createService(intent, window)
-      .then(service => {
+    const { intentId } = this.props
+
+    let service, intent
+    cozy.client.intents.createService(intentId, window)
+      .then(_service => {
+        service = _service
```
enguerran: I don't like this variable naming.
Maybe `service` should be named `serviceAnchor` as its purpose is to be used to `terminate` it.
and `_service` should be named `service` as the function `createService` is resolved to a `service`.

src/components/FileViewer.jsx
```diff
@@ -12,17 +12,32 @@ class FileViewer extends React.Component {
   }
 
   componentDidMount () {
-    const intent = this.props.intent
-    cozy.client.intents.createService(intent, window)
-      .then(service => {
+    const { intentId } = this.props
+
+    let service, intent
+    cozy.client.intents.createService(intentId, window)
```
enguerran: Could you use async/await syntax in this function to let the code be more expressive?

src/components/FileViewer.jsx
```diff
@@ -12,17 +12,32 @@ class FileViewer extends React.Component {
...
+          break
+          case 'GET_URL':
+          service.terminate({ url })
+          break
+        }
       }).catch(error => {
         this.setState({ error, loading: false })
+
+        if (this.intent.attributes.action == 'GET_URL') {
+          service.terminate({ error })
```
enguerran: When it catches an error, `service` may be undefined and `service.terminate` may throw an error. Could you check the `service` content before using its function?

src/components/FileViewer.jsx
enguerran: LGTM

src/components/FileViewer.jsx
ptbrowne: Using async/await removes this problem 👍

src/components/FileViewer.jsx
ptbrowne: Good catch !

src/components/FileViewer.jsx
ptbrowne: Done

manifest.webapp
```diff
@@ -1,4 +1,4 @@
-{
+  {
```
kosssi: not necessary ?

ptbrowne: 👍

src/components/FileViewer.jsx
```diff
@@ -12,17 +12,32 @@ class FileViewer extends React.Component {
...
         return cozy.client.files.getDownloadLinkById(id)
       })
       .then(link => `${cozy.client._url}${link}`)
       .then(url => {
-        this.setState({ url, loading: false })
+        switch (intent.attributes.action) {
+          case 'OPEN':
+          this.setState({ url, loading: false })
+          break
+          case 'GET_URL':
```
gregorylegarec: Thanks for that, I will reuse it !

ptbrowne:
ping

gregorylegarec:
@ptbrowne I did not review the full PR but as we will need the 'GET_URL' intent soon I am pretty up to merge.

PR #271, opened 2017-05-29
refactor: Reset all state with a rootReducer :hammer:

kosssi:
This feature reset all state with a rootReducer to unlink a user.

I find this solution [here](https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992) by Dan Abramov.

### DIFF
--- a/mobile/src/actions/unlink.js
+++ b/mobile/src/actions/unlink.js
@@ -1,12 +1,12 @@
 /* global cozy */
 
-import { initializeState } from '../../../src/actions/settings'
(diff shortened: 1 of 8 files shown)


kosssi:
[[FIL] Suite à une révocation du device, se réenregistrer ne vide pas la base locale](https://trello.com/c/XTWyoBzJ/315-fil-suite-a-une-revocation-du-device-se-reenregistrer-ne-vide-pas-la-base-locale)

PR #273, opened 2017-05-30
refactor: Piwik wrapper

y-lohse:
This is a proposal for a module dedicated to Piwik, which could be externalized to cozy-ui or a separate library.

The module exposes 2 things:
- access to a unique Piwik instance object; this is important because otherwise every instance has to be configured with the user id, the app name, etc.
- a redux middleware (see comments in code for more on that)

The module's function pull default values from our usual setup (global variables and the root element's dataset), so they need little configuration. However all these defaults can be overriden, so the module should work in any setup and all parts are optional. The only dependency is [piwik-react-router](https://github.com/joernroeder/piwik-react-router) and, of course, Piwik.

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -26,5 +26,6 @@
   data-cozy-locale="{{.Locale}}"
   data-cozy-app-name="{{.AppName}}"
   data-cozy-app-editor="{{.AppEditor}}"
+  data-cozy-tracking="{{.Tracking}}"
   data-cozy-icon-path="{{.IconPath}}"
 >
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -14,9 +14,11 @@
   <meta name="format-detection" content="telephone=no">
   <script src="cordova.js" defer></script>
   <% } else if (__STACK_ASSETS__) { %>
-  <script src="//{{.Domain}}/assets/js/piwik.js" defer></script>
   {{.CozyClientJS}}
   {{.CozyBar}}
+  <% } %>
+  <%  if (__STACK_ASSETS__ || __DEVELOPMENT__) { %>
+  <script src="//{{.Domain}}/assets/js/piwik.js" defer></script>
   <% } %>
 </head>
 <div
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -58,22 +58,25 @@ document.addEventListener('DOMContentLoaded', () => {
   })
 
   let history = hashHistory
-  try {
-    var PiwikReactRouter = require('piwik-react-router')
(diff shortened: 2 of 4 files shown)


src/index.ejs
```diff
@@ -14,10 +14,12 @@
   <meta name="format-detection" content="telephone=no">
   <script src="cordova.js" defer></script>
   <% } else if (__STACK_ASSETS__) { %>
-  <script src="//{{.Domain}}/assets/js/piwik.js" defer></script>
   {{.CozyClientJS}}
   {{.CozyBar}}
   <% } %>
+  <%  if (__STACK_ASSETS__ || __DEVELOPMENT__) { %>
```
y-lohse: Made piwik available in dev.

src/middlewares/EventTracker.js
```diff
@@ -1,4 +1,3 @@
-/* global Piwik */
```
y-lohse: This middleware has changed: instead of having a single middleware that decides if an action should be tracked and tracking it, there are now 2 separate middlewares. This one is custom for each application and only sets a flag in the action, which will be picked up by the Piwik middleware later in the chain.

src/index.ejs
kosssi: can you exclude mobile ?

src/main.jsx
```diff
@@ -60,22 +49,23 @@ document.addEventListener('DOMContentLoaded', () => {
...
-    }))
-    piwikTracker.push(['enableHeartBeatTimer'])
-    let userId = data.cozyDomain
-    let indexOfPort = userId.indexOf(':')
-    if (indexOfPort >= 0) userId = userId.substring(0, indexOfPort)
-    piwikTracker.push(['setUserId', userId])
-    piwikTracker.push(['setCustomDimension', __PIWIK_DIMENSION_ID_APP__, data.cozyAppName])
+  let middlewares = [thunkMiddleware, loggerMiddleware]
+
+  if (shouldEnableTracking() && getTracker()) {
```
kosssi: est ce que ca vaut pas le coup de mettre tout le if dans une fonction du lib/tracker ? et de mettre à la place :

    middlewares = addTracker(middlewares)

y-lohse: > J'imagine que tracker va devenir par la suite une lib partagé avec l'ensemble des projets non ?

Je pense oui. Garder la condition en dehors de la lib permet au dev de choisir s'il veux connecter le `history`du routeur, le middleware d'événements, ou les 2. Ou aucun des 2. Ou d'autres modules qui viendraient se greffer plus tard. Ca donne aussi de la flexibilité sur la config à utiliser pour chacune de ces options.

PR #274, opened 2017-06-01
chore: Remove source map on mobile prod build :fire:

kosssi:


config/webpack.config.prod.js
```diff
@@ -1,12 +1,14 @@
...
 const webpack = require('webpack')
+const {target} = require('./webpack.vars')
+const devtool = target === 'mobile' ? false : '#cheap-module-source-map'
 
 module.exports = {
   output: {
     filename: 'app.[hash].min.js'
   },
-  devtool: '#cheap-module-source-map',
+  devtool: devtool,
```
m4dz: I think we can totally disable sourcemaps in production, so explicitely set it to `false` here, whatever the context. I invoke @goldoraf @gregorylegarec @nono for their opinion about having sourcemaps for webapps in production.

kosssi: you can read more documentation [here](http://webpack.github.io/docs/configuration.html#devtool)

PR #275, opened 2017-06-01
feat: enable piwik tracking on mobile

y-lohse:
As discussed, I've updated the tracker lib so that it can be used to inject the piwik script instead of relying on it to be already there.

On the mobile side of things, i've added a file in lib that can active and deactivate the tracking, and I've added the middlewares. the activate function is called once on start up, and when the setting is changed.

### DIFF
--- a/mobile/src/lib/tracker.js
+++ b/mobile/src/lib/tracker.js
@@ -1,0 +1,31 @@
+/* global __PIWIK_TRACKER_URL__ __PIWIK_SITEID__ */
+import { getTracker, configureTracker } from '../../../src/lib/tracker'
+import { cozyAppName } from './cozy-helper'
+
+// We'll need access to the app's router history at some point
+let appHistory
+
+export const useHistoryForTracker = (history) => {
+  appHistory = history
+}
+
+export const startTracker = (cozyServerUrl = '') => {
+  // start the tracker, inject the script
+  const trackerInstance = getTracker(__PIWIK_TRACKER_URL__, __PIWIK_SITEID__, false, true)
+
+  // configure the options that aren't in webpack variables
+  let url = new URL(cozyServerUrl)
+
+  configureTracker({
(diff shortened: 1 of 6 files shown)


mobile/src/main.jsx
```diff
@@ -54,6 +55,9 @@ const renderAppWithPersistedState = persistedState => {
     if (navigator && navigator.splashscreen) navigator.splashscreen.hide()
   }, false)
 
+  useHistoryForTracker(hashHistory)
+  if (store.getState().mobile.settings.analytics) startTracker(store.getState().mobile.settings.serverUrl)
```
y-lohse: @kosssi : this is the part I'm the least sure about. Is this the right place to boot the tracking code, or should it be in the `deviceReady` callback or something like that?

mobile/src/lib/tracker.js
```diff
@@ -0,0 +1,31 @@
...
+export const startTracker = (cozyServerUrl = '') => {
+  // start the tracker, inject the script
+  const trackerInstance = getTracker(__PIWIK_TRACKER_URL__, __PIWIK_SITEID__, false, true)
+
+  // configure the options that aren't in webpack variables
+  let url = new URL(cozyServerUrl)
+
+  configureTracker({
+    userId: url.hostname || cozyServerUrl,
+    app: cozyAppName
```
kosssi: Are you sure you want 'Drive'? no 'io.cozy.drive.mobile' ?

mobile/src/main.jsx
kosssi: 👍 it's good

y-lohse:
[(1) [GLOB] Ajouter le tracking dans l'application mobile](https://trello.com/c/cCS63vid/314-(1)-%5Bglob%5D-ajouter-le-tracking-dans-l%27application-mobile)

PR #278, opened 2017-06-02
fix: use a different piwik instance for mobile

y-lohse:


y-lohse:
[(1) [GLOB] Ajouter le tracking dans l'application mobile](https://trello.com/c/cCS63vid/314-(1)-%5Bglob%5D-ajouter-le-tracking-dans-l%27application-mobile)

PR #293, opened 2017-06-14
feat: Remove error when is already on trash :tennis:

kosssi:


src/actions/index.js
```diff
@@ -217,12 +217,25 @@ export const trashFiles = files => {
         trashed.push(await cozy.client.files.trashById(file.id))
       }
     } catch (err) {
-      return dispatch({
-        type: TRASH_FILES_FAILURE,
-        alert: {
-          message: 'alert.try_again'
+      const isAlreadyInTrash = err => {
```
y-lohse: Is there any specific reason why the function is declared inside another function? If not I feel like it should be declared outside, to avoid recreating it every time.

kosssi: @y-lohse it's better?

PR #302, opened 2017-06-22
fix: select the first file scrolled up the page

enguerran:
**bug**
On the first file selection, the view scrolled up to the top.
It was due to the `<SelectionBar />` addition in the DOM.

**fix**
We always have the `<SelectionBar />` in the DOM, but when no files are selected, the wrapping `<div />` HTMLElement is `display: none`. As soone as the first file is selected, we don't mutate the DOM but displaying the wrapping `<div />` HTMLElement.

### DIFF
--- a/src/components/FolderView.jsx
+++ b/src/components/FolderView.jsx
@@ -89,7 +89,9 @@ class FolderView extends Component {
         </Topbar>
         <div role='contentinfo'>
           {__TARGET__ === 'mobile' && <UploadProgression />}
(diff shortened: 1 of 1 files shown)


enguerran:
[[FIL] Sélection un fichier après avoir scrollé la liste fait remonter la liste tout en haut](https://trello.com/c/TNyRp6Zq/382-fil-s%C3%A9lection-un-fichier-apr%C3%A8s-avoir-scroll%C3%A9-la-liste-fait-remonter-la-liste-tout-en-haut)

PR #303, opened 2017-06-23
Publish new release

jsilvestre:
I tested all the changes on int and it appears to work like a charm!

Here is the changelog : 
- (mobile) a specific message is now displayed when the user tries to use this app with a Cozy v2
- fix date format of files
- various styles improvement


enguerran:
@jsilvestre I think we could use a PR to make a checklist of what is needed for a release, example:

- [x] write the changelog
- [ ] update the package.json version number
- [ ] write a blog post to celebrate
- [ ] deploy

Something like https://github.com/Kinto/kinto/pull/1199

jsilvestre:
Yes, something like:
- write the changelog and hand it over the engagement team (PO)
- update the package.json version number
- wait for the build & deploy

PR #308, opened 2017-06-27
fix: upload queue close button

y-lohse:


y-lohse:
[(3) [FIL-44] ETU je peux suivre l'upload de plusieurs fichiers DBD les stocker dans mon cloud](https://trello.com/c/RSrXnWi5/265-3-fil-44-etu-je-peux-suivre-lupload-de-plusieurs-fichiers-dbd-les-stocker-dans-mon-cloud)

src/ducks/upload/styles.styl
```diff
@@ -31,13 +31,17 @@
     margin           0
     padding          em(8px) em(16px) em(8px)
 
+    > .upload-queue-header-inner
```
GoOz: now I'm questioning the need for the children selector `>`. It did have a purpose when you used a `div` instead of a class but now… it just give more unnecessary weight to the selector.

y-lohse: You're right, i'm removing it.

PR #312, opened 2017-06-28
fix: file action menu icon

y-lohse:
The mime type classe names I altered the other day were also used in a different module so that one broke. As far as I can tell these are the only two places they were used.


y-lohse:
[[FIL] bug sur l'App mobile](https://trello.com/c/DOTuR8bz/415-fil-bug-sur-lapp-mobile)

PR #313, opened 2017-06-28
fix: upload to specific folders

y-lohse:
The folder's `id` attribute is now named `id`, not `_id`.

### DIFF
--- a/src/ducks/files/Toolbar.jsx
+++ b/src/ducks/files/Toolbar.jsx
@@ -63,7 +63,7 @@ const mapStateToProps = (state, ownProps) => ({
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   uploadFiles: (files, displayedFolder) => {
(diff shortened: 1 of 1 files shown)


src/ducks/files/Toolbar.jsx
```diff
@@ -63,7 +63,7 @@ const mapStateToProps = (state, ownProps) => ({
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   uploadFiles: (files, displayedFolder) => {
-    dispatch(addToUploadQueue(files, displayedFolder._id, file => uploadedFile(file), (loaded, quotas, conflicts, errors) => {
+    dispatch(addToUploadQueue(files, displayedFolder.id, file => uploadedFile(file), (loaded, quotas, conflicts, errors) => {
```
enguerran: That's a problem we will face very often: `_id` vs `id`...

PR #316, opened 2017-06-28
Fix/extension folder

GoOz:
So some points I'd like to discuss…

First of all, I removed the chunk `const dotIdx = filename.lastIndexOf('.') - 1 >>> 0` but couldn't figure out why it was there in the first place. 
@goldoraf it was yours, so maybe you can explain and eventually tell me if I should put it back or not.

Finally, I tried to write the function like this:
```jsx
export const splitFilename = file => {
  isDirectory(file)
  ? { filename: file.name, extension: '' }
  : {
    extension: file.name.slice(file.name.lastIndexOf('.') + 1),
    filename: file.name.slice(0, file.name.lastIndexOf('.') + 1)
  }
}
```
But somehow, I really don't get why, it won't work. I'm told that _splitFilename is not function… I just can't get it, if someone has a clue, be my guest and talk!

Thank you for your attention, fellows.

### DIFF
--- a/src/components/File.jsx
+++ b/src/components/File.jsx
@@ -12,16 +12,12 @@ import Spinner from 'cozy-ui/react/Spinner'
 
 import { getFolderUrl } from '../reducers'
 
-export const splitFilename = file => {
-  if (isDirectory(file)) {
-    return { filename: file.name, extension: '' }
-  } else {
-    return {
-      extension: file.name.slice(file.name.lastIndexOf('.') + 1),
-      filename: file.name.slice(0, file.name.lastIndexOf('.') + 1)
-    }
+export const splitFilename = file => isDirectory(file)
+  ? { filename: file.name, extension: '' }
+  : {
+    extension: file.name.slice(file.name.lastIndexOf('.') + 1),
+    filename: file.name.slice(0, file.name.lastIndexOf('.') + 1)
   }
-}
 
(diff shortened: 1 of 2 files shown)


GoOz:
[(0.5) [forum-4723] Ne pas griser l'extension pour les dossiers](https://trello.com/c/YjXvu3GI/394-05-forum-4723-ne-pas-griser-lextension-pour-les-dossiers)

enguerran:
@GoOz you have to be careful with arrow function:

```jsx
const sayHello = (people) => people.age > 30
  ? 'Hi ' + people.name
  : 'Yo ' + people.name

// is not the same as

const sayHello = (people) => {
  people.age > 30
    ? 'Hi ' + people.name
    : 'Yo ' + people.name

// as this function is equivalent to

function sayHello (people) {
  people.age > 30
    ? 'Hi ' + people.name
    : 'Yo ' + people.name
}

// instead of

function sayHello (people) {
  return people.age > 30
    ? 'Hi ' + people.name
    : 'Yo ' + people.name
}
```

Long story short: if an arrow function is only one return instruction, then you can write it without brackets and without `return` keyword, but if you use brackets, you need to write the `return` keyword.

GoOz:
@y-lohse fixed the fileActionMenu call to function
@enguerran fixed the arrow function, now it works. thanks for the tip.

PR #318, opened 2017-07-04
feat: add jpeg preview on filelist

GoOz:
So, I don't replace the mime type icon per say, I just add a new div with the preview image as background on top of that type icon.
Easier way to deal with it than replacing and in that case, when an error occurs, no preview is displayed and the icon type remains visible. Same thing for long loading time.

Although, it would be nice to have even smaller preview image from the stack. Even the small image is still way too big for what we're showing which means way too much bits are loaded, unused and unnecessary bits.

### DIFF
--- a/src/components/File.jsx
+++ b/src/components/File.jsx
@@ -145,7 +145,7 @@ class File extends Component {
     const { filename, extension } = splitFilename(attributes)
     return (
       <div className={classes}>
-        { attributes.mime === 'image/jpeg' && <Preview thumbnail={`${cozy.client._url}${attributes.links.small}`} /> }
+        { attributes.links && <Preview thumbnail={`${cozy.client._url}${attributes.links.small}`} /> }
         {isRenaming
           ? <RenameInput />
           : <div>
--- a/src/components/File.jsx
+++ b/src/components/File.jsx
@@ -1,3 +1,4 @@
+/* global cozy */
 import React, { Component } from 'react'
 import classNames from 'classnames'
 import filesize from 'filesize'
@@ -9,6 +10,7 @@ import { translate } from 'cozy-ui/react/I18n'
 import RenameInput from '../ducks/files/RenameInput'
 import { isDirectory } from '../ducks/files/files'
 import Spinner from 'cozy-ui/react/Spinner'
+import Preview from '../components/Preview'
 
 import { getFolderUrl } from '../reducers'
 
@@ -143,6 +145,7 @@ class File extends Component {
     const { filename, extension } = splitFilename(attributes)
     return (
       <div className={classes}>
+        { attributes.mime === 'image/jpeg' && <Preview thumbnail={`${cozy.client._url}${attributes.links.small}`} /> }
         {isRenaming
           ? <RenameInput />
           : <div>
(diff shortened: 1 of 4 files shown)


GoOz:
[(2) [FIL] Afficher la miniature de l'image en guise d'icône](https://trello.com/c/j2Hu592L/402-2-fil-afficher-la-miniature-de-limage-en-guise-dic%C3%B4ne)

src/components/File.jsx
```diff
@@ -143,6 +145,7 @@ class File extends Component {
     const { filename, extension } = splitFilename(attributes)
     return (
       <div className={classes}>
+        { attributes.mime === 'image/jpeg' && <Preview thumbnail={`${cozy.client._url}${attributes.links.small}`} /> }
```
y-lohse: Is this really the only mime type for which we have previews?

src/styles/table.styl
```diff
@@ -119,6 +120,17 @@
 
 mime-types('fil-file-', '../assets/icons/')
 
+.fil-file-preview
+    content ''
+    position absolute
+    left 0
+    top 50%
+    transform translateY(-50%)
```
y-lohse: I don't like this positioning very much, but I don't have a better idea either ¯\_(ツ)_/¯

src/components/File.jsx
GoOz: afaik yes. We discussed it while grooming, I asked about other image format like SVG and it was said that it won't be supported (so far)

src/styles/table.styl
GoOz: What bothers you ? the `transform` ? Because it's the best solution for centering elements in CSS except flexbox but it requires a heavier structure. So for single element centering, it's still the best method (and even more in this case) to this day.

src/components/File.jsx
GoOz: So I asked back devs and apparently png and gif are supported too. I'll take a look.

src/styles/table.styl
y-lohse: Yeah, the pushing it down with `top` then pushing it up again with `transform` feels hacky, but alright. How does it compare to something [like this](https://codepen.io/anon/pen/BZxgBJ)? The same but less verbose?

src/components/File.jsx
```diff
@@ -143,6 +145,7 @@ class File extends Component {
     const { filename, extension } = splitFilename(attributes)
     return (
       <div className={classes}>
+        { attributes.links && <Preview thumbnail={`${cozy.client._url}${attributes.links.small}`} /> }
```
enguerran: I understand we need the server's url here, but I think we should add a abstraction layer to avoid the strong coupling between the external lib and the component. May the thumbnail's url be loaded via a Props to the `<File />` dumb component?

enguerran: Or may the `<Preview />` component be connected to load the required information?

src/styles/table.styl
GoOz: Oh yeah this technique used to break on older browsers, I honestly don't remember which ones though. That's why I prefer the transform technique as I'm sure it works fine everywhere. But I guess I could give it a try.

PR #322, opened 2017-07-05
feat: a public page listing shared folder's files

enguerran:
I tried to reuse existing components.
I faced two things:

1. I don't know how `<LightFolderView />` is getting its props
1. I don't know how to let the user download files

So, I guess I'll update this very PR when I'll get some answers.

I want to believe

### DIFF
--- a/src/targets/public/main.jsx
+++ b/src/targets/public/main.jsx
@@ -6,7 +6,7 @@ import { Router, Route, Redirect, hashHistory } from 'react-router'
 import { I18n } from 'cozy-ui/react/I18n'
 
 import configureStore from '../..//store/configureStore'
-import Layout from '../../components/Layout'
+import PublicLayout from '../../components/PublicLayout'
 
 import LightFolderView from '../../components/LightFolderView'
 
@@ -59,7 +59,7 @@ function init () {
       <I18n lang={lang} dictRequire={(lang) => require(`../../locales/${lang}`)}>
         <Provider store={store}>
           <Router history={hashHistory}>
-            <Route component={Layout}>
+            <Route component={PublicLayout}>
               <Route path='files(/:folderId)' component={LightFolderView} />
             </Route>
             <Redirect from='/*' to={`files/${id}`} />
--- a/src/targets/public/main.jsx
+++ b/src/targets/public/main.jsx
@@ -1,0 +1,66 @@
+/* global cozy */
+import React from 'react'
+import { render } from 'react-dom'
+import { Provider } from 'react-redux'
+import { Router, Route, Redirect, hashHistory } from 'react-router'
+import { I18n } from 'cozy-ui/react/I18n'
+
+import configureStore from '../..//store/configureStore'
+import Layout from '../../components/Layout'
+
+import LightFolderView from '../../components/LightFolderView'
+
+document.addEventListener('DOMContentLoaded', init)
+
+const arrToObj = (obj = {}, varval = ['var', 'val']) => {
+  obj[varval[0]] = varval[1]
+  return obj
+}
+
+const getQueryParameter = () => window
+  .location
+  .search
+  .substring(1)
+  .split('&')
+  .map(varval => varval.split('='))
+  .reduce(arrToObj, {})
+
+function init () {
+  const lang = document.documentElement.getAttribute('lang') || 'en'
+  const root = document.querySelector('[role=application]')
+  const data = root.dataset
+  const { id, sharecode } = getQueryParameter()
+
+  if (data.cozyDomain) {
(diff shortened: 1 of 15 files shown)


goldoraf:
@enguerran `LightFolderView` is getting its props by its `FileExplorer` parent, which is a connected component (container). As you don't need a full-featured file list, you may want to write a new, specific container that dispatches an `openFolder` call on component mounting, simplifying the routing.

Regarding the download, only the download of a shared folder is required. You may need to write a new action creator for that, as we only have an action for the download of a selection of files right now.

enguerran:
I'll need https://github.com/cozy/cozy-ui/pull/154 to have a button in small devices' selectionbar.

enguerran:
@goldoraf @y-lohse 
Now this PR includes:

- a share button for selected file
- a public page for shared folder
- a public page for directdownload (aka for shared file)

I still have a problem with permission sets. If you are familiar with it, maybe you can give me a hint on that.

y-lohse:
On the *sharer* side, everything works as expected, all good.

On the sharing side, not so much. The single file download isn't working as you explained, but it also gets blocked by the popup blocker. Not sure if we can work around this :/

For the public folder page, the files are visible, but I can't select them. There's also no breadcrumb, and the main menu is there (so I can click on "Trash", but of course I can't actually go there). I don't think the menu should be there at all on the public page?

enguerran:
@y-lohse there is not so much details in our PO's story to handle sharing side. So I do the best with the minimum effort. For instance _"Ouverture "directe" d'un fichier partagé"_ is not very clear of what we want but a _direct_ download.
Speaking of shared folder, I understand we want the folder's content and we want recipients to be able to download or view files (even in subdirectory), but it is not clear which tools we allow for recipients.

Notes:

1. I forget to remove the trash folder...
1. I cannot add a breadcrumb as recipients don't have to know the absolute path to the shared folder, it is a lack of privacy for the user who shares the folder.

y-lohse:
I understand and agree, the card is a little vague. In terms of UI, I think we should do something like the public page for a photo album, so just the content, full screen, no menu.
And yes, the breadcrumb shouldn't show the full path, but it should display the folder's name and (if it work s out of the box) the subfolders once you start navigating in it, as mentioned in the trello card.

enguerran:

enguerran:
[(3) [FIL] Partage par lien d'un fichier](https://trello.com/c/boX2EmWH/359-3-fil-partage-par-lien-dun-fichier)

src/targets/public/main.jsx
```diff
@@ -0,0 +1,73 @@
...
+    cozy.client.init({
+      cozyURL: `//${data.cozyDomain}`,
+      token: sharecode
+    })
+  }
+
+  if (directdownload) {
+    console.log('direct download')
+    cozy.client.files.getDownloadLinkById(id).then(
+      link => window.open(link, '_SELF')
```
enguerran: @y-lohse with the `'_SELF'` argument, it opens in the same exact tab.

- [x] find a workaround for _«it also gets blocked by the popup blocker. Not sure if we can work around this :/»_

enguerran:
I added two commits and now, look at that



No Side Bar with Trash and a BreadCrumb (public folder is on the right).

y-lohse:
That's really cool! This is pretty much how I saw things. 
i still noticed a problem: on a larger screen (with no media query applied) the folder's content doesn't take 100% of the width, see [this screenshot](https://www.dropbox.com/s/ijv7ezmaxni786f/Capture%20d%27%C3%A9cran%202017-07-10%2010.00.31.png?dl=0). It's also still not possible to select any file or folder.

src/targets/public/main.jsx
```diff
@@ -0,0 +1,72 @@
...
+  return obj
+}
+
+const getQueryParameter = () => window
+  .location
+  .search
+  .substring(1)
+  .split('&')
+  .map(varval => varval.split('='))
+  .reduce(arrToObj, {})
```
y-lohse: I believe you could use [`URL.searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) here?

enguerran: We could refactor later, but as it is an _experimental technology_ I will not change that in this PR regarding there is this very code in the cozy-photos public page too.

enguerran:
I do manage the **share one file** permissions issue (it came from a bad url).
And I do not have idea for the **share a folder** feature's files permissions, as for now cozy-stack developers told me:

> c'est fort possible que l'on ne gère pas des droits sur les fichiers à l'intérieur d'un dossier pour cette route

PR #323, opened 2017-07-07
fix: extract file attributes after upload

y-lohse:
This fixes the issues with the selection right after an upload, because without the function call `_id` wasn't converted to `id`.


y-lohse:
[[FIL] Impossible de supprimer un fichier juste après sa création](https://trello.com/c/MvHu5t8J/421-fil-impossible-de-supprimer-un-fichier-juste-apr%C3%A8s-sa-cr%C3%A9ation)

PR #326, opened 2017-07-10
feat: download folder button

GoOz:
Added download button for displayed folder.

### DIFF
--- a/src/ducks/files/Toolbar.jsx
+++ b/src/ducks/files/Toolbar.jsx
@@ -4,6 +4,7 @@ import classNames from 'classnames'
 import React from 'react'
 import { connect } from 'react-redux'
 import { translate } from 'cozy-ui/react/I18n'
+import { ROOT_DIR_ID } from '../../constants/config'
 
 import UploadButton from '../../components/UploadButton'
 import ShareButton from '../../components/ShareButton'
@@ -13,7 +14,7 @@ import { alert } from '../../lib/confirm'
 import { alertShow } from 'cozy-ui/react/Alerter'
 
 import { addToUploadQueue } from '../upload'
-import { uploadedFile } from '../../actions'
+import { uploadedFile, downloadFiles } from '../../actions'
 import { ShareModal } from '../../ducks/sharing'
 
 const toggleShowShareModal = state => ({ ...state, showShareModal: !state.showShareModal })
@@ -28,7 +29,8 @@ class Toolbar extends React.Component {
   }
 
   render () {
-    const { t, disabled, displayedFolder, actions, onSelectItemsClick, uploadFiles } = this.props
+    const { t, disabled, displayedFolder, actions, onSelectItemsClick, uploadFiles, downloadAll } = this.props
+    const notRootfolder = displayedFolder && displayedFolder.id !== ROOT_DIR_ID
     return (
       <div className={styles['fil-toolbar-files']} role='toolbar'>
         <UploadButton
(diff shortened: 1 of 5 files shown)


src/ducks/files/Toolbar.jsx
```diff
@@ -29,6 +30,8 @@ class Toolbar extends React.Component {
 
   render () {
     const { t, disabled, displayedFolder, actions, onSelectItemsClick, uploadFiles } = this.props
+    const notRootfolder = displayedFolder !== undefined && displayedFolder !== null && displayedFolder.dir_id === ROOT_DIR_ID ? true : false
```
enguerran: You can write it that way:

```js
const notRootfolder = displayedFolder && displayedFolder.dir_id !== ROOT_DIR_ID
```

Be ware that here, you called the variable `notRootfolder` that is `true` if `displayedFilder.di_id` is the root folder id...

enguerran: That is more or less the message travis gave us:

![][(http://i.imgur.com/FgxMouX.png)

GoOz: actually ROOT_DIR_ID works oddly, its value is `io.cozy.files.root-dir`and If you're in the root folder the value of `displayedFolder.dir_id` attribute is an empty string `""`, if you're in a level 1 folder it's `io.cozy.files.root-dir` and if if you're in a level 3 or more folder it's an hashed id. 
So right now I got back to the test of `displayedFolder.path` value which is a simpler test to do.

src/ducks/files/Toolbar.jsx
```diff
@@ -29,6 +29,8 @@ class Toolbar extends React.Component {
 
   render () {
     const { t, disabled, displayedFolder, actions, onSelectItemsClick, uploadFiles } = this.props
+    const notRootfolder = displayedFolder && displayedFolder.path !== '/'
+    console.log(displayedFolder);
```
y-lohse: There's still a `console.log` here ⚠️
And so it's not possible to use the root folder id to do this test?

src/ducks/files/Toolbar.jsx
enguerran: @GoOz `dir_id` is the id of the parent folder.

src/ducks/files/Toolbar.jsx
GoOz: yeah I know for the log, i'll remove it next commit.
Yeah I messed up, I compared the `ROOT_DIR_ID` with `displayedFolder.dir_id` and it wouldn't work (obviously), but it works way better when i compare with `displayedFolder.id`
Commit incoming

src/ducks/files/Toolbar.jsx
```diff
@@ -107,7 +118,8 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
 
       return action
     }))
-  }
+  },
+  downloadAll: folder => dispatch(downloadSelection(folder))
```
enguerran: We should rename `downloadSelection` to express the real behavior.

src/locales/en.json
```diff
@@ -23,6 +23,7 @@
     "menu_upload": "Upload files",
     "menu_new_folder": "New folder",
     "menu_select": "Select items",
+    "menu_download_all": "Download folder",
```
enguerran: The key and the value are not saying the same. It will be confusing later if it is not yet.

src/ducks/files/Toolbar.jsx
GoOz: oh right I forgot about that

PR #329, opened 2017-07-13
fix: ui in sharing public page

enguerran:
Here it is:


enguerran:
I did the minimum to get it right as in Minimum Viable Product 😉

enguerran:

PR #332, opened 2017-07-19
feat: warn shared file no longer available

enguerran:


src/targets/public/main.jsx
```diff
@@ -41,7 +42,21 @@ function init () {
...
       link => window.open(`${cozy.client._url}${link}`, '_SELF')
-    )
+    ).catch(e => {
+      cozy.bar.init({
+        appName: data.cozyAppName,
+        appEditor: data.cozyAppEditor,
+        iconPath: data.cozyIconPath,
+        lang: data.cozyLocale,
+        replaceTitleOnMobile: true
+      })
```
y-lohse: Is there a way to avoid duplicating the cozy bar init?

enguerran: I don't see how.
```
IF A and ¬C THEN initBar
IF ¬A THEN initBar
```
Even more so this is an only one-lined instruction .

y-lohse: I was mostly thinking of moving it to [l.41](https://github.com/enguerran/cozy-drive/blob/33916c81bb6066f2c2b72d5c173fcc5ba3d88ef1/src/targets/public/main.jsx#L41), when the client is initialized too — since we'll want the bar regardless of what happens?

enguerran: We don't.
For shared files, we don't need the cozy-bar (at least not as long as we don't use the file viewer for shared files).
We may still init the cozy-bar before, but it will take time.

I think what we should do is to use a static html/css page for such error. But in that case, we won't have i18n, at least as long as we don't do something without JavaScript.
Maybe the backend could send us such a page.

y-lohse: Ah I see, that's what I was missing. Thanks for clearing it up.

PR #335, opened 2017-07-24
fix: add link in error message

y-lohse:
AFAIK, it was either `dangerouslySetInnerHTML`, or installing `react-markdown` which seemed like a lot of hassle just for a link. All error messages are generated from the translations so this should be safe.


y-lohse:
[(0,5) [FIL] Améliorer le message d'erreur spécifique lorsqu'on essaye de se connecter à un Cozy v2 via l'app mobile v3](https://trello.com/c/KInItutk/422-05-fil-am%C3%A9liorer-le-message-derreur-sp%C3%A9cifique-lorsquon-essaye-de-se-connecter-%C3%A0-un-cozy-v2-via-lapp-mobile-v3)

PR #340, opened 2017-07-28
feat: display a menu more in mobile view

enguerran:

PR #341, opened 2017-07-31
fix: toolbar display

enguerran:
A previous commit (see PR https://github.com/cozy/cozy-drive/pull/340) was hiding the menu more in desktop view (large screen).
Here is the fix.


enguerran:

PR #342, opened 2017-08-01
fix: defensive code guarding against ill-formatted contacts

y-lohse:
Existing contacts with a wrong format were messing up the import.


y-lohse:
[(5) [FIL] Importer les contacts du téléphone Android et iOS](https://trello.com/c/akcF4891/350-5-fil-importer-les-contacts-du-t%C3%A9l%C3%A9phone-android-et-ios)

y-lohse:
I added a commit that prevents problems the other way around: contacts from the device may have fields set to `null`, which was messing things up.

PR #344, opened 2017-08-03
feat: Request contacts permissions during onboarding

y-lohse:


y-lohse:
[(1) [FIL]📱 Améliorer l'expérience de la fonctionnalité de sauvegarde des contacts](https://trello.com/c/MtfXeclB/450-1-fil%F0%9F%93%B1-am%C3%A9liorer-lexp%C3%A9rience-de-la-fonctionnalit%C3%A9-de-sauvegarde-des-contacts)

src/targets/mobile/containers/onboarding/Files.jsx
```diff
@@ -4,7 +4,7 @@ import OnBoarding from '../../components/OnBoarding'
 
 export const Files = ({ onActivate }) =>
 (
-  <OnBoarding onActivate={onActivate} stepName='files' currentStep={1} totalSteps={3} />
+  <OnBoarding onActivate={onActivate} stepName='files' currentStep={1} totalSteps={4} />
```
enguerran: I know that this magic number did not appear in your PR, but can we find a way to avoid this?
I guess it is used for the breadcrumb.

y-lohse: Probably yeah, I'll take a look.

enguerran:
IDK where the magic comes from, I just read it fast, but it is **exactly** what I meant.
Just upgrade jest snapshots if it's ok.

PR #345, opened 2017-08-03
feat: rating dialog

y-lohse:
The user gets prompted after starting the apps a few times to give his/her feedback.

I suggest starting the review with `mobile/src/containers/RatingModal.jsx`, it's the component that holds the main logic.

### DIFF
--- a/src/targets/mobile/containers/RatingModal.jsx
+++ b/src/targets/mobile/containers/RatingModal.jsx
@@ -1,0 +1,248 @@
+/* global cozy AppRate */
+import React, { Component } from 'react'
+import Modal, { ModalContent } from 'cozy-ui/react/Modal'
+import { connect } from 'react-redux'
+import withPersistentState from '../lib/withPersistentState'
+import { softwareID, softwareName } from '../lib/cozy-helper'
+
+import styles from '../styles/rating'
+
+const SCREEN_ENJOY = 'SCREEN_ENJOY'
+const SCREEN_FEEDBACK = 'SCREEN_FEEDBACK'
+const ALERT_RATING = 'ALERT_RATING'
+
+const BUTTON_INDEX_RATE = 1
+const BUTTON_INDEX_LATER = 2
+
+const FEEDBACK_EMAIL = 'support@cozycloud.cc'
+const PROMPT_AFTER_BOOTS = 5
+
+// RatingModal is the base component
+class RatingModal extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      screen: SCREEN_ENJOY
+    }
+  }
+
+  async onUserReply (enjoyed) {
+    if (enjoyed) {
+      try {
+        const { t } = this.context
+        const buttonIndex = await promptRating({
+          title: t('mobile.rating.rate.title'),
+          message: '',
+          yes: t('mobile.rating.rate.yes'),
+          no: t('mobile.rating.rate.no'),
+          later: t('mobile.rating.rate.later'),
+          softwareID,
+          softwareName
+        })
+
+        if (buttonIndex === BUTTON_INDEX_LATER) {
+          this.props.alert('mobile.rating.alert.later')
+          this.props.showLater()
+        } else if (buttonIndex === BUTTON_INDEX_RATE) {
+          this.props.alert('mobile.rating.alert.rated')
+          this.props.dontShowAgain()
+        } else {
+          this.props.alert('mobile.rating.alert.declined')
+          this.props.dontShowAgain()
+        }
+      } catch (e) {
+        this.props.dontShowAgain()
+        throw e
+      }
+    } else {
+      this.setState({ screen: SCREEN_FEEDBACK })
+    }
+  }
+
+  onCloseFeedback (gaveFeedback) {
+    if (gaveFeedback) this.props.alert('mobile.rating.alert.feedback')
+    this.props.dontShowAgain()
+  }
+
+  render () {
+    return this.state.screen === SCREEN_ENJOY ? <EnjoyCozy onReply={this.onUserReply.bind(this)} /> : <FeeedbackForm onClose={this.onCloseFeedback.bind(this)} />
+  }
+}
+
+// sub-components
+class EnjoyCozy extends Component {
+  render () {
+    const { onReply } = this.props
+    const { t } = this.context
+    return (
+      <Modal
+        title={t('mobile.rating.enjoy.title')}
+        withCross={false}
+      >
+        <ModalContent>
+          <div className={styles['button-block']}>
+            <button className='coz-btn coz-btn--secondary' onClick={() => onReply(false)}>
+              { t('mobile.rating.enjoy.no') }
+            </button>
+            <button className='coz-btn coz-btn--regular' onClick={() => onReply(true)}>
+              { t('mobile.rating.enjoy.yes') }
+            </button>
+          </div>
+        </ModalContent>
+      </Modal>
+    )
+  }
+}
+
+class FeeedbackForm extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      sending: false
+    }
+  }
+
+  async onSubmit (e) {
+    e.preventDefault()
+
+    const { t } = this.context
+    const mailData = {
+      mode: 'from',
+      to: [
+        { name: 'Support', email: FEEDBACK_EMAIL }
+      ],
+      subject: t('mobile.rating.email.subject'),
+      parts: [
+        { type: 'text/plain', body: this.textarea.value.toString() }
+      ]
+    }
+
+    try {
+      this.setState({ sending: true })
+      await cozy.client.jobs.create('sendmail', mailData)
+    } catch (e) {
+      // Sending the email failed; this can happen because of insuficient permissions for example. Not a big deal either in this context.
+    }
+
+    this.setState({ sending: false })
+    this.props.onClose(true)
+  }
+
+  registerElement (element) {
+    this.textarea = element
+  }
+
+  render () {
+    const { onClose } = this.props
+    const { sending } = this.state
+    const { t } = this.context
+    return (
+      <Modal
+        title={t('mobile.rating.feedback.title')}
+        withCross={false}
+      >
+        <ModalContent className={styles['feedback-form']}>
+          <form className={'coz-form'} onSubmit={this.onSubmit.bind(this)}>
+            <textarea className={styles['feedback-text']} ref={this.registerElement.bind(this)}>
+              { t('mobile.rating.email.placeholder') }
+            </textarea>
+            <div className={styles['button-block']}>
+              <button className='coz-btn coz-btn--secondary' type='button' onClick={() => onClose(false)}>
+                { t('mobile.rating.feedback.no') }
+              </button>
+              <button className='coz-btn coz-btn--regular' type='submit' aria-busy={sending}>
+                { t('mobile.rating.feedback.yes') }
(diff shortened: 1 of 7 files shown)


y-lohse:
[(3) [FIL] Demander aux utilisateurs de l'application mobile de noter notre application](https://trello.com/c/TGAfDGQB/424-3-fil-demander-aux-utilisateurs-de-lapplication-mobile-de-noter-notre-application)

y-lohse:
Yeah so I'm using redux for the information that needs to be persisted (number of application boots, if the user has been prompted and for what version). 
We could persist this information outside of redux, but since there's already code in place for this sort of things, I think it makes sense to use it here as well. What do you think?

src/targets/mobile/main.jsx
```diff
@@ -69,6 +70,7 @@ const renderAppWithPersistedState = persistedState => {
   document.addEventListener('deviceready', () => {
     pingOnceADayWithState()
     store.dispatch(backupImages())
+    store.dispatch(incrementBootCount())
```
enguerran: If you need to access the global state only to persist the value with such kind of instruction, ,maybe you can add that behavior in the stateful component. Or in a HOC responsible to persist the state of a component:

```jsx
const withPersistency = WrappedComponent => class extends React.Component {
  constructor (props) {
    super(props)
    this.state = undefined
  }

  componentWillMount () {
    loadStateFormLocalStorage().then(state => this.setState(() => state))
  }

  componentWillUnmout () {
    saveStateToLocalStorage(this.state)
  }

  render () {
    <WrappedComponent {...this.state} />
  }
}
```

or something similar.

src/targets/mobile/actions/rating.js
```diff
@@ -0,0 +1,28 @@
+/* global __APP_VERSION__ */
```
enguerran: If you use redux, you should create a single file with everything for the redux state:

```js
// constant may be used only in this file
const ADD = 'my-app/users/ADD';

// default export is the reducer
export default function reducer(state = {}, action = {}) {
  switch (action.type) {
    ADD: return {...state, [action.payload.id]: action.payload}
    default: return state;
  }
}

// Action Creators
export function addUser(payload) {
  return { type: ADD, payload };
}

// Selectors
export function getUsers (state) {
 return state
}

export function getUser (state, id) {
  return state[id]
}
```

src/targets/mobile/containers/RatingModal.jsx
```diff
@@ -0,0 +1,178 @@
...
+      parts: [
+        { type: 'text/plain', body: this.textarea.value.toString() }
+      ]
+    }
+
+    try {
+      this.setState({ sending: true })
+      await cozy.client.jobs.create('sendmail', mailData)
+    } catch (e) {
+      // Sending the email failed; this can happen because of insuficient permissions for example. Not a big deal either in this context.
```
y-lohse: The state gets updated just below. We've decided to silently ignore this error as addressing it requires making the user sign in again, which isn't worth it for this feature.

src/targets/mobile/actions/rating.js
y-lohse: This is not how the code for the mobile app is structured. 
It's either

```
- actions
  - feature1
  - feature2
- reducers
  - feature1
  - feature2
```

or

```
- feature1
  - redux
  - other files
- feature2
  - redux
  - other files
```

Currently the code related to the mobile app follows the first pattern. We can change this, but that's not what this PR is about.

enguerran: I talked about baby steps to reach http://slides.com/goldoraf/preact-redux#/1

src/targets/mobile/containers/RatingModal.jsx
```diff
@@ -0,0 +1,236 @@
...
+  render () {
+    const { screen } = this.state
+
+    return screen === SCREEN_ENJOY ? <EnjoyCozy onReply={this.onUserReply.bind(this)} /> : <FeeedbackForm onClose={this.onCloseFeedback.bind(this)} />
+  }
+}
+
+// sub-components
+class EnjoyCozy extends Component {
+  render ({ onReply }) {
```
enguerran: Do we want to stay React compatible? If yes, we cannot use parameter in `render()`.

src/targets/mobile/containers/RatingModal.jsx
```diff
@@ -0,0 +1,236 @@
...
+        title={t('mobile.rating.feedback.title')}
+        withCross={false}
+      >
+        <ModalContent className={styles['feedback-form']}>
+          <form className={'coz-form'} onSubmit={this.onSubmit.bind(this)}>
+            <textarea className={styles['feedback-text']} ref={this.registerElement.bind(this)}>
+              { t('mobile.rating.email.placeholder') }
+            </textarea>
+            <div className={styles['button-block']}>
+              <button className='coz-btn coz-btn--secondary' type='button' onClick={() => onClose(false)}>
```
enguerran: Is there a reason you use this pattern `onClick={() => onClose(false)}` and not the same as above `onClick={this.props.onClose.bind(this, false)}`?

enguerran: I prefer the version with arrow function even if it creates a new bind on each render. See https://medium.freecodecamp.org/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56 for more details.

src/targets/mobile/containers/RatingModal.jsx
```diff
@@ -0,0 +1,236 @@
...
+              </button>
+            </div>
+          </form>
+        </ModalContent>
+      </Modal>)
+  }
+}
+
+// promptRating is not a component because the native UI is used
+const promptRating = ({ title, message, yes, no, later, softwareName, softwareID }) => {
```
enguerran: Do you think it looks a way more better, and if there was no AppRate plugin in it, this function could be easily tested because it is given everything it needs.

src/targets/mobile/containers/RatingModal.jsx
```diff
@@ -0,0 +1,236 @@
...
+    super(props)
+    this.state = {
+      bootCount: 0,
+      prompted: false
+    }
+  }
+
+  componentStateRestored () {
+    this.setState(prevState => ({
+      bootCount: ++prevState.bootCount
```
enguerran: I cannot understand the name `componentStateRestored` and the implementation that do not take any restored state, but increment the current state.

src/targets/mobile/lib/withPersistentState.js
```diff
@@ -0,0 +1,20 @@
...
+
+// The withPersistentState HOC automatically saves the wrappped components state and restores it when the component is mounted again.
+// Since the saved state will not be available immediately when the wrapped component is mounted, a `componentStateRestored` hook will be called once the state is restored
+export const withPersistentState = (WrappedComponent, persistenceKey) => class extends WrappedComponent {
+  async componentWillMount () {
+    if (super.componentWillMount) super.componentWillMount()
+    const persistedState = await localforage.getItem(persistenceKey)
+    this.setState(prevState => (persistedState || prevState), () => {
+      if (super.componentStateRestored) super.componentStateRestored()
+    })
```
enguerran: I didn't know `setState` could accept a second parameter. I just read:

> This makes reading this.state right after calling setState() a potential pitfall. Instead, use componentDidUpdate or a setState callback (setState(updater, callback)), either of which are guaranteed to fire after the update has been applied.

enguerran: What is `super.componentStateRestored`? And what is for?

enguerran: Further, [documentation](https://facebook.github.io/react/docs/react-component.html#setstate) says:

> The second parameter to setState() is an optional callback function that will be executed once setState is completed and the component is re-rendered. Generally we recommend using componentDidUpdate() for such logic instead.

enguerran: Ok, I saw this after my interrogation:

> Since the saved state will not be available immediately when the wrapped component is mounted, a `componentStateRestored` hook will be called once the state is restored

If I understand well, you suggest to implement such a function to let the component do something once the state is set.
It appears that React already have such a function in the component's lifecycle called `componentDidUpdate` (which is called after the `render()`) or `componentWillUpdate` (which is called before the `render()`).
Do you thing this `componentStateRestored` is mandatory and in which usecase?

More details in the documentation: https://facebook.github.io/react/docs/react-component.html#updating

enguerran:
Could you `git rebase -i origin/master` to squash unnecessary commits?

y-lohse:
@enguerran I'm afraid your previous review came a bit early, I wasn't quite finished. I am now and there were other small changes, FYI.

src/targets/mobile/lib/withPersistentState.js
y-lohse: I believe it is *generally* recommended to use `componentDidUpdate` because *generally*, you want to do things regardless of why and how the component updated.
However in this case, we want to call the `componentStateRestored` hook, and we want to call it only once, specifically after the state has been restored from local storage. If we used a generic `componentDidUpdate` instead, we would need to add some detection code to make sure the component updated because of the restoration we just did and not something else, which is going to be messy. I'm fairly sure this is a valid use case for the second parameter.

y-lohse: This is a valid concern and I actually tried this at first.

> If I understand well, you suggest to implement such a function to let the component do something once the state is set.

Specifically once it's set by the state restoration, yes. The use case here being that we want to increment a counter once the state has been restored, and only then.

While you *can* detect the state restoration in the wrapped component using `componentWillUpdate` or `componentDidUpdate`, it requires adding code to detect that the changes are coming from the restoration, and not a `setState` somewhere else. It also makes the whole code quite hairy because you end up with multiple `componentWillUpdate` hooks that feed into each other, to a point where it wasn't working at all. 

I'm sure it's possible, but:
- the extra hook provides a clean and easy way to achieve what we want
- it's completely optional

So yes, I think this is the best approach.

src/targets/mobile/containers/RatingModal.jsx
y-lohse: I agree :)

src/targets/mobile/containers/RatingModal.jsx
y-lohse: Fair enough :)

y-lohse: The reason is parameters :)
`bind()` is nice because you don't have to pass parameters along, like you would with an arrow function. So I'm only using arrow functions when the context doesn't matter and there are parameters to be set explicitly.

enguerran: You can pass parameters through a bound function. `bind(thisArg, ...args)` accepts `thisArg` as the bound context and `...args` as the parameters for the bound function.

src/targets/mobile/lib/withPersistentState.js
enguerran: good point

enguerran: ok

src/targets/mobile/containers/RatingModal.jsx
y-lohse: Oh yes I know you *can*, but the nice thing is you don't have to. 
`onClick={this.handler.bind(this)}` doesn't care how many or what parameters `handler` should receive. `onClick={(e) => this.handler(e)}` forces you to be explicit about `e` in the template, and again in the `handler` definition.

PR #346, opened 2017-08-03
Upgrade our configuration

enguerran:
- upgrade to webpack 3
- move `mobile/src` to `src/targets/mobile`

Now we can keep on working :tada:


config/webpack.target.mobile.js
```diff
@@ -1,27 +1,27 @@
...
-const webpack = require('webpack')
+const { DefinePlugin, ProvidePlugin } = require('webpack')
 
 const {production} = require('./webpack.vars')
 const pkg = require(path.resolve(__dirname, '../package.json'))
 
 module.exports = {
   entry: {
-    app: path.resolve(__dirname, '../src/targets/mobile/main')
+    app: [path.resolve(__dirname, '../src/targets/mobile/main')]
```
CPatchane: Why do you need to put it in an array here?

enguerran: IDK, I copy/paste from yours: https://github.com/CPatchane/cozy-hello-world/blob/master/config/webpack.target.mobile.js#L12

enguerran:
I really rather to put test files in collocation with src files.

PR #353, opened 2017-08-10
fix: support for application/* mime types

y-lohse:


y-lohse:
[(0.5) [FIL] Certains types de fichiers n'ont pas d'icônes spécifiques dans la liste](https://trello.com/c/h33IXgL5/444-05-fil-certains-types-de-fichiers-nont-pas-dic%C3%B4nes-sp%C3%A9cifiques-dans-la-liste)

src/components/File.jsx
```diff
@@ -25,7 +25,21 @@ export const getClassFromMime = (attrs) => {
...
+  } else if (type === 'application') {
+    // these mime types don't always match the exact class names
+    if (/pdf/.test(subtype)) className = 'pdf'
+    else if (/zip/.test(subtype)) className = 'zip'
+    else if (/spreadsheet/.test(subtype) || /excel/.test(subtype)) className = 'spreadsheet'
+    else if (/presentation/.test(subtype) || /powerpoint/.test(subtype)) className = 'slide'
+  }
+
+  return styles['fil-file-' + className]
 }
```
enguerran: As you use [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) to check the `subtype`, you could simplify the code by using [`String.prototype.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) with a correct RegExp:

```js
function getClass(subtype) {
   const result = subtype.match(/pdf|zip|spreadsheet|presentation)
   return result ? result[0] : ''
}
```

On the other hand, you need some particular rules for `application/spreadsheet` and `application/excel`. I guess a `switch()/case` could be a better readable solution:

```js
switch (subtype) {
  case 'spreadsheet':
  case 'excel':
    return 'spreadsheet'
  case 'presentation':
  case 'powerpoint':
    return 'slide'
  default:
    return subtype || 'files'
}
```

Or you could get both of this two worlds with a correlation table:

```js
const correlationTabe = {
  zip: 'zip',
  pdf: 'pdf',
  spreadhseet: 'spreadsheet',
  excel: 'spreadsheet',
  presentation: 'silde',
  powerpoint: 'slide'
}
const reg = new RegExp(Object.keys(correlationTable).join('|'))
const result = subtype.match(reg)
return result ? correlationTable[result[0]] : 'files'
```

Working example [here](https://jsbin.com/peviwow/edit?js,console)

y-lohse: Cool, I like the correlation table approach. I'll update the PR.

src/components/File.jsx
```diff
@@ -25,7 +25,30 @@ export const getClassFromMime = (attrs) => {
...
+  let className = 'files'
+
+  if (styles['fil-file-' + type]) {
+    className = type
+  } else if (type === 'application') {
+    // for mime types with an `application` type, we need to consider the subtype
+    const correlationTable = {
+      zip: 'zip',
+      pdf: 'pdf',
+      spreadhseet: 'spreadsheet',
```
ptbrowne: typo : `spreadhseet`

src/components/File.jsx
```diff
@@ -25,7 +25,30 @@ export const getClassFromMime = (attrs) => {
...
+      excel: 'spreadsheet',
+      presentation: 'silde',
```
ptbrowne: typo : `silde`

enguerran:
@ptbrowne do you mind creating a Pull Request to fix these typos?

PR #365, opened 2017-08-23
Recipient default avatar

GoOz:
Added default avatar on recipient with initial of the name and a stable color using [ColorHash](https://github.com/gweax/ColorHash)

Used Recipient component in 'who has access' & 'autocomplete' components.

### DIFF
--- a/src/photos/ducks/sharing/colorhash.js
+++ b/src/photos/ducks/sharing/colorhash.js
@@ -5,7 +5,7 @@ const ColorHash = (str) => {
     ]
   }
 
-  function hashCode (str) {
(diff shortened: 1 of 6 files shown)


GoOz:
[(2) [GLOB] Afficher un avatar par défaut pour les contacts dans l'auto-complétion et le "who has access" dans Drive et Photos](https://trello.com/c/WP2VDgK0/376-2-glob-afficher-un-avatar-par-d%C3%A9faut-pour-les-contacts-dans-lauto-compl%C3%A9tion-et-le-who-has-access-dans-drive-et-photos)

PR #371, opened 2017-08-29
fix: fix chrome mobile on windows view and perhaps may more bugs but …

GoOz:
…future will tell

I'm not 100% sure what this option`disableHeight` does or doesn't do but it stops calculating the height of the list's container and so it fixes the visual bug on Chrome mobile on Windows at least. 

Could be the answer to our 'little' layout issue on mobile as well, we'll see.


GoOz:
[[FIL] 🐛 Problème de scroll sur safari mobile](https://trello.com/c/hO0d4QBy/463-fil-%F0%9F%90%9B-probl%C3%A8me-de-scroll-sur-safari-mobile)

PR #383, opened 2017-09-06
refactor: cozy-redux-client

enguerran:
@goldoraf what do you think? Isn't it cleaner?

### DIFF
--- a/src/lib/redux-cozy-client/CozyAPI.js
+++ b/src/lib/redux-cozy-client/CozyAPI.js
@@ -15,13 +15,21 @@ export default class CozyAPI {
     try {
       // WARN: if no document of this doctype exist, this route will return a 404,
       // so we need to try/catch and return an empty response object in case of a 404
-      const resp = await cozy.client.fetchJSON('GET', `/data/${doctype}/_all_docs?include_docs=true`)
+      const resp = await cozy.client.fetchJSON(
+        'GET',
+        `/data/${doctype}/_all_docs?include_docs=true`
+      )
       // WARN: the JSON response from the stack is not homogenous with other routes (offset? rows? total_rows?)
       // see https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#list-all-the-documents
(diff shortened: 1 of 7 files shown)


enguerran:
Do you think I could update coverage?

screenshot at 13-20-43

enguerran:
I will insist on incorrect usage of `async` keyword because babel generates [a lot of extra code to handle it](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=GYVwdgxgLglg9mABBANgQwM4ZhRAKASkQG8AoRRAJwFMoRKkByOAa0YG5SBfU0zAT0iJQkWAkQAjNABNUmbLkIlyVWvSZg4URKw7cgA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=es2015%2Cstage-0&prettier=false&targets=&version=6.26.0):

```js
function classic () {
  return 'ok';
}

async function badclassic () {
  return 'not ok';
}
```

becomes

```js
'use strict';

var badclassic = function () {
  var _ref = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee() {
    return regeneratorRuntime.wrap(function _callee$(_context) {
      while (1) {
        switch (_context.prev = _context.next) {
          case 0:
            return _context.abrupt('return', 'not ok');

          case 1:
          case 'end':
            return _context.stop();
        }
      }
    }, _callee, this);
  }));

  return function badclassic() {
    return _ref.apply(this, arguments);
  };
}();

function _asyncToGenerator(fn) { return function () { var gen = fn.apply(this, arguments); return new Promise(function (resolve, reject) { function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { return Promise.resolve(value).then(function (value) { step("next", value); }, function (err) { step("throw", err); }); } } return step("next"); }); }; }

function classic() {
  return 'ok';
}
```

enguerran:
I run `prettier` on the files to fit the ongoing #402

PR #386, opened 2017-09-07
Fix stuff

y-lohse:
You should probably review the commits individually.

This PR fixes several things:

- removes trashed files from the "recent" view
- fixes opening files in the recent view, incidentally making the related functions simpler
- fixes the trash view by adding the `isAvailableOffline` prop
- displays the files last update timestamp instead of the creation date, which is what the column was supposed to be anyway


y-lohse:
[[DRI] 🐛  La corbeille affiche une erreur à l'ouverture](https://trello.com/c/2TZYWBIZ/503-dri-%F0%9F%90%9B-la-corbeille-affiche-une-erreur-%C3%A0-louverture)

PR #394, opened 2017-09-14
feat: provide suggestions for the search bar

y-lohse:
### 1. Providing suggestions for the search bar

When the search bar asks for suggestions from Drive, the app will load the list of files from the server and try to find matches in it. This is all done on the client side and has obviously a massive downside: if there are many files on the cozy, we'll spend a lot of time and ressources fetching them all (the actual memory footprint, event with many files, isn't very high).
[Here's the code for that](https://github.com/cozy/cozy-drive/pull/394/files#diff-ab8f134c72c2b24476f993fcb9f5c5e2R5).

We could eventually get rid of that by having pouchdb on the web version as well. As far as I can tell, it is not possible to [let the stack run these queries at the moment](http://docs.couchdb.org/en/2.0.0/api/database/find.html#condition-operators).

The actual search is done by a library called [FuzzyPathSearch](https://github.com/cozy/cozy-drive/pull/394/files#diff-c769d0f032f544fe19d5044035d6295fR4).

### 2. Refactoring the [IntentHandler](https://github.com/cozy/cozy-drive/pull/394/files#diff-63c57d180f87d151796dc5b1a366b69dL7)

There's now a duck for services / intent handling. It has a [main component](https://github.com/cozy/cozy-drive/pull/394/files#diff-ccfb11fef287f9616f7aa1a57b1209a5R8) which loads the intent data, and dispatches the actual work to specialized sub-components ([here](https://github.com/cozy/cozy-drive/pull/394/files#diff-e4988eb68039bb85e671a7853e0ca568R6) and [here](https://github.com/cozy/cozy-drive/pull/394/files#diff-436a0fe8ab49647a8ebf075c6eeb7b89R4))


y-lohse:
[(3) [DRI] On peut chercher des fichiers et des dossiers grâce à une chaîne de caractères depuis n'importe quelle application sur desktop](https://trello.com/c/En9oAONQ/509-3-dri-on-peut-chercher-des-fichiers-et-des-dossiers-gr%C3%A2ce-%C3%A0-une-cha%C3%AEne-de-caract%C3%A8res-depuis-nimporte-quelle-application-sur-desk)

src/drive/ducks/services/lib.js
```diff
@@ -0,0 +1,6 @@
+/* global cozy */
+
+export const getFileDownloadUrl = async (id) => {
```
goldoraf: I think this helper already exists in the `files` duck, it'd be better to not have this one here

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -0,0 +1,79 @@
+/* global cozy */
+import React from 'react'
+import FuzzyPathSearch from '../FuzzyPathSearch'
+
+class SuggestionProvider extends React.Component {
```
goldoraf: Without a render method, I don't see the point of inheriting from React.Component here...

src/drive/ducks/services/components/URLGetter.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { getFileDownloadUrl } from '../lib'
+
+class URLGetter extends React.Component {
```
goldoraf: Same remark here

src/drive/ducks/services/components/SuggestionProvider.jsx
y-lohse: I think it's necessary because we're using other parts of the lifecycle, specifically `componentDidMount`. So I think a regular class wouldn't work, and a functional component will be called too often. But maybe I'm missing something?

src/drive/ducks/services/lib.js
y-lohse: Aah good catch, there's *almost* the same thing in the general actions file. I'll use that one instead.

src/drive/ducks/services/components/SuggestionProvider.jsx
goldoraf: I think the code in the `componentDidMount` could be put in a regular constructor of a regular class, but maybe it's me that's missing something :p

y-lohse: It could, but then you can't use that class as a component [here](https://github.com/cozy/cozy-drive/pull/394/files#diff-ccfb11fef287f9616f7aa1a57b1209a5R55) 🤷‍♂️

PR #401, opened 2017-09-19
Multi-contact sharing

goldoraf:


src/lib/redux-cozy-client/CozyClient.js
```diff
@@ -54,14 +54,18 @@ export default class CozyClient {
...
+  async createSharing (permissions, contactIds, sharingType, description) {
+    return this.api.createSharing(permissions, contactIds, sharingType, description)
   }
 
   async revokeSharing (sharingId) {
     return this.api.revokeSharing(sharingId)
   }
 
+  revokeSharingForClient (sharingId, clientId) {
+    return this.api.revokeSharingForClient(sharingId, clientId)
```
enguerran: What is the difference between `revokeSharing` and `revokeSharingForClient`?

src/lib/redux-cozy-client/slices/sharings.js
```diff
@@ -24,11 +29,18 @@ const documents = (state = [], action) => {
     case REVOKE_SHARING:
       const idx = state.findIndex(s => s.attributes.sharing_id === action.sharingId)
       if (idx === -1) return state
-      return [
-        ...state.slice(0, idx),
-        {...state[idx], attributes: {...state[idx].attributes, revoked: true}},
-        ...state.slice(idx + 1)
-      ]
+      const sharing = state[idx]
+      const loneRecipient = sharing.attributes.recipients.length === 1
```
enguerran: In this `REVOKE_SHARING` reducer, we check if there is only one recipient or a list of recipients...

src/lib/redux-cozy-client/slices/sharings.js
```diff
@@ -105,24 +117,29 @@ export const fetchSharings = (doctype, id = null, options = {}) => ({
...
     sharingId: sharing.attributes.sharing_id,
-    // TODO: right now, we create one sharing for each recipient, so we can just
-    // delete the sharing, but when we'll have many recipients for one sharing,
-    // we'll need to use another route:
-    // https://github.com/cozy/cozy-stack/blob/master/docs/sharing.md#delete-sharingssharing-idrecipientclient-id
-    promise: client => client.revokeSharing(sharing.attributes.sharing_id)
+    recipientId: recipient._id,
+    promise: client => loneRecipient
+      ? client.revokeSharing(sharing.attributes.sharing_id)
+      : client.revokeSharingForClient(
```
enguerran: In this `unshare`, we check if there is only one recipient or a list of recipients. As the case may be, we call `revoleSharing` or `revokeSharingForClient`. (see other comments)

I find this very confusing, maybe a naming problem, maybe a factoring problem...

src/lib/redux-cozy-client/slices/sharings.js
```diff
@@ -105,24 +117,29 @@ export const fetchSharings = (doctype, id = null, options = {}) => ({
...
+        sharing.attributes.sharing_id,
+        sharing.attributes.recipients.find(r => r.recipient.id === recipient._id).Client.client_id
```
enguerran: What is this `Client` property with a capital _«C»_?

src/sharing/components/ShareAutocomplete.jsx
```diff
@@ -23,39 +23,70 @@ export default class ShareAutocomplete extends Component {
...
 
-  onSuggestionsClearRequested () {
-    this.setState(state => ({
-      ...state,
-      suggestions: []
-    }))
+  onPick = (value) => {
+    this.props.onPick(value)
+    this.setState(state => ({...state, inputValue: ''}))
+    setTimeout(() => this.input.focus(), 1) // don't ask...
```
enguerran: What's that?

src/sharing/components/ShareAutocomplete.jsx
```diff
@@ -23,39 +23,70 @@ export default class ShareAutocomplete extends Component {
...
   }
 
-  onChange (event, { newValue }) {
-    const email = getPrimaryEmail(newValue) || newValue
-    const url = getPrimaryCozy(newValue)
-    const id = newValue._id
-    this.props.onChange(email, url, id)
+  onRemove = (value) => {
+    this.props.onRemove(value)
+    setTimeout(() => this.input.focus(), 1)
```
enguerran: Yet another `don't ask...` mystery. What's that?

src/photos/locales/en.json
```diff
@@ -125,6 +125,7 @@
         "email": "email",
         "emailPlaceholder": "Enter email address or name of your recipient",
         "send": "Send",
+        "genericSuccess": "You sent an invite to %{count} contacts.",
```
y-lohse: This translation probably needs a singular form as well, if there is only one contact?

src/sharing/components/ShareAutocomplete.jsx
```diff
@@ -23,39 +23,70 @@ export default class ShareAutocomplete extends Component {
...
+    if (typeof newValue === 'object') {
+      this.onPick(newValue)
+    } else {
+      this.setState(state => ({...state, inputValue: newValue}))
+    }
+  }
+
+  onKeyPress = (event) => {
+    // The user wants to add an unknown email
+    if (event.key === 'Enter' && this.state.inputValue !== '') {
```
y-lohse: It look like `event.key` is [not supported in safari](https://developer.mozilla.org/fr/docs/Web/API/KeyboardEvent/key), so it's probably safer to use the keycode instead.

src/sharing/components/autosuggest.styl
```diff
@@ -32,3 +32,46 @@
...
+    height           em(20px)
+    width            em(20px)
+    line-height      em(16px)
+    margin           0 -8px 0 8px
+    background-color cool-grey
+    border           none
+    border-radius    50%
+    &:after
+        color   silver
+        content 'x'
```
y-lohse: Meeeeeeh. Why not use [the icon](https://github.com/cozy/cozy-drive/blob/master/src/drive/assets/icons/icon-cross.svg) as a background?

src/lib/redux-cozy-client/slices/sharings.js
goldoraf: I'm not responsible for properties naming on the server side </passive-agressive>

src/lib/redux-cozy-client/CozyClient.js
goldoraf: The first deletes the sharing, the second only revokes it for one of many recipients of the sharing.

src/sharing/components/ShareAutocomplete.jsx
goldoraf: Don't ask!

src/sharing/components/ShareAutocomplete.jsx
goldoraf: I said: don't ask! :p

src/photos/locales/en.json
goldoraf: No, because it's only displayed when there are many contacts.

src/sharing/components/ShareAutocomplete.jsx
goldoraf: More seriously: this is a weird React behavior I discovered yesterday, see https://stackoverflow.com/questions/35522220/react-ref-with-focus-doesnt-work-without-settimeout-my-example

src/sharing/components/ShareAutocomplete.jsx
goldoraf: You're right, fixed.

src/sharing/components/autosuggest.styl
goldoraf: Well, I tried with the icon but it didn't quite look right, so I reversed to the 'x'. Well, a "painting" pass will be needed anyway...

src/lib/redux-cozy-client/slices/sharings.js
```diff
@@ -15,6 +15,11 @@ const REVOKE_SHARING = 'REVOKE_SHARING'
 const RECEIVE_SHARING_REVOKE = 'RECEIVE_SHARING_REVOKE'
 const RECEIVE_ERROR = 'RECEIVE_ERROR'
 
+const removeRecipient = (recipients, recipientId) => {
+  const idx = recipients.findIndex(r => r.recipient.id === recipientId)
+  return [...recipients.slice(0, idx), ...recipients.slice(idx + 1)]
```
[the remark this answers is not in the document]
goldoraf: Hmmm... it should never happen in theory, so I'm not sure adding this condition is a good idea : it'd better to fail fast, right? I'm not sure TBH...

src/lib/redux-cozy-client/slices/sharings.js
goldoraf: Having 2 separate actions would be better, that's right. I had a refactoring in mind (that I'm implementing right now) that would allow to declare actions in a lighter way, so I wanted to separate these 2 actions later.

src/sharing/components/ShareAutocomplete.jsx
enguerran: I know, that's exactly why I asked!
;-)

src/lib/redux-cozy-client/CozyClient.js
enguerran: Ohh, so the first one `deleteSharing` and the second `removeSharingRecipient`?
Maybe we could rename them this way, unless I am the only one who did not understand.

src/lib/redux-cozy-client/slices/sharings.js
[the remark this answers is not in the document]
enguerran: Let's see what will happen.

goldoraf:
[(3) [PHO] Partager avec plusieurs contacts](https://trello.com/c/y4Vs2fQd/488-3-pho-partager-avec-plusieurs-contacts)

PR #402, opened 2017-09-21
chore: use prettier

enguerran:
Prettier [helps formatting](https://github.com/prettier/prettier#how-does-it-compare-to-eslint-or-tslint-stylelint) code homogeneously.
Standard [helps linting](https://standardjs.com/rules.html) code (no-unused-var, eqeqeq, etc.).
`lint-staged` and `husky` run prettier on `git commit`.

Be my guest to discuss and comment this PR.

[EDIT] more info here http://jlongster.com/A-Prettier-Formatter and https://youtu.be/hkfBvpEfWdA and https://youtu.be/0Q4kUNx85_4

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "prewatch:browser": "npm run clean:browser",
     "prewatch:mobile": "npm run clean:mobile",
     "lint": "npm-run-all --parallel 'lint:*'",
-    "lint:js": "standard --parser babel-eslint 'src/**/*.js' 'src/**/*.jsx' 'test/**/*.js'",
     "lint:styles": "stylint src/drive/styles --config ./.stylintrc",
     "test": "node_modules/.bin/jest --coverage --config=jest.config.json",
     "watch:drive": "npm run watch:drive:browser",
@@ -67,7 +66,7 @@
   "devDependencies": {
     "autoprefixer": "^6.5.0",
     "babel-core": "^6.0.0",
-    "babel-eslint": "^7.2.1",
+    "babel-eslint": "^8.0.0",
     "babel-jest": "^18.0.0",
     "babel-loader": "^6.0.0",
     "babel-plugin-transform-class-properties": "^6.23.0",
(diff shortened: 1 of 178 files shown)


package.json
```diff
@@ -165,5 +187,38 @@
...
+    },
+    "env": {
+      "browser": true,
+      "jest": true
+    }
+  },
+  "prettier": {
+    "semi": false,
+    "singleQuote": true
   }
```
ptbrowne: IMHO it is better to use a file .eslintrc for this config, it will be easier to copy it to all our other apps and works better with `git diff` etc..

enguerran:
We were using `standardjs` to avoid bike shedding discussion regarding code formatting.
`standardjs` is just an opinionated `eslint`'s configuration.
`eslint` is good at warning developers when a rule isn't respected, but when you use `--fix` it does no miracle and leaves a lot of code written in a subjective way.

That's is the very purpose of prettier. As the creator say:
> Prettier gets rid of all original styling and guarantees consistency by parsing JavaScript into an AST and pretty-printing the AST. Unlike eslint, there aren't a million configuration options and rules. But more importantly: everything is fixable.

And **consistency** is very valuable in large team.

package.json
enguerran: It ain't eslint configuration, but prettier configuration.
I've written those rules inside `package.json` because I prefer to centralize things.
I am not a big fan of having multiple `.dotfiles` in the project's root.
As it is subjective, I can discuss on it:
https://www.loomio.org/p/UxBcK5nb/prettier-s-configuration-goes-to-prettierrc-or-in-package-json

enguerran:
TIL: https://eslint.org/docs/rules/no-return-await

> Inside an async function, return await is useless. Since the return value of an async function is always wrapped in Promise.resolve, return await doesn’t actually do anything except add extra time before the overarching Promise resolves or rejects. This pattern is almost certainly due to programmer ignorance of the return semantics of async functions.

package.json
ptbrowne: > It ain't eslint configuration, but prettier configuration.

Sorry I did not point the exact line, the eslint configuration is just above. For me there is one point that is not subjective : with a file, you can diff easily and copy without any problem. With a property in a JSON file, you cannot do that. In addition, the file can be read by multiple software, here it is coupled to webpack.

PR #407, opened 2017-09-25
fix: Viewer on iOS couldn't be closed

GoOz:
iOS Safari was catching tap events that was meant to the toolbar, so I transfered all the gesture logic to the Photo wrapper instead of the Viewer wrapper.

### DIFF
--- a/src/photos/components/Viewer.jsx
+++ b/src/photos/components/Viewer.jsx
@@ -269,10 +269,10 @@ export class Viewer extends Component {
       display: isImageLoading ? 'none' : 'initial'
     }
     return (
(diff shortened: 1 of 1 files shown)


GoOz:
[[DRI] 🐛📱 Le bouton de fermeture de la visionneuse photos ne fonctionne pas sur mobile](https://trello.com/c/9pzwy9nA/523-dri-%F0%9F%90%9B%F0%9F%93%B1-le-bouton-de-fermeture-de-la-visionneuse-photos-ne-fonctionne-pas-sur-mobile)

PR #410, opened 2017-09-27
fix: handle 409 CONFLICT errors during syncs

y-lohse:


src/drive/mobile/actions/mediaBackup.js
```diff
@@ -100,6 +102,18 @@ const uploadPhoto = (dirID, photo) => async (dispatch, getState) => {
...
     dispatch(successMediaUpload(photo))
   } catch (err) {
+    if (Array.isArray(err) && err.length > 0) {
+      const firstError = err[0]
+      if (
+        firstError.status &&
+        parseInt(firstError.status) === ERROR_CODE_CONFLICT
+      ) {
+        // since it's a conflict error, the file *is* actually on the server, so we consider it a success
+        dispatch(successMediaUpload(photo))
```
enguerran: Not sure it's a good idea to merge both behavior in the same function: `successMediaUpload` is supposed to dispatch a _successful media upload_ when `409` confict code is supposed to dispatch a _already uploaded media_.

If we want to merge both behavior, we should rename the action `mediaOnTheServer` or something like that.

y-lohse: I think it makes sense to keep it like this; when we dispatch `successMediaUpload `, we're saying "this media has been uploaded, don't do it again". Whether the media was uploaded *right now* or has been on another occasion and we didn't know until now doesn't matter.

PR #412, opened 2017-09-27
Update the client's last sync date on the server

y-lohse:
I did some refactoring in 1b04b6bcacd77923c410a4b5b82cbdc70a86929a first — stopping the success / error callbacks to go all the way down the function chain. They probably shouldn't even be  in the replication lib at all, but that will be for another day.

The 2 other commits are just calling the new route.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
   "dependencies": {
     "classnames": "^2.2.0",
     "cozy-bar": "4.1.4",
-    "cozy-client-js": "0.3.18",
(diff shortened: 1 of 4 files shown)


y-lohse:
[(1) [DRI] 📱 Pinger la route de "date de dernière synchro" lorsqu'on termine une synchro](https://trello.com/c/PgypJbuV/520-1-dri-%F0%9F%93%B1-pinger-la-route-de-date-de-derni%C3%A8re-synchro-lorsquon-termine-une-synchro)

PR #417, opened 2017-10-09
fix: import file only once

enguerran:
see https://github.com/cozy/cozy-ui/pull/201


enguerran:
This PR is blocking #405 so it is quite primary!

PR #420, opened 2017-10-10
chore: add sentry native and cocoapods

y-lohse:
This PR adds sentry to our native apps, meaning that exception occuring in android and IOS should be reported in our Sentry. The plugin is automatically started so there's nothing else to do.

On iOS, it requires to install the official sentry cocoapod package. Since we don't keep these files in version control, I've added a plugin that takes care of regenerating them.

Fair warning: this is a bit experimental. If we don't get any useful information out of it, we'll revert this PR. Sadly we need to actually publish a version of the app with it in order to really test it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "android:drive:publish": "(cd targets/drive/mobile && fastlane supply)",
     "ios:drive:run": "(cd targets/drive/mobile && cordova run ios --device)",
     "ios:drive:run:emulator": "(cd targets/drive/mobile && cordova run ios --emulator)",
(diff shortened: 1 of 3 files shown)


y-lohse:
[[DRI] 🐛📱 La sauvegarde de photos iOS fait planter l'application](https://trello.com/c/36lYqngr/511-dri-%F0%9F%90%9B%F0%9F%93%B1-la-sauvegarde-de-photos-ios-fait-planter-lapplication)

enguerran:
What about publishing a custom release not attached to the official release workflow?
As in `npm ios:build:drive` and use whatever iOS development tool to let beta testers use this release instead of the official release one.

y-lohse:
Yeah actually you're right, that makes way more sense. I'll reopen this PR if the results are any good.

PR #423, opened 2017-10-12
feat: new route to display files

ptbrowne:
Adds a new route `/show/:fileId` that shows a modal with the given file in an iframe. It is useful for the bank mobile application since the mobile intents do not work yet.

* When the user closes the modal, the displayed folder is the parent folder of the file.
* Some mobiles do not display the PDF in an iframe and instead try to download the file. Since we cannot know whether the iframe is displaying or downloading the file, a fallback message is shown 2s after the modal is shown "You can close this if you have downloaded the file".

On a mobile whose modal is not displaying the file : 

On a mobile displaying the file

This PR is a first step, we should have later on a proper viewer (maybe PDF.js) to display PDFs on all mobiles.

### DIFF
--- a/src/drive/ducks/files/styles.styl
+++ b/src/drive/ducks/files/styles.styl
@@ -1,4 +1,28 @@
-.fileOpener iframe {
+.fileOpener__iframe
     width 80vw
     height 80vw
-}
+    position relative // new stacking context to be above fallback
+
+
+.fileOpener__fallback
+    // vertical-align center
+    position absolute
+    top 0
+    bottom 0
+    height 3rem // sufficient to hold all the text
+    margin auto
+
+    text-align center
+
+    // display after 2s (wait for iframe to load)
+    animation display-after-1s ease 0s
+    animation-delay 2s
+    animation-fill-mode both
+    width 100%
+
+@keyframes display-after-1s
+    0%
+        color transparent
+    100%
+        color inherit
+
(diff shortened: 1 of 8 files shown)


y-lohse:
I also think it's more elegant if there's a single URL for files and folders. But I wouldn't say it's too important, cozy URLs are private and don't get shared, as long as they are explicit I think we're good. 
So I would say, if the impact on the code is small, merge the 2 routes, if it's not trivial, then it can stay like this.

src/drive/ducks/files/styles.styl
```diff
@@ -0,0 +1,28 @@
+.fileOpener__iframe
+    width 80vw
```
y-lohse: We've had some really bad times with `vw` units, I don't knwo what @GoOz thinks about it but at this point I'd be more confident in using `%`.

enguerran:
Well, if we want explicit routes, we shall rename `/files/:folderId` to `/folder/:folderId` and use `/file/:fileId`.

ptbrowne:
I agree that having the two routes merged would be better. I was going for that at the beginning but by looking at the code, it was non trivial. I spoke yesterday to @goldoraf on this issue and he was fine having an additional route.

src/drive/ducks/files/styles.styl
GoOz: Actually it's `vh` the problem, `vw` not so much.

ptbrowne:
OK, I'm making the change :)

PR #424, opened 2017-10-12
 🚑  Fix the complicated triple tap to show the support button on iOS

GoOz:
Use Hammer to handle tap gesture instead of basic onClick

### DIFF
--- a/src/drive/mobile/components/Settings.jsx
+++ b/src/drive/mobile/components/Settings.jsx
@@ -16,16 +16,16 @@ import Unlink from '../containers/settings/Unlink'
 
 class Settings extends Component {
   state = {
-    countClick: 0
+    tapCount: 0
   }
 
-  handleClick = () => {
-    this.setState(state => ({ countClick: state.countClick + 1 }))
+  incrementTapCount = () => {
+    this.setState(state => ({ tapCount: state.tapCount + 1 }))
   }
 
   render() {
     const { t } = this.props
-    const isDebug = this.state.countClick >= 3
+    const isDebug = this.state.tapCount >= 3
     return (
       <Main>
         <Topbar>
@@ -36,7 +36,7 @@ class Settings extends Component {
             <MediaBackup />
             <Contacts />
             <Support isDebug={isDebug} />
-            <About onClick={this.handleClick} />
+            <About onTap={this.incrementTapCount} />
             <Unlink />
 
             {__DEVELOPMENT__ && [<hr />, <h3>Debug Zone</h3>, <DebugTools />]}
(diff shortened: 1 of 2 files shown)


GoOz:
[[GLOB] 🐛📱Les boutons ne sont pas réactifs sur iOS, particulièrement le faux bouton version pour envoyer les logs](https://trello.com/c/h3dNaizn/530-glob-%F0%9F%90%9B%F0%9F%93%B1les-boutons-ne-sont-pas-r%C3%A9actifs-sur-ios-particuli%C3%A8rement-le-faux-bouton-version-pour-envoyer-les-logs)

src/drive/mobile/components/Settings.jsx
```diff
@@ -16,16 +16,16 @@ import Unlink from '../containers/settings/Unlink'
 
 class Settings extends Component {
   state = {
-    countClick: 0
+    tapCount: 0
```
enguerran: thanks for renaming this, it hurted my ears. ;-)

GoOz: Honestly, @y-lohse told me it would make you happy to do so, so I did :) but i didn't even think about it myself.

src/drive/mobile/containers/settings/About.jsx
```diff
@@ -1,26 +1,48 @@
...
-        onClick
-      }
-    ]}
-  />
-)
+class About extends Component {
+  componentDidMount() {
+    this.gesturesHandler = new Hammer.Manager(this.versionNumber)
+    this.gesturesHandler.add(new Hammer.Tap({ event: 'singletap' }))
+    this.gesturesHandler.on('singletap', this.props.onTap)
```
enguerran: Il n'y a pas un moyen d'enregistrer directement un event `{ event: 'tripletap' }` ?

GoOz: It… could make sense I guess but I don't know how to do that. @y-lohse what do you think?

y-lohse: Something like this should to the trick : 

```
this.gesturesHandler.add(new Hammer.Tap({ event: 'tripletap', taps: 3 }))
this.gesturesHandler.on('tripletap', this.props.onTap)
```

Haven't tried it but I guess it would work.

enguerran: My thoughts are: if we use a third-party library to handle _tap_ and/or _touch_ events, it could be more stable to deport the responsibility to the library. But if there is no built-in support for _triple tap_, then we can handle the tap sequence in our code.

PR #430, opened 2017-10-17
fix: uploads on android

y-lohse:
There's a problem in our upload code for android and I can't find a proper fix for it, so instead I'm fixing it here.

Essentially, uploading a file that already exists will produce an error in Android, but that particular error (`Broken pipe`) could also happen for other reasons. 
My proposal is to check if a file exists before trying to upload it. It's a lot of overhead during the first run, but it's the best I have.

### DIFF
--- a/src/drive/mobile/actions/mediaBackup.js
+++ b/src/drive/mobile/actions/mediaBackup.js
@@ -18,8 +18,6 @@ export const MEDIA_UPLOAD_SUCCESS = 'MEDIA_UPLOAD_SUCCESS'
 export const MEDIA_UPLOAD_CANCEL = 'MEDIA_UPLOAD_CANCEL'
 export const CURRENT_UPLOAD = 'CURRENT_UPLOAD'
 
-const ERROR_CODE_CONFLICT = 409
-
 const startMediaUpload = () => ({ type: MEDIA_UPLOAD_START })
 const endMediaUpload = () => ({ type: MEDIA_UPLOAD_END })
 const successMediaUpload = media => ({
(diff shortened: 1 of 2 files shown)


src/drive/mobile/actions/mediaBackup.js
```diff
@@ -98,20 +96,22 @@ export const startMediaBackup = (dir, force = false) => async (
...
 }
 
-const uploadPhoto = (dirID, photo) => async (dispatch, getState) => {
+const uploadPhoto = (dirName, dirID, photo) => async (dispatch, getState) => {
+  let fileAlreadyExists = false
+  try {
+    await cozy.client.files.statByPath(dirName + '/' + photo.fileName)
+    dispatch(successMediaUpload(photo))
+    return
+  } catch (_) {} // if an exception is throw, the file doesn't exist yet and we can safely upload it
```
y-lohse: Pre-emptive answer: this code is here and not in the part where we compute the list of files to upload because that would require checking all the files *before* starting the first upload, which can take a while.

y-lohse:
[[DRI] 🐛📱 Des photos sont détectées comme sauvegardable en permanence](https://trello.com/c/Ngu7GGFt/506-dri-%F0%9F%90%9B%F0%9F%93%B1-des-photos-sont-d%C3%A9tect%C3%A9es-comme-sauvegardable-en-permanence)

src/drive/mobile/actions/mediaBackup.js
enguerran: Puisque c'est mergé mais qu'on a discuté à l'oral, je note ici ma remarque : je pense que plus d'expressivité dans le code n'aurait fait de mal à personne.

PR #434, opened 2017-10-19
Merged old redux-cozy-client with new cozy-client

goldoraf:


src/photos/ducks/timeline/index.js
```diff
@@ -41,6 +41,7 @@ export const deletePhotos = ids => async dispatch => {
       const file = { ...rawFile, id: rawFile._id, type: rawFile._type }
       if (file.referenced_by) {
         for (const ref of file.referenced_by) {
+          console.log(ref)
```
y-lohse: Leftover debug ?

goldoraf: Yep, good catch ;)

goldoraf:
@y-lohse I'm still in the process of re-testing the whole apps, so I think it's not necessary for you to test it locally.

PR #438, opened 2017-10-24
feat: display message during first long sync

y-lohse:


y-lohse:
[(2) [DRI] 📱 Afficher un message d'information lors de la première sauvegarde s'il y a un grand nombre de photos](https://trello.com/c/fA8TrFhK/524-2-dri-%F0%9F%93%B1-afficher-un-message-dinformation-lors-de-la-premi%C3%A8re-sauvegarde-sil-y-a-un-grand-nombre-de-photos)

src/drive/mobile/containers/FirstUploadModal.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+const mapStateToProps = state => ({
+  uploading:
+    state.mobile.mediaBackup.uploading &&
+    state.mobile.mediaBackup.currentUpload &&
+    state.mobile.mediaBackup.currentUpload.messageData &&
+    state.mobile.mediaBackup.currentUpload.messageData.total_upload &&
+    state.mobile.mediaBackup.currentUpload.messageData.total_upload >=
+      MINIMUM_LONG_UPLOAD_FILES_COUNT
+})
```
enguerran: It is really hard to catch the intention here. `uploading` is a boolean, its value is `true` if and only if all other booleans are `true`. All other booleans are:
- `uploading`: ok, so to be `true` local `uploading` needs global `uploading` to be true ;
- `currentUpload`... I guess if `currentUpload` is set (surely not as a boolean), global `uploading` is true by the way ;
- `messageData`... its a `currentUpload`'s property... ohhhh, I see something !

In fact, what we want for local `uploading` to be true is only two things:
1. global `uploading` to be true
1. `total_upload` to be greater than `MINIMUM_LONG_UPLOAD_FILES_COUNT`

but as `total_upload` is a subsubsubproperty, we need to check the existence of parent properties. Destructuring is a very cool feature for such a thing:

```js
const mapStateToProps = state => {
    const { uploading = false, currentUpload = { messageData: { total_upload: 0 } } } = obj2.state.mobile.mediaBackup
    const { messageData: { total_upload } } = currentUpload
    return {
        uploading: uploading && total_upload >= MINIMUM_LONG_UPLOAD_FILES_COUNT
    }
}
```

src/drive/mobile/containers/FirstUploadModal.jsx
```diff
@@ -0,0 +1,74 @@
...
+    markAsShown = () => {
+      this.setState({ shown: true })
+    }
+
+    render() {
+      const { uploading } = this.props
+      const { shown } = this.state
+
+      if (uploading && !shown)
+        return <WrappedComponent {...this.props} onClose={this.markAsShown} />
```
enguerran: Maybe we should rename the `shown` state to `viewed` or `read` like in _mark as read_ or _mark as viewed_. I've never seen _mark as shown_, don't we?

src/drive/mobile/containers/FirstUploadModal.jsx
```diff
@@ -0,0 +1,74 @@
...
+      else return null
+    }
+  }
+}
+
+const OnlyOnceFirstUploadModal = withPersistentState(
+  showOnlyOnce(FirstUploadModal),
```
enguerran: I cannot understand the `showOnlyOnce` HOC as it uses an internal state controlled by a `onClose` button.
I guess the wanted behavior is to use `withPersistentState` to persist the internal state to actually show only once the `FirstUploadModal`. So the `showOnlyOnce` HOC takes its name because we use it with `withPersistentState`.

src/drive/mobile/containers/FirstUploadModal.jsx
```diff
@@ -0,0 +1,74 @@
...
+          </ul>
+          <p>{t('mobile.first_sync.result')}</p>
+          <button onClick={onClose} className={styles['btn--full-width']}>
+            {t('mobile.first_sync.button')}
+          </button>
+        </ModalContent>
+      </Modal>
+    )
+  }
+}
```
enguerran: Can't we use a functional component?

y-lohse: I'd like to, but is there a way to get `t` from `context` with a functional component?

src/drive/mobile/containers/FirstUploadModal.jsx
y-lohse: Well you're right, `viewed` is probably more usual.

src/drive/mobile/containers/FirstUploadModal.jsx
y-lohse: This is all correct, so I don't understand what problem you are pointing out here. Can you clarify please?

src/drive/mobile/containers/FirstUploadModal.jsx
y-lohse: That is cool, I didn't knew you could destructure with default values. 👍

src/drive/mobile/containers/FirstUploadModal.jsx
enguerran: Are you sure `t` is not in `props` passed as parameter?

src/drive/mobile/containers/FirstUploadModal.jsx
enguerran: `showOnlyOnce` utilisé tout seul ne fait pas vraiment ce qu'il prétend faire:

```jsx
const Hello = ({ name, onClose }) => (
  <div>
    <div>Hello {name}</div>
    <button onClick={onClose}>close</button>
  </div>
)
const ShowOnlyOnceHello = showOnlyOnce(Hello)
reactDOM.render(<ShowOnlyOnceHello />, document.getElementById('root')
```

c'est plutôt un truc qui dit qu'on peut le fermer.

C'est avec `withPersistentState` (un HOC un peu bizarre puisqu'il casse un peu le lifecycle des composants (p)react) qu'on peut dire qu'il montre seulement une fois.

src/drive/mobile/containers/FirstUploadModal.jsx
y-lohse: It's not, but actually I can inject it.

src/drive/mobile/containers/FirstUploadModal.jsx
y-lohse: Je l'ai renommé, mais c'est pas beaucoup mieux — et je n'ai pas de meilleure idée.

PR #439, opened 2017-10-24
Fixed some sharing locales and added a (sort of) migration to contacts

goldoraf:


src/lib/cozy-client/slices/sharings.js
```diff
@@ -238,6 +238,38 @@ const createSharingLink = document => ({
   promise: client => client.createSharingLink(getPermissionsFor(document, true))
 })
 
+// TODO: this is a poor man's migration in order to normalize contacts
+// and should be removed after a few weeks in prod
```
enguerran: maybe we could add an explicit due date

src/sharing/components/Recipient.jsx
```diff
@@ -35,34 +36,55 @@ export const UserAvatar = ({ name, url }) => (
...
+class Recipient extends Component {
+  state = {
+    revoking: false
+  }
+
+  onUnshare = () => {
+    this.setState({ revoking: true })
+    this.props
+      .onUnshare(this.props.contact)
+      .then(() => this.setState({ revoking: false }))
```
enguerran: As the `setState` is used in an asynchronous function, we should use, as always, the function version of `setState`:

```js
.then(() => this.setState(state => ({ revoking: false })))
```

src/sharing/components/Recipient.jsx
```diff
@@ -35,34 +36,55 @@ export const UserAvatar = ({ name, url }) => (
...
+  render() {
+    const { t, contact, status, type, documentType } = this.props
+    const { revoking } = this.state
+    const name = getPrimaryEmail(contact)
+    const url = getPrimaryCozy(contact)
+    return (
+      <div className={styles['pho-recipient']}>
+        <Avatar name={name} />
+        <Identity name={name} url={url} />
         <div className={styles['pho-recipient-status']}>
```
enguerran: It will be clearer to refactor this underlying component (everything under `<div className={styles['pho-recipient-status']}>`) into its own component definition with a HOC to display the spinner or the component.

src/lib/cozy-client/slices/sharings.js
goldoraf: Problem is: I don't know :/ It depends on when this patch will be released to all users

src/sharing/components/Recipient.jsx
goldoraf: Yep!

src/sharing/components/Recipient.jsx
goldoraf: OK for refactoring into a Status component, but I can't find a cool API for a HOC like you suggest. Any idea?

src/lib/cozy-client/slices/sharings.js
enguerran: We could add

```js
// Note for future-self: If you have no idea of what it means, please, erase this code.
```

src/lib/cozy-client/slices/sharings.js
```diff
@@ -238,6 +238,38 @@ const createSharingLink = document => ({
...
+              .updateDocument({
+                ...contact,
+                email: [
+                  {
+                    address: contact.email,
+                    primary: true
+                  }
+                ]
+              })
+              .then(resp => resp.data[0])
```
y-lohse: I think we should avoid mixing `async/await` and `then()`, if possible.

src/sharing/components/Recipient.jsx
y-lohse: I don't understand what difference it makes here. Whether you call `setState` with `{ revoking: false }` or `state => ({ revoking: false })`, the result will always be the same.

src/sharing/components/recipient.styl
```diff
@@ -68,4 +68,5 @@
 
 .pho-action-unshare
     color pomegranate !important
+    text-decoration none !important
```
y-lohse: Why do we need the `!important` here?

src/sharing/components/Recipient.jsx
y-lohse: I don't understand what difference it makes here. Whether you call `setState` with `{ revoking: false }` or `state => ({ revoking: false })`, the result will always be the same.

src/sharing/components/recipient.styl
goldoraf: Because of what I think are $@# global link styles I couldn't override...

src/lib/cozy-client/slices/sharings.js
goldoraf: :+1:

src/lib/cozy-client/slices/sharings.js
goldoraf: I agree, but in that case it would mean a less FP-oriented style

y-lohse: Well, it would still be FP. Something like this :

```
response.data.map(async (contact) => {
  if (typeof contact !== 'string') {
    return contact
  } else {
    const resp = await client.updateDocument(…)
    return resp.data[0]
  }
})
```

I know it gets rid of the ternary which is quite elegant here, but I think the more verbose version adds some clarity. It becomes obvious that what is returned is `contact` or `return resp.data[0]`, instead of `contact` or a promise that resolves.

But I can live with both! It's easy to compare both versions now so pick whatever you feel is best :)

src/sharing/components/recipient.styl
y-lohse: That's a bit worrysome. Do you know where these global styles are defined?

src/sharing/components/Recipient.jsx
enguerran: Maybe with something like that: https://codesandbox.io/s/zr49prv66p

```jsx
const withSpinner = (isLoading, BaseComponent) => props =>
  isLoading(props) ? <Spinner /> : <BaseComponent {...props} />;
```

src/photos/containers/AlbumPhotos.jsx
```diff
@@ -50,6 +50,7 @@ export class AlbumPhotos extends Component {
       return null
     }
     const { album, photos, shared } = this.props
+    console.log(shared)
```
enguerran: oups!

src/sharing/components/recipient.styl
goldoraf: I believe it's because of this: https://github.com/cozy/cozy-drive/blob/master/src/sharing/share.styl#L11

src/lib/cozy-client/slices/sharings.js
```diff
@@ -238,6 +238,41 @@ const createSharingLink = document => ({
   promise: client => client.createSharingLink(getPermissionsFor(document, true))
 })
 
+// TODO: this is a poor man's migration in order to normalize contacts
+// and should be removed after a few weeks in prod
+// Note for future-self: If you have no idea of what it means, please, erase this code.
+// Note for time-travelers: please travel a little more back in time in order to advise us
+// to get contacts right the first time
```
enguerran: 👍 awesome 👍

src/lib/cozy-client/slices/sharings.js
enguerran: I have to admit that current code is not very clear.

PR #444, opened 2017-10-25
Various fixes

goldoraf:


src/lib/cozy-client/reducer.js
```diff
@@ -314,15 +314,15 @@ export const trashFile = (file, actionOptions = {}) => ({
 
 export const addReferencedFiles = (doc, ids) => ({
   type: ADD_REFERENCED_FILES,
-  collection: `${doc.type}/${doc.id}#files`,
+  collection: `${doc._type}/${doc._id}#files`,
```
enguerran: I guess I trust you on this one.

goldoraf:
@enguerran `/targets` folder is not in the list of folders to be reformatted by prettier, so it wasn't modified when you added prettier, but as I was modifying that file, I machinally hit Ctrl+Shift+i ;) Looking better now

src/lib/cozy-client/reducer.js
goldoraf: Previously, `cozy-client` relied on the `type` property, which is used by some doctypes like `io.cozy.files`. So now we rely on the `_type` property.

PR #447, opened 2017-10-25
feat: add an IntentButton Component

Narkfr:


Narkfr:
[(8) [DRI-1] ETU je peux configurer un connecteur depuis Drive](https://trello.com/c/62xlnDeq/640-8-dri-1-etu-je-peux-configurer-un-connecteur-depuis-drive)

src/drive/components/IntentButton.jsx
```diff
@@ -0,0 +1,68 @@
+import styles from '../styles/intentbutton'
+
+/* global cozy */
+import React from 'react'
+import classNames from 'classnames'
+
+import Modal from 'cozy-ui/react/Modal'
+
+class IntentButton extends React.Component {
```
CPatchane: This component seems to be very generic (not specific to an intent) and it could be great to have it directly in cozy-ui 🎉

src/drive/components/IntentButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+class Intent extends React.Component {
+  componentDidMount() {
+    const { data, intent, closeModal } = this.props
+    cozy.client.intents
+      .create(intent.action, intent.docType, {
+        ...data,
+        exposeIntentFrameRemoval: true
+      })
+      .start(document.getElementById('intentViewer'))
+      .then(() => {
```
CPatchane: If you put `exposeIntentFrameRemoval` to `true`, you can get `removeIntentFrame` which allow to remove the intent HTML node after calling `closeModal()` (and avoid blank page inside the modal)

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -45,6 +46,22 @@ class Toolbar extends React.Component {
     const notRootfolder = displayedFolder && displayedFolder.id !== ROOT_DIR_ID
     return (
       <div className={styles['fil-toolbar-files']} role="toolbar">
+        <IntentButton
+          className={classNames(
+            styles['c-btn'],
+            styles['c-btn--regular'],
+            styles['u-hide--mob']
+          )}
+          intent={{
```
gregorylegarec: I love the idea of a having a component for intent, however I think we should keep the signature of the method [[https://github.com/cozy/cozy-client-js/blob/master/docs/intents-api.md#intents|`cozy.client.intents.create`]]
Two reasons:
* It's clearer to follow the parameter
* The given intent here is not an existing intent from DB, it does not have and id for example. I think we should keep the concept of an intent resulting from action, doctype and data, and not mocked object with only a part of its attributes.
So I suggest to do:
```jsx
<IntentButton
  action='CREATE'
  doctype='io.cozy.accounts'
  data={{
    dataType: 'bill'
  }}
>
```
What do you think ?
What do @CPatchane and @y-lohse think ?

src/drive/styles/intentbutton.styl
```diff
@@ -0,0 +1,17 @@
+@require 'cozy-ui'
+
+// This ugly rules hide the default close button of the modal
```
gregorylegarec: Is there really no other alternative ? Does the service really need to provide a close button ? What happen if for example the service is called within another container than an iframe ?

src/drive/locales/en.json
```diff
@@ -31,6 +31,9 @@
     "empty_trash": "Empty trash",
     "share": "Share"
   },
+  "intent": {
```
gregorylegarec: I am not sure that "intent" should be in the locale key, it's providing an technical information about how we implement the feature. IMHO locale keys should only describe functionnal informations.

src/drive/ducks/files/Toolbar.jsx
y-lohse: 👍 that makes sense

src/drive/components/IntentButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+    modalIsOpen: false
+  }
+
+  openModal() {
+    this.setState({
+      modalIsOpen: true
+    })
+  }
+
+  closeModal = () => {
```
y-lohse: You use two different method definition syntaxes here, any particular reason?

src/drive/ducks/files/Toolbar.jsx
CPatchane: Agree, it's more coherent too 👍

src/drive/components/IntentButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <span className={classNames(styles['intentButton'])}>
+        {modalIsOpen && (
+          <Modal secondaryAction={() => this.closeModal()}>
+            <Intent data={data} intent={intent} closeModal={this.closeModal} />
+          </Modal>
+        )}
+        <button
+          className={className}
+          onClick={() => this.openModal(data, intent)}
```
y-lohse: I think you forgot to remove `data, intent` from the function call, as it doesn't appear to be needed :)

src/drive/locales/en.json
CPatchane: I would suggest `service` here

src/drive/components/IntentButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+class Intent extends React.Component {
+  componentDidMount() {
+    const { data, intent, closeModal } = this.props
+    cozy.client.intents
+      .create(intent.action, intent.docType, {
+        ...data,
+        exposeIntentFrameRemoval: true
+      })
+      .start(document.getElementById('intentViewer'))
```
y-lohse: I think it's probably better to use React's `ref` rather than relying on `document.getElementById`.

gregorylegarec: Yep, we are trying to avoid at maximum DOM querying into React components.

src/drive/ducks/files/Toolbar.jsx
Narkfr: :+1:

Narkfr:
@y-lohse @gregorylegarec @CPatchane thx for the review everyone ! As @CPatchane said, I think this component could be exported to cozy-ui, or to the future lib cozy-intent, maybe with a little more improvement.

src/drive/styles/intentbutton.styl
gregorylegarec: What about removing the close button from service ?

Narkfr: I want to keep the close button on the service to terminate correctly the intent.
+ removing the close button will probably have cross effect on other intents

gregorylegarec:
Thanks @Narkfr 👍

enguerran:
Dis @Narkfr, cette modif https://github.com/cozy/cozy-drive/commit/aebca65e5eb14d38f24aec55b7709a12681d895c#diff-496c2803b6373b077bd43538e90ef51cL26
me semble étrange vu les résultats de cette requête https://github.com/cozy/cozy-drive/search?utf8=%E2%9C%93&q=item_more&type=Code
Peux-tu corriger cela ?

src/drive/styles/intentbutton.styl
kosssi: Dsl de revenir sur ceci mais je trouve ça vraiment pas terrible de mon coté.
Pourquoi ne pas ajouter une option au composant Modal ?
Nous avons une carte pour améliorer la Modal de cozy-ui je m'en occuperai à ce moment là.

PR #449, opened 2017-10-26
Fixed menus styling issues related to the last cozy-ui refacto

goldoraf:
En essayant de fixer ce qui était à la base un petit pb de styling des menus de statut d'un partage, j'ai constaté que Drive et Photos avaient tous les 2 un composant Menu quasi identique... Selon la règle du boy-scout, je me suis donc efforcé d'extraire un seul composant dans un répertoire `common`, et au passage j'ai fait de même avec le `MoreButton`.


src/common/components/Button/index.jsx
```diff
@@ -0,0 +1 @@
+export { default as MoreButton } from './MoreButton'
```
y-lohse: What is the purpose of the `Button` file? What is the advantage over importing things from `MoreButton`?

src/common/components/Button/index.styl
```diff
@@ -0,0 +1,12 @@
+@require 'cozy-ui'
+
+@media (max-width: (1023/basefont)rem)
```
y-lohse: Don't we have variables or mixins for media queries? It feels weird  to have these values hard-coded here.

src/photos/styles/toolbar.styl
```diff
@@ -5,6 +5,8 @@
     margin-left auto
 
     a
+        &.pho-action-newalbum
+            background embedurl("../assets/icons/icon-album-add.svg") em(16px) center no-repeat
```
y-lohse: That's a different thing than the menu rewrite isn't it?

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -70,7 +71,7 @@ class Toolbar extends React.Component {
           title={t('toolbar.item_more')}
           disabled={disabled}
           className={styles['fil-toolbar-menu']}
-          buttonClassName={styles['fil-toolbar-more-btn']}
+          button={<MoreButton />}
```
y-lohse: I'm not sure I understand this change. The `Menu` component already had a `button` prop (apparently?) but it was a boolean. Now you pass in a component but I haven't seen it being used. Did I miss it?

src/common/components/Button/index.styl
GoOz: Might as well use the new syntax for MQs.
`+medium-screen()` instead of `@media (max-width: (1023/basefont)rem)`

src/common/components/Button/index.styl
```diff
@@ -0,0 +1,12 @@
...
+    .dri-btn--more
+        display inline-flex
+
+@media (max-width: (768/basefont)rem)
```
GoOz: Same here with `+small-screen()`

src/common/components/Button/index.jsx
enguerran: Didn't see this...
Is that kind of weird? I mean to pass a variable without changing anything, is that a complexity or an abstraction?

goldoraf:
@y-lohse Well, *my* logic behind the monorepo was that `src` should *in the end* only contain ducks, aka something like `files`, `trash`, `timeline`, `albums`, `sharing`, `collect`, `settings`, `onboarding`, etc. But these ducks will necessarily share some low-level components, thus this new `common` folder. Currently, we have these `drive` and `photos` folders in `src` because I didn't take the time to explode them into ducks folders, but it's planned (in my mind). Is it clearer for you?

src/common/components/Button/index.jsx
goldoraf: My goal was to centralize in the `Button` "namespace" all things button-related (and these things will probably be integrated in cozy-ui one day, when it's ready), so you could do something like:
```js
import Button, { MoreButton, UploadButton } from 'common/components/Button'
```
For now, I just wanted to commonalize `MoreButton`, but my goal is to push here all our buttons and to evolve into more easy-to-use `Button` components.

src/common/components/Button/index.styl
goldoraf: We have these all over the place... Didn't knew we got mixins for that. But as we are more and more adverse to stylus, do we really want to go all-in with stylus mixins now?

src/drive/ducks/files/Toolbar.jsx
goldoraf: It's used here: https://github.com/cozy/cozy-drive/pull/449/files#diff-d660e9521573770d4df9c5faf824ef34R62

src/photos/styles/toolbar.styl
goldoraf: Right, but this button needed to be corrected too because of the last cozy-ui refactoring

src/common/components/Button/index.styl
goldoraf: Didn't knew these mixins exist. But https://github.com/cozy/cozy-drive/pull/449#discussion_r147335530 ?

goldoraf:
@GoOz yes, `MoreButton` should probably be integrated into cozy-ui *in the future*. But my goal right now is to extract all these common little components, to mature them (make them rock solid and easy to use), and then to integrate into cozy-ui.

src/drive/ducks/files/Toolbar.jsx
y-lohse: Aaah yes, missed that. All good then.

src/photos/styles/toolbar.styl
y-lohse: 👍

src/common/components/Button/index.styl
y-lohse: I absolutely think so. I don't share the adversion to stylus all and but regardless, duplicating magic numbers is a bad thing. We hopefully wouldn't do this without stylus either!

src/common/components/Button/index.jsx
y-lohse: I must say I'm not too overwhelmed by this. The eventual idea is good, but we don't know when we'll add another button to this list, so I find it premature. But I can live with it.

y-lohse:
> Well, my logic behind the monorepo was […] Is it clearer for you?

Yes, thank you, it's clearer… but I stand by my comment.
I think we should do either one of these 2 things:

* src
* * common
* * * components
* * * ducks
* * * lib
* * drive
* * photos

Or

* src
* * components
* * lib
* * duck1
* * duck2
* * …
* * drive
* * photos

But **not**

* src
* * common
* * * components
* * lib
* * duck1
* * duck2
* * drive
* * photos

I think number 1 is better overall, but number 2 conforms to what we have at the moment. So I would recommend we follow number 2 for this PR, and if we decide we want to use number 1 (or whatever else) we do the change in a separate PR. What do you folks think?

goldoraf:
@y-lohse wokay... honestly I can't see why you want to put ducks into `common` like in proposition 1: ducks are not commons to me, they're features, and because of the fact that "ducks" are not a well-know organization pattern, I'd prefer to not see a folder named `ducks`, but nevermind.

I concur that `common` would really only contain components (non-component code would probably go into `lib`), and consequently a folder named `common` should not appear in the tree. So proposition 2 looks like the best option to me.

But don't forget we still have a `targets` folder outside of `src` and some people like @enguerran seem bothered by that. If we manage to refactor the contents of the actual `drive` and `photos` folders into more ducks, perhaps we could move `targets` into `src`. We would have something like:

- src
  - components
  - lib
  - duck1
  - duck2
  - duck3
  - targets
    - drive
      -web
      - mobile
    - photos
      - web
      - public

What do you think of this org as an ultimate goal? If everybody agree with that, I'd remove the `common` folder from this PR.

y-lohse:
Yep, I think that's  a clear structure, works for me 👍

enguerran:
Faut voir ce que propose @CPatchane avec cozy-create-app/create-cozy-app ... ;)

enguerran:
BTW
@goldoraf wrote
> But don't forget we still have a targets folder outside of src and some people like @enguerran seem bothered by that.

What really bother me is the `main.jsx` and other source code files that are not in `src/`.

PR #450, opened 2017-10-30
feat: add path to recent files 

y-lohse:
We fetch the recent files with a mango index (to sort them by `updated_at`), but this API doesn't return the file's path as those are not stored in CouchDB. 
I've added a call to the server to fetch the parent folders of all recent files, as *they* have a `path` attribute. @GoOz has updated the `File` component to allow displaying a file's path next to it's name.

### DIFF
--- a/src/drive/components/File.jsx
+++ b/src/drive/components/File.jsx
@@ -2,7 +2,7 @@
 import React, { Component } from 'react'
 import classNames from 'classnames'
 import filesize from 'filesize'
-import { withRouter } from 'react-router'
+import { withRouter, Link } from 'react-router'
 import Hammer from 'hammerjs'
 
 import styles from '../styles/table'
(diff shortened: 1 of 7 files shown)


y-lohse:
[[DRI] (1) Ajouter le chemin des fichiers dans l'onglet récent pour éviter la confusion avec un dossier](https://trello.com/c/5wNzpROP/545-dri-1-ajouter-le-chemin-des-fichiers-dans-longlet-r%C3%A9cent-pour-%C3%A9viter-la-confusion-avec-un-dossier)

src/drive/components/File.jsx
```diff
@@ -124,8 +124,8 @@ class File extends Component {
     }
   }
 
-  render(
-    {
+  render() {
```
enguerran: :+1:

y-lohse:
Oh right @GoOz , I missed that. Do you want to do it or should I take care of it?

PR #451, opened 2017-10-30
fix: use cozy client and pouch to search for files

y-lohse:
Worked like a charm..? It seems almost to easy.


y-lohse:
[(3) [DRI] Améliorer les performances de la recherche avec PouchDB](https://trello.com/c/tOTH9FN3/548-3-dri-am%C3%A9liorer-les-performances-de-la-recherche-avec-pouchdb)

goldoraf:
hmmm... you don't use Pouch here, you need to declare an `offline` option to the client

y-lohse:
Mmh ok, I'll investigate more. But somehow the search times are wayyyy down already.

y-lohse:
I'll re-open this once it's ready.

PR #453, opened 2017-11-02
feat: add responsive date format

GoOz:
Changes the date format according to the vewport's size. 
1200+ pixels wide means a full written month, otherwise it's an abbreviated one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,7 @@
     "classnames": "^2.2.0",
     "cozy-bar": "4.1.4",
     "cozy-client-js": "0.3.19",
-    "cozy-ui": "4.0.3",
+    "cozy-ui": "4.0.4",
     "date-fns": "^1.22.0",
     "diacritics": "^1.3.0",
     "filesize": "^3.3.0",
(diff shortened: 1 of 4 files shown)


GoOz:
@y-lohse that's prettier I guess. Pretty sure, actually.

y-lohse:
But why does prettier change them *now*? They've been around for a while.

GoOz:
Mmmmh maybe I don't have the same option in my prettier plugin in my editor…

enguerran:
I don't know why this commit https://github.com/cozy/cozy-drive/pull/453/commits/b729ff615d05c504767e41cb8bcc1bba2b2a9937 is failing and I see that the next commit https://github.com/cozy/cozy-drive/pull/453/commits/fc2acd1cf2e482643de95274c489abac6f32837e pass the travis build.

I don't think it's a good thing to have failing commits: every commit must be stable.

GoOz:
It's failing because its dependency was out of date. I always do like that actually but I can change if needed.

PR #454, opened 2017-11-02
Fast search

y-lohse:
So the search now uses Cozy Client. I've made the following changes to the client:

- Make sure the cozy url has a protocol.
- Since the doctype `io.cozy.files` can't be written *to*, we only want to replicate *from* the cozy instance. I've added methods that allow to do one way syncs.

### DIFF
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -12,6 +12,8 @@ class SuggestionProvider extends React.Component {
       const { query } = event.data
       this.provideSuggestions(query, intent)
     })
+
+    this.context.client.startReplicationFrom(() => {}) // the sync functions take a redux-style `dispatch` function as a callback, but we don't handle the replication status at the moment
   }
 
   async provideSuggestions(query, intent) {
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -1,4 +1,3 @@
-/* global cozy */
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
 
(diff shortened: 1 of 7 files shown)


src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -40,23 +41,11 @@ class SuggestionProvider extends React.Component {
   // fetches pretty much all the files and preloads FuzzyPathSearch
```
enguerran: Shall we keep this comment?

src/lib/cozy-client/DataAccessFacade.js
```diff
@@ -18,7 +22,13 @@ export default class DataAccessFacade {
   }
 
   setup(cozyUrl, options) {
-    const config = { cozyURL: cozyUrl, ...options }
+    // prepend the protocol if there isn't one, required for pouchdb
+    if (/^http(s):/.test(cozyUrl) === false) {
```
enguerran: `test` is already returning a boolean, so I guess we should write
```js
if (!/^http(s):/.test(cozyUrl) {
```

enguerran: What is the difference between `/^http(s):/` and `/^https:/`?

src/lib/cozy-client/DataAccessFacade.js
```diff
@@ -18,7 +22,13 @@ export default class DataAccessFacade {
   }
 
   setup(cozyUrl, options) {
```
enguerran: Shall we change the call of `setup(cozyUrl, options)` to be sure the `cozyUrl` contains a protocol?

src/lib/cozy-client/DataAccessFacade.js
y-lohse: You're right about the `false` comparison :). `/^http(s):/` allows for both `http` and `https` protocol.

src/drive/ducks/services/components/SuggestionProvider.jsx
y-lohse: I would say yes? I feel like it's a good one line summary of the next 30 lines of code.

PR #455, opened 2017-11-02
feat: upload status with progress bar

enguerran:
The target visuals are https://marvelapp.com/507f02i/screen/32722205

I tried to refactor some code and to put things together:

- [x] explicit configureStore (https://github.com/cozy/cozy-drive/pull/455/commits/26c56a83061fd1747d41c67f60ef6d43c9dcf3ea) (see #440)
- [x] mediabackup is a duck (reducer and action creators in the same file)
- [x] create a new component with a progress bar
- [x] use a thumbnail with **android**
- [ ] ~use a thumbmail with **iOS** (must update the cordova plugin)~
- [ ] ~use a bytes progress bar for each photo (unable to get already uploaded bytes from the cordova plugin)~
- [x] beautiful css (I will ask for help)

Below is a animated screenshot
ezgif com-optimize


enguerran:
[(3) [DRI] 📱 Améliorer le retour visuel lors de la sauvegarde de photo](https://trello.com/c/7uFiqx4D/526-3-dri-%F0%9F%93%B1-am%C3%A9liorer-le-retour-visuel-lors-de-la-sauvegarde-de-photo)

src/drive/mobile/containers/UploadProgression.jsx
```diff
@@ -1,13 +1,66 @@
...
+    height: `${2}px`,
+    borderRadius: '0 1px 1px 0',
+    transition: `${0.4}s width, ${0.4}s background-color`,
+    backgroundColor: color || 'darkblue',
+    ...style
+  }
+  return (
+    <div style={containerStyle}>
+      <div style={barStyle} />
+    </div>
```
y-lohse: Shouldn't we use the `progress` element for the markup, and css for styling it ?

src/drive/mobile/containers/UploadProgression.jsx
```diff
@@ -1,13 +1,66 @@
...
+  )
+}
+
+const UploadProgression = ({ t, messageData, media }) => {
```
y-lohse: I think it would be clearer if the props here were `upload_counter, total_upload , filePath` instead of passing 2 objects.

src/drive/mobile/lib/network.js
```diff
@@ -20,9 +20,13 @@ const hasCordovaPlugin = () =>
 export const getConnectionType = () =>
   hasCordovaPlugin() ? navigator.connection.type : WIFI
 
-export const onWifi = connection => connection === WIFI
+const isWifi = connection => connection === WIFI
 
+// TODO: rename this function
+// There is no reason for a function in a `network.js` file
+// to be called something about `backup`...
```
y-lohse: The name isn't tooo bad either, but maybe it should be somewhere else then?

src/drive/mobile/store/configureStore.js
```diff
@@ -1,63 +1,6 @@
...
-      },
-      availableOffline: store.getState().availableOffline
-    })
-  )
-
-  return store
-}
+const configureStore = __DEVELOPMENT__
+  ? require('./configureStore.dev').default
+  : require('./configureStore.prod').default
```
y-lohse: Meh. Having 2 files for the store makes it very hard to spot the differences, but requires updating 2 files if we change the store. I think it's worth it if the 2 configs are very different, but I don't think they are?
Pinging in @goldoraf since I believe he had a strong opinion on this too.

src/drive/mobile/store/configureStore.dev.js
```diff
@@ -0,0 +1,45 @@
...
+  const ravenMiddleWare = RavenMiddleWare(
+    ANALYTICS_URL,
+    getReporterConfiguration()
+  )
+  const middlewares = [thunkMiddleware, ravenMiddleWare]
+  middlewares.push(loggerMiddleware)
+
+  // Enable Redux dev tools
+  const composeEnhancers =
+    (__DEVELOPMENT__ && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) || compose
```
y-lohse: Testing `__DEVELOPMENT__ ` is unnecessary here since the file is only loaded when `__DEVELOPMENT__ ` is truthy :)

src/drive/mobile/styles/uploadprogression.styl
```diff
@@ -1,12 +1,18 @@
-@require 'cozy-ui'
+.coz-progress {
+    height: 48px;
+    padding: 5px 16px;
+    box-sizing: border-box;
+    border-bottom: 1px solid #d6d8da;
+    transition: bottom 0.5s ease-out, top 0.5s ease-out, opacity 0.2s ease-out;
+    display: flex;
+}
```
y-lohse: We're not using colons and semi-colons elsewhere in the css codebase, so for consistency I think they should be removed here.

src/drive/mobile/containers/UploadProgression.jsx
enguerran: TIL https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

enguerran: Not sure it's very applicable as
- hard to style
- not very standard
(cf. https://css-tricks.com/html5-progress-element/)

y-lohse: We use them as part of the onboarding for mesinfos/cozy-beta, and they seem to work aright :) They are styled too!

src/drive/mobile/lib/network.js
enguerran: yes, that is true

src/drive/mobile/store/configureStore.js
enguerran: Ok, I shall remove this commit.

src/drive/mobile/styles/uploadprogression.styl
enguerran: I don't know what is the stylus syntax, that's such a shame.
But I think you are right.

enguerran:
@y-lohse Here is a new proposal:
- removed `configureStore.dev` and `configureStore.prod`
- change props value for the component
- add a 'uptodate' component to display the 'uptodate' message

@GoOz do you think you could do something for:
- [ ] prettify the css files
- [ ] add the icon `icon-upload-OK.svg` on the 'uptodate' component like below

src/drive/mobile/containers/UploadProgression.jsx
```diff
@@ -1,13 +1,87 @@
...
+    </div>
+  )
+}
+
+const UploadProgression = ({ t, current, total, media }) => {
+  return (
+    <div className={styles['coz-upload-status']}>
+      <Progress percent={percent(current, total)} />
+      {media.filePath ? (
+        <img className={styles['coz-progress-pic']} src={media.filePath} />
```
y-lohse: you didn't want to pass the `filePath` as a prop instead of the `media` object? I also just noticed the `img` tag has no `alt`. We don't have any very useful alt title available, but I suppose the file's name is better than noting.

src/drive/mobile/containers/UploadProgression.jsx
```diff
@@ -1,13 +1,87 @@
...
+      ) : (
+        <div
+          className={styles['coz-progress-pic']}
+          style={{ border: '1px solid darkgray', backgroundColor: 'lightGray' }}
```
y-lohse: Why are we using a combo of classes and inline styles here?

src/drive/mobile/ducks/mediaBackup/index.js
```diff
@@ -86,32 +67,39 @@ export const startMediaBackup = (dirName, force = false) => async (
...
+    dispatch(mediaUploadSucceed(photo))
     return
   } catch (_) {} // if an exception is throw, the file doesn't exist yet and we can safely upload it
 
   try {
-    const maxBackupTime = 1000 * 60 * 5 // 5 minutes
+    const MILLISECOND = 1
+    const SECOND = 1000 * MILLISECOND
+    const MINUTE = 60 * SECOND
+    const maxBackupTime = 5 * MINUTE
```
y-lohse: 🤣

src/drive/mobile/styles/uploadprogression.styl
```diff
@@ -1,12 +1,22 @@
-@require 'cozy-ui'
+.coz-upload-status {
+    height 48px
+    padding 5px 16px
+    box-sizing border-box
+    border-bottom 1px solid #d6d8da
+    transition bottom 0.5s ease-out, top 0.5s ease-out, opacity 0.2s ease-out
+    display flex
+}
```
y-lohse: There are still curly braces which we don't have elsewhere :(

GoOz:
@enguerran you want me to checkout your PR and do the changes, is that right? (true question)

src/drive/mobile/containers/UploadProgression.jsx
enguerran: I don't know, if `alt` is mandatory. What will be its purpose? A11y? Display `alt` if the `filepath` does not contain a image file?

src/drive/mobile/ducks/mediaBackup/index.js
enguerran: With good variable's names, I can remove comments.
I think this code is way more expressive than 
```js
const maxBackupTime = 1000 * 60 * 5 // 5 minutes
```
Maybe there is no need to be as explicit as it is, maybe a simple code like this is enough:
```js
const FIVEMINUTES = 300000
const maxBackupTime = FIVEMINUTES
```
But I want to show that a good naming worth any comments.

src/drive/mobile/styles/uploadprogression.styl
enguerran: I can't understand why there is such difference between css syntax and stylus syntax. What is the purpose? Where is the automatic format tool?

enguerran:
@GoOz As you can see in this PR, I am lousy at tweaking CSS and design rules.
And I would like to add an icon with a `` tag for the case we have to display the "up-to-date" component. But I don't know how to display this icon, I saw in the code base we use `background-image` to display filetype icons. I think I should do the same, but it's not working. And I don't even know why.

I thought you could take some times to 
- [ ] prettify the css files (check if everything is correct)
- [ ] add the icon `icon-upload-OK.svg` on the 'uptodate' component like below

GoOz:
@enguerran Ok then, wil do 😄

src/drive/mobile/containers/UploadProgression.jsx
y-lohse: > If we want thumbnails, it is not extraordinary to get everything about the file and not only the filepath. That is a small decision with no side effect.

Ok, that's good enough for me! And yes I was thinking of the `alt` for for accessibility reasons.

src/drive/mobile/containers/UploadProgression.jsx
y-lohse: But… `coz-progress-pic` is also not a global css rule, is it? It's part of the local css module.

On principle, I absolutely agree that this should be how we do things (albeit with utility classes and not inline styles). But it's not what we've been doing so far, am I don't think it's a good idea to start changing it here, without any global effort or even discussion.

> Let's say we want to change this square with a placekitten picture or any other default image with a real binary file. Do we want to change a css file? Or do we want to have everything in the same place?

The thing is, we're going to start looking in the css files, because that's where all of our styles are. I think the correct approach (because it would be consistent) here is to introduce a modifier class (`.coz-progress-pic.--placeholder` or something similar) in the css.

src/drive/mobile/ducks/mediaBackup/index.js
y-lohse: Yes I understood that :). I just think in this instance it's taking things too far, as

```
const maxBackupTime = 1000 * 60 * 5 // 5 minutes
```

…is more readable to me than

```
const MILLISECOND = 1
const SECOND = 1000 * MILLISECOND
const MINUTE = 60 * SECOND
const maxBackupTime = 5 * MINUTE
```

…but I don't mind it :)

src/drive/mobile/styles/uploadprogression.styl
y-lohse: I'm not even sure it's stylus per se, since stylus totally understands regular css syntax as well. I personally dislike it too — it doesn't change anything in terms or readability, and it actually takes me more time to write it like that.
You're right though, there should be an automatic formatting tool for this.

GoOz: The purpose is to write less syntax, more code. But I agree I'v never been fond of this syntax myself. Besides, my editor kinda do the CSS syntax automatically for me…

GoOz:
@enguerran I cleaned some things up, but it may need some changes when you'll add some feature. Let me know then.

src/drive/mobile/ducks/mediaBackup/index.js
enguerran: I can't agree more, it is way more readable.
But, what about:
```js
const maxBackupTime = 1000 * 60 * 60 * 5 // 5 minutes
```
or
```js
const maxBackupTime = 1000 * 60 * 15 / 2 // 5 minutes
```
or
```js
const maxBackupTime = 600 * 5 // 5 minutes
```

Such numbers are called *magic number*. Why magic? Because it comes from nowhere and need a comment to explain its existence. But comments are pledges you can't believe with closed eyes. If not, then you need time to understand what means this magic number and how the comment helps or not.

So I think it is less readable at the end of the day, because you can't trust what you read.

enguerran:
@GoOz I have no more features to add.

src/drive/mobile/ducks/mediaBackup/index.js
enguerran: Entre les étiquettes qu'on trouve sur les boîtes dans une cuisine (riz, pâtes, lentilles, etc.), les étiquettes qu'on trouve sur les sonnettes des immeubles, voire même la pancarte "peinture fraîche" sur un banc public. Je confirme : les commentaires sont des promesses.

enguerran:
It is very beautiful!

@GoOz Do we have _placeholder image_ for missing thumbnail?

Here are the screenshots:

localhost_8084_ galaxy s5 1
localhost_8084_ galaxy s5

enguerran:
Here with default image:

localhost_8084_ galaxy s5 2

enguerran:
With the correct height and the dodger-blue
localhost_8084_ galaxy s5 3

PR #458, opened 2017-11-06
Refactored sync code and added refetchCollections() action

goldoraf:


enguerran:
s/[tech] Refactored sync code and added reftechCollections() action/[tech] Refactored sync code and added refetchCollections() action/

src/lib/cozy-client/adapters/PouchdbAdapter.js
```diff
@@ -38,21 +44,45 @@ export default class PouchdbAdapter {
...
-    for (let doctype of this.doctypes) {
+  async startSync(dispatch, direction = SYNC_BIDIRECTIONAL) {
+    try {
+      const infos = await this.sync(dispatch, direction)
+      this.scheduleNextSync(dispatch, direction)
+      return infos
+    } catch (err) {
+      this.scheduleNextSync(dispatch, direction)
+      throw err
+    }
```
y-lohse: You could use a `finally` block here to ` this.scheduleNextSync(dispatch, direction)`

src/lib/cozy-client/adapters/PouchdbAdapter.js
```diff
@@ -38,21 +44,45 @@ export default class PouchdbAdapter {
...
+      )
+    )
+  }
+
+  async syncDoctype(doctype, replicationUrl, dispatch, direction) {
+    try {
       const seqNumber = await this.getSeqNumber(doctype)
-      await dispatch(startDoctypeSync(doctype, seqNumber))
-      this.syncDoctype(doctype, `${baseUrl}${doctype}`, dispatch, direction)
+      dispatch(startDoctypeSync(doctype, seqNumber))
```
y-lohse: Have you considered keeping the `dispatch` calls in `sync` instead of passing them all the way down? I feel like this function would be cleaner if it just returned a promise instead of doing that *and* the `dispatch` calls, but it's more or less the same anyway.

goldoraf: Yep, but it posed a number of problems... I agree that it's not cool to dispatch from here though, I'll find a solution later.

PR #460, opened 2017-11-07
chore: add cordova to dev dependencies

enguerran:
Two advantages:
1. fix what is the used version
1. do not use global install

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "babel-runtime": "^6.0.0",
     "btoa": "^1.1.2",
     "copy-webpack-plugin": "^4.0.1",
+    "cordova": "6",
     "css-loader": "^0.26.0",
     "css-mqpacker": "^4.0.1",
     "csswring": "^5.1.1",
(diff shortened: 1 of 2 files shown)


ptbrowne:
IMHO it would be better to specify more precisely (=pin) the cordova version. Otherwise you can have several contributors on different cordova versions and it will lead to difficult to catch bugs.

enguerran:
@ptbrowne if we trust semantic versioning of cordova, there is no risk. Especially if you use `yarn`: https://github.com/cozy/cozy-drive/pull/460/files#diff-8ee2343978836a779dc9f8d6b794c3b2R2070

ptbrowne:
> if we trust semantic versioning of cordova, there is no risk

Exactly, I have been bitten enough by Cordova not to trust them too much :)

ptbrowne:
It is true though that the lock file has the pinned version so we will know if it change. Good for me !

enguerran:
If we want we could upgrade npm to 5+ and use package.json.lock to be more explicit.

PR #476, opened 2017-11-10
Better sync selectors

goldoraf:


src/lib/cozy-client/slices/synchronization.js
```diff
@@ -36,14 +36,38 @@ const doctypeSync = (state = doctypeSyncInitialState, action) => {
   }
 }
 
-const synchronization = (state = {}, action) => {
+const syncInitialState = {
+  started: false,
+  initialStatus: 'pending',
```
y-lohse: Does it make sense to call it `initialStatus ` and not `status`?

src/lib/cozy-client/slices/synchronization.js
```diff
@@ -36,14 +36,38 @@ const doctypeSync = (state = doctypeSyncInitialState, action) => {
...
   switch (action.type) {
+    case START_SYNC:
+      return {
+        ...state,
+        started: true
+      }
+    case INITIAL_SYNC_OK:
+      return {
+        ...state,
+        initialStatus: 'ok'
```
y-lohse: I think we should define internal constants for `ok` and `error`.

src/lib/cozy-client/slices/synchronization.js
goldoraf: yes, because the status of some doctypes could change with the subsequent syncs

src/lib/cozy-client/slices/synchronization.js
goldoraf: It makes sense, but I find it can be harder to read in cases like this. Will do, though

src/lib/cozy-client/slices/synchronization.js
y-lohse: Aaaah, I see. Ok then!

src/lib/cozy-client/slices/synchronization.js
```diff
@@ -75,23 +99,19 @@ export const syncDoctypeError = (doctype, error) => ({
   error
 })
 
-// TODO: not ideal because it will return false if no offline doctypes have been declared in the client config
-export const hasSyncStarted = state =>
-  Object.keys(state.cozy.synchronization).length !== 0
+export const hasSyncStarted = state => state.cozy.synchronization.started
```
enguerran: Not sure we want to expose external reducers organization in this internal reducer.
One simple example:

```js
// for illustration this reducer is dumb and returns a specific state
const children = (state, action) => [
  {
    name: 'Bob',
    age: 8
  },
  {
    name: 'Alice',
    age: 10
  }
]

// the children reducer is combined somewhere in the external organization
const parent = combineReducers({ children })

// Then come the selectors
const getAgeFromChildrenPointOfView = (state, index) => state[index].age

const getAgeFromParentPointOfView = (state, index) => state.children[index].age
```

Both selectors are ok, but not written in the same level. If you put the selector in the parent's reducer scope, it makes sens to take in consideration the state organization (`children` is a property of the `state`). But if you put the selector in the children's reducer scope, then, it doesn't make sense to illustrate the external state organization.

We should rather put selectors in different levels with every level responsible to access the sublevel selector:

```js
const getAgeFromChildrenPOV = (state, index) => state[index].age

const getAgeFromParentPOV = (state, index) => getAgeFromChilrenPOV(state.children, index)
```

goldoraf: Right, I agree the `.cozy` here is a smell. We'll have to clean this up because we have this problem in other reducers, but it's outside the scope of this PR.

ptbrowne: I think the utility function `proxy` is useful :

enguerran: je ne vois pas bien l'avantage d'avoir
```js
const getAgeFromParentPOV = proxy(getAgeFromChildrenPOV, 'children')
```
plutôt que ça
```js
const getAgeFromParentPOV = (state, index) => getAgeFromChilrenPOV(state.children, index)
```

mis à part de se poser la question de ce que c'est que `proxy`.

ptbrowne: `getAgeFromParentPOV` n'est pas couplé à `state` dans ce cas.

ptbrowne: L'avantage n'est pas évident si on a seulement une méthode. Si on a plusieurs méthodes à proxier ca peut être intéressant.

```js
import customReducer, * as fromMyCustomReducer from 'reducers/custom'

export default combineReducers({
  custom: customReducer
})

const proxyAll = (methods, path) => mapValues(methods, method => proxy(method, path))

export { ...proxyAll(myCustomReducer, 'custom') } 
```

PR #479, opened 2017-11-13
configureStore merge

goldoraf:


src/drive/mobile/lib/reporter.js
```diff
@@ -1,21 +1,19 @@
 /* global __SENTRY_TOKEN__, __DEVELOPMENT__ */
 import Raven from 'raven-js'
 
-let isEnabled = false
```
enguerran: Remove this causes an error.

enguerran:
Aside from a full bench of tests, I cannot see if this PR causes regressions.

y-lohse:
Same here, too many changes to really understand what's going on with only the diff. I would even have missed the `isEnabled` thing that @enguerran pointed out.
I did a quick check out and of the PR and toured the app, didn't see anything broken.

enguerran:
J'ai cette erreur : 


Pour reproduire : 

```
git checkout 6def9001
rm -rf node_modules
yarn
yarn watch:drive:standalone
open localhost:8084
```

PR #481, opened 2017-11-13
feat: upload pause notification

y-lohse:


y-lohse:
[(3) 📱 Prévenir l'utilisateur que la sauvegarde est en pause lorsqu'il quitte l'application pendant une sauvegarde sur iOS](https://trello.com/c/IwPP3HKk/525-3-%F0%9F%93%B1-pr%C3%A9venir-lutilisateur-que-la-sauvegarde-est-en-pause-lorsquil-quitte-lapplication-pendant-une-sauvegarde-sur-ios)

src/drive/mobile/lib/i18n.js
```diff
@@ -0,0 +1,12 @@
+import { getLang } from './init'
+import { _polyglot, initTranslation } from 'cozy-ui/react/I18n/translation'
+
+// used when translating things outside of components, eg. native UIs
+export const getTranslateFunction = () => {
```
enguerran: Later we may choose to put this code inside `cozy-ui/react/I18n/translation`, don't we?

y-lohse: Yeah maybe, but then it would require `lang` and `dictRequire ` as arguments, which would make it nearly the same as `initTranslation`. But maybe we can find an elegant way to do this.

PR #485, opened 2017-11-14
Update deps

ptbrowne:
1. After `yarn`, some dependencies were updated.

EDIT: had to remove pouch and pouchdb-find from my app dependencies

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2086,11 +2086,12 @@ cozy-bar@4.3.7:
     react-redux "^5.0.6"
     redux "^3.7.2"
 
-cozy-client-js@0.4.0:
-  version "0.4.0"
-  resolved "https://registry.yarnpkg.com/cozy-client-js/-/cozy-client-js-0.4.0.tgz#8317d9d076dfad9bb1eb6f97e113d0fa26a49d57"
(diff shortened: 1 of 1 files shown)


y-lohse:
Do you have a way to reproduce these errors? Since we're not using these modules directly, we shouldn't have to include them in our dependencies 😕

ptbrowne:
In bank, I am working on `cozy-client` so I linked and got these errors.

y-lohse:
Something is weird here, cc-ing @goldoraf 

These seem to be dependencies of `pouchdb-find`, so they should be installed automatically. IMO we shouldn't add them to our dependency list if they aren't used.

If somehow cozy-client needs them, they should be added to the [cozy-client dependency list](https://github.com/cozy/cozy-drive/blob/master/src/lib/cozy-client/package.json), not the main one.

I also note that the version of `pouchdb` and `pouchdbfind` are not the same in [cozy-client's package.json](https://github.com/cozy/cozy-drive/blob/master/src/lib/cozy-client/package.json) and [the main package.json](https://github.com/cozy/cozy-drive/blob/master/package.json#L164). Not sure why that is the case but it may be linked to this problem.

PR #486, opened 2017-11-14
🚑 Fix upload previews on mobile

GoOz:
Resize, crop & fit image previews while uploading them.

### DIFF
--- a/src/drive/mobile/styles/uploadprogression.styl
+++ b/src/drive/mobile/styles/uploadprogression.styl
@@ -33,9 +33,11 @@
         position absolute
         top 50%
         left 50%
-        max-width 3rem
-        max-height 3rem
         transform translateX(-50%) translateY(-50%)
+        width 3rem
+        height 3rem
+        object-fit cover
+        object-position center
 
 .coz-upload-status-content
     flex auto


enguerran:
[(3) [DRI] 📱 Améliorer le retour visuel lors de la sauvegarde de photo](https://trello.com/c/7uFiqx4D/526-3-dri-%F0%9F%93%B1-am%C3%A9liorer-le-retour-visuel-lors-de-la-sauvegarde-de-photo)

src/drive/mobile/styles/uploadprogression.styl
```diff
@@ -33,9 +33,11 @@
         position absolute
         top 50%
         left 50%
-        max-width 3rem
-        max-height 3rem
         transform translateX(-50%) translateY(-50%)
+        width 3rem
+        height 3rem
+        object-fit cover
```
y-lohse: TIL. But are you sure about this change? It looks like `cover` might [crop the image](https://developer.mozilla.org/fr/docs/Web/CSS/object-fit)?

GoOz: That's the point. It's just a preview to know which image we're talking about but that's pretty much it and any image that is not a square will be cropped.

y-lohse: Oooh sorry, of course. Somehow in my mind this was a fix for the viewer in photo so I was a bit concerned. My bad!

PR #482, opened 2017-11-13
fix: remove unused errorType

enguerran:
As I would like to fix a [label bug](https://trello.com/c/IzDE90L5/550-pho-%F0%9F%90%9B-libell%C3%A9s-non-traduits-sur-l%C3%A9cran-de-partage-public-dalbums), I discover duplication.

For now, I moved all related files into one global folder.
But I hope we could refactor components' and style's code into more simple one.

PR #488, opened 2017-11-14
fix: misuse of empty labels

enguerran:
I misread the information when I worked on https://github.com/cozy/cozy-drive/pull/482

Here is a revert and a fix for empty shared album:

- revert: use `photosContext`
- fix: add 2 i18n keys for this specific case
- remove: the HOC was too noisy for this component, I removed it

_Note: I think one should rethink the abstraction of the _Empty_ component (The _Error_ one is similarly affected). What is an `<Empty />` component responsible for? It displays one big picture, one h2 title and one paragraph with details. As for now, it receives a single prop called `emptyType`, that means this component is responsible for steering the rendering towards the correct _«type»_ (what is a type? that is too mysterious), but he does not steer anything toward anywhere. We should change that._

### DIFF
--- a/src/components/Error/Empty-photos.jsx
+++ b/src/components/Error/Empty-photos.jsx
@@ -3,18 +3,13 @@ import styles from './empty-photos'
 import React from 'react'
 import { translate } from 'cozy-ui/react/I18n'
 
-export const Empty = ({ t }) => {
+const Empty = ({ t, emptyType }) => {
   return (
     <div className={styles['pho-empty']}>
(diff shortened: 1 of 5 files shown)


enguerran:
[[PHO] 🐛 Libellés non traduits sur l’écran de partage public d’albums](https://trello.com/c/IzDE90L5/550-pho-%F0%9F%90%9B-libell%C3%A9s-non-traduits-sur-l%C3%A9cran-de-partage-public-dalbums)

PR #490, opened 2017-11-15
Feat/better onboarding

GoOz:
Few improvements:

- [x] Added a `label` before the Cozy address' `input`.
- [x] More explicite example & errors.
- [x] The whole compliant with Cozy's styleguide.
- [x] ~~Added a "**https://**" prefix in front of the `input` and removed the auto-updated URL value to make the protocol implicit.~~
- [x] Select & focus the field when there's an error.
- [x] Added a spinner while we're fetching the server and its availability.

Basically it looks now like this:
onboarding

### DIFF
--- a/src/drive/mobile/containers/onboarding/SelectServer.jsx
+++ b/src/drive/mobile/containers/onboarding/SelectServer.jsx
@@ -7,10 +7,19 @@ import { translate } from 'cozy-ui/react/I18n'
 
 import { registerDevice, setUrl } from '../../actions/settings'
 import styles from '../../styles/onboarding'
+import Spinner from 'cozy-ui/react/Spinner'
 
 export class SelectServer extends Component {
   componentDidMount() {
     this.serverInput.focus()
+  }
+
+  componentDidUpdate() {
+    if (this.props.error) {
+      this.serverInput.focus()
+      this.serverInput.select()
+    }
+    console.log(this.state)
   }
 
   render() {
@@ -22,6 +31,7 @@ export class SelectServer extends Component {
       serverUrl,
       error
     } = this.props
+    const { fetching } = this.state
     return (
       <div className={classNames(styles['wizard'], styles['select-server'])}>
         <header className={styles['wizard-header']}>
@@ -37,7 +47,7 @@ export class SelectServer extends Component {
           >
             <div className={styles['cozy-logo-white']} />
           </div>
-          <label className={styles['label']}>
+          <label className={styles['coz-form-label']}>
             {t('mobile.onboarding.server_selection.label')}
           </label>
           <input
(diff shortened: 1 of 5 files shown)


src/drive/mobile/containers/onboarding/SelectServer.jsx
```diff
@@ -37,39 +64,46 @@ export class SelectServer extends Component {
...
           )}
         </div>
         <footer className={styles['wizard-footer']}>
           <button
             role="button"
             className={classNames(styles['c-btn'], styles['c-btn--regular'])}
-            onClick={() => selectServer(serverUrl)}
+            onClick={() => this.setServerUrl(serverUrl)}
             disabled={error || !serverUrl}
+            aria-busy={fetching ? true : false}
```
GoOz: I first wrote `aria-busy={fetching}` but I wondered if it was explicit enough. You tell me.

src/drive/mobile/containers/onboarding/SelectServer.jsx
```diff
@@ -9,10 +9,36 @@ import { registerDevice, setUrl } from '../../actions/settings'
 import styles from '../../styles/onboarding'
 
 export class SelectServer extends Component {
+  constructor(props) {
```
y-lohse: I forgot about this. Actually, you can just do

```
export class SelectServer extends Component {
state = {fetching: false}
```

and then you don't even need a constructor.

src/drive/mobile/containers/onboarding/SelectServer.jsx
```diff
@@ -9,10 +9,36 @@ import { registerDevice, setUrl } from '../../actions/settings'
...
+  }
+
   componentDidMount() {
     this.serverInput.focus()
   }
 
+  componentDidUpdate() {
+    if (this.props.error) {
+      this.serverInput.focus()
+      this.serverInput.select()
```
y-lohse: What does `.select()` do?

src/drive/mobile/containers/onboarding/SelectServer.jsx
GoOz: oh right ok then :)

src/drive/mobile/containers/onboarding/SelectServer.jsx
GoOz: It selects the text in the input. Unless I was mislead, `.focus()` just put the cursor inside the field, ready to write and `select()` selects the string, ready to delete. 
I was asked to add this feature. :)

y-lohse: Cool, makes sense then!

GoOz:
[(1 + 0,5j) 📱 Améliorer l'écran d'erreur de la saisie de l'adresse du Cozy pour indiquer qu'il est possible de modifier l'adresse](https://trello.com/c/9Hun2kIg/563-1-05j-%F0%9F%93%B1-am%C3%A9liorer-l%C3%A9cran-derreur-de-la-saisie-de-ladresse-du-cozy-pour-indiquer-quil-est-possible-de-modifier-ladresse)

PR #493, opened 2017-11-17
Added sharing features to Drive

goldoraf:


src/sharing/components/SharedBadge.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import classNames from 'classnames'
+import styles from './badge'
+
+const SharedBadge = ({ byMe, className, small, xsmall }) => (
```
enguerran: Je ne suis pas fan de cette abstraction pour gérer les styles.
Finalement on a 3 propriétés, dont 2 mutuellement exclusives.

src/lib/cozy-client/CozyClient.js
```diff
@@ -61,9 +80,9 @@ export default class CozyClient {
     return this.facade.getAdapter(doctype)
   }
 
-  async fetchCollection(name, doctype, options = {}, skip = 0) {
+  async fetchDocuments(queryName, doctype, options = {}, skip = 0) {
```
y-lohse: I think you didn't rename all the previous calls? For exemple there's one in the `SuggestionProvider` that hasn't been changed I think.

goldoraf: Good catch, thanks!

PR #497, opened 2017-11-20
fix: respect network condition setting

y-lohse:
Most of the changes are just renaming variables for clarity, but the behavior should be identical. I'll highlight the one part that really changed below.

### DIFF
--- a/src/drive/mobile/ducks/mediaBackup/index.js
+++ b/src/drive/mobile/ducks/mediaBackup/index.js
@@ -27,32 +27,35 @@ const getDirID = async path => {
   return _id
 }
 
-const canBackup = (force, getState) => {
+const canBackup = (isManualBackup, getState) => {
   const backupOnAnyNetworkType = !getState().mobile.settings.wifiOnly
 
   return (
-    force ||
-    (getState().mobile.settings.backupImages &&
-      (isWifi() || backupOnAnyNetworkType))
+    (isManualBackup || getState().mobile.settings.backupImages) &&
+    (isWifi() || backupOnAnyNetworkType)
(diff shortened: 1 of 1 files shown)


src/drive/mobile/ducks/mediaBackup/index.js
```diff
@@ -27,32 +27,35 @@ const getDirID = async path => {
...
-const canBackup = (force, getState) => {
+const canBackup = (isManualBackup, getState) => {
   const backupOnAnyNetworkType = !getState().mobile.settings.wifiOnly
 
   return (
-    force ||
-    (getState().mobile.settings.backupImages &&
-      (isWifi() || backupOnAnyNetworkType))
+    (isManualBackup || getState().mobile.settings.backupImages) &&
+    (isWifi() || backupOnAnyNetworkType)
```
y-lohse: This is the part that really changes. Before, when the user ordered a manual back up, we didn't respect the network setting (ie. if there was no wifi, it would still back up). Now in order to do the backup, you need:

1. Triggered by the automatic setting or a manual command
2. A WiFi connection, or enable backups without WiFi

y-lohse:
[(1) 📱 Afficher un message lorsque le wifi n'est pas activé et que des photos sont à sauvegarder](https://trello.com/c/Jzh4Bsoa/564-1-%F0%9F%93%B1-afficher-un-message-lorsque-le-wifi-nest-pas-activ%C3%A9-et-que-des-photos-sont-%C3%A0-sauvegarder)

enguerran:
We get rid of the `force` variable :hurray:

PR #498, opened 2017-11-20
fix: Share status column is now hidden on mobile

GoOz:
And I merged shared status and offline status into one column actually to avoid potential layout issues.


GoOz:
[🐛📱 Un "-" est affiché sur la liste des fichiers mobiles](https://trello.com/c/zwIx1BWU/593-%F0%9F%90%9B%F0%9F%93%B1-un-est-affich%C3%A9-sur-la-liste-des-fichiers-mobiles)

PR #505, opened 2017-11-21
fix: hide cozy to cozy sharing behind a flag

y-lohse:
- Only shows the share by email tab if the URL contains the magic `sharingiscaring` word
- Make the link sharing the default tab


y-lohse:
[(1) Cacher le partage Cozy à Cozy derrière un flag](https://trello.com/c/IHz8ilSz/590-1-cacher-le-partage-cozy-%C3%A0-cozy-derri%C3%A8re-un-flag)

PR #507, opened 2017-11-21
Quickfixed reporting error on web

goldoraf:


src/drive/mobile/lib/reporter.js
```diff
@@ -1,14 +1,19 @@
...
-const PROD_ANALYTICS_URL = `https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/6`
-const DEV_ANALYTICS_URL = `https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/2`
-export const ANALYTICS_URL = __DEVELOPMENT__
-  ? DEV_ANALYTICS_URL
-  : PROD_ANALYTICS_URL
-
 let isEnabled = false
 
+const getAnalyticsUrl = () => {
+  if (window.__SENTRY_TOKEN__ === undefined) {
```
y-lohse: Do we need to use `window` here? `__SENTRY_TOKEN__ ` is provided as a global variable by webpack, but it's not *really* related to window.

y-lohse:
👍

PR #508, opened 2017-11-21
Fixed sharing link on mobile

goldoraf:
On mobile, the sharing link was broken because `window.location.origin` returns `file:///, obviously. I'm not fan of composing the url in the UI component, but unfortunately we have no access to the client instance in the selectors.

### DIFF
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -12,7 +12,6 @@ export const getLang = () =>
 export const initClient = url => {
   const offline = { doctypes: ['io.cozy.files'] }
   if (isIos()) offline.options = { adapter: 'cordova-sqlite' }
-  console.log('test')
   return new CozyClient({
     cozyURL: url,
     oauth: {
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -12,6 +12,7 @@ export const getLang = () =>
 export const initClient = url => {
   const offline = { doctypes: ['io.cozy.files'] }
   if (isIos()) offline.options = { adapter: 'cordova-sqlite' }
+  console.log('test')
   return new CozyClient({
     cozyURL: url,
     oauth: {
@@ -29,6 +30,7 @@ export const initClient = url => {
         scopes: [
           'io.cozy.files',
           'io.cozy.contacts',
+          'io.cozy.apps',
           'io.cozy.jobs:POST:sendmail:worker'
         ]
       }
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -29,7 +29,7 @@ export const initClient = url => {
         scopes: [
           'io.cozy.files',
           'io.cozy.contacts',
-          'io.cozy.apps',
+          'io.cozy.apps:GET',
           'io.cozy.jobs:POST:sendmail:worker'
         ]
       }
--- a/src/lib/cozy-client/DataAccessFacade.js
+++ b/src/lib/cozy-client/DataAccessFacade.js
@@ -13,11 +13,13 @@ const FILES_DOCTYPE = 'io.cozy.files'
 
 export default class DataAccessFacade {
   constructor() {
+    this.url = null
     this.stackAdapter = new CozyStackAdapter()
     this.pouchAdapter = new PouchdbAdapter()
   }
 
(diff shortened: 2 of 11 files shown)


src/lib/cozy-client/slices/sharings.js
```diff
@@ -425,7 +425,7 @@ const getSharingRecipients = (state, sharings) =>
     .reduce((a, b) => a.concat(b), [])
 
 const buildSharingLink = (id, doctype, sharecode) =>
-  `${window.location.origin}/public?sharecode=${sharecode}&id=${id}${
+  `/public?sharecode=${sharecode}&id=${id}${
```
y-lohse: Would it be possible tu return the full link here ? It seems weird that `buildSharingLink ` returns a relative link, but I don't know what's possible.

goldoraf: In this function, which is called by a selector, we have no access to the client instance and though to the cozy url. I can't see a quick, better way of doing.

goldoraf:
@y-lohse I agree it's not cool to not return the full link, so I made another try. This is better, even if we access a slice of store's state we don't own, because at least the problem is encapsulated in cozy-client. The part where we transform the host url into the app url is not ideal too, but we'll see in the future.

src/lib/cozy-client/slices/sharings.js
```diff
@@ -424,8 +424,35 @@ const getSharingRecipients = (state, sharings) =>
...
+}
+
+const getAppUrl = (state, doctype) => {
+  // TODO: here we access a slice of state we don't "own" and that's a pb
+  // we should store the cozy url in the store ourselves
+  if (state.mobile && state.mobile.settings) {
+    const host = state.mobile.settings.serverUrl
+    const app = getAppForDoctype(doctype)
+    return host.replace(
+      /^(https?:\/\/)(\w+)\.(\w+)\.(\w+)/g,
```
y-lohse: I think this regex is fragile, because it assumes the server url follows the form `https://{name}.{provider}.{tld and the rest}`. While this is true for the cozy we provide, it's may not always be like this, especially for self-hosted users. But I don't know enough about how the apps URL are handled, maybe it would be safer to check this with the back-end team.

enguerran: If I serve cozy-stack directly on the main domain of mine: `http://ticabri-cozy.com` it breaks.

src/lib/cozy-client/DataAccessFacade.js
```diff
@@ -37,6 +39,10 @@ export default class DataAccessFacade {
     }
   }
 
+  getUrl() {
+    return this.url
+  }
```
enguerran: Est-ce qu'on a vraiment besoin d'un getter ?

src/lib/cozy-client/slices/sharings.js
```diff
@@ -424,8 +424,35 @@ const getSharingRecipients = (state, sharings) =>
...
+    case 'io.cozy.photos.albums':
+      return 'photos'
+    default:
+      throw new Error(
+        `Sharing link: don't know which app to use for doctype ${doctype}`
+      )
+  }
+}
+
+const getAppUrl = (state, doctype) => {
```
enguerran: Est-ce que cette fonction ne retourne pas simplement `window.location.origin` ?

goldoraf:
@y-lohse I rewrote the thing using the list of apps retrieved from the stack. The challenging part was to ensure this list is fetched and data is available when needed, so I modified the middleware in order to allow a new action property `dependencies` with a list of required queries. This is not perfect and will probably be removed in the future, but it was the simpler and less intrusive way I found.

src/lib/cozy-client/DataAccessFacade.js
goldoraf: Dans la mesure où c'est une classe et non un objet litéral, j'aime autant avoir un getter qu'accéder aux propriétés directement, mais ça se discute ;)

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -29,6 +29,7 @@ export const initClient = url => {
         scopes: [
           'io.cozy.files',
           'io.cozy.contacts',
+          'io.cozy.apps',
```
y-lohse: Maybe `io.cozy.apps:GET`, that's what the manifest has, and we really don't want permissions to write on that doctype^^

goldoraf: Fixed

PR #509, opened 2017-11-21
feat: display an icon in the search bar results

enguerran:
I changed the procedure:
1. load every `icon-type` file with [file-loader](https://github.com/webpack-contrib/file-loader)
1. build a mapping object : `name => icon-file-name`
1. map the mime-type to an icon file
1. warn in the console no mapping found
1. build an url
1. pass the url to the searchBar's `renderSuggestion` function

# 1. load every `icon-type`

```js
const iconsContext = require.context(
  '../../../assets/icons/',
  false,
  /icon-type-.*.svg$/
)
```

# 2. build an object

```js
const icons = iconsContext.keys().reduce(
  (acc, item) => ({
    ...acc,
    [item.replace(/\.\/icon-type-(.*)\.svg/, '$1')]: iconsContext(item)
  }),
  {}
)
```

# 3. map `mime-type => icon-file`

```js
const mapMimeIcon = {
  'image/jpeg': 'image',
  'image/png': 'image',
  'application/pdf': 'pdf'
}
```

# 4. warn if mapping's missing

_and add a default icon_

```js
const keyIcon =
  mapMimeIcon[mimetype] ||
  console.warn(
    `No icon found, you may need to add a mapping for ${mimetype}`
  ) ||
  'files'
```

# 5. and 6. are easy to understand

# Screenshot

_example with pdf mapping missing_

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "eslint-plugin-react": "^7.3.0",
     "eslint-plugin-standard": "^3.0.1",
     "extract-text-webpack-plugin": "^3.0.0",
-    "file-loader": "^0.9.0",
+    "file-loader": "^1.1.5",
     "git-directory-deploy": "^1.5.0",
     "html-webpack-plugin": "^2.30.1",
     "husky": "^0.14.3",
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -31,7 +31,8 @@ class SuggestionProvider extends React.Component {
           title: result.name,
           subtitle: result.path,
           term: result.name,
-          onSelect: 'open:' + result.url
+          onSelect: 'open:' + result.url,
+          icon: result.icon
         }))
       },
       intent.attributes.client
@@ -66,7 +67,8 @@ class SuggestionProvider extends React.Component {
             id: file._id,
             name: file.name,
             path,
-            url: window.location.origin + '/#/folder/' + dirId
+            url: window.location.origin + '/#/folder/' + dirId,
+            icon: getIconUrl(file.mime)
           }
         })
 
@@ -75,6 +77,33 @@ class SuggestionProvider extends React.Component {
       resolve()
     })
   }
+}
+
+const iconsContext = require.context(
+  '../../../assets/icons/',
+  false,
+  /icon-type-.*.svg$/
+)
+const icons = iconsContext.keys().reduce((acc, item) => {
+  acc[item.replace(/\.\/icon-type-(.*)\.svg/, '$1')] = iconsContext(item)
+  return acc
+}, {})
+
+const mapMimeIcon = {
+  'image/jpeg': 'image',
+  'image/png': 'image',
+  'application/pdf': 'pdf'
+}
+
+function getIconUrl(mimetype) {
+  const keyIcon =
+    mapMimeIcon[mimetype] ||
+    console.warn(
+      `No icon found, you may need to add a mapping for ${mimetype}`
+    ) ||
+    'files'
+
+  return `${window.location.origin}/${icons[keyIcon]}`
 }
 
 export default SuggestionProvider
--- a/config/webpack.config.pictures.js
+++ b/config/webpack.config.pictures.js
@@ -17,7 +17,7 @@ module.exports = function(production) {
           exclude: /(sprites|icons)/,
           loader: 'file-loader',
           options: {
-            path: 'img',
+            outputPath: 'img',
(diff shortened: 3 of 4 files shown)


src/drive/containers/File.jsx
```diff
@@ -16,7 +16,10 @@ import breakpointsAware from 'cozy-ui/react/helpers/breakpoints'
 import { SharedBadge } from 'sharing'
 import { getSharingDetails } from 'cozy-client'
 
+import iconTypeFiles from '../assets/icons/icon-type-files.svg'
+
 import { getFolderUrl } from '../reducers'
+console.log(iconTypeFiles)
```
y-lohse: This file shoudn't have change, should it?

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -66,7 +68,8 @@ class SuggestionProvider extends React.Component {
             id: file._id,
             name: file.name,
             path,
-            url: window.location.origin + '/#/folder/' + dirId
+            url: window.location.origin + '/#/folder/' + dirId,
+            icon: iconTypeFiles.id
```
y-lohse: Not a review comment at all : what does `iconTypeFiles.id` contain ?

enguerran:

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -77,4 +79,34 @@ class SuggestionProvider extends React.Component {
...
+    /icon-type-.*.svg$/
+  )
+  const icons = iconsContext.keys().reduce(
+    (acc, item) => ({
+      ...acc,
+      [item.replace(/\.\/icon-type-(.*)\.svg/, '$1')]: iconsContext(item)
+    }),
+    {}
+  )
+
```
ptbrowne: I am not a fan of this construct. IMHO it is nor very legible, nor performant (you are creating one object and trashing it at every iteration of the reduce). Even if performance is not an issue here, I would prefer if we would not create pressure on the GC during loading of the application.

lodash has a `fromPairs` function that is useful in this case. It is easy to recode it if you want to avoid the dependency on lodash.

```js
const fromPairs = (arr, pairs) => {
  const out = {}
  pairs.forEach(([k, v]) => out[k] = v)
  return out
}

const icons = fromPairs(iconsContext.keys().map(item =>
  [item.replace(/\.\/icon-type-(.*)\.svg/, '$1'), iconsContext(item)])
```

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -77,4 +79,34 @@ class SuggestionProvider extends React.Component {
...
+  const mapMimeIcon = {
+    'image/jpeg': 'image',
+    'image/png': 'image',
+    'application/pdf': 'pdf'
+  }
+
```
ptbrowne: `mapMimeIcon`, the `iconsContext` and `icons` variables should be outside the function otherwise you build them every time you need an icon.

src/drive/ducks/services/components/SuggestionProvider.jsx
enguerran: You are very right, I did not see it with _the nose on the steering_.
I first wrote: 
```js
const icons = iconsContext.keys().reduce((acc, item) => {
  acc[item.replace(/\.\/icon-type-(.*)\.svg/, '$1')] = iconsContext(item)
  return acc
}, {})
```

That seems enough to avoid what you remark.

ptbrowne: Yes, I am ok with this 👍

src/drive/ducks/services/components/SuggestionProvider.jsx
enguerran: Crap, I wanted to isolate the code to be easily movable.

package.json
```diff
@@ -125,7 +125,7 @@
     "stylint": "^1.5.9",
     "stylus": "^0.54.0",
     "stylus-loader": "^2.4.0",
-    "svg-sprite-loader": "^0.0.31",
+    "svg-sprite-loader": "3.4.1",
```
y-lohse: Do we still need this loader?

enguerran: No we don't, but I did not remove it in this PR because we need to do some work about tooling, like using create-cozy-app with cozy-scripts to manage dependencies, for instance.

enguerran: I update it anyway, because it was very outdated.

enguerran: Finally I reseted it... because it changed `yarn.lock` with a lot of new lines...

PR #510, opened 2017-11-22
fix: client links on mobile

y-lohse:
It's easier to review the 3 commits one after the other :

- the first one moves the `pushClient` component to the new shared `component` folder
- the second does some minor fixes
- the thrid one fixes the actual problem

What happened was, on mobile, the button would very quickly appear in the DOM, probably *under* the "later" button in the onboarding. So at the end of the onboarding, you press "later", but if you click again quickly or whatever, it also clicks the link and moves you to the download page.
I know there's a class with a media query and it shouldn't even appear on mobile, but somehow it did.
Anyway, I've changed the default value so that the button doesn't appear until it's sure it should, at which point everything is fine.


y-lohse:
[🐛📱La page de téléchargement de client s'ouvre à la fin de l'onboarding mobile](https://trello.com/c/seoxrVg1/603-%F0%9F%90%9B%F0%9F%93%B1la-page-de-t%C3%A9l%C3%A9chargement-de-client-souvre-%C3%A0-la-fin-de-lonboarding-mobile)

PR #512, opened 2017-11-23
fix: FirstUploadModal not displayed

enguerran:
In https://github.com/cozy/cozy-drive/pull/455 we renamed variables without modifying this call. Maybe we could use selectors here. But for now, let's fix the bug.

- https://github.com/cozy/cozy-drive/pull/455/files#diff-d173561dcc55260e5a4050bb22efe48eL32
- https://github.com/cozy/cozy-drive/pull/455/files#diff-d173561dcc55260e5a4050bb22efe48eL33

### DIFF
--- a/src/drive/mobile/containers/FirstUploadModal.jsx
+++ b/src/drive/mobile/containers/FirstUploadModal.jsx
@@ -54,13 +54,12 @@ const OnlyOnceFirstUploadModal = withPersistentState(
 const mapStateToProps = state => {
   const {
     uploading = false,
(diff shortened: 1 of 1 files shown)


enguerran:
[🐛📱 La modale de début de sauvegarde des médias n'est plus visible](https://trello.com/c/udilXh8I/609-%F0%9F%90%9B%F0%9F%93%B1-la-modale-de-d%C3%A9but-de-sauvegarde-des-m%C3%A9dias-nest-plus-visible)

PR #521, opened 2017-11-24
Fix upload modal flash

goldoraf:
This fixes the issue, but it's not very clean, I've tried to avoid the render if the state has not been restored in the `withPersistentState` HOC, but I couldn't make it work strangely...


enguerran:
See #522

PR #523, opened 2017-11-24
feat: searchbar uses icons from File.jsx

enguerran:
I refactor code from `File.jsx` to be able to get icon for `SuggestonProvider.jsx`.

### DIFF
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
+import { getTypeFromMimeType } from '../../../containers/File'
 
 class SuggestionProvider extends React.Component {
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


enguerran:

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
+import { getTypeFromMimeType } from '../../../containers/File'
```
y-lohse: Maybe we should move this function somewhere else? Not sure where, but it feels weird to import a utility from a component.

enguerran: I asked myself the same question. And I concluded with this: It's not a component, it's a `File` component's file with a component and tools about its competence. I am not sure this should be move somewhere related to its category rather than related to its responsability/competence.

y-lohse: > I am not sure this should be move somewhere related to its category rather than related to its responsability/competence.

I understand, although our code tens to be organized by category. But I don't have a better idea and you've thought about it, so that's ok for me.

goldoraf: Totally agree with @enguerran

PR #527, opened 2017-11-24
fix: some mimetypes are superset of existing mimetype

enguerran:
- application/vnd.oasis.opendocument.spreadsheet
- application/vnd.openxmlformats-officedocument.wordprocessingml.document
- application/octet-stream
- application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
- application/vnd.openxmlformats-officedocument.presentationml.presentation

### DIFF
--- a/src/drive/containers/File.jsx
+++ b/src/drive/containers/File.jsx
@@ -27,6 +27,7 @@ export const splitFilename = file =>
       }
 
 const mappingMimetypeSubtype = {
+  word: 'text',
   zip: 'zip',
   pdf: 'pdf',
   spreadsheet: 'sheet',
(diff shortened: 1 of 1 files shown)


enguerran:

src/drive/containers/File.jsx
```diff
@@ -41,7 +42,8 @@ export const getTypeFromMimeType = (collection, prefix = '') => mimetype => {
     return type
   }
   if (type === 'application') {
-    return mappingMimetypeSubtype[subtype]
+    const key = subtype.match(Object.keys(mappingMimetypeSubtype).join('|'))[0]
```
y-lohse: This is were a comment would be good :). The code is not explicit at all, but after reading it a few times I get what it does and why.

enguerran: Well:

1. we join the keys of `mappingMimetypeSubtype` with a `OR`
1. we try to match the string `subtype` and the joined string
1. we grab the first item of the result as it contains the matching substring

That is almost straightforward but it uses unusual JavaScript functions.

PR #528, opened 2017-11-24
fix: manage undefined mapping icons

enguerran:
Display `fil-file-file` if no mapping found (+ a message in the console)

### DIFF
--- a/src/drive/containers/File.jsx
+++ b/src/drive/containers/File.jsx
@@ -42,9 +42,12 @@ export const getTypeFromMimeType = (collection, prefix = '') => mimetype => {
     return type
   }
   if (type === 'application') {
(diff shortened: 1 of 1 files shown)


enguerran:

src/drive/containers/File.jsx
```diff
@@ -53,7 +55,12 @@ export const getClassFromMime = attrs => {
   }
 
   return styles[
-    'fil-file-' + getTypeFromMimeType(styles, 'fil-file-')(attrs.mime)
+    'fil-file-' +
+      (getTypeFromMimeType(styles, 'fil-file-')(attrs.mime) ||
+        console.warn(
```
goldoraf: Clever trick, maybe too clever? (but still fine!)

enguerran: Not easy to read for a new comer, but this code is not central and it worth be concise.

PR #525, opened 2017-11-24
Added resetStore() to cozy-client

goldoraf:

PR #530, opened 2017-11-27
The Onboarding — Part I

y-lohse:
At the moment, the code that handles authentication is very coupled with the rest of the app. Our goal is to have a standalone module that could also be used in other apps, and this PR is a first step in that direction.

I'm adding some more comments for easier reviewing below.


src/authentication/Authentication.jsx
```diff
@@ -0,0 +1,83 @@
+import React, { Component } from 'react'
```
y-lohse: This component is displayed when the user needs to log in. Right now it switches between the "Welcome" screen and the "Log in" screen, but soon it will host all the other steps for registering a new cozy account.

src/authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
y-lohse: This is the main component of the module, the "public API" so to speak. It takes the app's "normal" routes and information about the user's current state (`isAuthenticated `, `isRevoked`, etc) ans toggles between different views.

src/authentication/Revoked.jsx
```diff
@@ -0,0 +1,45 @@
+/* global cozy */
```
y-lohse: Replaces the former `RevokableWrapper`. Note that it doesn't render the routes as childrens anymore as it xas a bit buggy.

src/authentication/steps/SelectServer.jsx
```diff
@@ -0,0 +1,162 @@
+/* global __ALLOW_HTTP__ */
```
y-lohse: Almost the same file as we already had, allows the user to enter the URL of the cozy. In this version, most of the URL validation happens inside the component and not in a redux action.

src/authentication/styles.styl
```diff
@@ -0,0 +1,109 @@
+@require 'cozy-ui'
```
y-lohse: This style is a subset of the `onboarding.styl` file, with only the parts useful for the auth screen. I'm planning on rewriting a large part of it later.

src/drive/components/AppRoute.jsx
```diff
@@ -12,15 +14,23 @@ import {
...
+  <Route>
+    <Route component={Layout}>
+      <Redirect from="/files/:folderId" to="/folder/:folderId" />
+      <Route component={FileExplorer}>
+        <Redirect from="/" to="folder" />
+        <Route path="folder(/:folderId)" component={Folder} />
+        <Route path="recent" component={Recent} />
+        <Route path="trash(/:folderId)" component={Trash} />
+      </Route>
+      {__TARGET__ === 'mobile' && (
```
y-lohse: The main router now directly includes the mobile views. Although I think this could be changed as we'll see later.

src/drive/mobile/actions/settings.js
```diff
@@ -46,80 +46,10 @@ export const setBackupContacts = backupContacts => ({
...
-    .catch(err => {
-      if (err.message === 'REGISTRATION_ABORT') {
-        cozy.client._storage.clear()
-      } else {
-        dispatch(wrongAddressError())
-        logException(err)
-      }
-      throw err
-    })
+export const saveCredentials = (client, token) => (dispatch, getState) => {
```
y-lohse: This action creator is only called [here](https://github.com/cozy/cozy-drive/pull/530/files#diff-8278b5d865d4e4819d89588ae0ea7788R47), I wonder if we should just dispatch the individual actions.

src/drive/mobile/containers/DriveMobileRouter.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from 'react'
```
y-lohse: This file binds the Authentication module (which is application-agnostic) and the Cozy Drive app.

src/drive/mobile/containers/DriveMobileRouter.jsx
```diff
@@ -0,0 +1,53 @@
...
+        onAuthenticated={this.afterAuthentication}
+        onLogout={this.afterLogout}
+        allowRegistration={false}
+      />
+    )
+  }
+}
+
+const DriveMobileRouterWithRoutes = props => (
+  <DriveMobileRouter {...props} appRoutes={AppRoute} />
```
y-lohse: I think we could inject the mobile-specific routes here ?

src/drive/mobile/containers/OnBoarding.jsx
```diff
@@ -1,24 +1,28 @@
-import React from 'react'
```
y-lohse: The new onboarding only shows things specific to cozy-drive.

src/drive/mobile/containers/onboarding/Analytics.jsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
 import { connect } from 'react-redux'
-import OnBoarding from '../../components/OnBoarding'
```
y-lohse: We had 2 components called `OnBoarding`, one dumb and one container. I renamed the dumb component.

src/drive/mobile/styles/onboarding.styl
```diff
@@ -25,36 +25,6 @@
             color           cool-grey
             margin-left     auto
 
-        .close-button
```
y-lohse: I removed the parts that are used for the `Welcome` and `SelectServer` screens.

targets/drive/mobile/main.jsx
```diff
@@ -122,10 +110,9 @@ const renderAppWithPersistedState = (persistedState = {}) => {
       dictRequire={lang => require(`drive/locales/${lang}`)}
     >
       <CozyProvider store={store} client={client}>
-        <Router
+        <DriveMobileRouter
           history={hashHistory}
```
y-lohse: I'm not very happy that we pass the `history` prop here and all other props in [DriveMobileRouter](https://github.com/cozy/cozy-drive/pull/530/files#diff-8278b5d865d4e4819d89588ae0ea7788R1)…

y-lohse:
### Auth

There is a new module called `authentication` that is very close to what we have on Cozy Bank. It's role is to authenticate the user whenever needed. At the moment, it only initializes Cozy Client — it doesn't configure it, doesn't care about the cozy-bar or how informations are stored.

### Log in

There are 3 places where we log in the user:

- if we have the right infos, in the [main.jsx](https://github.com/y-lohse/cozy-drive/blob/4d87e6c03fd2df3cd845da022f6885dd23703209/targets/drive/mobile/main.jsx#L57)
- if we just learned the server url, we do it [here](https://github.com/cozy/cozy-drive/pull/530/files#diff-57d5a0284415891d91dbcc4f3dc528c1R41)
- if the user was revoked and logs back in, we do it [here](https://github.com/cozy/cozy-drive/pull/530/files#diff-684c1f3266cb71afe9fa22f6b9f96806R17)

### Init cozy bar

- again, if the user is already logged in, we init the bar in the [main.jsx](https://github.com/y-lohse/cozy-drive/blob/4d87e6c03fd2df3cd845da022f6885dd23703209/targets/drive/mobile/main.jsx#L60)
- otherwise we di it at [the end of the onboarding](https://github.com/cozy/cozy-drive/pull/530/files#diff-bc42a1dc47630dd5658bca30af4614dcR16)

### Unregister

There are 2 situations where we need to reset the app:

1. The user taps the button in settings -> we call [`resetClient`](https://github.com/cozy/cozy-drive/pull/530/files#diff-ab1bd3d381dde64a9940cd3f2da52522R13) and update the redux store
2. The user is revoked and decides to log out : we do [the same thing](https://github.com/cozy/cozy-drive/pull/530/files#diff-8278b5d865d4e4819d89588ae0ea7788R48)

`resetClient` is [almost the same](https://github.com/cozy/cozy-drive/pull/530/files#diff-87c5ff644d44594b5bc0fc5cac02e376R55), but now all calls are in the same place, called from the same action creator, no one is messing around with the routes or what not. We still need to integrate #525 after this PR but it should be much easier.

y-lohse:
[(1) [DRI] 📱 Afficher un message d'erreur spécifique lorsque l'url contient "cosy"](https://trello.com/c/XXggEI0C/519-1-dri-%F0%9F%93%B1-afficher-un-message-derreur-sp%C3%A9cifique-lorsque-lurl-contient-cosy)

src/drive/components/AppRoute.jsx
enguerran: This will cause a merge conflict with #534

PR #531, opened 2017-11-27
fix: cozy bar's apps on drive

GoOz:
Temporary fix was still here even though it wasn't needed anymore and actually it was messing with app list in the cozy-bar.


GoOz:
@y-lohse Well, it was. There was comments. Do you have another way in mind?

y-lohse:
No I mean, saying *what* it was fixing. eg. `temporary fix for cozy bar not showing in safari3` or something like that :)

GoOz:
aaaaah yeah ok, you're right :)

PR #534, opened 2017-11-28
Use Photos viewer in Drive

goldoraf:
In this PR, I extracted the viewer from Photos and made it more generic so that it can handle all file types and be used in Drive. Routing is not the responsibility of the viewer anymore: there is a wrapper per app that handle routing and prepare the files for consumption by the viewer (https://github.com/cozy/cozy-drive/pull/534/files#diff-9c5af223a322fc50792bb6ddec261a18 for Drive).

Gestures that were handled by the main component are now handled by the ImageViewer, but a `withSwipe` HOC should be used by other viewers to allow swipe on all file types.


enguerran:
[(5) Ajouter la visionneuse d'images dans Drive](https://trello.com/c/1QG5e74x/592-5-ajouter-la-visionneuse-dimages-dans-drive)

enguerran:
[🐛[PHO] 🐛 Ouvrir une page directement sur la visionneuse ne marche pas](https://trello.com/c/dchBRISi/462-pho-%F0%9F%90%9B-ouvrir-une-page-directement-sur-la-visionneuse-ne-marche-pas)

src/drive/components/AppRoute.jsx
```diff
@@ -16,7 +17,12 @@ const AppRoute = (
     <Redirect from="/files/:folderId" to="/folder/:folderId" />
     <Route component={FileExplorer}>
       <Redirect from="/" to="folder" />
-      <Route path="folder(/:folderId)" component={Folder} />
+      <Route path="folder" component={Folder}>
+        <Route path="file/:fileId" component={FileViewer} />
+      </Route>
+      <Route path="folder/:folderId" component={Folder}>
```
y-lohse: I think we avoid the duplication here by using an optional parameter, such as `<Route path="folder/(:folderId)" component={Folder}>`

src/drive/components/FolderView.jsx
```diff
@@ -113,6 +119,15 @@ class FolderView extends Component {
       </Main>
     )
   }
+
+  renderViewer(children) {
```
y-lohse: Sort of secondary, but why is this called `renderViewer `? Is there a viewer involved?

src/drive/components/FileViewer.jsx
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    if (this.props.files.length === 0) return null
+    const files = this.props.files.filter(f => f.type !== 'directory')
+    const { params, router } = this.props
+    const currentIndex = files.findIndex(f => f.id === params.fileId)
+    return (
+      <Viewer
+        files={files}
+        currentIndex={currentIndex}
+        onChange={nextFile =>
```
y-lohse: Neat 👍

src/viewer/Viewer.jsx
```diff
@@ -0,0 +1,129 @@
...
+      this.props.onChange(nextFile, nextIndex)
+    }
+  }
+
+  render() {
+    const { files, currentIndex, onClose } = this.props
+    const currentFile = files[currentIndex]
+    const fileCount = files.length
+    const hasPrevious = currentIndex > 0
+    const hasNext = currentIndex < fileCount - 1
```
y-lohse: I may be wrong, but I believe the current behavior is to keep rotating the photos, ie. if you click "next" while on viewing the last file in a collection, is goes to the first.

src/viewer/withSwipe.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+const withSwipe = WrappedComponent => {
+  return class Wrapper extends Component {
+    componentDidMount() {
+      this.gesturesHandler = new Hammer(React.findDOMNode(this.wrapped))
+      this.gesturesHandler.on('swipe', this.onSwipe)
+    }
+
+    onSwipe = e => {
+      console.log(e)
```
y-lohse: Extra log here

src/drive/components/AppRoute.jsx
goldoraf: I tried, but it doesn't work: apparently an optional parameter must be on a leaf route and can't have children

src/drive/components/FolderView.jsx
goldoraf: As the viewer is instantiated by a child route, it is passed to this component as a child and the `renderViewer` method passes the `files` prop required by the viewer. I'm not sure I'm clear here...

src/viewer/Viewer.jsx
goldoraf: I don't think so, I'll ask the PO ;)

src/viewer/withSwipe.jsx
goldoraf: Woops...

src/drive/components/FolderView.jsx
y-lohse: Yeah ok I understand now, but the code doesn't make it obvious. Since technically the `children` could be anything, I think we shouldn't refer to a viewer here. Or maybe adding a comment about what's going on.

src/photos/components/AppRoute.jsx
```diff
@@ -5,21 +5,21 @@ import Layout from './Layout'
...
+import PhotosViewer from '../components/PhotosViewer'
 import { PhotosPicker } from '../ducks/albums'
 
 export const ComingSoon = () => <p style="margin-left: 2em">Coming soon!</p>
 
 const AppRoute = (
   <Route component={Layout}>
     <Route path="photos" component={Timeline}>
-      <Route path=":photoId" component={Viewer} />
+      <Route path=":photoId" component={PhotosViewer} />
```
enguerran: Can't we have the same `FileViewer` for `photos` app?
Or the `PhotosViewer` for the `drive` app?

src/photos/components/PhotosViewer.jsx
```diff
@@ -0,0 +1,36 @@
...
+            query: router.location.query
+          })
+        }
+        {...this.props}
+      />
+    )
+  }
+}
+
+export default withRouter(PhotosViewer)
```
enguerran: Not sure we need `withRouter` see other comment.

src/photos/components/PhotosViewer.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react'
```
enguerran: This is very similar to `FileViewer`, can't we factorize a lot of code?

src/drive/components/FolderView.jsx
goldoraf: I'll add a comment to make it clearer, but tbh, I'm not fan of this trick (that we used in Photos before) and I'd like a better solution, but it'd require a lot of work

src/viewer/Viewer.jsx
```diff
@@ -0,0 +1,129 @@
...
+    this.setState({ controlsHidden: false })
+    document.removeEventListener('mousemove', this.showControls)
+  }
+  hideControls = () => {
+    this.setState({ controlsHidden: true })
+    document.addEventListener('mousemove', this.showControls)
+  }
+
+  componentDidMount() {
+    document.addEventListener('keydown', this.onKeyDown, false)
```
enguerran: I think it should listen to `keyup` rather than `keydown` or we need to be sure that keeping keydown pressed renders the correct behavior: fast browsing files.

src/components/Image/index.js
```diff
@@ -0,0 +1 @@
+export { default as ImageLoader } from './ImageLoader'
```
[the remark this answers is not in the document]
goldoraf: Why? It's a common pattern to have a module reexports some of its interesting components

src/photos/components/AppRoute.jsx
goldoraf: They have a slightly different behavior, that could diverge even more. That's why I preferred having separate wrappers

src/photos/components/PhotosViewer.jsx
goldoraf: See ^^

src/viewer/Viewer.jsx
goldoraf: It works better with `keyup` :+1:

src/components/Image/index.js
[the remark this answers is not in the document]
enguerran: I know it's common, but it's weird to have a oneline file to redirect. How about moving ImageLoader content into this file. But well, okay, does not matter. Forget it.

src/photos/components/AppRoute.jsx
enguerran: Look good to me, but can we rename with the app name with one of these:
- s/`FileViewer`/`DriveViewer/` and `PhotosViewer`
- s/`FileViewer`/`FilesViewer/` and `PhotosViewer`
- `FileViewer` and s/`PhotosViewer`/`PhotoViewer/`

src/viewer/Viewer.jsx
```diff
@@ -0,0 +1,147 @@
...
+    this.setState({ controlsHidden: false })
+    document.removeEventListener('mousemove', this.showControls)
+  }
+  hideControls = () => {
+    this.setState({ controlsHidden: true })
+    document.addEventListener('mousemove', this.showControls)
+  }
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.onKeyUp, false)
```
enguerran: `keyup` is a better event to catch a `onNext` behavior. But we cannot fast browse. Maybe we could add an event listener to accept fast browsing with `keydown` pressed for at least 1000 ms and that stops on the image when keydown is released... (I can explain it in french if I am not clear.)

goldoraf: I see what you mean, but I'm afraid it'd take a long time to implement properly (just a few hours probably, but I'd like to validate this story tonight), so... next time? :(

src/photos/components/AppRoute.jsx
goldoraf: ok for `FilesViewer` and `PhotosViewer`, but what about make their names even more explicit like `FilesViewerWrapper` or something?

src/drive/containers/File.jsx
```diff
@@ -12,7 +12,7 @@ import RenameInput from '../ducks/files/RenameInput'
 import { isDirectory } from '../ducks/files/files'
 import Spinner from 'cozy-ui/react/Spinner'
 import Preview from '../components/Preview'
-import breakpointsAware from 'cozy-ui/react/helpers/breakpoints'
+import { withBreakpoints } from 'cozy-ui/react'
```
enguerran: I miss this breaking change, no cozy-ui update available?

PR #536, opened 2017-11-29
fix: server selection improvements

y-lohse:
- changes the input type so the browser doesn't require a protocol
- fix an inversion in error messages
- use an internal `fetching` state because some async checks happen internally and were introducing a visible delay


y-lohse:
[(1) [DRI] 📱 Afficher un message d'erreur spécifique lorsque l'url contient "cosy"](https://trello.com/c/XXggEI0C/519-1-dri-%F0%9F%93%B1-afficher-un-message-derreur-sp%C3%A9cifique-lorsque-lurl-contient-cosy)

PR #539, opened 2017-11-29
fix: decide on download type at runtime

y-lohse:
This PR may cause problems for our next release. The commit we're merging into prod here is already on master so theoretically it's all ok but I'm not super confident.


y-lohse:
[🐛 Le partage par lien d'un fichier ne fonctionne pas](https://trello.com/c/8s3lX2Ym/627-%F0%9F%90%9B-le-partage-par-lien-dun-fichier-ne-fonctionne-pas)

PR #546, opened 2017-11-30
feat: display message when quota is reached

y-lohse:
Most changes are just because I split [one file into multiples](https://github.com/cozy/cozy-drive/pull/546/files#diff-d5df49947f20daea222c0c595047f1dbL1). 
The actual changes are :

- [a new component](https://github.com/cozy/cozy-drive/pull/546/files#diff-c5adba55890ddbfa6b6fb8bf48dd7538R1)
- [detecting the quota error](https://github.com/cozy/cozy-drive/pull/546/files#diff-d173561dcc55260e5a4050bb22efe48eR11)
- [minor css additions](https://github.com/cozy/cozy-drive/pull/546/files#diff-8c4e81d6be3ab034b04dfe94f956b64cR1)

### DIFF
--- a/src/drive/mobile/styles/feedback.styl
+++ b/src/drive/mobile/styles/feedback.styl
@@ -3,6 +3,9 @@
 .feedback-form
     @extend $form
 
+    .coz-form
+        margin-top 0
+
 .feedback-text
     box-sizing border-box
     width 100%
@@ -13,3 +16,9 @@
 
 .button-block
     text-align right
+
+.c-btn--paperplane
+    background-image embedurl('../assets/icons/icon-paperplane-white.svg')
+    background-position .5rem center
+    background-repeat no-repeat
+    padding-left 2rem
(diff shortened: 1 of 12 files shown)


y-lohse:
[(3) 📱 Afficher un message lorsque le quota disque du Cozy est atteint](https://trello.com/c/P1djYIvQ/565-3-%F0%9F%93%B1-afficher-un-message-lorsque-le-quota-disque-du-cozy-est-atteint)

y-lohse:
I slightly rewrote the [code of the two modals](https://github.com/cozy/cozy-drive/pull/546/files#diff-c5adba55890ddbfa6b6fb8bf48dd7538R53) but it's mostly cosmetic changes.

src/drive/mobile/styles/feedback.styl
```diff
@@ -13,3 +16,9 @@
 
 .button-block
     text-align right
+
+.c-btn--paperplane
```
GoOz: I guess this should belong in Cozy-UI at some point.

y-lohse: Weeeell I don't know, probably not. Have I had more time, I would just have used the `Icon` component which is a better solution for this IMO. But it involves changing the webpack configuration and I don't feel confident doing that without breaking anything just before the sprint end.

GoOz: Either way, it's ok to let this as is, don't worry. ;)

PR #551, opened 2017-12-01
Reset store 2

y-lohse:
Nearly the same thing as #525 — just fixed the conflicts.


y-lohse:
Woops, messed that up.

PR #557, opened 2017-12-05
Bar api

y-lohse:
The cozy-bar now offers 3 components to decide what appears there — this PR moves some code around to use them.


src/drive/ducks/files/Toolbar.jsx
```diff
@@ -55,11 +58,93 @@ class Toolbar extends Component {
...
       uploadFiles,
       downloadAll,
-      leaveFolder
+      leaveFolder,
+      breakpoints: { isMobile }
     } = this.props
     const notRootfolder = displayedFolder && displayedFolder.id !== ROOT_DIR_ID
     const hasWriteAccess =
       canUpload && (!shared.withMe || shared.sharingType === 'master-master')
+
```
y-lohse: This looks like a lot of changes, but in reality this section of the component was already there — I'm just putting it in a variable so I can easily wrap it (or not) later.

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -111,83 +196,9 @@ class Toolbar extends Component {
...
-                className={classNames(styles['fil-action-delete'])}
-                onClick={() => leaveFolder(displayedFolder)}
-              >
-                {t('toolbar.leave')}
-              </a>
-            </Item>
-          )}
-        </Menu>
+
+        {isMobile ? <BarRight>{MoreMenu}</BarRight> : MoreMenu}
```
y-lohse: This is the wrapped part — since `MoreMenu` is now a variable we avoid a lot of duplication.
I almost turned it into a standalone Component but it requires almost every prop in here.

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -49,7 +49,10 @@ export const initBar = () => {
     appEditor: 'Cozy',
     iconPath: require('../../../../targets/drive/vendor/assets/app-icon.svg'),
     lang: getLang(),
-    replaceTitleOnMobile: true
+    cozyURL: client.getUrl(),
```
y-lohse: So why did I provide the URL and the token to cozy-bar in this PR?

On mobile, every view has something in the `BarRight`, except the Settings views, where there's nothing. When there's nothing to display, the fallback is the [Nav](https://github.com/cozy/cozy-bar/blob/master/src/components/Bar.jsx#L106), but trying to render that component without the list of apps breaks everything. So with the cozy url and the token, the bar can fetch the apps and not disply anything at all in the `BarRight` section when needed.

Why didn't we care until now? Because [`displayMobile`](https://github.com/cozy/cozy-bar/blob/master/src/components/Bar.jsx#L105) was set to false.

We have another upcoming card on this subject this sprint to tidy things up.

src/drive/styles/toolbar.styl
```diff
@@ -17,41 +17,35 @@
             margin-left      0
             margin-right     1rem
 
-    .fil-toolbar-menu
+.fil-toolbar-menu
```
y-lohse: Only reduced the specificity here.

y-lohse:
[(2) 📱 Utiliser la nouvelle API de la cozy-bar pour piloter du contenu](https://trello.com/c/qDpTkLyf/598-2-%F0%9F%93%B1-utiliser-la-nouvelle-api-de-la-cozy-bar-pour-piloter-du-contenu)

src/drive/ducks/files/Toolbar.jsx
enguerran: I do believe that based-components architecture needs some abstraction to stay organized. I faced the same need with `<FileList />` (which is a collection) and `<File />` (which are items): almost every `FileList.props` are passed to `<File />`, then I am sure it is a matter of abstraction. For instance `<File />` must be a functionnal component and must accept every props needed to be rendered at the lowest level: 
1. display select box or not
1. icon
1. filename
1. etc.

Currently it is a function inside the `FileList` component because it can access `FileList`'s `this.props`. I am not sure it could stay as is, unless items can't be used outside the collection. And maybe that is the very case in my example.

Is it in yours?

y-lohse: I would say that no, they can't be used outside of here.

Inside `MoreMenu`, there is a `Menu` and a few `Item` — and [those are already "dumb components"](https://github.com/cozy/cozy-drive/blob/master/src/components/Menu/index.jsx). 
The only thing we could abstract away even more is this very specific configuration of a `Menu`, which isn't going to be used elsewhere. It might improve readability a little, but the resulting Component would not be useful anywhere else.

PR #559, opened 2017-12-05
WIP feat: ability to choose which adapter is used when using fetchDocument

ptbrowne:
Ability to choose which adapter will be used based on `policy`.


y-lohse:
What is the status of this PR? Does it sill need a review from @goldoraf ?

ptbrowne:
Nope. It is outdated. Thanks for noticing !

PR #561, opened 2017-12-05
fix: use our brand font on mobile

y-lohse:
Slightly surprised this is all it took:

- added font-face instructions on mobile
- added the font files
- configured webpack's `file-loader` to handle font files


y-lohse:
[🐛📱 Corriger la police de caractère qui n'est plus Lato sur iOS](https://trello.com/c/SZXkiiFn/658-%F0%9F%90%9B%F0%9F%93%B1-corriger-la-police-de-caract%C3%A8re-qui-nest-plus-lato-sur-ios)

config/webpack.target.mobile.js
```diff
@@ -15,6 +15,17 @@ module.exports = function(production, app) {
       path: path.resolve(__dirname, `../targets/${app}/mobile/www`),
       filename: '[name].js'
     },
+    module: {
+      rules: [
+        {
+            test: /\.(eot|svg|ttf|woff|woff2)$/,
```
enguerran: Why did you add `svg` file?

enguerran: Same question for `ttf`, `eot` ;)

y-lohse: I felt like it would be a good idea to cover all possible font formats, but I think I should at least remove svg now that you say it. We use it for loads of different things, better not mix up the loaders.

enguerran: > Don’t write code in anticipation of features that haven’t been requested. You aren’t going to need it (YAGNI).

PR #566, opened 2017-12-07
Added PDF viewer

goldoraf:
This PR adds an embedded PDF viewer on desktop and fixes 1 bug and an usability issue (next/prev controls interfered with audio/video controls).

In agreement with our PO, I modified the CSS so that action buttons get aligned to the right and hide their labels. As the CSS is extending `$selectionbar`, this makes the CSS a bit messy but I can't find a better way without spending a lot of time on it...

### DIFF
--- a/src/viewer/Viewer.jsx
+++ b/src/viewer/Viewer.jsx
@@ -65,6 +65,8 @@ export default class Viewer extends Component {
     const fileCount = files.length
     const hasPrevious = currentIndex > 0
     const hasNext = currentIndex < fileCount - 1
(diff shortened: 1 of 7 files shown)


src/viewer/Viewer.jsx
```diff
@@ -96,6 +101,8 @@ export default class Viewer extends Component {
         return AudioViewer
       case 'video':
         return isMobile() ? NoViewer : VideoViewer
+      case 'pdf':
+        return isMobile() || isCordova() ? NoViewer : PdfViewer
```
y-lohse: Why the double check? Isn't `isCordova` enough?

PR #571, opened 2017-12-07
fix: perms.byMe is undefined

enguerran:
with minified source it is
r.byMe is undefined


enguerran:
[Drive est cassé sur recette](https://trello.com/c/RsmrQbWc/667-drive-est-cass%C3%A9-sur-recette)

PR #579, opened 2017-12-08
refactor: use getFolderIdFromRoute

enguerran:
on `root_dir`, `params.folderId` is `undefined` this is handled by `getFolderIdFromRoute`


enguerran:
[(3) [forum-4890] 🐛 Affichage incomplet de la liste des fichiers](https://trello.com/c/OWw9dnyU/480-3-forum-4890-%F0%9F%90%9B-affichage-incomplet-de-la-liste-des-fichiers)

PR #582, opened 2017-12-11
fix: delete icons before regenerating them

y-lohse:


y-lohse:
[Navigation cassée dans l'app mobile v0.9.0](https://trello.com/c/YWdnCVAq/673-navigation-cass%C3%A9e-dans-lapp-mobile-v090)

PR #584, opened 2017-12-11
fix: unify file opening in recent and trash views

y-lohse:
Opening files from the trash was broken. For now, we'll use the same behavior on the recent and trash views — open files in new tabs.


y-lohse:
[(3) 📱 Ajouter la visionneuse PDF sur Android et iOS](https://trello.com/c/RIzCPLSu/430-3-%F0%9F%93%B1-ajouter-la-visionneuse-pdf-sur-android-et-ios)

PR #585, opened 2017-12-11
pdf viewer fixes

y-lohse:
- Our toolbar overlaps the pdf viewer's native toolbar, making it unusable. I disabled pointer events on the wrapper element, but re-enabled them on the actual buttons.
- Especially in firefox, loading large pdf's introduces a notable lag and the screen stays dark for a while. There is no cross-browser compatible way to detect when the `object` is loaded, but we can put a waiting background underneath it.


y-lohse:
[(3) Ajouter la visionneuse PDF web](https://trello.com/c/BtJ2a5iG/642-3-ajouter-la-visionneuse-pdf-web)

PR #594, opened 2017-12-12
Feat/new layout

GoOz:
I added the new layout to allow mobile views to use minimal-ui.

Due to this change, some components had to be updated:
- [x] `<UploadQueue />` is positioned over the content instead of pushing it. (OK with Design team)
- [x] Refactored `.fil-content-file` from tables so that whatever the number of lines (filename and/or path) the height of this element stays the same.
- [x] Removed `<AutoSizer />` from `react-virtualized` and used `react-measure` instead to set the width of PhotoList because `<AutoSizer />` was using a layout method incompatible with the new layout.

~~⚠️ Needs a new Cozy-UI release~~ **UI released**


GoOz:

y-lohse:
Related to https://github.com/cozy/cozy-ui/pull/247

src/drive/ducks/upload/styles.styl
```diff
@@ -158,16 +158,19 @@ progress.upload-queue-progress::-moz-progress-bar
     $coz-bar-border-size = 1px
 
     .upload-queue
-        position static
+        top $coz-bar-size
```
y-lohse: There's no position anymore ?

src/photos/ducks/upload/styles.styl
```diff
@@ -156,16 +156,19 @@ progress.upload-queue-progress::-moz-progress-bar
     $coz-bar-border-size = 1px
 
     .upload-queue
-        position static
+        top $coz-bar-size
```
y-lohse: Same thing, no position ?

src/drive/ducks/upload/styles.styl
GoOz: there is but it keeps the fixed position set for desktop
https://github.com/GoOz/cozy-drive/blob/603682218a6fc3d18963c5bc406fe562b4283093/src/drive/ducks/upload/styles.styl#L6

src/photos/ducks/upload/styles.styl
GoOz: same answer :)

PR #595, opened 2017-12-12
work on the mobile viewer

enguerran:
I would like to enhance the viewer behavior when we're on mobile as requested in https://trello.com/c/t7lvpy6G

- [x] refactor MoreButton to use `cozy-ui/React/Button`. It still throw an proptypes error that will be fix with the next release of cozy-ui as it will include https://github.com/cozy/cozy-ui/pull/241/commits/ad4cb4466b6ec0adea31d6de2960c62fed84d1d3
- [x] very simple simplification of `withFileUrl` code
- [x] rename `NoNetwork` component to `NoNetworkViewer` because this is the Viewer that is displayed when there is no network.
- [x] remove `Promise` when handling promise in `filesystem.js`

In `NoViewer` I added a `Spinner` when fetching/saving remote file. I guess we could refactor the different `Button` that are used in `NoViewer`.

I also added a `Alerter` error when there is no app or if something happens.

### DIFF
--- a/src/components/Button/MoreButton.jsx
+++ b/src/components/Button/MoreButton.jsx
@@ -1,23 +1,19 @@
 import React from 'react'
 import classNames from 'classnames'
 import { translate } from 'cozy-ui/react/I18n'
+import { Button } from 'cozy-ui/react'
 
 import styles from './index.styl'
 
-const MoreButton = ({ t, disabled, onClick }) => (
-  <button
-    role="button"
-    className={classNames(
-      styles['c-btn'],
-      styles['c-btn--secondary'],
-      styles['c-btn--more'],
-      styles['dri-btn--more']
-    )}
+const MoreButton = ({ t, disabled, onClick, children }) => (
+  <Button
+    className={classNames(styles['c-btn--more'], styles['dri-btn--more'])}
+    theme="secondary"
     disabled={disabled}
     onClick={onClick}
   >
-    <span className={styles['u-visuallyhidden']}>{t('Toolbar.more')}</span>
-  </button>
+    <span className={styles['u-visuallyhidden']}>{children}</span>
+  </Button>
 )
 
 export default translate()(MoreButton)
--- a/src/viewer/NoViewer.jsx
+++ b/src/viewer/NoViewer.jsx
@@ -1,26 +1,54 @@
 import React from 'react'
 import classNames from 'classnames'
 import { downloadFile } from 'cozy-client'
-import { isCordova } from 'drive/mobile/lib/device'
-import { openOfflineFile, openOnlineFile } from 'drive/mobile/lib/filesystem'
+import { logException } from '../drive/mobile/lib/reporter'
 import { translate } from 'cozy-ui/react/I18n'
 import Button from 'cozy-ui/react/Button'
+import Alerter from 'photos/components/Alerter'
+import { isCordova } from 'drive/mobile/lib/device'
+import {
+  openOfflineFile,
+  createTemporaryLocalFile,
+  openFileWithCordova
+} from 'drive/mobile/lib/filesystem'
 
 import styles from './styles'
 
-const OpenWithCordova = ({ t, file }) => (
-  <Button
-    theme="regular"
-    className={styles['pho-viewer-noviewer-download']}
-    onClick={() =>
-      file.isAvailableOffline ? openOfflineFile(file) : openOnlineFile(file)
-    }
-  >
-    {t('Viewer.noviewer.openWith')}
-  </Button>
-)
+class OpenWithCordovaButton extends React.Component {
+  state = {
+    loading: false
+  }
 
-const OpenWithWeb = ({ t, url }) => (
+  openFile = async file => {
+    this.setState(state => ({ ...state, loading: true }))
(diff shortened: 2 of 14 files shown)


enguerran:
[(5) 📱 Améliorer la visionneuse sur mobile](https://trello.com/c/t7lvpy6G/640-5-%F0%9F%93%B1-am%C3%A9liorer-la-visionneuse-sur-mobile)

src/components/Button/MoreButton.jsx
```diff
@@ -1,23 +1,19 @@
...
-  <button
-    role="button"
-    className={classNames(
-      styles['c-btn'],
-      styles['c-btn--secondary'],
-      styles['c-btn--more'],
-      styles['dri-btn--more']
-    )}
+const MoreButton = ({ t, disabled, onClick, children }) => (
+  <Button
```
goldoraf: :+1:

src/viewer/NoViewer.jsx
```diff
@@ -1,26 +1,55 @@
...
+    loading: false
+  }
+
+  onClick = () => {
+    const { file, onError } = this.props
+    file.isAvailableOffline
+      ? openOfflineFile(file)
+      : (async () => {
+          this.setState(state => ({ ...state, loading: true }))
+          const response = await cozy.client.files.downloadById(file.id)
```
goldoraf: What do you think about moving this code in a new helper in `src/drive/mobile/lib/filesystem.js` ?

y-lohse: Same here, i think a function that does `cozy.client.files.downloadById(file.id).then(response => response.blob())` would help in several places.

src/viewer/NoViewer.jsx
```diff
@@ -1,26 +1,55 @@
...
+          const blob = await response.blob()
+          const localFile = await temporarySave(blob, file.name)
+          this.setState(state => ({ ...state, loading: false }))
+          return openFileWithCordova(localFile.nativeURL, file.mime).catch(
+            onError
```
y-lohse: Maybe the `onError` should be the result of a `catch` block that also wraps the download and saving parts, just in case.

src/viewer/NoViewer.jsx
enguerran: Moving pure functions in a specific file and folder is just a matter of organization.
I admit this is the worst place and I leave it inline because I knew it will catch everybody's attention.
My first idea was to put it in `/src/drive/mobile/lib/filesystem.js` but this file clusters everything about cordova and device's I/O. So I don't know if it is its best place.

What do you thing?

1. move it to `/src/drive/mobile/lib/filesystem.js`?
1. move it to somewhere else?

goldoraf: 1. for now

src/viewer/NoViewer.jsx
```diff
@@ -1,26 +1,54 @@
...
+    this.setState(state => ({ ...state, loading: false }))
+    return openFileWithCordova(localFile.nativeURL, file.mime)
+  }
 
-const OpenWithWeb = ({ t, url }) => (
+  onClick = () => {
+    const { file, onError } = this.props
+    file.isAvailableOffline
+      ? openOfflineFile(file)
+      : this.openFile(file, onError).catch(onError)
```
y-lohse: J'aurais eu tendance à gérer l'erreur dans `openFile`, car ca permet de catcher d'autres erreurs que juste le résultat du `openFileWithCordova`.
Est-ce qu'il ne faudrait pas aussi gérer cette erreur lors d'un `openOfflineFile` ?

src/viewer/NoViewer.jsx
```diff
@@ -40,24 +68,44 @@ const Download = ({ t, file }) => (
...
-        styles['pho-viewer-noviewer'],
-        styles[`pho-viewer-noviewer--${file.class}`]
-      )}
-    >
-      <p className={styles['pho-viewer-filename']}>{file.name}</p>
-      <h2>{t('Viewer.noviewer.title')}</h2>
-      {action}
-    </div>
-  )
+const NoViewerButton = ({ file, fallbackUrl, t, onError }) => {
```
y-lohse: This is good, i should have done that.

y-lohse:
Argh, it doesn't seem to work for me :/ When I try to open a `pptx` file, I get this error : 
`Activity not found: No Activity found to handle Intent` and no alert at all.

y-lohse:
It all works as intended, from what I can see. So just the comment about catching the error from `openOfflineFile` me.

src/viewer/NoViewer.jsx
enguerran: Oui, surement, après je t'avoue que la gestion des exceptions dans `filesystem.js` est à revoir. On peut dire qu'on laisse comme ça et qu'on changera si ça pose problème. N'overengineerons pas trop tôt. ;-)
Ça te va ?

y-lohse: Ca me va pour le premier point. Pour le second je pense qu'il faut vraiment ajouter un `.catch(onError)` à `openOfflineFile`, sinon on n'a pas l'alerter quand le fichier est hors ligne. C'est un cas super rare, mais c'est aussi un fix rapide :)

PR #597, opened 2017-12-13
feat: load more files on scroll

y-lohse:
Pretty straightforward : i'm using the [InterSectionObserver](https://developer.mozilla.org/fr/docs/Web/API/IntersectionObserver) to detect when the "More" button comes into view, and then I fetch more files.
The only "complexity" comes from observing the button in case it gets re-rendered.

### DIFF
--- a/src/drive/components/FileList.jsx
+++ b/src/drive/components/FileList.jsx
@@ -4,11 +4,43 @@ import { translate } from 'cozy-ui/react/I18n'
 import File, { FilePlaceholder } from '../containers/File'
 import LoadMore from './LoadMore'
 
+require('intersection-observer') // polyfill for safari
+
 const LIMIT = 30
 
 class FileList extends PureComponent {
   state = {
     isLoading: false
+  }
+
+  intersectionObserver = null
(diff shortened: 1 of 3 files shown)


y-lohse:

src/drive/components/FileList.jsx
```diff
@@ -4,13 +4,46 @@ import { translate } from 'cozy-ui/react/I18n'
...
+    }
+  }
+
+  updateLoadMoreElement = element => {
+    if (this.loadMoreElement) {
+      this.intersectionObserver.unobserve(this.loadMoreElement)
+    }
+
+    if (element && element.getDOMNode) {
+      this.loadMoreElement = element.getDOMNode()
```
enguerran: You shall use https://reactjs.org/docs/react-dom.html#finddomnode
as `getDOMNode` is deprecated since 0.14

goldoraf: Ne devrait-on pas plutôt utiliser `React.findDOMNode()` maintenant ?

PR #602, opened 2017-12-14
fix: make mimetype to file type function standalone

y-lohse:
- We had an error on our `services` page, the one used for intents. `Cannot read property BarCenter of undefined`.
- `BarCenter` is a component exposed by the cozy bar, and there is no cozy bar on the `services` page because we don't need it. So why is `BarCenter` called at all?
- When we provide suggestions for the search bar, we also provide icons. To get the correct icon, we import a function that was defined in the `File` component.
- The `File` component itself imports half of our code base, including things that eventually call `BarCenter`

So

- I put the function in a standalone file that doesn't import anything else.
- No calling `BarCenter` anymore.
- The `services.js` file is 2x smaller now.

What I still don't like : the `Services` duck relies on a `lib` file that is not part of the duck itself. Not a big deal imo.

### DIFF
--- a/src/drive/containers/File.jsx
+++ b/src/drive/containers/File.jsx
@@ -15,6 +15,7 @@ import Preview from '../components/Preview'
 import { withBreakpoints } from 'cozy-ui/react'
 import { SharedBadge } from 'sharing'
 import { getSharingDetails } from 'cozy-client'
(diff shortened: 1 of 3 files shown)


y-lohse:
[🐛 La page des intents est cassée](https://trello.com/c/wmR4HJOY/676-%F0%9F%90%9B-la-page-des-intents-est-cass%C3%A9e)

PR #603, opened 2017-12-14
Fix/layout ios

GoOz:
- [x] Fix filename and path position on iOS 9-10
- [x] Update UI to 6.0.1 to fix layout on iOS 9-10

### DIFF
--- a/src/drive/styles/table.styl
+++ b/src/drive/styles/table.styl
@@ -143,6 +143,7 @@ mime-types('fil-file-', '../assets/icons/')
 .fil-file
     display flex
     flex-direction column
+    justify-content center
 
 .fil-file-preview
     content ''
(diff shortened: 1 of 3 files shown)


GoOz:

src/drive/styles/table.styl
```diff
@@ -157,10 +158,10 @@ mime-types('fil-file-', '../assets/icons/')
     background-position center center
 
 .fil-file-filename
-    flex 1 1 50%
+    flex 0 0 1.3em
 
 .fil-file-path
-    flex 1 1 1.5em
+    flex 0 0 1.2em
```
y-lohse: 1. These changes aren't related to the layout problems, are they?
2. Are you sure about this change :D ? Don't we want to allow shrinking ?

GoOz: 1. not directly no, but the first time I changed the layout I noticed an issue about this, that added an unwanted scrollbar, that bug was here but wasn't noticable before the layout change.
2. No we don't. As the flex flow is in column, what would shrink would be the height of the path. and we definitively don't want that. :) As for the width, there's an ellipsis so no problem here.

PR #606, opened 2017-12-15
Fix/spinners

GoOz:
- [x] Updated Cozy-UI to 6.0.3 version to fix the layout scroll issue
- [x] Fix spinners position due to a way too much wide selector

### DIFF
--- a/src/drive/styles/table.styl
+++ b/src/drive/styles/table.styl
@@ -204,7 +204,7 @@ mime-types('fil-file-', '../assets/icons/')
         max-width  100vw
 
     .fil-content-body
-        & > [class^='coz-spinner']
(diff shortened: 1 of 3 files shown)


GoOz:

src/drive/styles/table.styl
```diff
@@ -204,7 +204,7 @@ mime-types('fil-file-', '../assets/icons/')
         max-width  100vw
 
     .fil-content-body
-        [class^='coz-spinner']
+        & > [class^='coz-spinner']
```
y-lohse: I don't think the `&` selector is useful here ?

GoOz: Indeed. It's optional. I'll remove it.

PR #639, opened 2017-12-19
chore: remove unused configuration

enguerran:
since https://github.com/cozy/cozy-photos-v3/pull/79 and the monorepo fusion
we use a config file `jest.config.json`


y-lohse:
What do we do with this PR? I think we should fix the conflicts and merge it.

PR #641, opened 2017-12-19
Sentry exception reporting during the onboarding

goldoraf:
In the first commit, I removed the flag that allowed to disable Sentry, because we want to be able to report exceptions in all cases.

In the second commit, I added exception logging to the various errors that can occur when the user is entering its server url.

### DIFF
--- a/src/drive/mobile/actions/settings.js
+++ b/src/drive/mobile/actions/settings.js
@@ -1,7 +1,6 @@
 import { startReplication as startPouchReplication } from '../lib/replication'
 import { setClient, setFirstReplication } from '../../actions/settings'
 import { openFolder, getOpenedFolderId } from '../../actions'
-import { configureReporter } from '../lib/reporter'
 import { startTracker, stopTracker } from '../lib/tracker'
 import { revokeClient as reduxRevokeClient } from './authorization'
 import { resetClient } from '../lib/cozy-helper'
@@ -26,7 +25,6 @@ export const setAnalytics = (analytics, source = 'settings') => (
 ) => {
   dispatch({ type: SET_ANALYTICS, analytics })
   const state = getState()
-  configureReporter(analytics)
   if (analytics && state.mobile) {
     startTracker(state.mobile.settings.serverUrl)
   } else if (analytics === false) {
--- a/src/drive/mobile/containers/settings/Support.jsx
+++ b/src/drive/mobile/containers/settings/Support.jsx
@@ -40,11 +40,7 @@ export const Support = ({
         onClick: async () => {
           if (isOnline()) {
             try {
-              await logInfo(
-                t('mobile.settings.support.logs.title'),
-                serverUrl,
-                true
-              )
+              await logInfo(t('mobile.settings.support.logs.title'), serverUrl)
               success()
             } catch (e) {
               failure()
--- a/src/drive/mobile/lib/reporter.js
+++ b/src/drive/mobile/lib/reporter.js
@@ -22,9 +22,11 @@ export const configureReporter = () => {
   Raven.config(ANALYTICS_URL, getReporterConfiguration()).install()
 }
 
-export const logException = err => {
+export const logException = (err, extraContext = null) => {
   return new Promise(resolve => {
-    Raven.captureException(err)
+    extraContext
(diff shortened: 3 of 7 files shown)


src/drive/mobile/actions/settings.js
```diff
@@ -26,7 +26,7 @@ export const setAnalytics = (analytics, source = 'settings') => (
 ) => {
   dispatch({ type: SET_ANALYTICS, analytics })
   const state = getState()
-  configureReporter(analytics)
+  configureReporter()
```
y-lohse: I think this line can be completely removed. The reporter's configuration never changes now, and it is initialized elsewhere, so this line is now useless as far as I can tell.

src/drive/mobile/containers/settings/Support.jsx
```diff
@@ -40,11 +40,7 @@ export const Support = ({
         onClick: async () => {
           if (isOnline()) {
             try {
-              await logInfo(
-                t('mobile.settings.support.logs.title'),
-                serverUrl,
-                true
-              )
+              await logInfo(t('mobile.settings.support.logs.title'), serverUrl)
```
enguerran: do we still have to `await` the loginfo instruction's execution?

src/drive/mobile/lib/reporter.js
```diff
@@ -16,41 +14,36 @@ const getAnalyticsUrl = () => {
...
 })
 
-export const configureReporter = enable => {
-  isEnabled = enable
+export const configureReporter = () => {
   Raven.config(ANALYTICS_URL, getReporterConfiguration()).install()
 }
 
-export const logException = err => {
+export const logException = (err, extraContext = null) => {
```
enguerran: as `Raven` call it `extra`, shall we stay on this naming `extra` instead of `extraContext`?

src/drive/mobile/lib/reporter.js
```diff
@@ -16,41 +14,36 @@ const getAnalyticsUrl = () => {
...
   return new Promise(resolve => {
-    Raven.captureException(err)
+    extraContext
+      ? Raven.captureException(err, { extra: extraContext })
+      : Raven.captureException(err)
```
y-lohse: What happens if you just call `Raven.captureException(err, { extra: null })`?

src/drive/mobile/lib/reporter.js
```diff
@@ -1,8 +1,6 @@
 /* global __SENTRY_TOKEN__, __DEVELOPMENT__ */
 import Raven from 'raven-js'
 
-let isEnabled = false
```
y-lohse: \o/

src/drive/mobile/lib/reporter.js
```diff
@@ -16,41 +14,36 @@ const getAnalyticsUrl = () => {
...
+      ? Raven.captureException(err, { extra: extraContext })
+      : Raven.captureException(err)
     console.warn('Raven is recording exception')
     console.error(err)
     resolve()
   })
 }
 
-const logMessage = (message, serverUrl, level = 'info', force) => {
+const logMessage = (message, serverUrl, level = 'info') => {
```
enguerran: good move to remove this weird parameter: if I call `logMessage`, I want it to log a message. Using a force parameter let me believe `logMessage` may be called `pleaseAreYouAbleToLogMessage` ;)

src/drive/mobile/lib/reporter.js
goldoraf: I called it `extraContext` in the hope it would make its purpose clearer, not sure what's better here

src/drive/mobile/containers/settings/Support.jsx
goldoraf: As logInfo returns a promise, I guess it's better to await it, but honestly I'm not sure it was necessary to return a promise from the `logInfo` function...

src/drive/mobile/lib/reporter.js
goldoraf: I don't know, that's why I preferred to handle the two cases differently ;)

y-lohse: Ok! I don't mind either way, but I think it would be easier to set the default value to `{}` and remove the ternary.

PR #642, opened 2017-12-19
chore: upgrade file opener plugin

y-lohse:
I found a strange bug on iOS - some file types like `pptx` or `docx` would not open. They were downloaded, but when we tried to open them we received an error `File does not exist`.

Not knowing where to look, I upgraded [the plugin](https://github.com/pwlin/cordova-plugin-file-opener2) — there's no changelog, but it fixed the problem and everything else works as before, as far as I can tell.


y-lohse:
[(0,5) 🐛📱 Corriger le message d'erreur lorsque aucune app ne peut ouvrir un fichier sur iOS](https://trello.com/c/mrcNMwzA/685-05-%F0%9F%90%9B%F0%9F%93%B1-corriger-le-message-derreur-lorsque-aucune-app-ne-peut-ouvrir-un-fichier-sur-ios)

PR #646, opened 2017-12-19
fix: manage url input

enguerran:
I revamp URL check instructions to explicit the codebase and to describe the intent.

We still have an anti-pattern in `componentWillReceiveProps`:

```js
componentWillReceiveProps(nextProps) {
  this.setState(state => ({
    ...state,
    error: nextProps.externalError ? ERR_WRONG_ADDRESS : null,
    fetching: nextProps.fetching
  }))
}
```

Putting a props in the internal state is an anti-pattern.

I added `propTypes` and `defaultProps`.

### DIFF
--- a/src/authentication/steps/SelectServer.jsx
+++ b/src/authentication/steps/SelectServer.jsx
@@ -21,81 +21,94 @@ export class SelectServer extends Component {
   }
 
   componentDidMount() {
-    this.serverInput.focus()
+    this.input.focus()
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({
+    this.setState(state => ({
+      ...state,
       error: nextProps.externalError ? ERR_WRONG_ADDRESS : null,
       fetching: nextProps.fetching
-    })
+    }))
   }
 
   componentDidUpdate() {
     if (this.state.error) {
-      this.serverInput.focus()
-      this.serverInput.select()
+      this.input.focus()
+      this.input.select()
     }
   }
 
-  validateValue(e) {
-    this.setState({ value: e.target.value, error: null })
+  onChange(value) {
+    if (this.state.error) {
+      this.setState(state => ({ ...state, error: null }))
+    }
+    this.setState(state => ({ ...state, value: value.trim() }))
   }
 
   onSubmit = async e => {
     e.preventDefault()
-    if (!this.state.value) return
+    const value = this.state.value
+    let error
 
-    let url = this.state.value
-
-    if (url.indexOf('@') > -1) {
-      this.setState({ error: ERR_EMAIL })
-      return
+    if (this.hasAtSign(value)) {
+      error = ERR_EMAIL
+    }
+    if (this.hasMispelledCozy(value)) {
+      error = ERR_COSY
     }
 
-    // prepend the protocol if ommited
-    if (/^http(s)?:\/\//.test(url) === false) url = 'https://' + url
-
-    let parsedURL
-    try {
-      parsedURL = new URL(url)
-
-      if (parsedURL.protocol === 'http:' && !__ALLOW_HTTP__) {
-        throw new Error('Only https protocol is allowed')
-      }
-
-      if (/\..*cosy.*\..+$/.test(parsedURL.host)) {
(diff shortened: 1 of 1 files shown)


src/authentication/steps/SelectServer.jsx
```diff
@@ -21,81 +22,92 @@ export class SelectServer extends Component {
...
+      this.input.focus()
+      this.input.select()
     }
   }
 
-  validateValue(e) {
-    this.setState({ value: e.target.value, error: null })
+  onChange(value) {
+    this.setState(state => ({ ...state, error: null }))
+    this.setState(state => ({ ...state, value: value.trim() }))
```
y-lohse: Why do you `setState` twice here?

src/authentication/steps/SelectServer.jsx
```diff
@@ -172,4 +182,17 @@ export class SelectServer extends Component {
   }
 }
 
+SelectServer.propTypes = {
```
goldoraf: :+1:

src/authentication/steps/SelectServer.jsx
```diff
@@ -21,81 +22,92 @@ export class SelectServer extends Component {
...
+    const url = this.getUrl(value)
+    if (
+      !(url instanceof URL) ||
+      (url.protocol === 'http:' && !__ALLOW_HTTP__)
+    ) {
+      error = ERR_WRONG_ADDRESS
+    }
+    if (error) {
+      this.setState(state => ({ ...state, error }))
+      logException(error, {
```
goldoraf: I saw in Sentry docs that it's better to pass an Error instance instead of a string to `Raven.captureException` (which is used by `logException`): https://docs.sentry.io/clients/javascript/usage/#raven-js-reporting-errors

enguerran:
[(0,5) 📱 Supprimer les espaces lors de la saisie par l'utilisateur lors de la connexion de l'app mobile](https://trello.com/c/KcmwRYy4/702-05-%F0%9F%93%B1-supprimer-les-espaces-lors-de-la-saisie-par-lutilisateur-lors-de-la-connexion-de-lapp-mobile)

src/authentication/steps/SelectServer.jsx
enguerran: Ok, je modifie ça. J'ai du merger les conflits à la main suite à ta modif sur le reporter.

src/authentication/steps/SelectServer.jsx
```diff
@@ -21,81 +22,92 @@ export class SelectServer extends Component {
...
 
   componentDidMount() {
-    this.serverInput.focus()
+    this.input.focus()
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({
+    this.setState(state => ({
+      ...state,
```
ptbrowne: Why do you use a setState callback here ?

enguerran: I always use `setState` in its functional form.
See https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b
And https://twitter.com/dan_abramov/status/824308413559668744

src/authentication/steps/SelectServer.jsx
enguerran: Because this is more or less uncoupled.
In fact, setting error state to null should not be done `onChange` but `onBlur` or `onEnter`.
In the other hand, the `trim()` function may occur on every key press because we don't want space character at all.

So... I will move the setting error state to null in another event handler. Thanks for your comment.

src/authentication/steps/SelectServer.jsx
ptbrowne: Phew such a long article for a simple information :D TLDR : if your state update relies on the previous state, use a functional setState since setState applies its updates asynchronously; by doing `setState(updates)`, you can have a race condition.

From the tweet, the advantage I see is from a testing point of view since you can test each individual test updater.

Understood, thanks for the pointers.

enguerran: And a conclusion: always use the functional `setState` as it does not cost much more and it is a very good and cleaner code with pure functions and race condition handled.

PR #647, opened 2017-12-20
fix: init bar when offline

y-lohse:
When we are offline, we can't get a token to pass to the cozy-bar. But we still want to init the bar! Even without a token.

To test this, you can start the app in `standalone` mode, and then block all queries to `cozy.tools:8080` with the devtools. Without this PR, the cozy bar doesn't show up at all. With this PR, it appears as usual.

### DIFF
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -44,13 +44,14 @@ export const initBar = async client => {
   if (document.getElementById('coz-bar')) {
     return
   }
+
   cozy.bar.init({
     appName: 'Drive',
     appEditor: 'Cozy',
     iconPath: require('../../../../targets/drive/vendor/assets/app-icon.svg'),
     lang: getLang(),
     cozyURL: client.getUrl(),
-    token: await getToken(),
+    token: getTokenWithNoException(),
     replaceTitleOnMobile: false,
     displayOnMobile: true
   })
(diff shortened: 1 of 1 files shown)


y-lohse:
[(3) 📱 Naviguer dans les autres applications](https://trello.com/c/1Nac0Gko/596-3-%F0%9F%93%B1-naviguer-dans-les-autres-applications)

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -44,13 +44,19 @@ export const initBar = async client => {
   if (document.getElementById('coz-bar')) {
     return
   }
+
+  let token = null
+  try {
+    token = await getToken()
+  } catch (_) {}
```
enguerran: Are we sure we want to completly silent the catched exception?

To avoid the let declaration:

```js
// somewhere
getTokenWithNoException = async () => {
    try {
        return await getToken()
    } catch (_) {}
}

// and in your function

const token = await getTokenWithNoException()
cozy.bar.init({...options, token})
```

y-lohse: The separate function makes sense here, I'll do that! But yes we want to silence the exception, not being abble to fetch a token for the bar has no real consequences — it will work offline as best as it can, and that's it.

y-lohse:
Merging this since I've update the code after the last review and it's needed for the release we're doing right now!

PR #652, opened 2017-12-21
Use UI components

GoOz:
- [x] New Spinner with className props and breakpoints
- [x] Use of `<Sidebar />`
- [x] Use of `<Button />`
- [x] Clean up MQs
- [x] Clean up `font-size()` deprecation
- [x] Clean up `em` in favour of `rem`
- [x] **[Bonus]** Fix ellipsis on filename
- [x] **[Bonus]** Fix layout with long filename
- [x] **[Bonus]** Fix topbar's height on Chrome

~~⚠️ **Will need a new release of Cozy-UI**~~


GoOz:
[[Drive] Use of `<Sidebar />`, `<Spinner />` & `<Button />` components](https://trello.com/c/3hMU3RYP/2-drive-use-of-sidebar-spinner-button-components)

src/authentication/steps/SelectServer.jsx
```diff
@@ -170,14 +171,9 @@ export class SelectServer extends Component {
           )}
         </div>
         <footer className={styles['wizard-footer']}>
-          <button
-            role="button"
-            className={classNames(styles['c-btn'], styles['c-btn--regular'])}
-            disabled={error || !value || fetching}
-            aria-busy={fetching}
-          >
+          <Button disabled={error || !value || fetching} aria-busy={fetching}>
```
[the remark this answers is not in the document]
GoOz: HAHA oui ce serait tellement cool que c'est déjà implémenté en fait. j'ai juste oublié de l'appliquer…
On a déjà un param `busy`du coup qui rajoute le spinner ainsi que l'aria-busy et un `disabled` qui… bah disable le bouton mais c'est séparé l'un et l'autre. Je vas changer ça dans Drive quand même.

src/components/Button/index.styl
```diff
@@ -1,27 +1,21 @@
...
-    @extend $button--more
-
-.c-btn-extra
-    @extend $button-extra
-
 .u-visuallyhidden
     @extend $visuallyhidden
 
-@media (max-width: (1023/basefont)rem)
+.dri-btn--more
```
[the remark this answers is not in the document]
GoOz: Oui j'y ai pensé aussi. J'ai hésité à faire un nouveau bouton ou faciliter la création de ce bouton encore plus. Mais je m'étais dit que ce serait pour plus tard. Note que cette PR est une première (grosse) passe sur cette histoire de bouton et probablement pas la dernière.

src/components/Error/empty.styl
```diff
@@ -1,70 +1,32 @@
-@require 'cozy-ui'
+@import 'cozy-ui/'
```
y-lohse: Why `import` and not `require`?

src/components/Menu/index.styl
```diff
@@ -9,10 +9,10 @@
     position   absolute
     @extend    $popover
     right      0
-    padding    5px 0
+    padding    .3125rem 0
     margin     0
-    margin-top 1px
-    min-width  220px
+    margin-top .0625rem
```
y-lohse: I guess we need a padding and margin scale next :p

src/components/pushClient/Banner.jsx
```diff
@@ -39,35 +39,38 @@ class BannerClient extends Component {
...
-        </a>
+        </ButtonLink>
         <p className={styles['coz-banner-text']}>
           <span>{t('Nav.banner-txt-client')}</span>
-          <a
+          <ButtonLink
             href={t(desktopLink)}
-            target="_blank"
-            className={classNames(styles['c-btn'])}
+            target
```
y-lohse: I think this prop has the wrong name, it should be `blank` or something like that. It's not very explicit what `target` does, and technically there are other values than `_blank` even though they are rarely used.

src/drive/mobile/containers/RatingModal.jsx
```diff
@@ -153,26 +144,12 @@ class FeeedbackForm extends Component {
...
-              <button
-                className={classNames(
-                  styles['c-btn'],
-                  styles['c-btn--regular']
-                )}
-                type="submit"
-                aria-busy={sending}
-              >
+              </Button>
+              <Button aria-busy={sending}>
```
y-lohse: Another `aria-busy` here that should be `busy` ;)

src/components/pushClient/Banner.jsx
GoOz: None of the `target` values should be used actually. And besides `_blank`, every others are kinda useful only with frameset which is never. 
Anyway I asked myself the question of the naming but I thought that `target` speaks more for a link than `blank`, but maybe I'm wrong.

src/components/Menu/index.styl
GoOz: What do you mean?

src/components/Error/empty.styl
GoOz: Not quite sure exactly but apparently, due to our Stylus management, `@require` does not do what it's supposed to do. @ptbrowne knows more as it is one of its recommandations.

GoOz:
@y-lohse about the `px` to `rem` border-size, it actually makes more sense. If we use `px`, they won't grow if the user increase the font-size, as you pointed out, but with `rem` it will and that will ensure the consistency of the design. True, everything will be bigger, but everything will be proportional that way.

src/components/Error/empty.styl
y-lohse: i'd be interested in knowing more about this. IIRC `import` creates a copy, which is what was causing massive build times. And we use `require` everywhere else.

ptbrowne: Mmh, I recommended to use `require` instead of `import`, did you experience problems with `require` @GoOz  >

PR #654, opened 2017-12-22
Updated cozy-bar so that its token can be renewed if the current one lasks apps permission

goldoraf:


src/drive/mobile/styles/main.styl
```diff
@@ -18,7 +18,6 @@ html {
 }
 
 :global
-    [role=banner] .coz-bar-btn.coz-bar-burger{
-        visibility hidden
-        flex 0 0 .22em
+    [role=banner] .coz-drawer--settings > div > :nth-child(-n+3) {
```
y-lohse: Lol. Do you mind adding a comment saying what this does please? Or making the selector more explicit^^

PR #660, opened 2017-12-26
Compatibility with React

ptbrowne:
Since we test with React, we must be compatible with React.

EDIT : More context : 

For React to pass context properties down to children components, contextTypes must be defined with the properties that are passed. In preact, access to the context is not restricted like this. 

https://reactjs.org/docs/context.html

> If contextTypes is not defined, then context will be an empty object.

- [x] React compatibility

EDIT: Removed : 

- [ ] Fix collection "creation" with a receive data

### DIFF
--- a/src/lib/cozy-client/connect.jsx
+++ b/src/lib/cozy-client/connect.jsx
@@ -4,6 +4,12 @@ import PropTypes from 'prop-types'
 import { applySelectorForAction, enhancePropsForActions } from '.'
 import { mapValues, filterValues } from './utils'
 
+const storeShape = PropTypes.shape({
+  subscribe: PropTypes.func.isRequired,
+  dispatch: PropTypes.func.isRequired,
+  getState: PropTypes.func.isRequired
+})
+
 const connect = (
   mapDocumentsToProps,
   mapActionsToProps = null
@@ -32,7 +38,7 @@ const connect = (
   }
 
   Wrapper.contextTypes = {
-    store: PropTypes.object.isRequired
+    store: storeShape.isRequired
   }
 
   const makeMapStateToProps = (initialState, initialOwnProps) => {
--- a/src/lib/cozy-client/connect.jsx
+++ b/src/lib/cozy-client/connect.jsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import { connect as reduxConnect } from 'react-redux'
-
+import PropTypes from 'prop-types'
 import { applySelectorForAction, enhancePropsForActions } from '.'
 import { mapValues, filterValues } from './utils'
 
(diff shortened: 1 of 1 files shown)


src/lib/cozy-client/reducer.js
```diff
@@ -198,6 +198,7 @@ const collection = (state = collectionInitialState, action) => {
       const response = action.response
       return {
         ...state,
+        type: state.type || action.doctype,
```
goldoraf: Why this? Could you please give me some context?

ptbrowne: If you send a receive data with no fetch beforehand, the created collection has no doctype.

goldoraf: I got that, but why (and where) is a RECEIVE_DATA dispatched with no FETCH_COLLECTION beforehand? Is it in your app code?

ptbrowne: @goldoraf in case of, I edited my message as you were writing your reply ^

goldoraf: OK. You shouldn't have to do that kind of hack in your components' tests. As I understand it, what you want is to test the UI behavior of your component, is it really necessary to manipulate the "data layer" in order to test your component (it shouldn't)? Couldn't you test the "unconnected" version of the component?

ptbrowne: > Couldn't you test the "unconnected" version of the component?

Something like : 

```jsx
const UnconnectedReimbursements = function ({ bills }) {
  return ...
}

export UnconnectedReimbursements
export default cozyConnect(UnconnectedReimbursements)
```

and then in my tests do

```jsx
const bills =  data: fixtures['io.cozy.bills']
const dom = render(<UnconnectedReimbursements bills={{ data: bills, fetchStatus: 'loaded' }}, domNode)
expect(dom.querySelector('.bill').length).toBe(bills.length)
```

ptbrowne: Implemented in https://gitlab.cozycloud.cc/labs/cozy-sante/merge_requests/45/diffs?commit_id=f7cf11d385b10c5af4d5664a36942e6e9d17b5d1

I removed my change on the reducer 👍 Thanks !

src/lib/cozy-client/connect.jsx
```diff
@@ -31,6 +31,10 @@ const connect = (
     }
   }
 
+  Wrapper.contextTypes = {
+    store: PropTypes.object.isRequired
```
enguerran: Est-ce que `<Wrapper store={} />` est suffisant ?
Sinon, il faut ajouter des `prop-types` plus spécifique pour bien expliquer ce qu'il faut.

ptbrowne: J'ai mis la bonne shape maintenant.

ptbrowne:
ping :)

PR #674, opened 2018-01-08
fix: add protocol to server url

y-lohse:
A while ago, we started using the URL API to parse the server URL.

We noticed afterwards that this API tends to add an extra `/` at the end, eg. `https://my.cozy.cloud` becomes `https://my.cozy.cloud/`, and then some of our calls end up having double slashes (eg. `https://mYcozy.cloud//status`).

We fixed it, but fixed badly, using the unparsed URL. This PR fixes things properly, I hope.

It should now work when entering a URL with or without protocol, and there should never be an extra slash at the end.


y-lohse:

PR #675, opened 2018-01-08
Fix/buttons everywhere

GoOz:
Fix some broken buttons :(

### DIFF
--- a/src/photos/styles/toolbar.styl
+++ b/src/photos/styles/toolbar.styl
@@ -29,6 +29,12 @@
     margin-left auto
 
     a
+        &.pho-action-upload
+            padding-left 1rem
+
+            svg
+                margin-right .5rem
+                vertical-align text-top
         &.pho-action-newalbum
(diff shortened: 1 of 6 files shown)


src/photos/styles/toolbar.styl
```diff
@@ -29,6 +29,12 @@
     margin-left auto
 
     a
+        &.pho-action-upload
+            padding-left 1rem
+
+            svg
+                margin-right .5rem
+                vertical-align text-top
```
y-lohse: Are you sure about the `text-top` align? Maybe it's because I'm paying attention, but it looks weird to me : 
capture d ecran 2018-01-09 11 00 54

GoOz: It's not centered also without it. This bit needs a good ol' refactoring but… 
Or maybe I could try something else… wait.

GoOz:
@y-lohse Should be better now, isn't it?

GoOz:

y-lohse:
Nooot really but nevermind. It's probably just the icon.

PR #676, opened 2018-01-09
fix: taps on more button

y-lohse:
On touch devices, taping the "more" button on the right of a file would sometimes not work, depending on what element exactly was the `target`. When it didn't work, the file was opened in the viewer instead.

This fixes the problem, so tapping the "more" button should always bring up the FileActionMenu now.


y-lohse:

PR #653, opened 2017-12-22
fix: adapt share process with new stack

enguerran:
Multiple changes to feat the new cozy-stack share API.
tir-sharemodal

PR #679, opened 2018-01-10
fix: sharing infos

y-lohse:
This PR does 2 things:

- re-enables the icons for shared folders (folders that *I* share should have a green icon, folders that are shared *with me* a purple one.)
- The "share" button is different for folders that are shared, and if a folder is shared, the upload button is moved to the side menu

capture d ecran 2018-01-10 13 26 11

The second part required some heavy work on the `Toolbar` component. I'm quite happy [with the current API](https://github.com/cozy/cozy-drive/pull/679/files#diff-12918a0e2886ac9526cde1534cc4ca9fR108) but unhappy with [the implementation](https://github.com/cozy/cozy-drive/pull/679/files#diff-cd8ecf10d1445222404e50d7f56d783fR10). I'll add some review comments to point out things where I'd like feedback. 

edit: actually I've reworked the implementation a bit and am relative happy with it. It could be extended a lot more but I think it's a good start!

### DIFF
--- a/src/components/Toolbar/index.jsx
+++ b/src/components/Toolbar/index.jsx
@@ -19,9 +19,23 @@ class Toolbar extends Component {
       breakpoints: { isMobile }
     } = this.props
 
-    const actions = React.Children.toArray(children)
-    const highlightedActions = actions.filter(child => child.props.highlighted)
-    const secondaryActions = actions.filter(child => !child.props.highlighted)
+    const sortActions = (sortedActions, action) =>
+      action.props.highlighted
+        ? {
+            secondaryActions: [...sortedActions.secondaryActions],
+            highlightedActions: [...sortedActions.highlightedActions, action]
+          }
+        : {
+            secondaryActions: [...sortedActions.secondaryActions, action],
+            highlightedActions: [...sortedActions.highlightedActions]
+          }
+
+    const { highlightedActions, secondaryActions } = React.Children.toArray(
+      children
+    ).reduce(sortActions, {
+      highlightedActions: [],
+      secondaryActions: []
+    })
 
     const MoreMenu = (
       <Menu
--- a/src/components/Toolbar/index.jsx
+++ b/src/components/Toolbar/index.jsx
@@ -1,0 +1,121 @@
+/* global cozy */
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import Menu, { Item } from 'components/Menu'
+import styles from './index.styl'
+
+import { withBreakpoints } from 'cozy-ui/react'
+
+const { BarRight } = cozy.bar
+
+class Toolbar extends Component {
+  render() {
+    const {
+      children,
+      moreMenuTitle,
+      moreMenuDisabled,
+      moreMenuClassName,
+      moreMenuButton,
+      breakpoints: { isMobile }
+    } = this.props
+
+    const actions = React.Children.toArray(children)
+    const highlightedActions = actions.filter(child => child.props.highlighted)
+    const secondaryActions = actions.filter(child => !child.props.highlighted)
+
+    const MoreMenu = (
+      <Menu
+        title={moreMenuTitle}
+        disabled={moreMenuDisabled}
+        className={moreMenuClassName}
+        button={moreMenuButton}
+      >
+        {highlightedActions.map(component =>
+          this.wrapInItem(
+            React.cloneElement(component, {
+              highlighted: false,
+              isSecondaryAction: false
+            })
+          )
+        )}
+        {secondaryActions.map(component =>
+          this.wrapInItem(
+            React.cloneElement(component, {
+              highlighted: false,
+              isSecondaryAction: true
+            })
+          )
+        )}
+      </Menu>
+    )
+
+    return (
+      <div role="toolbar">
+        {highlightedActions.map(component =>
+          React.cloneElement(component, {
+            highlighted: true,
+            isSecondaryAction: false
+          })
+        )}
+        {isMobile ? <BarRight>{MoreMenu}</BarRight> : MoreMenu}
+      </div>
+    )
+  }
+
+  wrapInItem(component) {
+    return <Item>{component}</Item>
+  }
+}
+
+Toolbar.propTypes = {
+  moreMenuTitle: PropTypes.string.isRequired,
+  moreMenuDisabled: PropTypes.bool,
+  moreMenuClassName: PropTypes.string,
+  moreMenuButton: PropTypes.element.isRequired,
+  children: PropTypes.node,
+  breakpoints: PropTypes.object
+}
+
+Toolbar.defaultProps = {
+  moreMenuDisabled: false,
+  moreMenuClassName: ''
+}
+
+export class ToolbarAction extends Component {
+  render() {
+    try {
+      const { children, visible, highlighted, isSecondaryAction } = this.props
+
+      if (!visible) return null
+
+      const child = React.cloneElement(React.Children.only(children), {
+        highlighted
+      })
+
+      if (isSecondaryAction) return child
+      else {
+        const hiddenStyle = highlighted
+          ? styles['u-hide--mob']
+          : styles['u-hide--desk']
+        return <span className={hiddenStyle}>{child}</span>
+      }
+    } catch (_) {
+      return null
+    }
+  }
+} // needs to extend Component to have `props.children`
+
+ToolbarAction.propTypes = {
+  visible: PropTypes.bool,
+  highlighted: PropTypes.bool,
+  isSecondaryAction: PropTypes.bool,
+  children: PropTypes.element.isRequired
+}
+
+ToolbarAction.defaultProps = {
+  visible: true,
+  highlighted: false,
+  isSecondaryAction: true
+}
+
+export default withBreakpoints()(Toolbar)
--- a/src/drive/ducks/files/Container.jsx
+++ b/src/drive/ducks/files/Container.jsx
@@ -39,7 +39,7 @@ const mapStateToProps = (state, ownProps) => ({
 
 const mapDispatchToProps = (dispatch, ownProps) => {
   const hasWriteAccess =
-    !ownProps.shared.withMe || ownProps.shared.sharingType === 'master-master'
+    !ownProps.shared.withMe || ownProps.shared.sharingType === 'two-way'
   return {
     actions: Object.assign({}, ownProps.actions, {
       list: {
--- a/src/drive/ducks/files/Toolbar.jsx
+++ b/src/drive/ducks/files/Toolbar.jsx
@@ -24,6 +24,7 @@ import { uploadedFile, downloadFiles, trashFiles } from '../../actions'
 import {
   ShareButton,
   SharedWithMeButton,
+  SharedByMeButton,
   ShareModal,
   SharingDetailsModal
 } from 'sharing'
@@ -40,7 +41,57 @@ const ALERT_LEVEL_INFO = 'info'
 const ALERT_LEVEL_ERROR = 'error'
 const ALERT_LEVEL_SUCCESS = 'success'
 
-class Toolbar extends Component {
+const ActionShare = props => {
+  const { highlighted, shared, onClick } = props
+
+  let label, component
+
+  if (shared.withMe) {
+    label = 'Files.share.sharedWithMe'
+    component = SharedWithMeButton
+  } else if (shared.byMe) {
+    label = 'Files.share.sharedByMe'
+    component = SharedByMeButton
+  } else {
+    label = 'toolbar.share'
+    component = ShareButton
+  }
+
+  return highlighted ? (
+    <component
+      label={t(label)}
+      onClick={onClick}
+      className={styles['u-hide--mob']}
+    />
+  ) : (
+    <a
+      className={classNames(styles['fil-action-share'], styles['u-hide--desk'])}
+      onClick={onClick}
+    >
+      {t(label)}
+    </a>
+  )
+}
+
+const ActionUpload = props => {
+  const { highlighted, disabled, onUpload, visible } = props
+  return highlighted ? (
+    <UploadButton
+      disabled={disabled}
+      onUpload={onUpload}
+      label={t('toolbar.item_upload')}
+      className={classNames(styles['c-btn'], styles['u-hide--mob'])}
+    />
+  ) : (
+    <UploadButton
+      onUpload={onUpload}
+      label={t('toolbar.menu_upload')}
+      className={styles['fil-action-upload']}
+    />
+  )
+}
+
+class FilesToolbar extends Component {
   state = {
     showShareModal: false
   }
@@ -63,7 +114,16 @@ class Toolbar extends Component {
     } = this.props
     const notRootfolder = displayedFolder && displayedFolder.id !== ROOT_DIR_ID
     const hasWriteAccess =
-      canUpload && (!shared.withMe || shared.sharingType === 'master-master')
+      canUpload && (!shared.withMe || shared.sharingType === 'two-way')
+
+    const commonActions = []
+    const conditionalActions = []
+
+    return (
+      <div className={styles['fil-toolbar-files']} role="toolbar">
+        <ActionUpload />
+      </div>
+    )
 
     const MoreMenu = (
       <Menu
@@ -73,7 +133,7 @@ class Toolbar extends Component {
         button={<MoreButton>{t('Toolbar.more')}</MoreButton>}
       >
         {notRootfolder &&
-          !shared.withMe && (
+          !shared.shared && (
             <Item>
               <a
                 className={styles['fil-action-share']}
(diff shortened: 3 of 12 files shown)


src/lib/cozy-client/slices/sharings.js
```diff
@@ -393,14 +391,14 @@ const getSharingForRecipient = (state, document, recipient) => {
...
+      perm.attributes.permissions.files.values.includes(id)
     ),
-    ...perms.withMe.filter(
-      perm => perm.attributes.permissions.rule0.values.indexOf(id) !== -1
+    ...perms.withMe.filter(perm =>
+      perm.attributes.permissions.rule0.values.includes(id)
     )
   ]
-    .map(p => getSharing(state, p.attributes.source_id))
+    .map(p => getSharing(state, p.attributes.source_id.split('/')[1]))
```
y-lohse: This is not great at all. @goldoraf are you working on this or should I work out something better here?

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -58,146 +97,97 @@ class Toolbar extends Component {
...
-            </Item>
-          )}
-        {shared.withMe && (
-          <Item>
-            <a
-              className={styles['fil-action-share']}
-              onClick={() => this.setState(toggleShowShareModal)}
+    return (
+      <div className={styles['fil-toolbar-files']}>
+        <Toolbar
```
y-lohse: So you want to create a Toolbar, which has one or more highlighted actions, and several more inside the "more" menu. On mobile, all actions are inside the "more" menu.
This is how you use the API:

- create a `Toolbar` container
- list your actions as `ToolbarAction`, indicating if they should be visible and/or highlighted
- provide a Component to render for the `ToolbarAction` as a child

For actions that are always in the "more" menu, it's very easy, you just do something like

```
<ToolbarAction>
  <a className onClick>{label}</a>
<ToolbarAction/>
```

This could be made even easier by adding props to the `ToolbarAction` like `icon`, `label`, etc.

Things are trickier for actions that are renderer both outside *and* inside the menu, since they might look quite different and the `ToolbarAction` has no idea. In those cases, the components receive a `highlighted` prop, which tells them if they are renderer inside or outside of the menu.

src/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+  moreMenuClassName: PropTypes.string,
+  moreMenuButton: PropTypes.element.isRequired
+}
+
+Toolbar.defaultProps = {
+  moreMenuDisabled: false,
+  moreMenuClassName: ''
+}
+
+export class ToolbarAction extends Component {} // needs to extend Component to have `props.children`
```
y-lohse: This is what bothers me the most with the current implementation — `ToolbarAction ` is just a lie.
We use it as a way to declare all the actions and [be able to filter them](https://github.com/cozy/cozy-drive/pull/679/files#diff-cd8ecf10d1445222404e50d7f56d783fR21). But they are [never even rendered](https://github.com/cozy/cozy-drive/pull/679/files#diff-cd8ecf10d1445222404e50d7f56d783fR59) because we are only interested in the children.

@enguerran suggested using a HOC, since really, it's a HOC. But eventually i didn't do it because it changes the API from:

```
<Toolbar>
  <ToolbarAction visible highlighted={false}>
    <MyComponent />
  </ToolbarAction>
</Toolbar>
```

which is nice and hierchical, to something like

```
<Toolbar>
  {asToolbarAction(<MyComponent />, true, false)}
</Toolbar>
```

y-lohse:
[(5) ETQ utilisateur je peux voir les éléments partagés dans mon arborescence quand je la parcours](https://trello.com/c/jCsgxyuI/687-5-etq-utilisateur-je-peux-voir-les-%C3%A9l%C3%A9ments-partag%C3%A9s-dans-mon-arborescence-quand-je-la-parcours)

src/lib/cozy-client/slices/sharings.js
enguerran: Those permissions and sharing properties are very hard to follow. See https://cozy.github.io/cozy-stack/sharing.html

If I look at the `p.attributes.source_id.split('/')[1]`, I'll find `ce8835a061d0ef68947afe69a0046722` which exists three times on the response's json:

```json
{
  "data": {
    "type": "io.cozy.sharings",
    "id": "ce8835a061d0ef68947afe69a0046722",
    "meta": {
      "rev": "1-4859c6c755143adf0838d225c5e97882"
    },
    "attributes": {
      "sharing_type": "one-shot",
      "description": "sharing test",
      "preview_path": "/sharings/preview",
      "app_slug": "cal",
      "owner": true,
      "created_at": "2018-01-04T12:35:08Z",
      "updated_at": "2018-01-04T13:45:43Z"
    },
    "links": {
      "self": "/sharings/ce8835a061d0ef68947afe69a0046722"
    },
    "relationships": {
      "permissions": {
        "data": {
          "id": "46b25ad6-d044-11e7-af96-579d7e6c689e",
          "type": "io.cozy.permissions"
        }
      },
      "recipients": {
        "data": [
          {
            "id": "2a31ce0128b5f89e40fd90da3f014087",
            "type": "io.cozy.contacts",
            "status": "pending"
          }
        ]
      }
    }
  },
  "included": [
    {
      "type": "io.cozy.permissions",
      "id": "46b25ad6-d044-11e7-af96-579d7e6c689e",
      "meta": {
        "rev": "1-fbed00fed407"
      },
      "attributes": {
        "type": "shared-by-me",
        "source_id": "io.cozy.sharings/ce8835a061d0ef68947afe69a0046722",
        "codes": {
          "yuot7NaiaeGugh8T": "2a31ce0128b5f89e40fd90da3f014087"
        },
        "expires_at": 1483951978,
        "permissions": {
          "tests": {
            "description": "test",
            "type": "io.cozy.tests",
            "values": ["test-id"],
            "verbs": ["GET"]
          }
        }
      },
      "links": {
        "self": "/permissions/46b25ad6-d044-11e7-af96-579d7e6c689e"
      }
    },
    {
      "type": "io.cozy.contacts",
      "id": "2a31ce0128b5f89e40fd90da3f014087",
      "meta": {
        "rev": "1-461114b45855dc6acdb9bdc5d67e1092"
      },
      "attributes": {
        "email": {
          "address": "toto@fr"
        },
        "cozy": {
          "url": "url.fr"
        }
      },
      "links": {
        "self": "/contacts/2a31ce0128b5f89e40fd90da3f014087"
      }
    }
  ]
}
```

src/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+          })
+        )}
+        {isMobile ? <BarRight>{MoreMenu}</BarRight> : MoreMenu}
+      </div>
+    )
+  }
+
+  wrapInItem(component) {
+    return <Item>{component}</Item>
+  }
```
enguerran: Can't we merge this code and the `cloneElement` one. I mean:

```jsx
const wrap = comp => <div>{comp}</div>

const WrappedComp = wrap(React.cloneElement(<Item />, { highlighted: true }))
```

is the same as

```jsx
const WrappedComp = props => (
  <div>
    <Item highlighted={true} />
  </div>
)
```

but the second one is simpler, doesn't it?

src/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+        const hiddenStyle = highlighted
+          ? styles['u-hide--mob']
+          : styles['u-hide--desk']
+        return <span className={hiddenStyle}>{child}</span>
+      }
+    } catch (_) {
+      return null
+    }
+  }
+} // needs to extend Component to have `props.children`
```
enguerran: If a component is only a `class` with a `render` function, does it make sense to just use a functional component?

src/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+export class ToolbarAction extends Component {
+  render() {
+    try {
+      const { children, visible, highlighted, isSecondaryAction } = this.props
+
+      if (!visible) return null
+
+      const child = React.cloneElement(React.Children.only(children), {
+        highlighted
+      })
```
enguerran: Why do you use both JSX and React API? I find it very confusing.

src/drive/ducks/files/Container.jsx
```diff
@@ -39,7 +39,7 @@ const mapStateToProps = (state, ownProps) => ({
 
 const mapDispatchToProps = (dispatch, ownProps) => {
   const hasWriteAccess =
-    !ownProps.shared.withMe || ownProps.shared.sharingType === 'master-master'
+    !ownProps.shared.withMe || ownProps.shared.sharingType === 'two-way'
```
enguerran: o_O wow, I miss this one on the #653

src/drive/ducks/files/Container.jsx
```diff
@@ -39,7 +39,7 @@ const mapStateToProps = (state, ownProps) => ({
...
   return {
     actions: Object.assign({}, ownProps.actions, {
```
enguerran: We use `spread` instead of `Object.assign`

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -24,27 +23,67 @@ import { uploadedFile, downloadFiles, trashFiles } from '../../actions'
...
-})
-
 const ALERT_LEVEL_INFO = 'info'
 const ALERT_LEVEL_ERROR = 'error'
 const ALERT_LEVEL_SUCCESS = 'success'
 
-class Toolbar extends Component {
+const ActionShare = props => {
+  const { highlighted, shared, onClick, t } = props
+  let label, Component
```
enguerran: Using `let` must warn you that you miss a refactoring somewhere.

Same remark for ugly `if(){} else if {} else`.

Hard to read, hard to understand which value is set. Amha avoinding `if` statement is a good practice.

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -24,27 +23,67 @@ import { uploadedFile, downloadFiles, trashFiles } from '../../actions'
...
+  }
+
+  return highlighted ? (
+    <Component label={t(label)} onClick={onClick} />
+  ) : (
+    <a className={styles['fil-action-share']} onClick={onClick}>
+      {t(label)}
+    </a>
+  )
+}
```
enguerran: We have a conditional assignement for `label` and `Component`. And a conditional `return`. We may find something clearer.

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -58,146 +97,97 @@ class Toolbar extends Component {
...
-          <SharedWithMeButton
-            label={t('Files.share.sharedWithMe')}
-            onClick={() => this.setState(toggleShowShareModal)}
-            className={styles['u-hide--mob']}
-          />
-        )}
-
-        {isMobile ? <BarRight>{MoreMenu}</BarRight> : MoreMenu}
+          </ToolbarAction>
+        </Toolbar>
```
enguerran: This very long `<Toolbar><ToolbarAction /></Toolbar>` is very hard to read.
Can't we find a way to refactor and minimize the code to clarify its reading?

src/lib/cozy-client/slices/sharings.js
```diff
@@ -393,14 +391,14 @@ const getSharingForRecipient = (state, document, recipient) => {
...
   return [
-    ...perms.byMe.filter(
-      perm => perm.attributes.permissions.files.values.indexOf(id) !== -1
+    ...perms.byMe.filter(perm =>
+      perm.attributes.permissions.files.values.includes(id)
     ),
-    ...perms.withMe.filter(
-      perm => perm.attributes.permissions.rule0.values.indexOf(id) !== -1
+    ...perms.withMe.filter(perm =>
+      perm.attributes.permissions.rule0.values.includes(id)
```
enguerran: :+1:

src/sharing/SharingDetailsModal.jsx
```diff
@@ -3,7 +3,7 @@ import styles from './share.styl'
 import React, { Component } from 'react'
 import { cozyConnect, fetchSharings } from 'cozy-client'
 import { UserAvatar } from './components/Recipient'
-import Modal from 'cozy-ui/react/Modal'
+import Modal, { ModalContent } from 'cozy-ui/react/Modal'
 
 const Owner = UserAvatar
```
enguerran: Not good, we shall have

```
import { UserAvatar as Owner } from './components/Recipient'
```

⚠️ remember that object are mutable in JS:

```js
const elephant = {
 type: 'elephant'
}
const mouse = elephant
mouse.type = 'mouse'

console.log(elephant) // { type: 'mouse' }
console.log(mouse) // { type: 'mouse' }
```
So it's confusing to see `const Owner = UserAvatar`.

src/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+        disabled={moreMenuDisabled}
+        className={moreMenuClassName}
+        button={moreMenuButton}
+      >
+        {highlightedActions.map(component =>
+          this.wrapInItem(
+            React.cloneElement(component, {
+              highlighted: false,
+              isSecondaryAction: false
+            })
```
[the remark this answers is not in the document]
y-lohse: Because `Component` is not actually a component but a component instance — so I should rename it.
Trying what you say gives an error like `Unable to createElement because tag name is [Object object]`, because `Component` is a VNode at this point. *Maybe* `<Component.type />` would work but is that really better?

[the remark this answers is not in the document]
y-lohse: I'm not seeing it 👀  where ?

src/components/Toolbar/index.jsx
y-lohse: It would be simpler, but we need the `React.cloneElement`.

src/components/Toolbar/index.jsx
y-lohse: I'm only using the React API to clone elements and manipulate `children`, things I can't do in JSX AFAIK.

src/components/Toolbar/index.jsx
y-lohse: I'll try that again, because yes, a functional component would make sense. But apparently [`props.children`](https://github.com/developit/preact-compat/issues/366#issuecomment-295822533) isn't always added by preact on functional components? It certainly wasn't there when I tested it, but I've changed a few things since.

src/drive/ducks/files/Toolbar.jsx
y-lohse: I thought you may say that, but I disagree. I find this entirely readable, way more than any alternative I can think of. Variables and `if` statements are part of the language, it's ok to use them now and then.

src/drive/ducks/files/Toolbar.jsx
y-lohse: Any suggestions? I think `Toolbar` is quite good, and the sub-component should be `ToolbarSomething`. i'm not too happy with `Action` though.

src/sharing/SharingDetailsModal.jsx
y-lohse: You are correct and I'm changing it, but it's not my code :-)

src/components/Toolbar/index.jsx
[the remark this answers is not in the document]
enguerran:
```js
const Component = props => <div>Hello</div>
const ComponentInstance = <div>Hello</div>
```

JSX covers both, doesn't it?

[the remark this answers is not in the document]
enguerran: `highlightedActions.map` lets me believe you build a list for highlighted actions, but you enhance components with `highlighted: false`.
On the other hand, `secondaryActions.map` lets me believe _[blablabla]_, so I understand the enhancement with `highlighted: false`.

src/components/Toolbar/index.jsx
enguerran: That is my very question:

Do we ever need `React.cloneElement` within JSX if the purpose is not to keep `ref` value?

src/components/Toolbar/index.jsx
enguerran:
```jsx
const cloneComponent = BaseComponent => props => <BaseComponent {...props} />
```
seems to be almost the same as `React.cloneElement`, except in the JSX version, `ref` value attached to the `BaseComponent` will not be maintained. 

https://codesandbox.io/s/q9lv0j2k6

src/drive/ducks/files/Toolbar.jsx
enguerran: I hear that:
> Variables and if statements are part of the language, it's ok to use them now and then.

But ;-)

If I take [this very good repo](https://github.com/ryanmcdermott/clean-code-javascript) as an example, I agree with this one

> When you have classes and functions that have if statements, you are telling your user that your function does more than one thing (https://github.com/ryanmcdermott/clean-code-javascript#avoid-conditionals)

For `let`, I don't say you can't use them, I say it's a smell, maybe you should [favor functional over imperative programing](https://github.com/ryanmcdermott/clean-code-javascript#favor-functional-programming-over-imperative-programming) or maybe it's a [side effect](https://github.com/ryanmcdermott/clean-code-javascript#avoid-side-effects-part-1) (with [part2](https://github.com/ryanmcdermott/clean-code-javascript#avoid-side-effects-part-1)), etc.

src/drive/ducks/files/Toolbar.jsx
enguerran: Without giving it too much time, I will go with a configuration solution:

1. create an object or a set of value
1. loop on this objet to create different items

nono:
@y-lohse what do you want to do with this PR?

y-lohse:
I need to resolve the conflicts, and I think I'll get rid of the change that is making the test fails. It's something related to the sharing, so it's not that important anymore.

y-lohse:
I'll reopen this when I find the time to rebase it and get rid of the failing test.

PR #682, opened 2018-01-12
chore: fonts.css asset is already imported by the stack

jinroh:
The `fonts.css` asset imported from the main domain is already injected by the stack, with an immutable hash. This PR removes the import made by the application to avoid wasting a network roundtrip.


y-lohse:
Thanks !

PR #685, opened 2018-01-12
feat: navigate when handling cozydrive:// url

ptbrowne:
We use the `history` to navigate when we handle an URL `cozydrive://mylink`.

### DIFF
--- a/targets/drive/mobile/main.jsx
+++ b/targets/drive/mobile/main.jsx
@@ -38,6 +38,9 @@ if (__DEVELOPMENT__) {
   require('preact/devtools')
 }
 
+// Register callback for when the app is launched through cozydrive:// link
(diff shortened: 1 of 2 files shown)


targets/drive/mobile/main.jsx
```diff
@@ -38,6 +38,8 @@ if (__DEVELOPMENT__) {
   require('preact/devtools')
 }
 
+window.handleOpenURL = require('drive/mobile/lib/handleOpenURL').default(hashHistory)
```
y-lohse: How is the `handleOpenURL` name defined? Can it be changed to something more explicit?

ptbrowne: It is hardcoded in the plugin : I cannot rename the attribute of window. I can rename the lib into "deeplink". What do you think ?

ptbrowne: 👍 Done

PR #691, opened 2018-01-15
Use viewer for file

ptbrowne:
## Background

From Bank, we need to be able to open Drive on a specific file. We cannot retrieve the parent of the file (we have no permission on io.cozy.files) so we cannot redirect to `/folder/$folderId/file/$fileId` from bank.

## Solution

I changed FileOpener which had already the role of fetching file information and displaying it to use the Viewer.

The `/file/$fileId` route now fetches the file and redirects (router.replace not to add an entry in the back button history) to `/folder/$fileFolderId/file/$fileId`.

### DIFF
--- a/src/drive/ducks/files/FileOpener.js
+++ b/src/drive/ducks/files/FileOpener.js
@@ -4,9 +4,7 @@ import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { flowRight as compose } from 'lodash'
 
-import { Spinner, Modal, translate } from 'cozy-ui/react'
-
-import { getFileDownloadUrl } from '../../actions'
+import { Spinner, translate } from 'cozy-ui/react'
 import styles from './styles'
 
 const withAlert = Wrapped =>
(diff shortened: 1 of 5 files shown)


src/drive/ducks/files/FileOpener.js
```diff
@@ -18,55 +16,34 @@ class FileOpener extends Component {
...
-      this.setState({ loading: false })
     }
   }
 
-  handleClose = async () => {
-    this.setState({ closing: true })
-    const { router, params: { fileId } } = this.props
-    const { relationships: { parent } } = await cozy.client.files.statById(
-      fileId
-    )
```
ptbrowne: I stumbled upon a case where `relationships.parent` was `undefined` whereas dir_id was defined. This is why I use `dir_id` [above](https://github.com/cozy/cozy-drive/pull/691/files#diff-cec592dc3fb6c108dee8815cc445e469R28). Any idea why ?

y-lohse: Not really. @jinroh can you think of a case where that would happen?

ptbrowne:
> If I understand correctly, the main difference now is that the file gets opened in the Viewer?

Exactly, this route was meant to be implemented with the Viewer but it was not ready yet.

src/drive/ducks/files/FileOpener.js
```diff
@@ -18,55 +16,35 @@ class FileOpener extends Component {
...
-      const url = await getFileDownloadUrl(fileId)
-      this.setState({ url })
+      const fileInfo = await cozy.client.files.statById(fileId)
+      // Go to the parent folder, we replace since we do not want
+      // to add a new history entry
+      router.replace(`/folder/${fileInfo.attributes.dir_id}/file/${fileId}`)
     } catch (e) {
-      router.push('/')
+      console.warn(e)
+      // Go to the parent folder, we replace since we do not want
```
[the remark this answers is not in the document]
ptbrowne: Done

PR #697, opened 2018-01-17
chore: activate version pining for dependencies

y-lohse:
Following up on #696, activate pinning for regular dependencies too. We'll probably get a couple more PRs from renovate, but we won't have any sneaky `yarn.lock` updates to deal with either.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,7 @@
   "extends": [
     "config:base"
   ],
+  "pinVersions": true,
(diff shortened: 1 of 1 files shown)


y-lohse:
[L'envoi des logs à Sentry par les utilisateurs ne fonctionne pas](https://trello.com/c/mY6DZWas/751-lenvoi-des-logs-%C3%A0-sentry-par-les-utilisateurs-ne-fonctionne-pas)

PR #698, opened 2018-01-17
Polyfill

y-lohse:
Foolow up on #694 : turns out `whatwg-polyfill` is meant for node and isn't really safe to use with webpack, unless we add some extra configuration. it also provides a lot of things we don't really need, so i'm falling back to the other, simpler polyfill.

I'm also bumping the version number of cozy-authentication for an upcoming release.

### DIFF
--- a/src/authentication/package.json
+++ b/src/authentication/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cozy-authentication",
-  "version": "1.0.7",
+  "version": "1.1.0",
   "description": "Component providing login to a Cozy",
   "main": "build/index.js",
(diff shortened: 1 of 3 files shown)


y-lohse:
[(1) 🐛📱 Adresse du Cozy indiquée comme ne correspondant pas à un Cozy alors que si](https://trello.com/c/BfO2NpIL/677-1-%F0%9F%90%9B%F0%9F%93%B1-adresse-du-cozy-indiqu%C3%A9e-comme-ne-correspondant-pas-%C3%A0-un-cozy-alors-que-si)

PR #703, opened 2018-01-19
Upload fixups

y-lohse:
- Turns out we'd rather interrupt the upload as soon as we get a quota error, otherwise the user never sees it
- When the upload starts, `mobile.mediaBackup.something.current` is `0` so if we just test `current` instead of `current !== undefined` the uploader doesn't show up.


y-lohse:
[🐛📱 Quand on dépasse la limite d'espace disque pendant une sauvegarde de phoos, on n'affiche pas le message](https://trello.com/c/5LavNTPD/756-%F0%9F%90%9B%F0%9F%93%B1-quand-on-d%C3%A9passe-la-limite-despace-disque-pendant-une-sauvegarde-de-phoos-on-naffiche-pas-le-message)

PR #711, opened 2018-01-23
fix: remove fake link in onboarding

y-lohse:
I disabled links in the onboarding so it looks like this again:

capture d ecran 2018-01-23 10 36 37

I didn't find a simpler way to disable that link — see https://github.com/rexxars/react-markdown/issues/139


y-lohse:
[🐛📱 Exemple considéré comme un lien dans l'onboarding mobile](https://trello.com/c/GWpLStAA/760-%F0%9F%90%9B%F0%9F%93%B1-exemple-consid%C3%A9r%C3%A9-comme-un-lien-dans-lonboarding-mobile)

PR #712, opened 2018-01-23
feat: use <Viewer /> in <Embeder />

drazik:
The embeder was using an <embed> tag to show files in other apps. Mobile browsers didn't show PDFs using this technique.

Since there is now a Viewer component, we can use it, so on mobile browsers the user can download the file.


src/drive/ducks/files/FileOpenerExternal.js
```diff
@@ -0,0 +1,86 @@
...
+    this.loadFileInfo()
+  }
+
+  async loadFileInfo() {
+    const { alert } = this.props
+    try {
+      await this.props.fetchFile()
+      // Go to the parent folder, we replace since we do not want
+      // to add a new history entry
+      // router.replace(`/folder/${fileInfo.attributes.dir_id}/file/${fileId}`)
```
ptbrowne: To remove ^

src/drive/ducks/files/FileOpenerExternal.js
```diff
@@ -0,0 +1,86 @@
...
+    } catch (e) {
+      console.warn(e)
+      // Go to the root folder, we replace since we do not want
+      // to add a new history entry
```
ptbrowne: To remove ^

targets/drive/web/services.jsx
```diff
@@ -41,13 +42,14 @@ document.addEventListener('DOMContentLoaded', () => {
 
   const client = new CozyClient({
     cozyURL: cozyUrl,
-    token: data.cozyToken,
-    offline: { doctypes: ['io.cozy.files'] }
+    token: data.cozyToken
```
ptbrowne: Since in the service, we fetch via cozy.client.fetchDocument we had to deactivate offline otherwise the adapter for `io.cozy.files` was not the right one (pouchdbAdapter) and we had a 404. Is this a problem @goldoraf ?

src/drive/ducks/files/FileOpenerExternal.js
```diff
@@ -0,0 +1,86 @@
+import React, { Component } from 'react'
```
ptbrowne: We renamed FileOpener into FileOpenerExternal since it is either used in  : 

* an intent handler (aka service)
* via cozydrive://

Maybe @drazik you could add a docstring at the top of the file explaining this ?

src/drive/ducks/files/FileOpenerExternal.js
```diff
@@ -0,0 +1,86 @@
...
+      // Go to the parent folder, we replace since we do not want
+      // to add a new history entry
+      // router.replace(`/folder/${fileInfo.attributes.dir_id}/file/${fileId}`)
+    } catch (e) {
+      console.warn(e)
+      // Go to the root folder, we replace since we do not want
+      // to add a new history entry
+      alert({
+        message: 'alert.could_not_open_file'
+      })
```
ptbrowne: What happens if the file does not exist ? I am not sure that the Alerter is mounted in the service, so maybe it will not work.. Maybe the simplest thing would be to ditch the `alert` mechanism and render an error message ?

drazik: Actually it seems we never enter in the `catch` block. So I don't really know how to catch the error that occurs.

src/lib/cozy-client/adapters/CozyStackAdapter.js
```diff
@@ -78,10 +91,14 @@ export default class CozyStackAdapter {
   }
 
   async fetchDocument(doctype, id) {
-    const doc = await cozy.client.data.find(doctype, id)
-    // we normalize again...
-    const normalized = { ...doc, id: doc._id, _type: doc._type }
-    return { data: [normalized] }
+    let doc
+    if (doctype === FILES_DOCTYPE) {
+      doc = normalizeDoc(await cozy.client.files.statById(id), doctype)
```
goldoraf: Here you should pass `false` as a 2d arg to `statById` so that we're sure to hit the stack instead of the Pouch when an offline config for files exist (this is the stack adapter after all). See https://github.com/cozy/cozy-client-js/blob/master/src/files.js#L206

targets/drive/web/services.jsx
goldoraf: It seems legit, the service doesn't need offline support. But see my above comment

PR #717, opened 2018-01-24
fix: inject cozy-bar in services

drazik:
The cozy-bar was not injected in the services part of cozy-drive, causing errors when components were relying on cozy-bar's `BarLeft`, `BarCenter` or `BarRight` components.

Currently I propose a quick fix that consists in injecting the bar. Maybe there is a better way to fix this problem we can discuss.

### DIFF
--- a/targets/drive/web/services.ejs
+++ b/targets/drive/web/services.ejs
@@ -12,6 +12,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <% if (__STACK_ASSETS__) { %>
   {{.CozyClientJS}}
+  {{.CozyBar}}
   <% } %>
 </head>
 <div


y-lohse:
But do we want the cozy bar in services? Drive has 3 services at the moment :

- Returning a download URL for a file — has no UI, so it doesn't need the bar
- Return search results — no Ui either
- The embeder / file viewer — this one has a UI, but should it display the cozy-bar ?

ptbrowne:
@y-lohse We understand this, but we feel this is an optimization. The bar is used like this

```
const { BarLeft } = cozy.bar
```

So if a component (like Breadcrumb) is imported, it will break. The better solution would be to find why is Breadcrumb is in the chain of imports of the services when we import the Viewer but we are not so keen on diving into Drive codebase to figure out why this is the case :/

y-lohse:
> The better solution would be to find why is Breadcrumb is in the chain of imports of the services when we import the Viewer but we are not so keen on diving into Drive codebase to figure out why this is the case :/

I understand, but it reminds me of this : https://github.com/cozy/cozy-drive/pull/602 i'm worried we end up in the same situation again. How urgent is this fix on your end?

drazik:
This is completely breaking the embed feature, so I guess this is very urgent for the launch.

y-lohse:
Alright then, I'll create an issue for the more permanent fix.

PR #723, opened 2018-01-25
Viewer in the public shared folder view

goldoraf:
Please note that for simplicity reasons I decided to not use routing to open/close the viewer, but used a local state instead. I also cleaned up a bit of code now that we don't have to open files in tabs anymore.


goldoraf:
[(1) 🐛 La visionneuse sur la page publique d'un partage ne fonctionne pas](https://trello.com/c/sLxiwHUg/772-1-%F0%9F%90%9B-la-visionneuse-sur-la-page-publique-dun-partage-ne-fonctionne-pas)

PR #730, opened 2018-01-29
Enable folders in search results

y-lohse:
[this is the actual fix](https://github.com/cozy/cozy-drive/pull/730/files#diff-ab8f134c72c2b24476f993fcb9f5c5e2R57) : folders don't have a `trashed` property, and `undefined === false` is falsy so the folders were filtered out.

After fixing *that*, the app crashed because we tried to get an icon based on the folders `mime` property which again, doesn't exist.
[I handled the case of folder icons](https://github.com/cozy/cozy-drive/pull/730/files#diff-ab8f134c72c2b24476f993fcb9f5c5e2R97) separately because `folder` isn't a mime type, technically speaking.
I also made sure that [`getFileTypeFromMime ` won't just crash if called with `undefined`](https://github.com/cozy/cozy-drive/pull/730/files#diff-b60035d97397e5f3b868c2846f4130c4R11).

### DIFF
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -2,6 +2,8 @@ import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
 import { getFileTypeFromMime } from 'drive/lib/getFileTypeFromMime'
 
+const TYPE_DIRECTORY = 'directory'
+
 class SuggestionProvider extends React.Component {
   componentDidMount() {
     const { intent } = this.props
@@ -49,12 +51,12 @@ class SuggestionProvider extends React.Component {
       )
       const files = allDocs.data
 
-      const folders = files.filter(file => file.type === 'directory')
+      const folders = files.filter(file => file.type === TYPE_DIRECTORY)
 
       const normalizedFiles = files
-        .filter(file => file.trashed === false)
+        .filter(file => file.trashed !== true)
         .map(file => {
-          const isDir = file.type === 'directory'
+          const isDir = file.type === TYPE_DIRECTORY
           const dirId = isDir ? file._id : file.dir_id
           let path
           if (isDir) {
@@ -69,7 +71,7 @@ class SuggestionProvider extends React.Component {
             name: file.name,
             path,
             url: window.location.origin + '/#/folder/' + dirId,
-            icon: getIconUrl(file.mime)
+            icon: getIconUrl(file)
           }
         })
 
@@ -90,9 +92,11 @@ const icons = iconsContext.keys().reduce((acc, item) => {
   return acc
 }, {})
 
-function getIconUrl(mimetype) {
+function getIconUrl(file) {
   const keyIcon =
-    getFileTypeFromMime(icons)(mimetype) ||
+    (file.type === TYPE_DIRECTORY
+      ? 'folder'
+      : getFileTypeFromMime(icons)(file.mime)) ||
     console.warn(
       `No icon found, you may need to add a mapping for ${mimetype}`
     ) ||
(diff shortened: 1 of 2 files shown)


y-lohse:
[(1) Afficher les dossiers dans les résultats de la recherche](https://trello.com/c/UfqEgszG/648-1-afficher-les-dossiers-dans-les-r%C3%A9sultats-de-la-recherche)

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -49,12 +51,12 @@ class SuggestionProvider extends React.Component {
       )
       const files = allDocs.data
 
-      const folders = files.filter(file => file.type === 'directory')
+      const folders = files.filter(file => file.type === TYPE_DIRECTORY)
 
       const normalizedFiles = files
-        .filter(file => file.trashed === false)
+        .filter(file => file.trashed !== true)
```
enguerran: What about `file.trashed && file.trashed === false`? Is not it more expressive?

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -90,13 +92,15 @@ const icons = iconsContext.keys().reduce((acc, item) => {
...
-    'files'
+    file.type === TYPE_DIRECTORY
+      ? 'folder'
+      : getFileTypeFromMime(icons)(file.mime) ||
+        console.warn(
+          `No icon found, you may need to add a mapping for ${file.mime}`
+        ) ||
+        'files'
 
   return `${window.location.origin}/${icons[keyIcon]}`
```
enguerran: Browsers will add automatically `window.location.origin` if an url starts with a `/`. For instance `/icon` is the same as `localhost/icon`.

src/drive/ducks/services/components/SuggestionProvider.jsx
y-lohse: The thing is, if `file.trashed` is `undefined` we want to return `true`, so that the file/folder is not filtered out. `file.trashed && file.trashed === false` would return `false` for a `file.trashed = undefined`.

src/drive/ducks/services/components/SuggestionProvider.jsx
y-lohse: I think you wrote that line 😛 . But its is correct! The URL we return will be displayed inside an iframe, with a different `window.location.origin`, so it's important to return an absolute URL here.

src/drive/ducks/services/components/SuggestionProvider.jsx
enguerran: Why do we want to return `true` if `file.trashed === undefined`? That is counterintuitive.
If we extract the filter function into a named function it could be more explicit to understand this counterintuitive behavior:

```js
const notInTrash = file => file.trashed !== true
const normalizedFiles = files.filter(notInTrash)
```

I build a jsbin to illustrate what could be used to return true if value is `undefined`: https://jsbin.com/wurodamupa/edit?js,console

```js
const arr = [ true, false, undefined, null]

console.log(arr.map(val => val))
// [true, false, undefined, null]
console.log(arr.map(val => !val))
// [false, true, true, true]
console.log(arr.map(val => !!val))
// [true, false, false, false]
console.log(arr.map(val => val === false))
// [false, true, false, false]
console.log(arr.map(val => val !== true))
// [false, true, true, true]
```

y-lohse: > Why do we want to return true if file.trashed === undefined?

If a file has no `trashed` attribute, it means it's not trashed. AFAIK this is the case for folders at least, who don't have this attribute. Really, the only time a file is trashed is if the `trashed` attribute is `true`.

I'll use a named function as it does make things a bit clearer 👍

PR #731, opened 2018-01-30
refactor: use a named filter function for clarity

y-lohse:


src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -49,31 +49,31 @@ class SuggestionProvider extends React.Component {
...
-
-          return {
-            id: file._id,
-            name: file.name,
-            path,
-            url: window.location.origin + '/#/folder/' + dirId,
-            icon: getIconUrl(file)
-          }
-        })
+      const notInTrash = file => file.trashed !== true
```
enguerran: Why did not you use a more classical approach as explain in #730 comment:

> I build a jsbin to illustrate what could be used to return true if value is `undefined`: https://jsbin.com/wurodamupa/edit?js,console

```js
const arr = [ true, false, undefined, null]

console.log(arr.map(val => val))
// [true, false, undefined, null]
console.log(arr.map(val => !val))
// [false, true, true, true]
console.log(arr.map(val => !!val))
// [true, false, false, false]
console.log(arr.map(val => val === false))
// [false, true, false, false]
console.log(arr.map(val => val !== true))
// [false, true, true, true]
```

To be more clear `val => !val` is returning the exact same result as the non straigth forward `val => val !== true` event on `null` and `undefined`.

But I may miss something?

y-lohse: You're not, I read it too quickly, my bad. There's another change I have to make, I'll incorporate that change in it.

PR #737, opened 2018-01-31
Search: result order and trashed dirs

y-lohse:
- I amended the code that filters out trashed files and folders — it now *really* discards anything that's in the trash
- We had the correct results, but not in the correct order. This was due to a combination of several small problems, which I've all corrected

### DIFF
--- a/src/drive/ducks/services/FuzzyPathSearch.js
+++ b/src/drive/ducks/services/FuzzyPathSearch.js
@@ -74,6 +74,7 @@ class FuzzyPathSearch {
     let wordsFromPreviousQuery = []
 
     for (let currentQuerySegment of query) {
+      let wasInPreviousQuery = false
       for (let previousQuerySegment of this.previousQuery) {
         if (currentQuerySegment.word.includes(previousQuerySegment.word)) {
           if (currentQuerySegment.word !== previousQuerySegment.word) {
@@ -84,12 +85,13 @@ class FuzzyPathSearch {
             wordsFromPreviousQuery.push(currentQuerySegment)
           }
 
+          wasInPreviousQuery = true
           continue
         }
       }
 
       // this segment wasn't included in any previous query segment so it's a new word and we prioritize it
-      priorizedWords.push(currentQuerySegment)
+      if (!wasInPreviousQuery) priorizedWords.push(currentQuerySegment)
     }
 
     return this.sortQuerybyLength(priorizedWords).concat(
@@ -106,6 +108,11 @@ class FuzzyPathSearch {
 
     files.forEach(file => {
       let fileScore = 0
+      const pathArray = removeDiacritics(
+        (file.path + '/' + file.name).toLowerCase()
+      )
+        .split('/')
+        .filter(pathChunk => !!pathChunk)
 
       for (let word of words) {
         // let the magic begin...
@@ -113,9 +120,6 @@ class FuzzyPathSearch {
         let wordScore = 0
         let wordOccurenceValue = 10000
         let firstOccurence = true
-        const pathArray = removeDiacritics(
-          (file.path + '/' + file.name).toLowerCase()
-        ).split('/')
         const maxDepth = pathArray.length
 
         for (let depth = 0; depth < maxDepth; ++depth) {
@@ -159,7 +163,7 @@ class FuzzyPathSearch {
     })
 
     suggestions.sort((a, b) => {
-      const score = a.score - b.score
+      const score = b.score - a.score
       return score !== 0 ? score : a.file.path.localeCompare(b.file.path)
     })
 
--- a/src/drive/ducks/services/FuzzyPathSearch.spec.js
+++ b/src/drive/ducks/services/FuzzyPathSearch.spec.js
@@ -1,0 +1,237 @@
+import FuzzyPathSearch from './FuzzyPathSearch'
+
+describe('simple search', () => {
+  const guitars = [
+    { name: 'fender stratocaster', path: '' },
+    { name: 'fender telecaster', path: '' },
+    { name: 'gibson SG', path: '' }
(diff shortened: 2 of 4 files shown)


src/drive/ducks/services/FuzzyPathSearch.js
```diff
@@ -84,12 +85,13 @@ class FuzzyPathSearch {
...
           }
 
+          wasInPreviousQuery = true
           continue
         }
       }
 
       // this segment wasn't included in any previous query segment so it's a new word and we prioritize it
-      priorizedWords.push(currentQuerySegment)
+      if (!wasInPreviousQuery) priorizedWords.push(currentQuerySegment)
```
y-lohse: Before this fix, if we ran the exact same query twice in a row, we would accidentally end up duplicating the query words, ie. `['admin']` would turn into `['admin', 'admin']`

src/drive/ducks/services/FuzzyPathSearch.js
```diff
@@ -106,16 +108,18 @@ class FuzzyPathSearch {
 
     files.forEach(file => {
       let fileScore = 0
+      const pathArray = removeDiacritics(
+        (file.path + '/' + file.name).toLowerCase()
+      )
+        .split('/')
+        .filter(pathChunk => !!pathChunk)
```
y-lohse: We were missing this last filter, which discards unimportant parts of the path.

src/drive/ducks/services/FuzzyPathSearch.js
```diff
@@ -159,7 +163,7 @@ class FuzzyPathSearch {
     })
 
     suggestions.sort((a, b) => {
-      const score = a.score - b.score
+      const score = b.score - a.score
```
y-lohse: Subtle, but kinda important.

y-lohse:
[(1) Afficher les dossiers dans les résultats de la recherche](https://trello.com/c/UfqEgszG/648-1-afficher-les-dossiers-dans-les-r%C3%A9sultats-de-la-recherche)

src/drive/ducks/services/FuzzyPathSearch.js
enguerran: a Set instead of an Array could let us add only once the same item: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

enguerran: I think the use of this variable is very ugly, but the double for loop and the double if statement are too. So that is not a prioritize.

src/drive/ducks/services/FuzzyPathSearch.js
enguerran: Yes:
> If compareFunction(a, b) is less than 0, sort a to an index lower than b, i.e. a comes first.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -53,7 +53,8 @@ class SuggestionProvider extends React.Component {
       const files = allDocs.data
       const folders = files.filter(file => file.type === TYPE_DIRECTORY)
 
-      const notInTrash = file => file.trashed !== true
+      const notInTrash = file =>
+        file.trashed || !/^\/\.cozy_trash/.test(file.path)
```
enguerran: if `file.trashed === true` then `notInTrash` will return `true`. Are you sure?

y-lohse: Jesus, no, thank you ><

y-lohse:
There are no tests, but it's very testable code. Let me see if I can add some today!

y-lohse:
I had never tried out jest, [this is ridiculously easy](https://facebook.github.io/jest/docs/en/getting-started.html). A decent test coverage of this file is definitely on the menu for this PR.

y-lohse:
I've added tests for a bunch of different things (simple and more complex cases, successive queries, result ordering)… They're definitely not complete and the structure could be better… but it's better than nothing.

src/drive/ducks/services/FuzzyPathSearch.spec.js
```diff
@@ -0,0 +1,237 @@
+import FuzzyPathSearch from './FuzzyPathSearch'
+
+describe('simple search', () => {
```
enguerran: On ne teste pas le cas ou une `query` ne retourne aucun résultat.

src/drive/ducks/services/FuzzyPathSearch.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    const result = fps.search(query)
+
+    expect(result).toBeInstanceOf(Array)
+    expect(result.length).toEqual(2)
+    expect(result.includes(guitars[0])).toBe(true)
+    expect(result.includes(guitars[1])).toBe(true)
+  })
+})
+
+describe('search with path', () => {
```
enguerran: il manque aussi peut-être le cas ou on n'a un path qui n'existe pas.

src/drive/ducks/services/FuzzyPathSearch.spec.js
```diff
@@ -0,0 +1,237 @@
...
+
+    const reversedResult = reversedFps.search(query)
+
+    expect(reversedResult).toBeInstanceOf(Array)
+    expect(reversedResult.length).toEqual(2)
+    expect(reversedResult[0]).toBe(guitars[1])
+    expect(reversedResult[1]).toBe(guitars[0])
+  })
+
+  it('should favor matches nearer the start of the path', () => {
```
enguerran: :+1:

PR #740, opened 2018-01-31
feat: :iphone: send feedback from settings

enguerran:
- 2b009667 (Enguerran, 23 seconds ago)
   feat: :iphone: send feedback from settings

- 61b92abe (Enguerran, 4 hours ago)
   refactor: ratingmodal components - remove binds - use class properties -
   update syntax

- e0446255 (Enguerran, 28 hours ago)
   chore: change scheme for pouchdb-adapter-cordova-sqlite dependence
   git+https:// to https://

# 1. Settings

Settings

# 2. Feedback Form

Feedback Form

# 3. Alert

Alert

### DIFF
--- a/src/drive/mobile/containers/RatingModal.jsx
+++ b/src/drive/mobile/containers/RatingModal.jsx
@@ -1,4 +1,4 @@
-/* global cozy AppRate */
+/* global cozy */
 import React, { Component } from 'react'
 import Modal, { ModalContent } from 'cozy-ui/react/Modal'
 import { Button } from 'cozy-ui/react'
@@ -20,14 +20,11 @@ const PROMPT_AFTER_BOOTS = 5
 
 // RatingModal is the base component
 class RatingModal extends Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      screen: SCREEN_ENJOY
-    }
+  state = {
+    screen: SCREEN_ENJOY
   }
 
-  async onUserReply(enjoyed) {
+  onUserReply = async enjoyed => {
     if (enjoyed) {
       try {
         const { t } = this.context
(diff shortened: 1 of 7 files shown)


enguerran:
I tried to refactor the code to improve it, but I failed and rollbacked to a less ambitious Pull Request.

But it will be very good to refactor:

- RatingModal and inner components
- Settings and its not very useful generation-by-configuration feature

Before refactoring, we must add `prop-types` and think about the best way to break the UI into component hierarchy and identify where state should live.

src/drive/mobile/containers/RatingModal.jsx
```diff
@@ -166,48 +100,43 @@ const promptRating = ({
...
-      AppRate.promptForRating()
-    } catch (e) {
-      reject(e)
     }
-  })
+
+    window.AppRate.promptForRating()
+  } else {
+    Promise.resolve()
+  }
```
y-lohse: I haven't tested the code yet, but pretty sure this change is breaking things, because `promptRating` doesn't return anything anymore, [but we want it to return](https://github.com/enguerran/cozy-drive/blob/2b00966701a7e8393d2d63054c2852035e58a436/src/drive/mobile/containers/RatingModal.jsx#L30) the result of `onButtonClicked`

src/drive/mobile/containers/RatingModal.jsx
```diff
@@ -219,29 +148,36 @@ const withBootDelay = (WrappedComponent, showAfterBoots) => {
...
-      ) : null
+      return (
+        visible && (
+          <WrappedComponent
+            dontShowAgain={this.dontShowAgain}
+            showLater={this.showLater}
+            {...this.props}
+          />
+        )
+      )
```
y-lohse: Why did you change this? As far as I can tell, they do the same, but [it looks to me like the preferred thing to return from a component when you don't want it to render anything](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) is `null` rather than `false`.

src/drive/mobile/containers/RatingModal.jsx
enguerran: Obviously I failed my rollback ;) Well catched.

src/drive/mobile/containers/RatingModal.jsx
enguerran: Cannot find the article that proposed to use condition instead of ternary operator. We do this often:

```jsx
return (
  <div>
    {isError && <Error error={error} />}
    Count: {count}
  </div>
);
```

src/drive/mobile/containers/RatingModal.jsx
```diff
@@ -1,10 +1,10 @@
-/* global cozy AppRate */
 import React, { Component } from 'react'
 import Modal, { ModalContent } from 'cozy-ui/react/Modal'
-import { Button } from 'cozy-ui/react'
+import Button from 'cozy-ui/react/Button'
 import { connect } from 'react-redux'
 import withPersistentState from '../lib/withPersistentState'
 import { SOFTWARE_ID, SOFTWARE_NAME } from '../lib/constants'
+import FeedbackForm from './FeedbackForm'
```
enguerran: updated!

src/drive/mobile/containers/RatingModal.jsx
```diff
@@ -60,115 +56,54 @@ class RatingModal extends Component {
...
+          </Button>
+        </div>
+      </ModalContent>
+    </Modal>
+  )
 }
 
 // promptRating is not a component because the native UI is used instead
-const promptRating = ({
+const promptRating = async ({
```
enguerran: updated!

src/drive/mobile/containers/RatingModal.jsx
enguerran: Do you agree? Or do you want I change it back?

enguerran:
@y-lohse I updated the commit, do you want to change your review?

src/drive/mobile/containers/RatingModal.jsx
y-lohse: No I don't really mind, I was just curious to know why you changed it since they both work the same.

I know we use ternary's *within* render functions, but in this case it's about the return value of the render function. Not quite the same. The react doc linked above talks about both cases :)

enguerran:
[(2) 📱Faciliter la remontée de suggestions et de bugs](https://trello.com/c/QCRP7vvB/785-2-%F0%9F%93%B1faciliter-la-remont%C3%A9e-de-suggestions-et-de-bugs)

PR #741, opened 2018-02-01
fix: hide settings on mobile

y-lohse:
Evolution of the previous fix : now the app menu can take all the available height, and nothing iss displayed on tablets.

Note that this stylus file is only used for native builds, so it doesn't affect the web version of Cozy Drive.

### DIFF
--- a/src/drive/mobile/styles/main.styl
+++ b/src/drive/mobile/styles/main.styl
@@ -1,3 +1,5 @@
+@require 'settings/breakpoints.styl'
+
 html {
     -webkit-user-select: none;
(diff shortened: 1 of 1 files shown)


y-lohse:
[📱 Le menu apps ne prend que 50% de la hauteur même s'il n'y a rien en dessous](https://trello.com/c/03TS0Iij/790-%F0%9F%93%B1-le-menu-apps-ne-prend-que-50-de-la-hauteur-m%C3%AAme-sil-ny-a-rien-en-dessous)

PR #763, opened 2018-02-12
fix: change sharing modal order and titles

y-lohse:


y-lohse:
[(0,5) Améliorer la modale de partage actuelle](https://trello.com/c/3CjoJMAY/797-05-am%C3%A9liorer-la-modale-de-partage-actuelle)

PR #766, opened 2018-02-12
fix: modal appears twice in the DOM

y-lohse:
We recently changed the `SharingModal` so that it renders using a Portal, as provided by cozy-ui. This caused a really weird bug where the Modal was rendered *twice*. And if you close the first modal, the second is still there and you can't change it.

I narrowed down the problem a little bit and found a workaround — if the component returning the Modal isn't connected to cozy-client, the bug doesn't happen. 
I will create an issue in cozy-ui tomorrow with everything I know, and search a bit more. But i already spent several hours on this, and it's very, very strange. I doubt we'll find a better solution soon.

### DIFF
--- a/src/sharing/ShareModal.jsx
+++ b/src/sharing/ShareModal.jsx
@@ -44,11 +44,10 @@ const ShareByEmailComingSoon = shunt(
   ComingSoon
 )
 
-export class ShareModal extends Component {
+class ModalContent extends Component {
   render() {
     const { t } = this.context
     const {
-      onClose,
       sharing,
       sharingDesc,
       share,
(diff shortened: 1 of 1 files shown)


y-lohse:
[(0,5) Améliorer la modale de partage actuelle](https://trello.com/c/3CjoJMAY/797-05-am%C3%A9liorer-la-modale-de-partage-actuelle)

enguerran:
I guess it is an old thing I spotted: we render some components too often. I can't remember where this comes from, but I know we have some bad usage of component lifecycle.

At least, your code decouple things, so I approve it if you say it fixes the problem.

PR #767, opened 2018-02-13
chore: update transifex config

y-lohse:
Attempting to fix transifex


y-lohse:
Nevermind, everything is ok here.

PR #786, opened 2018-02-26
fix: close banner button

y-lohse:


y-lohse:
[Corriger un bug de style](https://trello.com/c/bGRptvxN/824-corriger-un-bug-de-style)

PR #790, opened 2018-02-28
chore: switch to single sentry project

y-lohse:
As discussed, we don't need 2 sentry projects, we can do everything we want with the `environment` tag.


y-lohse:

PR #791, opened 2018-03-01
feat: add task to upload sourcemaps to sentry

y-lohse:
cc the banksters team since it might be of interest to you as well @kosssi @ptbrowne @drazik @doubleface

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -6,10 +6,11 @@ module.exports = {
   output: {
     filename: '[name].[hash].min.js'
   },
-  devtool: false,
+  devtool: 'source-map',
   plugins: [
     new optimize.UglifyJsPlugin({
       mangle: true,
+      sourceMap: true,
       compress: {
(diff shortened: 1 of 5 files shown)


y-lohse:

config/webpack.config.prod.js
```diff
@@ -6,10 +6,11 @@ module.exports = {
   output: {
     filename: '[name].[hash].min.js'
   },
-  devtool: false,
+  devtool: '#source-map',
```
goldoraf: Why the `#`? Are you sure it is necessary?

y-lohse: That's actually a good question, I just copied it from the `dev` config. I'll experiment a bit more [with that](https://webpack.js.org/configuration/devtool/), looks like there are better options.

y-lohse: You were correct, there's no need for the `#`. I also switched the dev sourcemaps to a faster version.

PR #796, opened 2018-03-02
fix: proper client revokation detection

y-lohse:
I've essentially inverted the logic — instead of treating everything like a revocation *except* offline errors, we know consider every error as "it's fine, we're just offline" *except* the ones that really tell us that the client has been revoked.


y-lohse:
[📱🐛 Gérer les erreurs de stack quand couch est mort](https://trello.com/c/FL6PXeiU/845-%F0%9F%93%B1%F0%9F%90%9B-g%C3%A9rer-les-erreurs-de-stack-quand-couch-est-mort)

PR #801, opened 2018-03-08
fix: upload sourcemaps on correct path

y-lohse:
This will only work for Android, still looking for a solution for iOS.


y-lohse:

PR #816, opened 2018-03-12
fix: polyfill search params

y-lohse:
Added missing polyfill for `URL.searchParams`, required for MS Edge (and others).


y-lohse:
[[Photos] La fenêtre de partage ne s’affiche pas sous Microsft Edge](https://trello.com/c/8hXeHb37/854-photos-la-fen%C3%AAtre-de-partage-ne-saffiche-pas-sous-microsft-edge)

PR #818, opened 2018-03-13
Button migration

y-lohse:
Updated the buttons to use the new API


y-lohse:
[(1) Corriger un problème de style sur les boutons avec icone](https://trello.com/c/s3TiWnzy/849-1-corriger-un-probl%C3%A8me-de-style-sur-les-boutons-avec-icone)

src/components/pushClient/Banner.jsx
```diff
@@ -59,7 +59,7 @@ class BannerClient extends Component {
             this.markAsSeen('banner')
           }}
         >
-          <span>{t('Nav.btn-client-mobile')}</span>
+          {t('Nav.btn-client-mobile')}
```
GoOz: Shouldn't it be in a `label` props ?

src/components/pushClient/Banner.jsx
```diff
@@ -81,9 +81,8 @@ class BannerClient extends Component {
           onClick={e => {
             this.markAsSeen('close')
           }}
-        >
-          <Icon icon="cross" width="24" height="24" />
-        </Button>
+          icon="cross"
```
GoOz: In this case, wouldn't it be best to write `icon={<Icon icon="cross" width="24" height="24" />}`?
Because we're losing the size there.

src/drive/containers/File.jsx
```diff
@@ -196,9 +196,7 @@ const FileAction = ({ onClick }) => (
       styles['fil-content-file-action']
     )}
   >
-    <Button theme="action" onClick={onClick} extension="narrow">
-      <Icon icon="dots" color="charcoalGrey" width="17" height="17" />
-    </Button>
+    <Button theme="action" onClick={onClick} extension="narrow" icon="dots" />
```
GoOz: I guess you should use `icon={<Icon icon="dots" color="charcoalGrey" width="17" height="17" />}` here too.

src/viewer/NoNetworkViewer.jsx
```diff
@@ -7,9 +7,11 @@ import styles from './styles'
 const NoNetworkViewer = ({ t, onReload }) => (
   <div className={styles['pho-viewer-canceled']}>
     <h2>{t('Viewer.loading.error')}</h2>
-    <Button theme="regular" onClick={onReload}>
-      {t('Viewer.loading.retry')}
-    </Button>
+    <Button
+      theme="regular"
```
GoOz: `theme="regular"` is no longer useful, as regular is the default theme. Not that it will break but it's just redundant.

src/components/pushClient/Banner.jsx
y-lohse: Well now, with version 7.7.2, yes ! Before that the `label` prop for `ButtonLinks` was broken. 👍

PR #822, opened 2018-03-16
chore: remove enzyme dep

y-lohse:
It doesn't look like enzyme is used at all : https://github.com/cozy/cozy-drive/pull/636

So bye bye !


y-lohse:
It *is* used after all! https://travis-ci.org/cozy/cozy-drive/jobs/354406416#L682

PR #834, opened 2018-03-20
fix: merge-skip bug

y-lohse:
I think a small bug slipped [here](https://github.com/cozy/cozy-drive/commit/696bc3b6d3b9700d60e00e7dfdb984056993854d) : after that change, in order to combine the results instead of replacing them, the action would need to have both `skip` and `merge` to `true`, but I'm pretty sure any one of them is enough.


y-lohse:
[Corriger le problème de chargement infini dans Photos](https://trello.com/c/AEn4kDvC/881-corriger-le-probl%C3%A8me-de-chargement-infini-dans-photos)

PR #837, opened 2018-03-26
feat: added create cozy button

y-lohse:
Adds a button on the public pages that invites the user to create a cozy. Easy.

### DIFF
--- a/src/components/Button/CozyHome.jsx
+++ b/src/components/Button/CozyHome.jsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'cozy-ui/react/I18n'
+import { ButtonLink } from 'cozy-ui/react'
+
+const CozyHome = ({ t, from }) => (
+  <ButtonLink
+    label={t('Share.create-cozy')}
+    href={`https://cozy.io/try-it?from=${encodeURIComponent(from)}`}
+  />
+)
+
+CozyHome.propTypes = {
+  t: PropTypes.func.isRequired,
+  from: PropTypes.string
+}
+
+CozyHome.defaultProps = {
+  from: ''
+}
+
+export default translate()(CozyHome)
--- a/src/drive/components/DownloadButton.jsx
+++ b/src/drive/components/DownloadButton.jsx
@@ -1,13 +1,20 @@
 import React from 'react'
 import { Button } from 'cozy-ui/react'
 
-const DownloadButton = ({ label, disabled = false, onDownload, className }) => (
+const DownloadButton = ({
+  label,
+  disabled = false,
+  onDownload,
+  className,
+  theme
+}) => (
   <Button
     disabled={disabled}
     className={className}
     onClick={() => onDownload()}
     icon="download"
     label={label}
+    theme={theme}
   />
 )
 
--- a/src/drive/components/DownloadButton.jsx
+++ b/src/drive/components/DownloadButton.jsx
@@ -1,21 +1,8 @@
 import React from 'react'
 import { Button } from 'cozy-ui/react'
 
-const DownloadButton = ({
-  label,
-  disabled = false,
-  onDownload,
-  className,
-  theme
-}) => (
-  <Button
-    disabled={disabled}
-    className={className}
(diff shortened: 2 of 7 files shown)


src/drive/locales/en.json
```diff
@@ -18,7 +18,7 @@
       "https://play.google.com/store/apps/details?id=io.cozy.drive.mobile",
     "link-client-ios":
       "https://itunes.apple.com/us/app/cozy-drive/id1224102389?mt=8",
-    "link-client-web": "https://cozy.io/fr/try-it?from=albumsharing"
+    "link-client-web": "https://cozy.io/fr/try-it"
```
y-lohse: Not really related with this PR, sorry, but I was too lazy to do a separate PR just for this.

y-lohse:
[(1) Créer un Cozy depuis la page publique de partage](https://trello.com/c/nky7QeNM/876-1-cr%C3%A9er-un-cozy-depuis-la-page-publique-de-partage)

src/components/Button/CozyHome.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'cozy-ui/react/I18n'
+import { ButtonLink } from 'cozy-ui/react'
+
+const CozyHome = ({ t, from }) => (
+  <ButtonLink
+    label={t('Share.create-cozy')}
+    href={`https://cozy.io/try-it?from=${encodeURIComponent(from)}`}
```
enguerran: There is some convention to detect sources in links, like __utc__ query parameters, no?

src/components/Button/CozyHome.jsx
```diff
@@ -0,0 +1,22 @@
...
+  />
+)
+
+CozyHome.propTypes = {
+  t: PropTypes.func.isRequired,
```
enguerran: I think you should get `t` from context rather than `props` or `props` should be at least optional with a default value in the button.

src/components/Button/CozyHome.jsx
```diff
@@ -0,0 +1,22 @@
...
+  from: PropTypes.string
+}
+
+CozyHome.defaultProps = {
+  from: ''
```
enguerran: If we omit `from`, then the link will still contain the query parameter `from`. 

```js
const url = `https://cozy.io/try-it${from ? `?from=${encodeURIComponent(from)}` : ''}`
```

or use the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL).

src/drive/components/DownloadButton.jsx
```diff
@@ -1,13 +1,20 @@
...
 import { Button } from 'cozy-ui/react'
 
-const DownloadButton = ({ label, disabled = false, onDownload, className }) => (
+const DownloadButton = ({
+  label,
+  disabled = false,
+  onDownload,
+  className,
+  theme
+}) => (
```
enguerran: We could be a little more flexible by writing:
```jsx
const DownloadButton = ({onDownload, ...props}) => (
  <Button
    onClick={() => props.onDownload()}
    icon="download"
    ...props
  />
)
```

src/drive/locales/en.json
enguerran: Shouldn't that be `"link-client-web": "https://cozy.io/en/try-it"`

src/components/Button/CozyHome.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'cozy-ui/react/I18n'
+import { ButtonLink } from 'cozy-ui/react'
+
+const CozyHome = ({ t, from }) => (
```
enguerran: This name does not reflect the fact that it is a button or a link.

src/drive/locales/en.json
y-lohse: Good catch, it shouldn't have a language at all actually.

src/components/Button/CozyHome.jsx
y-lohse: Sorry, I don't understand what you mean?

PR #842, opened 2018-03-28
fix: show revoked error for all types of errors

y-lohse:
There are other types of errors where we want to display the "revoked" screen — for example 404, or expired tokens.

### DIFF
--- a/src/drive/components/LightFolderView.jsx
+++ b/src/drive/components/LightFolderView.jsx
@@ -61,10 +61,7 @@ class DumbFolderView extends React.Component {
         getFolderIdFromRoute(this.props.location, this.props.params)
       )
       .then(e => {
-        if (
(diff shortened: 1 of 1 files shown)


y-lohse:
[(1) Le partage par lien n'est pas révoqué à la suppression d'un dossier](https://trello.com/c/5R1AjkNP/871-1-le-partage-par-lien-nest-pas-r%C3%A9voqu%C3%A9-%C3%A0-la-suppression-dun-dossier)

y-lohse:
I'm not *completely* sure, because I don't know all the errors… but I would say yes, at least for now. The problem is that we don't always get accurate errors from the stack either — for a file that doesn't exist for example, we get a 404, even if we don't have the permission to access it. 
I think it's better to do it like this, and maybe start adding exceptions if we find errors that we want to let through after all.

PR #848, opened 2018-03-29
fix: some annoying stuff

enguerran:
Work with @jsilvestre


src/drive/ducks/files/Toolbar.jsx
```diff
@@ -131,17 +135,27 @@ class Toolbar extends Component {
...
+          ) : (
+            <Item>
+              <a
+                className={classNames(styles['fil-action-delete'])}
+                onClick={() => leaveFolder(displayedFolder)}
+              >
+                {t('toolbar.trash')}
+              </a>
+            </Item>
+          ))}
```
y-lohse: If I'm not missing anything, it's only the label changing here, so we could do `{t(shared.withMe ? 'toolbar.leave' : 'toolbar.trash')}`and leave the rest all the time, no?

enguerran: I forgot the action that changes too.

enguerran: I would like us to sell the whole menu in a more functional way.
We can't have so many conditions and not care that it becomes unreadable and difficult to maintain, when in fact, functionally, it's quite simple: you have buttons that must be present under certain conditions.

PR #852, opened 2018-04-04
fix: external file opener

y-lohse:
I'm pretty sure pouch also uses `_id` and not `id` for documents, but somehow this old piece of code uses `id`.
I've at least one case where it's definitely `_id` and **not** `id` which causes a bug, so I've added some fallbacks but left the existing behavior as the default one.

(This caused a bug when opening a file in Drive coming from Banks)


y-lohse:
[La visionneuse mobile ne s'ouvre pas depuis l'app Bancaire](https://trello.com/c/l8GUpHuk/909-la-visionneuse-mobile-ne-souvre-pas-depuis-lapp-bancaire)

ptbrowne:
> I'm pretty sure pouch also uses _id and not id for documents, but somehow this old piece of code uses id.

It depends.

http://docs.couchdb.org/en/2.0.0/api/document/common.html#put--db-docid

Inside the document it is always `_id`, but when returning results about your operation, couch uses `id`. Same for Pouch.

PR #854, opened 2018-04-09
Sentry sourcemaps upload — now with better pathes

y-lohse:
I've copy-pasted some code from the internets.

### DIFF
--- a/src/drive/mobile/lib/reporter.js
+++ b/src/drive/mobile/lib/reporter.js
@@ -10,7 +10,7 @@ export const ANALYTICS_URL =
 const normalizeUrl = (url, pathStripRe) =>
   url.replace(/^file:\/\//, 'app://').replace(pathStripRe, '')
 
-const normalizeData = data => {
+export const normalizeData = data => {
   const PATH_STRIP_RE = /^.*\/[^.]+(\.app|CodePush|.*(?=\/))/
 
   if (data.culprit) {
--- a/src/drive/mobile/lib/reporter.js
+++ b/src/drive/mobile/lib/reporter.js
@@ -6,11 +6,34 @@ export const ANALYTICS_URL =
     ? ''
     : `https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/6`
 
+// normalize files path on mobile, see https://github.com/getsentry/sentry-cordova/blob/17e8b3395e8ce391ecf28658d0487b97487bb509/src/js/SentryCordova.ts#L213
+const normalizeUrl = (url, pathStripRe) =>
+  'app://' + url.replace(/^file\:\/\//, '').replace(pathStripRe, '')
+
+const normalizeData = data => {
+  const PATH_STRIP_RE = /^.*\/[^\.]+(\.app|CodePush|.*(?=\/))/
+
+  if (data.culprit) {
+    data.culprit = normalizeUrl(data.culprit, PATH_STRIP_RE)
+  }
+  const stacktrace =
+    data.stacktrace || (data.exception && data.exception.values[0].stacktrace)
+  if (stacktrace) {
+    stacktrace.frames.forEach(frame => {
+      if (frame.filename !== '[native code]') {
+        frame.filename = normalizeUrl(frame.filename, PATH_STRIP_RE)
+      }
+    })
+  }
+  return data
+}
+
 export const getReporterConfiguration = () => ({
   shouldSendCallback: true,
   environment: __DEVELOPMENT__ ? 'development' : 'production',
   release: __APP_VERSION__,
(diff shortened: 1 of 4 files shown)


y-lohse:
[(1) 📱 Améliorer notre gestion de Sentry](https://trello.com/c/oaQKEI6x/811-1-%F0%9F%93%B1-am%C3%A9liorer-notre-gestion-de-sentry)

src/drive/mobile/lib/reporter.js
```diff
@@ -6,11 +6,34 @@ export const ANALYTICS_URL =
     ? ''
     : `https://${__SENTRY_TOKEN__}@sentry.cozycloud.cc/6`
 
+// normalize files path on mobile, see https://github.com/getsentry/sentry-cordova/blob/17e8b3395e8ce391ecf28658d0487b97487bb509/src/js/SentryCordova.ts#L213
+const normalizeUrl = (url, pathStripRe) =>
+  'app://' + url.replace(/^file:\/\//, '').replace(pathStripRe, '')
```
enguerran: Can't we use the `url.replace` to _replace_ `file://` with `app://`:

```js
url.replace(/^file:\/\//, 'app://').replace(pathStripRe, '')
```

src/drive/mobile/lib/reporter.js
```diff
@@ -6,11 +6,34 @@ export const ANALYTICS_URL =
...
+
+const normalizeData = data => {
+  const PATH_STRIP_RE = /^.*\/[^.]+(\.app|CodePush|.*(?=\/))/
+
+  if (data.culprit) {
+    data.culprit = normalizeUrl(data.culprit, PATH_STRIP_RE)
+  }
+  const stacktrace =
+    data.stacktrace || (data.exception && data.exception.values[0].stacktrace)
```
enguerran: You should check every part of the subproperty:

```js
data.stacktrace || (
  data.exception &&
  data.exception.values &&
  data.exception.values[0] &&
  data.exception.values[0].stacktrace
)
```

src/drive/mobile/lib/reporter.js
```diff
@@ -6,11 +6,34 @@ export const ANALYTICS_URL =
...
+  if (stacktrace) {
+    stacktrace.frames.forEach(frame => {
+      if (frame.filename !== '[native code]') {
+        frame.filename = normalizeUrl(frame.filename, PATH_STRIP_RE)
+      }
+    })
```
enguerran: You should not hide side-effects:

```js
stacktrace.frames = stacktrace.frames.map(
  frame => frame.filename !== '[native code]' ?
  { ...frame, filename: normalizeUrl(frame.filename, PATH_STRIP_RE) } :
  frame
)
```

src/drive/mobile/lib/reporter.js
```diff
@@ -6,11 +6,34 @@ export const ANALYTICS_URL =
...
+  }
+  return data
+}
```
enguerran: More generally, you should add test coverage on this small function.

y-lohse:
Thanks for the review, all accurate comments… except for the need for tests maybe. 

They would take a lot of effort to set up, because the data object that the function receives comes from sentry, and it already seems to be doing some exception processing before that function. We'd mostly be testing raven.js which is not super useful.
Also this function is super isolated, so it would be quite hard to break. I don't see it evolving anytime soon either.
So yeah, I think adding tests here requires a lot of effort for very little gain.

All the other changes are done though 👍

enguerran:
I thought that we need to test the function itself. It takes an object as a parameter and returns a new object. We should test this.

```js
const normalizeData = data => {
  const PATH_STRIP_RE = /^.*\/[^.]+(\.app|CodePush|.*(?=\/))/

  if (data.culprit) {
    data.culprit = normalizeUrl(data.culprit, PATH_STRIP_RE)
  }
  const stacktrace =
    data.stacktrace ||
    (data.exception &&
      data.exception.values &&
      data.exception.values[0] &&
      data.exception.values[0].stacktrace)

  if (stacktrace) {
    stacktrace.frames = stacktrace.frames.map(
      frame =>
        frame.filename !== '[native code]'
          ? { ...frame, filename: normalizeUrl(frame.filename, PATH_STRIP_RE) }
          : frame
    )
  }
  return data
}
```

And a test could be
```js
test("test normalizeData", () => {
  const data = { /* some data */ };
  const expectedData = { /* expected data */ };
  expect(normalizeData(result)).toEqual(expectedData);
})
```

What do you think?

y-lohse:
So a fake object then? Sounds good to me. Still not sure it will be *that* useful, but t's low effort too so it's worth it!

PR #859, opened 2018-04-11
chore: change all versions at once

y-lohse:
We have to change versions inside:

- the config.xml files
- the drive manifest
- the photos manifest
- the package.json

Not sure if package.json should refer to Drive or Photos, but still.

With these new tasks, we can change the version in package.json, and then run `yarn version:drive` or `yarn version:photos` to change the other files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,11 @@
     "run:drive:ios": "(cd targets/drive/mobile && cordova run ios --device)",
     "run:drive:ios:emulator": "(cd targets/drive/mobile && cordova run ios --emulator)",
     "publish:drive:ios": "npm run build:drive:mobile && (cd targets/drive/mobile && fastlane ios pushtest)",
-    "sentry:drive:mobile": "sentry-cli releases -o sentry -p cozy-drive-v3 files $npm_package_version upload-sourcemaps ./targets/drive/mobile/www"
+    "sentry:drive:mobile": "sentry-cli releases -o sentry -p cozy-drive-v3 files $npm_package_version upload-sourcemaps ./targets/drive/mobile/www",
+    "version:drive:manifest": "replace '\\d+\\.\\d+\\.\\d+' $npm_package_version ./targets/drive/manifest.webapp",
+    "version:drive:config": "replace 'version=\"\\d+\\.\\d+\\.\\d+\"' 'version=\"'$npm_package_version'\"' ./targets/drive/mobile/config.xml",
+    "version:drive": "npm-run-all --parallel 'version:drive:*'",
+    "version:photos": "replace '\\d+\\.\\d+\\.\\d+' $npm_package_version ./targets/photos/manifest.webapp"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,7 +46,11 @@
     "run:drive:ios": "(cd targets/drive/mobile && cordova run ios --device)",
     "run:drive:ios:emulator": "(cd targets/drive/mobile && cordova run ios --emulator)",
     "publish:drive:ios": "npm run build:drive:mobile && (cd targets/drive/mobile && fastlane ios pushtest)",
-    "sentry:drive:mobile": "sentry-cli releases -o sentry -p cozy-drive-v3 files $npm_package_version upload-sourcemaps ./targets/drive/mobile/www"
+    "sentry:drive:mobile": "sentry-cli releases -o sentry -p cozy-drive-v3 files $npm_package_version upload-sourcemaps ./targets/drive/mobile/www",
+    "version:drive:manifest": "replace '\\d+\\.\\d+\\.\\d+' $npm_package_version ./targets/drive/manifest.webapp",
+    "version:drive:config": "replace 'version=\"\\d+\\.\\d+\\.\\d+\"' 'version=\"'$npm_package_version'\"' ./targets/drive/mobile/config.xml",
+    "version:drive": "npm run version:drive:manifest && npm run version:drive:config",
```
enguerran: You can use `npm-run-all --parallel 'version:drive:*'`

package.json
```diff
@@ -125,6 +129,7 @@
     "react-addons-test-utils": "15.6.2",
     "react-dom": "15.6.2",
     "react-test-renderer": "15.6.2",
+    "replace": "0.3.0",
```
enguerran: `sed` must be a better tool to _replace_ data with a regex and it is available on every platform!

y-lohse: Not on windows :) Aaaand… I don't really now how to use it? `sed "s/[0-9]+\./yolo/g" ./targets/drive/manifest.webapp` doesn't do anything, `\d` either… If your familiar with it and can just spit out the right command, I don't mind replacing it!

enguerran: I don't know either, but some might know
https://unix.stackexchange.com/questions/78625/using-sed-to-find-and-replace-complex-string-preferrably-with-regex#78626

y-lohse: Yeah no sorry, I just spent some time trying to make it work without success. The npm package does the job just fine and works on really all platforms, I don't think it's worth spending any more time on this. Feel free to improve it further if you feel like it though :)

ptbrowne: I wsa a bit curious :) The regex did not work because the plus sign is part of the extended regular expressions and you have to pass a flag to sed to enable their support. On BSD flavors, it is with '-E' and on Linux it is with '-r'.

```
$ sed -E "s/[0-9]+./yolo/g" ./targets/drive/manifest.webapp
```

works on my machine :)

> The Extended Regular Expressions or ERE flavor standardizes a flavor similar to the one used by the UNIX egrep command. "Extended" is relative to the original UNIX grep, which only had bracket expressions, dot, caret, dollar and star. An ERE support these just like a BRE. Most modern regex flavors are extensions of the ERE flavor. By today's standard, the POSIX ERE flavor is rather bare bones. The POSIX standard was defined in 1986, and regular expressions have come a long way since then.

https://www.regular-expressions.info/posix.html

https://stackoverflow.com/questions/22099623/sed-plus-sign-doesnt-work

PR #865, opened 2018-04-13
feat: file list placeholder

y-lohse:
Here's the preview:

preview

When the content of a folder is loading, we display a set of placeholders instead of a giant spinner, or nothing.

The code for the animation is complicated but I really don't see a better way to do it :/

### DIFF
--- a/src/drive/components/FileListPlaceholder.jsx
+++ b/src/drive/components/FileListPlaceholder.jsx
@@ -1,0 +1,93 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import classNames from 'classnames'
+import styles from '../styles/table'
+
+// using a seeded PRNG to prevent re-renders from changing the results
+const seededRandom = seed => {
+  const x = Math.sin(seed) * 10000
+  return x - Math.floor(x)
+}
+const seededRandomBetween = (min, max, seed) =>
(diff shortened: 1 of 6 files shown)


y-lohse:
[(2) Changer l'état de chargement du contenu d'un dossier](https://trello.com/c/42LU6Qao/905-2-changer-l%C3%A9tat-de-chargement-du-contenu-dun-dossier)

src/drive/components/FileListPlaceholder.jsx
```diff
@@ -0,0 +1,93 @@
...
+  index: PropTypes.number
+}
+
+FilePlaceholder.defaultProps = {
+  index: 1
+}
+
+const FileListPlaceholder = ({ rows }) => (
+  <div>
+    {Array.apply(null, Array(rows)).map((value, index) => (
```
enguerran: I had hard time to understand this code and I still can't understand why we have to use `Array.apply(null, new Array(5))`.

But I found a clearer write:

```js
[...new Array(rows)].map(createFilePlaceholder)
```

y-lohse: Oh that *is* better, nice. I tried some variations like that but couldn't find one that worked. Using that instead 👍

PR #867, opened 2018-04-13
feat: share files with drive on Android :mobile:

enguerran:
We want Android users to be able to _share_ any file from any app with Cozy Drive like shown below:

___insert screenshot here___

This is a three steps action:

1. register the app to receive intent by setting `<intent-filter>` in `AndroidManifest.xml` (this file is generated from `config.xml` with cordova application)
1. catch the incoming intent with `window.plugins.intentShim.onIntent` from darryncampbell/darryncampbell-cordova-plugin-intent
1. add the file to cozy


src/drive/ducks/upload/index.js
```diff
@@ -65,6 +65,7 @@ const processNextFile = (
   dirID
 ) => async (dispatch, getState) => {
   const item = getUploadQueue(getState()).find(i => i.status === PENDING)
+  console.log('upload', item)
```
enguerran: Rahhh, c'est encore un coup du git hook qui ne prend pas en compte mon `git add --patch` avec des morceaux de fichiers en _staging_ zone !

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
-      lang={getLang()}
-      dictRequire={lang => require(`drive/locales/${lang}`)}
-    >
-      <CozyProvider store={store} client={client}>
-        <DriveMobileRouter history={hashHistory} />
-      </CozyProvider>
-    </I18n>,
-    root
-  )
+function registerShareWithCozyDriveIntent(store) {
```
enguerran: Here we register a code to be able to get incoming intent.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
-        initBar(client)
-      } else {
-        console.warn('Your device is no more connected to your server')
-        store.dispatch(revokeClient())
-      }
+window.handleOpenURL = require('drive/mobile/lib/handleDeepLink').default(
+  hashHistory
+)
+
+function uploadIntent(intent, store) {
```
enguerran: Here is the code that might be refactored, it is ugly and to verbose.
It includes so many responsibilities that we don't understand what are the steps involved!
A new commit (only refactoring) is coming.

targets/drive/mobile/main.jsx
```diff
@@ -126,22 +133,91 @@ const isBackgroundServiceParameter = () => {
...
+  onDeviceReady: async function() {
+    if (!isBackgroundServiceParameter()) {
+      this.startApplication()
+    } else {
       startBackgroundService()
     }
+
+    // this.store.dispatch(backupImages())
+    if (navigator && navigator.splashscreen) navigator.splashscreen.hide()
+    registerShareWithCozyDriveIntent(this.store)
```
enguerran: Here `this.store` is not defined because of the asynchronous start of the application: we load state from local storage (indexedDB which is async).

targets/drive/mobile/main.jsx
```diff
@@ -126,22 +133,91 @@ const isBackgroundServiceParameter = () => {
...
+    document.addEventListener(
+      'deviceready',
+      this.onDeviceReady.bind(this),
+      false
+    )
+    document.addEventListener('resume', this.onResume.bind(this), false)
+    document.addEventListener('pause', this.onPause.bind(this), false)
+  },
+
+  startApplication: async function() {
```
enguerran: This function that starts the application is too big and must be refactored. There is too much thing to do.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+
+function uploadIntent(intent, store) {
+  if (intent['android.intent.extra.STREAM']) {
+    const contentFiles = Array.isArray(intent['android.intent.extra.STREAM'])
+      ? intent['android.intent.extra.STREAM']
+      : [intent['android.intent.extra.STREAM']]
+    contentFiles.forEach(content => {
+      window.FilePath.resolveNativePath(
+        content,
+        async filePath => {
```
enguerran: for each `content` of the intent (`content://`) we resolve the native file path.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+          console.log('Here is the filepath', filePath)
+          const dirEntry = await getEntry(filePath)
```
enguerran: with the native file path, we resolve a DirectoryEntry as defined by the cordova plugin.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+          console.log('Here is the directoryentry', dirEntry)
+          dirEntry.file(file => {
```
enguerran: with the `DirectoryEntry` we resolve the file.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+            file.lastModifiedDate = new Date(file.lastModifiedDate)
```
enguerran: the file is not well formated: `lastModifiedDate` is not a date, so we make it a `Date`

src/drive/ducks/upload/index.js
```diff
@@ -76,11 +77,11 @@ const processNextFile = (
...
-    let status
-
-    if (error.status === 400) status = CONFLICT
-    else if (error.status === 413) status = QUOTA
-    else status = FAILED
+    const statusError = {
+      400: CONFLICT,
+      413: QUOTA
+    }
+    const status = statusError[error.status] || FAILED
```
y-lohse: It's not blocking, but I'm not in favor of this change. It feels like we're going out of our way to not use `let`, making the mental gymnastic required to understand the code harder.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+            console.log('this is binary file', file)
+            store.dispatch(
+              addToUploadQueue(
```
enguerran: we upload everything by using the application upload action.

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
+                    )
+                  } else {
+                    action.alert = alertShow(
+                      'upload.alert.success',
+                      { smart_count: loaded.length },
+                      'success'
+                    )
+                  }
+
+                  return action
```
enguerran: Too many lines here, a big cargo cult from the `Toolbar.jsx` action.

targets/drive/mobile/main.jsx
y-lohse: Why are we doing this?

targets/drive/mobile/main.jsx
```diff
@@ -38,78 +43,80 @@ if (__DEVELOPMENT__) {
...
-      } else {
-        console.warn('Your device is no more connected to your server')
-        store.dispatch(revokeClient())
-      }
+window.handleOpenURL = require('drive/mobile/lib/handleDeepLink').default(
+  hashHistory
+)
+
+function uploadIntent(intent, store) {
+  if (intent['android.intent.extra.STREAM']) {
```
y-lohse: What happens if the condition fails here? Nothing as far as I can tell, but will that happen?

targets/drive/mobile/main.jsx
y-lohse: Sorry I just saw your comments, they didn't show up in the preview window.

y-lohse: Maybe we should add a comment in the code to explain that?

y-lohse:
Regarding the `package-lock.json`, I generally use `--save` and not `--fetch`. i also noticed the plugins are defined with a version range, whereas all the other plugins are pinned to a specific version. I think we should stay consistent and pin the versions on these as well, which also removes the need for a lock file.

targets/drive/mobile/main.jsx
enguerran: I don't know, this code must be refactored and live a bit to catch every intent application can send as Android's intents are free to customize.

src/drive/ducks/upload/index.js
enguerran: I knew you'd say that, but I tried anyway. Here's my opinion:

In the original code, we had 6 instructions that can be the origin of 6 bugs.
In the refactored code, we only have 2 instructions: one object assignment (0 bug) and one value assignment with a well-known default value syntax. No `if`, no `else if` and the variable `status` is assigned only once.

That seems to me an easier code to maintain, to refactor, to understand and to debug.

y-lohse: I could argue that there aren't 6 instructions in the original code, or that there are more than 2 in the new one, or explain the mental process when reading the old and new code… but I think we'll end up sitting here a long time, so let's not do that :)

I just wanted to point out that I don't think it's a good practice to refactor code that hasn't much to do with the current PR so that it suits our personal tastes. If we all start doing this, it's just going to be an endless edit war that won't benefit anyone ;)

enguerran:
[(5) 📱 Ajouter un fichier depuis une autre application du téléphone Android](https://trello.com/c/zXAm3DL4/398-5-%F0%9F%93%B1-ajouter-un-fichier-depuis-une-autre-application-du-t%C3%A9l%C3%A9phone-android)

enguerran:
I removed the refactoring part you can now find on #870 

I removed `package-lock.json` discussed here https://github.com/cozy/cozy-drive/pull/867#issuecomment-381138421 (we shall add dependencies in our main `package.json` file)

And I added a commit with a refactoring to help understand start-up stages.

PR #874, opened 2018-04-18
Don't close the viewer when trying to view a file lost in a big folder

goldoraf:
Because it breaks the search! This PR implements a contingency measure by loading the file attributes directly. It allows us to show the file in the viewer, but without the nav arrows.

### DIFF
--- a/src/drive/components/FilesViewer.jsx
+++ b/src/drive/components/FilesViewer.jsx
@@ -1,6 +1,7 @@
+/* global cozy */
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
-import Viewer from 'viewer'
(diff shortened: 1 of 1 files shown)


goldoraf:
[(1) Ouvrir la visionneuse lors de la sélection d'un fichier dans les résultats de recherche](https://trello.com/c/bmXRf44s/892-1-ouvrir-la-visionneuse-lors-de-la-s%C3%A9lection-dun-fichier-dans-les-r%C3%A9sultats-de-recherche)

goldoraf:
Yeah, I agree the `FilesViewer` is far from ideal, but I'm not sure about the direction we should take. Should it be "standalone", i.e. should it be able to handle all necessary fetchings? I'm not sure...

PR #875, opened 2018-04-18
Don't show the viewer cross when displayed in an intent modal

goldoraf:
I choosed to not show the cross when the `onClose` prop isn't provided. That's a simpler solution than what we thought about.


goldoraf:
[(1) Supprimer la double croix dans la modale inter app](https://trello.com/c/Cb0SSEAe/911-1-supprimer-la-double-croix-dans-la-modale-inter-app)

PR #876, opened 2018-04-19
Don't return null from the viewer if the folder contains a lot of folders and no files are fetched at first

goldoraf:
Another quick fix after testing on recette: if we click on a search result that comes from a folder with 30+ folders, no files are loaded at first, and so the viewer doesn't show. 

Merged without review to save time considering its triviality.


enguerran:
This could be a big time saving if the code were covered by tests to make sure there are no side effects to never `return null` when `files.length ==== 0`. And also travis waiting should be mandatory, in my opinion.

PR #877, opened 2018-04-19
Sorting fixes

goldoraf:
Sorting by name didn't work correctly on large folders because folders and files got mixed when fetching more. Because Couch and Pouch can't sort on 2 fields but different orders (it'd allow us to sort by type AND name and be done), I had to resort to fetch folders and files separately.

### DIFF
--- a/src/drive/components/FileList.jsx
+++ b/src/drive/components/FileList.jsx
@@ -49,6 +49,7 @@ class FileList extends PureComponent {
     // We're in /recent
     if (!this.props.displayedFolder) return false
     if (isCordova()) {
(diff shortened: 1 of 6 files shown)


src/drive/components/FileList.jsx
```diff
@@ -49,6 +49,7 @@ class FileList extends PureComponent {
     // We're in /recent
     if (!this.props.displayedFolder) return false
     if (isCordova()) {
+      console.log(this.props.files.length < LIMIT, !this.state.hasNoMoreRows)
```
enguerran: an oversight?

goldoraf: woops, good catch ;)

PR #878, opened 2018-04-19
Added header to the file opener intent modal

goldoraf:


goldoraf:
[(1) Supprimer la double croix dans la modale inter app](https://trello.com/c/Cb0SSEAe/911-1-supprimer-la-double-croix-dans-la-modale-inter-app)

PR #879, opened 2018-04-20
Allow files dn'd in Drive

goldoraf:
I used the excellent [react-dropzone](https://github.com/react-dropzone/react-dropzone) for this. Not much to add, I think the code is straightforward.


src/drive/styles/table.styl
```diff
@@ -90,6 +90,11 @@
 .fil-content-header-sortabledesc
 .fil-content-header-sortableasc
     cursor pointer
+.fil-content-dropzone
+    border 1px solid grey
+
+.fil-content-dropzone-active
+    background-color #F1F7FF
```
y-lohse: Can we use the named color please? [It's called `zircon`](https://docs.cozy.io/cozy-ui/styleguide/section-settings.html#kssref-settings-colors-blue).

src/drive/styles/dropzone.styl
```diff
@@ -0,0 +1,51 @@
+@require 'settings/z-index.styl'
+
+.fil-dropzone-teaser
+    position  absolute
+    z-index   $overlay-index
+    left      50%
+    transform translateX(-50%)
```
y-lohse: Maybe it's just me, but I find this way of centering content very weird. I believe this would be more standard:

```
left 0
right 0
```

and no transform. Works at least on Chrome.
@GoOz what do you think?

GoOz: No it's good like this. `left 50%` means the item's left border is placed at 50% of its parent's width and the `translateX(-50%)` moves back the item to the left at half its own size which results on a perfect centering.
This is a common technique, I use it a lot.
`right 0 left 0` wouldn't center it at all actually, it would make the item have a full width.

y-lohse: Well `right 0 left 0` def. works, though I thought it needed a fixed width, but apparently not.

But ok, cool if it's a legit way to do it, please ignore my comment!

src/drive/actions/index.jsx
```diff
@@ -235,6 +241,31 @@ export const uploadedFile = file => {
   }
 }
 
+export const uploadQueueProcessed = (loaded, quotas, conflicts, errors, t) => {
```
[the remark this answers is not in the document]
goldoraf: `intents.js` wasn't there when I worked on this, that's why I missed it, but done ;)

PR #883, opened 2018-04-25
release 1.4.1 with intent disabled

enguerran:
In release `1.4.0` we forgot to disable the "share to Cozy Drive" feature on Android.
Here is a fix for the previous release, `master` will be tagged `1.4.1`.


enguerran:
Whenever we'll need to reenable Android intents, just use `git revert 912cada`.

PR #889, opened 2018-04-26
Optimize the images weight

nono:


enguerran:
We should `filter-branch` ([nettoyer un depôt git](https://makina-corpus.com/blog/metier/2016/nettoyer-un-depot-git)) to clean up spaces in the git repository.

PR #892, opened 2018-04-30
feat: files fetch limit to 100

enguerran:
We could take more time to understand what are all those `limit`, but it is a little bit overengineered if our purpose is to inscrease this limit on fetching files.


enguerran:
In yellow, components that use a magic number to fetch files.

enguerran:
[(3) Augmenter le nombre d'éléments par page dans Drive](https://trello.com/c/Dd4GeAMn/937-3-augmenter-le-nombre-d%C3%A9l%C3%A9ments-par-page-dans-drive)

PR #893, opened 2018-04-30
Cozy-to-cozy sharing take #2

goldoraf:


goldoraf:
[Intégrer le partage Cozy à Cozy dans Drive](https://trello.com/c/NaNSLUw6/926-int%C3%A9grer-le-partage-cozy-%C3%A0-cozy-dans-drive)

src/photos/components/Photo.jsx
```diff
@@ -26,7 +25,7 @@ const getStyleFromBox = box => {
   return style
 }
 
-const Photo = props => {
+const Photo = (props, { client }) => {
```
enguerran: I can't understand how it is related to cozy-to-cozy sharing.

goldoraf: FYI, I cherry-picked the commit and merged it in another PR

src/sharing/index.js
```diff
@@ -15,11 +47,260 @@ export const getPrimaryCozy = contact =>
...
+    this.dispatch(revokeSelf(sharing))
+  }
+
+  shareByLink = async document => {
+    trackSharingByLink(document)
+    const resp = await this.context.client
+      .collection('io.cozy.permissions')
+      .createSharingLink(document)
+    this.dispatch(addSharingLink(resp.data))
+    return resp
```
y-lohse: We probably don't need the intermediate `resp` here.

src/sharing/index.js
```diff
@@ -15,11 +47,260 @@ export const getPrimaryCozy = contact =>
...
+
+  revokeSharingLink = document => {
+    // Because some duplicate links have been created in the past, we must ensure
+    // we revoke all of them
+    const perms = getDocumentPermissions(this.state, document.id)
+    return Promise.all(
+      perms.map(p =>
+        this.context.client.collection('io.cozy.permissions').destroy(p)
+      )
+    ).then(() => this.dispatch(revokeSharingLink(perms)))
```
y-lohse: Can we switch this to `async` / `await`? It'd be easier to read, and the rest of the file seems to be using `async`.

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -136,11 +122,15 @@ class Toolbar extends Component {
...
         {notRootfolder &&
-          (shared.withMe ? (
+          (isSharedWithMe ? (
             <Item>
               <a
                 className={classNames(styles['fil-action-delete'])}
-                onClick={() => leaveFolder(displayedFolder)}
+                onClick={() =>
+                  onLeave(displayedFolder).then(() =>
+                    trashFolder(displayedFolder)
```
y-lohse: Small thing, but I think inline functions shouldn't be more than a one-liner, so maybe we could have a standalone function here? This would also allow to use `async/await`.

src/sharing/components/Recipient.jsx
```diff
@@ -36,65 +41,79 @@ export const UserAvatar = ({ name, url }) => (
...
-          onUnshare={this.onUnshare}
-        />
+      <div className={styles['pho-recipient-status']}>
+        {revoking && <Spinner />}
+        {!revoking &&
+          status === 'owner' && <span>{t(`Share.status.${status}`)}</span>}
+        {!revoking &&
+          status !== 'owner' &&
+          !isMe &&
+          !isOwner && <span>{t(`Share.status.${status}`)}</span>}
```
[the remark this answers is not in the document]
goldoraf: No, if `isMe` is true we want to display the menu. This code is confusing, but I struggle to make it clearer :/

src/drive/ducks/files/Toolbar.jsx
enguerran: I don't know if inline (one extra redirection) will make the code more readable in this case.
Still, you can open a discussion in https://github.com/cozy/cozy-guidelines/issues

PR #895, opened 2018-05-03
chore: change user-agent of fetch headers

enguerran:
this is used on cozy-stack to enhance statistics
https://trello.com/c/mg9hDsFD/935-fiabiliser-les-stats-des-apps-mobile

### DIFF
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -91,3 +91,27 @@ const getTokenWithNoException = async () => {
 }
 
 export const getClientUrl = () => cozy.client._url
+
+const getUpdatedUserAgent = navigator => {
+  const deviceName = getDeviceName() || 'unknown device name'
+  const identifier = getIdentifier(navigator) || 'unknown identifier'
+  const version = getVersion(navigator) || 'unknown version'
+  const userAgent = getUserAgent(navigator) || 'unknown user-agent'
+  return `${deviceName} ${identifier}-${version} ${userAgent}`
+}
+const getIdentifier = navigator =>
+  navigator && navigator.appInfo && navigator.appInfo.identifier
+const getVersion = navigator =>
+  navigator && navigator.appInfo && navigator.appInfo.version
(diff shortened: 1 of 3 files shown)


enguerran:
[(?) Fiabiliser les stats des apps mobile](https://trello.com/c/mg9hDsFD/935-fiabiliser-les-stats-des-apps-mobile)

y-lohse:
The code LGTM, I'll try to run it to make sure it works for me to today.

enguerran:
LMK if it works for you.

enguerran:
> However, have you been able to confirm it actually changes the user agent header with a mobile device?

No, I haven't

enguerran:
It appears the first solution failed to be stable. I adopted another solution recommended by ionic developers: https://ionicframework.com/docs/native/user-agent/
It works better and should allow us to do some statistics

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -91,3 +91,27 @@ const getTokenWithNoException = async () => {
...
+const getUserAgent = navigator => navigator && navigator.userAgent
+
+export const updateUserAgent = () => {
+  try {
+    window.UserAgent.set(getUpdatedUserAgent(window.navigator))
+  } catch (err) {
+    console.error(
+      `Unable to update User-Agent with cordova plugin: ${err.message}`
+    )
+    throw err
```
y-lohse: Do we really want to `throw` here? There's no other catch block down the line, and this code runs at the start of the application, so everything will fail if the user agent can't be set. That seems a bit extreme no?

enguerran: We must fail fast to be as reliable as possible.

This is a library function called by someone. This someone should handle the exception if the user-agent were not updated. Where you are right is that we must handle this exception in the [`startApplication` function that makes the call](https://github.com/cozy/cozy-drive/pull/895/files/445f4bd2e3942280eb01527fc554c7283b6d43d8#diff-f112ba1b7dce001913124514eee437fbR135).

PR #896, opened 2018-05-07
feat: Open With Cozy Drive

enguerran:
Revert "feat: disable Android intents"
This reverts commit 0ae1864e973a9acece5fbf9da9e68b50d9a9dfd8.

And it's amended with feature enhancement.

It fully works with:
- Photos
- Gallery
- other apps with local files

It works without the real filename with:
- GMail
- and other apps with distant filepath

We use https://github.com/hiddentao/cordova-plugin-filepath to retrieve the local file data. But if this plugin cannot resolve the filepath, we fall back to https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-file/ which do not give a correct filename.


enguerran:
[(5)📱Ajouter un fichier depuis une autre application du téléphone Android](https://trello.com/c/zXAm3DL4/398-5%F0%9F%93%B1ajouter-un-fichier-depuis-une-autre-application-du-t%C3%A9l%C3%A9phone-android)

src/drive/mobile/lib/intents.js
```diff
@@ -0,0 +1,112 @@
...
+  if (extras['android.intent.extra.STREAM']) {
+    const contentFiles = Array.isArray(extras['android.intent.extra.STREAM'])
+      ? extras['android.intent.extra.STREAM']
+      : [extras['android.intent.extra.STREAM']]
+    try {
+      const files = await getFiles(contentFiles)
+      uploadFiles(files, store)
+    } catch (err) {
+      store.dispatch({
+        alert: alertShow('intents.alert.errors', null, 'error')
```
y-lohse: Just a head's up: the action needs a `type`, otherwise redux will throw an error. 

Also (and I didn't look into it) the error message literally says `'intents.alert.errors'` so I think a translation is missing ?

enguerran: I caught these two mistakes. Thanks a lot.

PR #897, opened 2018-05-09
Fixed thumbnails display on public page

goldoraf:
Thumbnails in a public page need some special permission given by the token and a cookie. The previous client could not send a authenticated request, so I updated the client to use an authenticated client instead.


enguerran:
Do not forget to remove this commit https://github.com/cozy/cozy-drive/pull/893/commits/a06d7a633eda51f1068eb18e8626457cd2f7ec0b

PR #898, opened 2018-05-09
New registry token

aenario:


y-lohse:
Thanks, but this isn't the travis file used for publication. Can you replace the secure in [this file](https://github.com/cozy/cozy-drive/blame/master/targets/.travis.yml#L12) please? That should work.

aenario:
Done in both.

PR #899, opened 2018-05-11
chore: fix android build

enguerran:
I faced multiples problems with `yarn run:drive:android`:

- `cordova-custom-config` needs `tostr` to be able to modify `config.xml` on build time
- and something about cordova and gradle I cargo-culted from https://stackoverflow.com/a/49589661 that works better than the previous solution with the script and the hook

What do you think?


y-lohse:
Big 👍 for the gradle fix.

I don't mind the `tostr` devdep too much, but maybe we can also PR `cordova-custom-config` to add it to their dependencies? It seems a bit weird to have a dependency in our package.json for a cordova plugin.

PR #902, opened 2018-05-15
feat: ios file sharing

y-lohse:
These changes allows the user to send a file to Cozy Drive on iOS. it relies on [this plugin](https://github.com/EternallLight/cordova-plugin-openwith-ios).

Further comments coming up 👇

(Looping in the gangsters @kosssi @ptbrowne @drazik in case you're interested, although I'm not sure it's too useful for Cozy Banks.)

### DIFF
--- a/src/drive/mobile/lib/intents.js
+++ b/src/drive/mobile/lib/intents.js
@@ -58,6 +58,18 @@ const getFiles = contentFiles =>
     })
   )
 
+const base64toBlob = (dataURI, type) => {
+  const byteString = atob(dataURI)
+  const ab = new ArrayBuffer(byteString.length)
+  const ia = new Uint8Array(ab)
+
+  for (let i = 0; i < byteString.length; i++) {
+    ia[i] = byteString.charCodeAt(i)
+  }
+
+  return new Blob([ab], { type })
+}
+
 const uploadFiles = (files, store) => {
   store.dispatch(
     addToUploadQueue(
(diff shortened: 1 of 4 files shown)


src/drive/mobile/lib/intents.js
```diff
@@ -58,6 +58,18 @@ const getFiles = contentFiles =>
...
+  const byteString = atob(dataURI)
+  const ab = new ArrayBuffer(byteString.length)
+  const ia = new Uint8Array(ab)
+
+  for (let i = 0; i < byteString.length; i++) {
+    ia[i] = byteString.charCodeAt(i)
+  }
+
+  return new Blob([ab], { type })
+}
```
y-lohse: The cordova plugin sends image files as base64 encoded strings, so we need to turn them into blobs.

src/drive/mobile/lib/intents.js
```diff
@@ -111,3 +123,48 @@ export const intentHandler = store => async ({
...
+export const intentHandlerIOS = store => async intent => {
+  const files = await Promise.all(
+    intent.items.map(async item => {
+      if (item.uti === 'public.image') {
+        const blob = base64toBlob(item.data, item.type)
+        blob.name = item.name
+        return blob
+      } else {
+        const dirEntry = await getEntry(item.data)
+        if (!dirEntry.type) dirEntry.type = getMimeTypeFromUTI(item.utis)
```
y-lohse: For files that are not images, we receive a link to the file, and *sometimes* a mime type, but mostly not. But we do get [UTIs](https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html#//apple_ref/doc/uid/TP40009259-SW1), so i've mapped the main ones to mime types.

targets/drive/mobile/config.xml
```diff
@@ -48,6 +45,7 @@
     </platform>
     <engine name="ios" spec="4.5.4" />
     <platform name="ios">
+        <hook src="scripts/configure-share-extension.js" type="after_plugin_install" />
```
y-lohse: There are a few things that need to be done in xcode after the plugin is installed, so i've scripted them.
However, the `after_prepare` hook runs *before* all the plugins are installed, and there's no `after_plugins` hook. So this script runs after *every* plugin install. That sucks, but keep reading.

targets/drive/mobile/scripts/configure-share-extension.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs')
+const path = require('path')
+const rootdir = process.env.PWD
+const installWithPluginId = 'com.web-mystery.cordova.openwith-ios'
+
+module.exports = function (context) {
+  if (context.opts.cordova.plugins.indexOf(installWithPluginId) < 0) return
```
y-lohse: To speed things up a bit, the script early-outs if it detects the plugin is not installed. This is the best I could come up with.

Note that it's ok if it runs multiple times anyway.

y-lohse:

y-lohse:
Ah yes of course, thanks for catching that. It's fixed now.

PR #904, opened 2018-05-21
fix: proper ios shared fle mime detection

y-lohse:
Files shared through iOS don't have a proper mime type, so we have to forcefully set it.

### DIFF
--- a/src/drive/mobile/lib/intents.js
+++ b/src/drive/mobile/lib/intents.js
@@ -4,18 +4,28 @@ import { addToUploadQueue } from '../../ducks/upload/index'
 import { ROOT_DIR_ID } from '../../constants/config'
 import { uploadedFile, uploadQueueProcessed } from '../../actions/index'
 
-const getFile = dirEntry =>
+const getFile = (dirEntry, type = '') =>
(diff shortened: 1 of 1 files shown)


y-lohse:

src/drive/mobile/lib/intents.js
```diff
@@ -130,8 +140,7 @@ export const intentHandlerIOS = store => async intent => {
         return blob
       } else {
         const dirEntry = await getEntry(item.data)
-        if (!dirEntry.type) dirEntry.type = getMimeTypeFromUTI(item.utis)
-        const file = await getFile(dirEntry)
+        const file = await getFile(dirEntry, getMimeTypeFromUTI(item.utis))
```
goldoraf: Is `getMimeTypeFromUTI` a global function? I couldn't find any reference on it.

y-lohse: No it's local to the file, I introduced it [in the previous commit](https://github.com/y-lohse/cozy-drive/blob/d945ff01fbc902e259cfa648655050fa4434688d/src/drive/mobile/lib/intents.js#L116-L132) ;)

PR #908, opened 2018-05-23
Added support for folder upload by DnD

goldoraf:
Because `react-dropzone` provides `onDrop` with a pre-digested array of files, I had to use the original event to get an array of items on which I can call `webkitGetAsEntry`: this non-standard API supported by Cr, FF, Edge and now Safari 11.1 allows us to see if a file is in fact a folder, and we can then use the File API to extract its contents.


goldoraf:
[Transfert de fichiers par drag and drop](https://trello.com/c/FQm5RH0Y/956-transfert-de-fichiers-par-drag-and-drop)

src/drive/ducks/upload/index.js
```diff
@@ -74,8 +74,9 @@ const processNextFile = (
     const uploadedFile = await cozy.client.files.create(file, { dirID })
     dispatch({ type: RECEIVE_UPLOAD_SUCCESS, file })
     // TODO: is the extractFileAttributes call really necessary?
-    dispatch(fileUploadedCallback(extractFileAttributes(uploadedFile)))
+    fileUploadedCallback(extractFileAttributes(uploadedFile))
   } catch (error) {
+    console.log(error)
```
enguerran: Shouldn't you be displaying a warning or an error in the console?

src/drive/ducks/upload/index.js
```diff
@@ -120,3 +121,62 @@ export const getUploadQueue = state => state[SLUG].queue
 export const getProcessed = state =>
   getUploadQueue(state).filter(f => f.status !== PENDING)
 export const getSuccessful = state => getLoaded(getUploadQueue(state))
+
+// DnD helpers for folder upload
+export const canHandleFolders = evt => {
```
enguerran: It might be easier to test this function if it took the `dataTransfer' object directly instead of an event, wouldn't it?

src/drive/ducks/upload/index.js
```diff
@@ -120,3 +121,62 @@ export const getUploadQueue = state => state[SLUG].queue
...
+        }
+      }
+      resolve()
+    }
+    dirReader.readEntries(entriesReader)
+  })
+  return results
+}
+
+export const getFilesFromItems = async evt => {
```
enguerran: It might be easier to test this function if it took the `dataTransfer' object directly instead of an event, wouldn't it?

src/drive/ducks/upload/index.js
```diff
@@ -120,3 +121,62 @@ export const getUploadQueue = state => state[SLUG].queue
...
+    const entriesReader = async entries => {
+      for (let i = 0; i < entries.length; i += 1) {
+        const entry = entries[i]
+        if (entry.isFile) {
+          results.push(await getFileFromEntry(entry))
+        } else if (entry.isDirectory) {
+          results = results.concat(await getFilesFromDirectory(entry))
+        }
+      }
+      resolve()
```
enguerran: You should put `results` in the function scope instead of using a closure and `resolve(results)` to return the promised value, shouldn't you?

src/drive/ducks/upload/index.js
```diff
@@ -120,3 +121,62 @@ export const getUploadQueue = state => state[SLUG].queue
...
+  const dt = evt.dataTransfer
+  return dt.items && dt.items.length && dt.items[0].webkitGetAsEntry != null
+}
+
+const getFileFromEntry = async entry => {
+  let result
+  await new Promise(resolve => {
+    entry.file(file => {
+      result = file
+      resolve()
```
enguerran: You should put `results` in the function scope instead of using a closure and `resolve(results)` to return the promised value, shouldn't you?

y-lohse: Or it can be rewritten like this :

```
return await new Promise(resolve => entry.file(resolve))
```

A bit extreme but you get the idea.

y-lohse: Actually `return new Promise(resolve => entry.file(resolve))` would be the same, and more concise.

enguerran: Good point!

src/drive/ducks/upload/index.js
goldoraf: That's what I wanted to do initially but hten I realized it would make `FolderView` handling some of the sordid DOM implementation details I tried to isolate, that's why I decided to pass the event object instead. Either way, I don't think that kind of DOM-oriented code is really testable sadly...

src/drive/ducks/upload/index.js
```diff
@@ -120,3 +121,62 @@ export const getUploadQueue = state => state[SLUG].queue
...
+  })
+  return results
+}
+
+export const getFilesFromItems = async evt => {
+  let results = []
+  const items = evt.dataTransfer.items
+  for (let i = 0; i < items.length; i += 1) {
+    const item = items[i]
+    if (item.webkitGetAsEntry != null && item.webkitGetAsEntry()) {
```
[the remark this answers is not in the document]
goldoraf: Yep, if the item isn't a file

goldoraf:
> Looks mostly good to me, but I also think we should avoid the let result; await Promise; pattern. Resolving the promise with the desired result is the proper way to do it imo.

Of course, I should have clean the code before submitting the PR ; the API being callback-oriented, it took a bit of work to make it promise/async/await oriented (and thus a bit more readable) and I failed to make the ultimate cleaning effort, sorry for that.

goldoraf:
Finally, I took the time to isolate everyting related to Dn'd support into a single file/component. I think it's definitely better like this ;)

PR #909, opened 2018-05-23
Removed unwanted Alerter call in NoViewer

goldoraf:
This call has been added by error during the `Alerter` transition, and being in the render method, it notifies the user repeatedly of an error when opening a PDF.


y-lohse:
That might actually have been me.

PR #928, opened 2018-05-29
Disabled sorting on mobile because of perf issues

goldoraf:


goldoraf:
@y-lohse, yes sorting works well when using the webapp on mobile, so removing it completely isn't necessary

y-lohse:
Ok, the name `MobileFileListHeader ` had me confused :)

PR #930, opened 2018-05-29
fix: reload after upload

y-lohse:
We need to call [this function](https://github.com/cozy/cozy-drive/blob/master/src/drive/actions/index.jsx#L231) to update the redux state after uploading a file, so that the file actually appears.

This function is used 3 times, but with 3 different forms. Only one of them actually worked. This PR fixes the two others.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ desktop.ini
 # Editors / IDEs
 .floo
 .flooignore
+.brackets.json
 
 # Mobile
 targets/*/mobile/hooks
--- a/src/drive/ducks/upload/index.js
+++ b/src/drive/ducks/upload/index.js
@@ -57,8 +57,6 @@ const queue = (state = [], action) => {
 }
 export default combineReducers({ queue })
 
-const extractFileAttributes = f => Object.assign({}, f, f.attributes)
-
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -30,6 +30,7 @@ desktop.ini
 # Editors / IDEs
 .floo
 .flooignore
+.brackets.json
```
y-lohse: Unrelated, but it's helpful for me. Just close your eyes and pretend it's not part of this commit.

src/drive/ducks/upload/index.js
```diff
@@ -73,8 +71,7 @@ const processNextFile = (
     dispatch({ type: UPLOAD_FILE, file })
     const uploadedFile = await cozy.client.files.create(file, { dirID })
     dispatch({ type: RECEIVE_UPLOAD_SUCCESS, file })
-    // TODO: is the extractFileAttributes call really necessary?
-    fileUploadedCallback(extractFileAttributes(uploadedFile))
+    fileUploadedCallback(uploadedFile)
```
y-lohse: Turns out `extractFileAttributes ` is not necessary because the callback actually does it.

PR #931, opened 2018-06-01
Fixed missing icons in the search suggestions

goldoraf:
The icons were missing because we sent wrong URLs. With that fix, we now send [data URIs](https://github.com/cozy/cozy-drive/compare/master...goldoraf:fix-search-icons?expand=1#diff-ab8f134c72c2b24476f993fcb9f5c5e2R129).

I also reverted back to using cozy-client-js for the file list fetching and the file opener intent because we want to get rid of the old cozy-client, but the PouchDB integration in the new cozy-client is not dry enough.

### DIFF
--- a/src/drive/ducks/services/components/SuggestionProvider.jsx
+++ b/src/drive/ducks/services/components/SuggestionProvider.jsx
@@ -1,3 +1,4 @@
+/* global cozy */
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
 import { getFileTypeFromMime } from 'drive/lib/getFileTypeFromMime'
@@ -13,10 +14,11 @@ class SuggestionProvider extends React.Component {
     window.addEventListener('message', event => {
       if (event.origin !== intent.attributes.client) return null
       const { query, id } = event.data
-      this.provideSuggestions(query, id, intent)
+      // sometimes we get messages with undefined query & id, no idea why
+      if (query && id) {
+        this.provideSuggestions(query, id, intent)
+      }
     })
-
-    this.context.client.startReplicationFrom(() => {}) // the sync functions take a redux-style `dispatch` function as a callback, but we don't handle the replication status at the moment
   }
 
   async provideSuggestions(query, id, intent) {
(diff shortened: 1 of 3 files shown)


targets/drive/web/services.jsx
```diff
@@ -40,9 +40,16 @@ document.addEventListener('DOMContentLoaded', () => {
   const { intent } = getQueryParameter()
 
   const client = new CozyClient({
-    cozyURL: cozyUrl,
+    uri: cozyUrl,
```
y-lohse: Do we want to change this now? I think after this PR we will still be using the old CozyClient, so it should be called `cozyURL`, no?

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -46,12 +48,24 @@ class SuggestionProvider extends React.Component {
...
+      do {
+        response = await cozy.client.data.query(index, {
+          selector: { _id: { $gt: null } },
+          limit: limit,
+          skip: page * limit,
+          wholeResponse: true
+        })
+        files = files.concat(response.docs)
+        ++page
+      } while (response.next && page < pageLimit)
```
y-lohse: I can't remember how this was performing, we'll have to keep an eye on it.

src/drive/ducks/services/components/SuggestionProvider.jsx
```diff
@@ -13,10 +14,11 @@ class SuggestionProvider extends React.Component {
     window.addEventListener('message', event => {
       if (event.origin !== intent.attributes.client) return null
       const { query, id } = event.data
-      this.provideSuggestions(query, id, intent)
+      // sometimes we get messages with undefined query & id, no idea why
+      if (query && id) {
+        this.provideSuggestions(query, id, intent)
+      }
```
y-lohse: This is... concerning? Do you know what the messages are?

targets/drive/web/services.jsx
goldoraf: Hmmm... with this PR, cozy-client is not used anymore by the intents, only cozy-client-js is used. So I should get rid of this client instantiation right?

src/drive/ducks/services/components/SuggestionProvider.jsx
goldoraf: Yeah, it's definitely not good, but I didn't want to investigate the cozy-bar at that time. The `intent` looks good, but `query` and `id` are undefined...

targets/drive/web/services.jsx
y-lohse: Oh right we're in the services `main.jsx`! Yeah ok then I guess we don't need it at all.

PR #932, opened 2018-06-01
Fixed Drive alerts

goldoraf:
With the recent migration to the `cozy-ui` `Alerter`, we failed to fix some alerts, essentially the alerts that are included in redux actions and that were handled by the previous redux alerter.

### DIFF
--- a/src/drive/reducers/view.js
+++ b/src/drive/reducers/view.js
@@ -90,8 +90,11 @@ const fileCount = (state = null, action) => {
     case TRASH_FILES_SUCCESS:
     case DESTROY_FILES_SUCCESS:
       return state - action.ids.length
+    // This looks counter-intuitive to decrement the file count here,
+    // but we're in the trash, so when we restore files from the trash,
+    // the file count of the trash decrements
     case RESTORE_FILES_SUCCESS:
-      return state + action.ids.length
+      return state - action.ids.length
     default:
       return state
   }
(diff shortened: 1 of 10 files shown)


src/drive/reducers/view.js
```diff
@@ -91,7 +91,7 @@ const fileCount = (state = null, action) => {
     case DESTROY_FILES_SUCCESS:
       return state - action.ids.length
     case RESTORE_FILES_SUCCESS:
-      return state + action.ids.length
+      return state - action.ids.length
```
goldoraf: This change fixes an issue when restoring a file: the LoadMore button was displayed just after the restore, and it went in a loop of trying to load more files...

goldoraf:
[Certaines alertes/notifs ne s'affichent plus dans Drive](https://trello.com/c/JZqWlL0j/974-certaines-alertes-notifs-ne-saffichent-plus-dans-drive)

src/drive/reducers/view.js
enguerran: This is hard to understand how `DESTROY_FILES_SUCCESS` and `RESTORE_FILES_SUCCESS` modify the state the same way.

BTW, I guess this change has nothing to do with the `Alerter` of `cozy/cozy-ui`?

goldoraf: That's right, it looks weird, but apart from adding a comment, I don't see how I can make it clearer...

And yes, I stumbled upon that bug during testing of the alert fixes, and didn't feel like creating a separate PR for a single char change.

y-lohse: Can we add a comment then? I don't understand how this bugfix work, some at least a hint in the code would be good :)

PR #943, opened 2018-06-04
fix: fetching file with cozy client js

y-lohse:
- tells cozy-client-js to not rely on offline data to fetch the file
- formats the data as expected by the viewer


y-lohse:
[(1) Corriger le bouton "télécharger le fichier" dans la modale interapp drive](https://trello.com/c/3qJPNdIT/908-1-corriger-le-bouton-t%C3%A9l%C3%A9charger-le-fichier-dans-la-modale-interapp-drive)

PR #944, opened 2018-06-04
fix: specific message when upload fails because of the network

y-lohse:
- Detects network errors
- Shows a specific message when they happen
- Regular, unspecified errors are now grey instead of bright red

### DIFF
--- a/src/drive/actions/index.jsx
+++ b/src/drive/actions/index.jsx
@@ -236,7 +236,14 @@ export const uploadedFile = file => {
   }
 }
 
-export const uploadQueueProcessed = (loaded, quotas, conflicts, errors, t) => {
+export const uploadQueueProcessed = (
+  loaded,
+  quotas,
+  conflicts,
+  networkErrors,
+  errors,
+  t
+) => {
   if (quotas.length > 0) {
     // quota errors have their own modal instead of a notification, if possible
     if (t) {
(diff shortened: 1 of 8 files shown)


y-lohse:
[(0,5) Changer le message d'erreur en cas d'upload s'il n'y a pas de connexion disponible](https://trello.com/c/EcHZVzQ4/966-05-changer-le-message-derreur-en-cas-dupload-sil-ny-a-pas-de-connexion-disponible)

src/drive/actions/index.jsx
```diff
@@ -249,8 +256,10 @@ export const uploadQueueProcessed = (loaded, quotas, conflicts, errors, t) => {
       smart_count: loaded.length,
       conflictNumber: conflicts.length
     })
+  } else if (networkErrors.length > 0) {
+    Alerter.info('upload.alert.network')
   } else if (errors.length > 0) {
-    Alerter.error('upload.alert.errors')
+    Alerter.info('upload.alert.errors')
```
goldoraf: Why an `info` instead of `error`?

y-lohse: `error` creates a bright red alert, which we should (as of now) only use when the user is responsible for the error. Here's it's more likely a script error somewhere, so we wanted a nicer looking message.

PR #947, opened 2018-06-05
fix: schedule replication 🐛

kosssi:
When we use pouchdb we want to replicate pouch/couch every 30 seconds but for now only two replicate is launch.


y-lohse:
I think you meant to do this PR on the [cozy-client](https://github.com/cozy/cozy-client) repo ? THis one won't be active for much longer.

ptbrowne:
With one or two test, I would be more confident but it seems nice and short :)

kosssi:
@y-lohse In cozy-banks we use this library but I will create 2 issues on cozy-client to explain how make that (maybe implement it too).

PR #948, opened 2018-06-05
fix: update user agent

y-lohse:
The plugin didn't really work with iOS, but the `AppendUserAgent` preference seems to be doing the job!


y-lohse:
[Corriger la fiabilisation des stats](https://trello.com/c/4TppAKaN/982-corriger-la-fiabilisation-des-stats)

PR #951, opened 2018-06-07
chore: renovate cozy-ui all the time

y-lohse:


GoOz:
Even with Breaking versions?

y-lohse:
It's just the PR, the merge is still manual. So we can take our time with breaking changes :)

PR #953, opened 2018-06-07
fix: mobile bug when selection an item

y-lohse:
When selecting an item, we had an error because the selection bar couldn't find the correct icon :(. The icon is now in cozy-ui and we're using it now, so everything should be ok.


y-lohse:
[(1) Erreur lors d'un appui long pour sélectionner sur Android](https://trello.com/c/tywfjzfH/958-1-erreur-lors-dun-appui-long-pour-s%C3%A9lectionner-sur-android)

src/drive/ducks/files/Container.jsx
```diff
@@ -71,7 +71,7 @@ const mapDispatchToProps = (dispatch, ownProps) => {
           displayCondition: selections =>
             hasWriteAccess && selections.length === 1
         },
-        availableOffline: {
+        'phone-download': {
```
goldoraf: Not fan of using that kind of string as an object key in JS code, couldn' we stay with camelCase, even if it is ugly in Stylus files?

PR #955, opened 2018-06-07
Added folder upload by DnD

goldoraf:
Folder upload by Dn'd didn't work correctly: first, it simply didn't work at all on FF because of mimetypes (fixed with the new CC version), and second, it uploaded a flat list of files instead of creating dirs and sub-dirs. That's why I moved the folder handling code [in the upload module](https://github.com/cozy/cozy-drive/compare/master...goldoraf:folder-dnd-ff?expand=1#diff-a8db4e1272c43f0485054c5491a45fc0R104).

I also needed to have access to a CC instance in the upload thunks, so I used the [withExtraArgument method of redux-thunk](https://github.com/cozy/cozy-drive/compare/master...goldoraf:folder-dnd-ff?expand=1#diff-967a91470826706845a91e50ee1316c2R18).

### DIFF
--- a/src/drive/ducks/upload/index.js
+++ b/src/drive/ducks/upload/index.js
@@ -1,4 +1,3 @@
-/* global cozy */
 import { combineReducers } from 'redux'
 
 import UploadQueue from './UploadQueue'
@@ -21,8 +20,8 @@ const CONFLICT = 'conflict'
 const QUOTA = 'quota'
 const NETWORK = 'network'
 
-const itemInitialState = file => ({
-  file,
+const itemInitialState = item => ({
+  ...item,
   status: PENDING
 })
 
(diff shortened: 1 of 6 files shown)


goldoraf:
[(2) Transfert de fichiers & et de dossiers par drag and drop](https://trello.com/c/FQm5RH0Y/956-2-transfert-de-fichiers-et-de-dossiers-par-drag-and-drop)

goldoraf:
[(2) Transfert de fichiers & et de dossiers par drag and drop](https://trello.com/c/FQm5RH0Y/956-2-transfert-de-fichiers-et-de-dossiers-par-drag-and-drop)

src/drive/ducks/upload/index.js
```diff
@@ -125,3 +172,22 @@ export const getUploadQueue = state => state[SLUG].queue
...
+      results.push({
+        file: item.getAsFile(),
+        isDirectory: entry.isDirectory === true,
+        entry
+      })
+    } else {
+      results.push({ file: item, isDirectory: false, entry: null })
+    }
+  }
+  return results
```
y-lohse: I think we could use a `map()` call here, to be more in line with the rest of the code base?

goldoraf: Not sure it's possible to use a map, because these APIs use some kind of weird iterators. Maybe with an `Array.from`...

PR #956, opened 2018-06-07
fix: display a folder even without redux folder id

y-lohse:


y-lohse:
Yeah fair enough. i've added a comment and add this to our technical debt listing. In hindsight maybe having a separate Container for the recent file view might be another, better option.

PR #957, opened 2018-06-08
Replaced Drive ActionMenu by the cozy-ui one

goldoraf:
We have some UI fixes to implement on the ActionMenu (swipe RTL instead of top-down, ...), but as we'd like Banks to benefit from that fixes too, it's better to implement them in cozy-ui. So this PR makes Drive use the cozy-ui ActionMenu.


goldoraf:
[(2) 📱 Plus moyen de fermer le menu sandwich de Hambourg](https://trello.com/c/0jAwWSVq/821-2-%F0%9F%93%B1-plus-moyen-de-fermer-le-menu-sandwich-de-hambourg)

PR #962, opened 2018-06-08
Transition from previous cozy-client on mobile

y-lohse:
I've done a few things in this PR to stabilize the connexion process, i'll comment them below.

### DIFF
--- a/src/authentication/Revoked.jsx
+++ b/src/authentication/Revoked.jsx
@@ -11,8 +11,9 @@ class Revoked extends Component {
   }
 
   async logBackIn() {
-    const url = this.props.url || cozy.client._url
-    if (cozy.client) {
+    const url =
+      this.props.url || cozy.client._url || this.context.client.options.uri
+    if (cozy.client && cozy.client._storage) {
       cozy.client._storage.clear()
     }
     try {
@@ -24,7 +25,9 @@ class Revoked extends Component {
         token,
         router: this.props.router
       })
-    } catch (_) {}
+    } catch (e) {
+      console.warn(e)
+    }
   }
 
   render({ t }) {
--- a/src/drive/mobile/containers/DriveMobileRouter.jsx
+++ b/src/drive/mobile/containers/DriveMobileRouter.jsx
@@ -6,7 +6,7 @@ import MobileRouter from 'authentication/MobileRouter'
 
 import { setUrl, saveCredentials } from '../actions/settings'
 import { unlink } from '../actions/unlink'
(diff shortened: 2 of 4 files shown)


src/authentication/Revoked.jsx
```diff
@@ -11,8 +11,9 @@ class Revoked extends Component {
   }
 
   async logBackIn() {
```
y-lohse: Un-break loging back in again with the new cozy client, missed that one last time.

src/drive/mobile/containers/DriveMobileRouter.jsx
```diff
@@ -15,6 +15,7 @@ class DriveMobileRouter extends Component {
     this.props.saveCredentials(clientInfo, token)
     await restoreCozyClientJs(url, clientInfo, token)
     if (wasRevoked) {
+      initBar(this.context.client)
```
y-lohse: Restarts the cozy-bar after we have been revoked and logged back in.

targets/drive/mobile/main.jsx
```diff
@@ -43,28 +44,48 @@ window.handleOpenURL = require('drive/mobile/lib/handleDeepLink').default(
   hashHistory
 )
 
+const getPreviousToken = async store => {
```
y-lohse: In versions 1.5.3 and earlier, the cozy token was stored by cozy-client-js. But now we want to store it ourselves.
This code should stay here for a few versions and then it will be safe to delete it.

targets/drive/mobile/main.jsx
```diff
@@ -43,28 +44,48 @@ window.handleOpenURL = require('drive/mobile/lib/handleDeepLink').default(
...
+    return token
+  }
+}
+
 const startApplication = async function(store, client) {
   configureReporter()
-  const { client: clientInfos } = store.getState().settings
-    const { token } = store.getState().mobile.settings
-  if (clientInfos && token) {
+
```
y-lohse: There were some problems with the next part, so I re-organized it:

- if the server is not responding or we're offline, we don't revoke ourselves
- but we do revoke ourself if there is no token or client information
- the cozy bar is always started if we're logged in, but not if we're on the authentication screen

y-lohse:
[📱 L'app ne marche plus quand on met à jour depuis la version de prod](https://trello.com/c/pflPRwSb/997-%F0%9F%93%B1-lapp-ne-marche-plus-quand-on-met-%C3%A0-jour-depuis-la-version-de-prod)

PR #972, opened 2018-06-11
feat: show file thumbnails even without links

y-lohse:
In this PR, I change the `Preview` component so that even if the file has no links yet, we:

1. try to fetch the links and display one of them
2. fallback to the download URL

This works pretty well.

However, as we discussed, we have similar behaviors in `ImageLoader` and `ImageViewer`. I think we could put all the logic inside a single component, and use that one everywhere. The API would essentially look like this:

```
<ImageLoader
  file={myFile}
  size='small'
  onError={handleError}
  render={(src) => (
    <img src={src} className='whatever' alt={file.name} />
  )}
/>
```

Using the render prop, we can render an `` or a `<div>` with a background image or whatever we want. And the `ImageLoader` component doesn't have to deal with classnames, alts, etc.

If that sounds good to you, I will implement this component and replace the `Preview` and old `ImageLoader` component, and use it in the `ImageViewer`.


y-lohse:
[(1) 📱 Afficher les miniatures sur mobile une fois que l'app fonctionne en mode PouchDB](https://trello.com/c/NrVBdgyl/765-1-%F0%9F%93%B1-afficher-les-miniatures-sur-mobile-une-fois-que-lapp-fonctionne-en-mode-pouchdb)

src/viewer/ImageViewer.jsx
```diff
@@ -235,32 +212,35 @@ export default class ImageViewer extends Component {
...
-          )}
+        {file && (
+          <ImageLoader
+            file={file}
+            size="large"
+            ref={photo => {
+              this.photo = React.findDOMNode(photo)
+            }}
+            onError={this.onImageError}
+            key={file.id}
```
y-lohse: This one is interesting : `ImageLoader` needs to reset it's internal state when a prop changes, so that the new image gets loaded. But 
- calling `setState` in `ComponentDidUpdate` is disallowed by our linter, even with a condition
- `componentWill*` lifecycle methods have been deprecated
- I ended up [reading this](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-2-reset-uncontrolled-component-with-an-instance-method) and using a `key`

PR #976, opened 2018-06-12
Various Photos fixes

goldoraf:
This PR fixes various issues, mainly related to the sharing public page. I think the commit messages are sufficiently explicit.


targets/photos/web/public/index.ejs
```diff
@@ -6,18 +6,6 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Public Cozy Photos</title>
-  <% _.forEach(htmlWebpackPlugin.files.js, function(file) {
```
y-lohse: wat

goldoraf: I don't even know why that code was here...

PR #979, opened 2018-06-12
Micro fixes

y-lohse:
Lots of small fixes, I'll describe them below.

### DIFF
--- a/src/components/Image/ImageLoader.jsx
+++ b/src/components/Image/ImageLoader.jsx
@@ -2,7 +2,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const TTL = 6000
+const TTL = 10000
 
 const PENDING = 'PENDING'
 const LOADING_LINK = 'LOADING_LINK'
--- a/src/components/Image/ImageLoader.jsx
+++ b/src/components/Image/ImageLoader.jsx
@@ -30,6 +30,7 @@ class ImageLoader extends React.Component {
     if (status === PENDING) this.loadLink()
     else if (status === LOADING_LINK) this.loadFallback()
     else if (status === LOADING_FALLBACK) {
+      console.warn('failed loading thumbnail', lastError)
       this.setState({ status: FAILED })
       this.props.onError(lastError)
     }
@@ -41,11 +42,14 @@ class ImageLoader extends React.Component {
       this.img.onload = resolve
       this.img.onerror = reject
       this.img.src = src
-      this.timeout = setTimeout(reject, TTL)
+      this.timeout = setTimeout(
+        () => reject(new Error('Loading image took too long')),
+        TTL
+      )
     })
   }
 
-  async getFileLink(file) {
+  async getFileLinks(file) {
     if (file.links) return file.links
     else {
       const response = await cozy.client.files.statById(
--- a/src/components/Image/ImageLoader.jsx
+++ b/src/components/Image/ImageLoader.jsx
@@ -46,6 +46,10 @@ class ImageLoader extends React.Component {
         () => reject(new Error('Loading image took too long')),
         TTL
       )
+    }).finally(() => {
+      clearTimeout(this.timeout)
+      this.img.onload = this.img.onerror = null
+      this.img = null
     })
   }
 
--- a/src/components/Image/ImageLoader.jsx
+++ b/src/components/Image/ImageLoader.jsx
@@ -20,6 +20,14 @@ class ImageLoader extends React.Component {
     this.loadNextSrc()
   }
 
+  componentWillUnmount() {
+    if (this.img) {
+      this.img.onload = this.img.onerror = null
+      this.img.src = ''
+      clearTimeout(this.timeout)
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


src/components/Image/ImageLoader.jsx
```diff
@@ -20,6 +20,14 @@ class ImageLoader extends React.Component {
...
   }
 
+  componentWillUnmount() {
+    if (this.img) {
+      this.img.onload = this.img.onerror = null
+      this.img.src = ''
+      clearTimeout(this.timeout)
+    }
+  }
+
```
y-lohse: Cleaning up when the component unmounts, most notably canceling the image loading requests that are still pending.

src/components/Image/ImageLoader.jsx
```diff
@@ -30,6 +38,7 @@ class ImageLoader extends React.Component {
     if (status === PENDING) this.loadLink()
     else if (status === LOADING_LINK) this.loadFallback()
     else if (status === LOADING_FALLBACK) {
+      console.warn('failed loading thumbnail', lastError)
```
y-lohse: I suspect we'll get some cases of "why don't I have thumbnails?", this will allow us to find answers in the logs. We can remove this once the feature is stable.

src/components/Image/ImageLoader.jsx
```diff
@@ -41,11 +50,19 @@ class ImageLoader extends React.Component {
...
-      this.timeout = setTimeout(reject, TTL)
+      this.timeout = setTimeout(
+        () => reject(new Error('Loading image took too long')),
+        TTL
+      )
+    }).finally(() => {
+      clearTimeout(this.timeout)
+      this.img.onload = this.img.onerror = null
+      this.img.src = ''
+      this.img = null
```
y-lohse: Cleaning up, whether the image did load or not. Prevents fetching the raw version even  though the thumbnail was fetched, among other things.

src/drive/actions/async.js
```diff
@@ -7,10 +7,11 @@ class Stack {
     const folder = await cozy.client.files.statById(folderId, false, {
       limit: FILES_FETCH_LIMIT
     })
+
     const parentId = folder.attributes.dir_id
     const parent =
       !!parentId &&
-      (await cozy.client.files.statById(parentId).catch(ex => {
+      (await cozy.client.files.statById(parentId, false).catch(ex => {
```
y-lohse: Forces `statById ` to use the stack, especially useful while the first replication is underway.

src/drive/actions/async.js
```diff
@@ -126,8 +127,8 @@ class Stack {
 }
 
 class PouchDB {
-  constructor({ folder, recent, sort }) {
-    this.folderIndex = folder
+  constructor({ folders, recent, sort }) {
+    this.folderIndex = folders
```
y-lohse: That was a typo, the index is actually called `folders`. Seemed to work without it though ?

src/drive/containers/File.jsx
```diff
@@ -108,16 +108,18 @@ const FileName = ({
...
-        file={attributes}
-        size="small"
-        render={src => (
-          <div
-            className={styles['fil-file-preview']}
-            style={`background-image: url(${src});`}
-          />
-        )}
-      />
+      {attributes.class === 'image' && (
```
y-lohse: We only want to preload the image for actual images, not PDFs / folders / etc.

src/drive/mobile/containers/DriveMobileRouter.jsx
```diff
@@ -11,9 +11,10 @@ import { restoreCozyClientJs, initBar } from 'drive/mobile/lib/cozy-helper'
 class DriveMobileRouter extends Component {
   afterAuthentication = async ({ url, clientInfo, token, router }) => {
     const wasRevoked = this.props.isRevoked
+    this.context.client.options.uri = url
```
y-lohse: `client.options.uri` is set at the start of the app, but the very first time the app starts, it's undefined. Since we use the variable in several places afterwards, it's a bit of a problem, so here we define it.

src/drive/mobile/containers/DriveMobileRouter.jsx
```diff
@@ -11,9 +11,10 @@ import { restoreCozyClientJs, initBar } from 'drive/mobile/lib/cozy-helper'
...
+    await restoreCozyClientJs(url, clientInfo, token)
```
y-lohse: This needs to be done *before* calling `saveCredentials`, because that ends up starting a replication which needs a valid cozy-client-js.

targets/drive/mobile/main.jsx
```diff
@@ -77,7 +77,7 @@ const startApplication = async function(store, client) {
       // the server is not responding, but it doesn't mean we're revoked yet
       shouldInitBar = true
     }
-    else {
+    else if (store.getState().mobile.settings.serverUrl) {
```
y-lohse: Prevents dispatching a `revoke` during the first boot, which would make you skip the onboarding.

PR #981, opened 2018-06-13
fix: ios app id for ratting

y-lohse:
The plugin now needs the app store id number rather than the string?


y-lohse:
[(1) Corriger le formulaire de notation sur iphone](https://trello.com/c/pE7hl8cA/961-1-corriger-le-formulaire-de-notation-sur-iphone)

PR #984, opened 2018-06-13
Show recents files first when displaying an Administrative folder or backuped photos

goldoraf:
As the condition for determining if the displayed folder should show recent files first depends on localized paths, I had to bring the `t` function into the redux context. I again used the `withExtraArgument` method of `react-redux` thus the changes related to the store configuration.

### DIFF
--- a/src/drive/actions/async.js
+++ b/src/drive/actions/async.js
@@ -2,8 +2,16 @@
 import { isCordova } from '../mobile/lib/device'
 import { TRASH_DIR_ID, FILES_FETCH_LIMIT } from '../constants/config.js'
 
+export const shouldShowRecentsFirst = (
+  folderPath,
+  parentPath,
+  specialFolders
+) =>
+  specialFolders.indexOf(folderPath) !== -1 ||
+  specialFolders.indexOf(parentPath) !== -1
+
 class Stack {
-  getFolder = async folderId => {
+  getFolder = async (folderId, specialFolders = []) => {
     const folder = await cozy.client.files.statById(folderId, false, {
       limit: FILES_FETCH_LIMIT
     })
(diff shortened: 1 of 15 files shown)


goldoraf:
[(2) Trier par défaut du plus récent au plus ancien les dossiers dans /Administratif et le répertoire de sauvegarde des photos](https://trello.com/c/iGsgxp4G/960-2-trier-par-d%C3%A9faut-du-plus-r%C3%A9cent-au-plus-ancien-les-dossiers-dans-administratif-et-le-r%C3%A9pertoire-de-sauvegarde-des-photos)

src/drive/actions/index.jsx
```diff
@@ -86,15 +90,28 @@ export const openFolder = folderId => {
       }
     })
     try {
+      const specialFolders = [
+        t('mobile.settings.media_backup.media_folder'),
+        `/${t('Nav.item_collect')}`
```
y-lohse: Why does one have a `/` and not the other ?

src/drive/actions/async.js
```diff
@@ -2,8 +2,16 @@
 import { isCordova } from '../mobile/lib/device'
 import { TRASH_DIR_ID, FILES_FETCH_LIMIT } from '../constants/config.js'
 
+export const shouldShowRecentsFirst = (
+  folderPath,
+  parentPath,
+  specialFolders
+) =>
+  specialFolders.indexOf(folderPath) !== -1 ||
+  specialFolders.indexOf(parentPath) !== -1
```
y-lohse: I wonder if it would be better to use regular expressions here, as it would avoid matching things like `/lol/Administrative`. Just a detail, but what do you think?

goldoraf: It's not necessary imo, this condition doesn't match `/lol/Administrative`, only `/Administrative` and sub-folders.

goldoraf:
Concerning the PouchDB part, as the sort is disabled on mobile for now for perfs reasons, I didn't want to force a default sort on some folders: it'd slow the app and give users the impression that some folders are sorted randomly. I've added a TODO here: https://trello.com/c/UxXUjiar/965-timebox-forum-5712-le-tri-est-lent

PR #985, opened 2018-06-13
Sharing fixes

goldoraf:
This PR fixes a few things:
 - when a sharing is revoked, if the corresponding `byId` entry is "empty" (no more sharings or perms), it is deleted ;
 - we filter out self-revoked sharings and sharings with all revoked recipients ;
 - UI fixes.

### DIFF
--- a/src/sharing/state.js
+++ b/src/sharing/state.js
@@ -14,8 +14,7 @@ export const receiveSharings = ({
   type: RECEIVE_SHARINGS,
   data: {
     sharings: sharings.filter(
-      // we filter sharings revoked by the recipient
-      s => s.attributes.owner || s.attributes.members[0].status !== 'revoked'
+      s => !areAllRecipientsRevoked(s) && !hasBeenSelfRevoked(s)
     ),
     permissions,
     apps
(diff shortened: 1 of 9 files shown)


src/sharing/state.js
```diff
@@ -25,10 +43,19 @@ export const revokeSharingLink = permissions => ({
...
+const updateByIdItem = (state, id, updater) => {
+  const { [id]: byIdState, ...rest } = state
+  const update = updater(byIdState || byIdInitialState)
+  return isItemEmpty(update)
+    ? rest
+    : {
+        ...rest,
+        [id]: update
+      }
+}
```
y-lohse: I get what this function does, but it took me a while. I wonder if just renaming things would be enough, ie.

- updateByIdItem -> updateByItemId
- byIdState -> stateById
- byIdInitialState -> initialItemState

what do you think?

goldoraf: I think updateByItemId is not better but worse, we're talking about a `byId` item, not a item ID. So I propose:
 - byIdState -> byIdItemState
 - byIdInitialState -> byIdItemInitialState

Does it look better to you?

y-lohse: > I think updateByItemId is not better but worse, we're talking about a byId item, not a item ID

Oooh ok. In that case I think the variables are properly named. And I don't really know how to make it more obvious, so let's go for this.

PR #992, opened 2018-06-18
fix: iphonex UI

y-lohse:
Corrects the layout problems on the iPhone X.

### DIFF
--- a/src/drive/styles/layout.styl
+++ b/src/drive/styles/layout.styl
@@ -2,6 +2,8 @@
 
 .fil-wrapper
     @extend    $app-2panes-sticky
+    padding-left env(safe-area-inset-left)
 
 .fil-sidebar
     justify-content  space-between
+    padding-bottom env(safe-area-inset-bottom)
--- a/targets/drive/mobile/config.xml
+++ b/targets/drive/mobile/config.xml
@@ -72,18 +72,7 @@
         <icon height="152" src="res/icons/ios/icon-76-2x.png" width="152" />
         <icon height="167" src="res/icons/ios/icon-83.5-2x.png" width="167" />
         <icon height="1024" src="store/icons/icon-1024.jpg" width="1024" />
-        <splash height="1136" src="res/screens/ios/screen-iphone-568h-2x.png" width="640" />
-        <splash height="480" src="res/screens/ios/screen-iphone-portrait.png" width="320" />
-        <splash height="960" src="res/screens/ios/screen-iphone-portrait-2x.png" width="640" />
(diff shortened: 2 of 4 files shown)


src/drive/styles/layout.styl
```diff
@@ -2,6 +2,8 @@
 
 .fil-wrapper
     @extend    $app-2panes-sticky
+    padding-left env(safe-area-inset-left)
 
 .fil-sidebar
     justify-content  space-between
+    padding-bottom env(safe-area-inset-bottom)
```
y-lohse: @GoOz I think it would be good to add these changes to cozy-ui, but I'm not familiar with the different layouts these. [More infos here](https://webkit.org/blog/7929/designing-websites-for-iphone-x/).

targets/drive/mobile/config.xml
```diff
@@ -72,18 +72,7 @@
...
-        <splash height="1334" src="res/screens/ios/screen-iphone-portrait-667h.png" width="750" />
-        <splash height="2208" src="res/screens/ios/screen-iphone-portrait-736h.png" width="1242" />
-        <splash height="1242" src="res/screens/ios/screen-iphone-landscape-736h.png" width="2208" />
-        <splash height="1024" src="res/screens/ios/screen-ipad-portrait.png" width="768" />
-        <splash height="2048" src="res/screens/ios/screen-ipad-portrait-2x.png" width="1536" />
-        <splash height="768" src="res/screens/ios/screen-ipad-landscape.png" width="1024" />
-        <splash height="1536" src="res/screens/ios/screen-ipad-landscape-2x.png" width="2048" />
-        <splash height="2048" src="res/screens/ios/screen-ipad-landscape-ipadpro.png" width="2732" />
-        <splash height="2732" src="res/screens/ios/screen-ipad-portrait-ipadpro.png" width="2048" />
+        <splash src="res/screens/ios/Default@2x~universal~anyany.png" />
```
y-lohse: Switching to [launch storyboard images](https://github.com/apache/cordova-plugin-splashscreen#launch-storyboard-images), we only provide one, very big splashscreen, that gets resized as needed.

y-lohse:
[(?) Corriger l'affichage sur iphone X](https://trello.com/c/u2vdprug/1000-corriger-laffichage-sur-iphone-x)

src/drive/styles/layout.styl
y-lohse: `safe-area-inset-bottom` is only provided on the iphone x, on anything else it just produces an invalid rule that is ignored :)

GoOz: Yeah I just saw your link and it took me time to read it. Sorry, we're good then. ;)

PR #996, opened 2018-06-20
Use photos orientation metadatum to correct image ratios

goldoraf:
We need to take into account EXIF orientation tags (present on some images) to dispaly the correct image ration in all cases.


goldoraf:
[(1) [forum-5717] Miniatures de photos portrait déformées](https://trello.com/c/cz6BcZ5z/962-1-forum-5717-miniatures-de-photos-portrait-d%C3%A9form%C3%A9es)

PR #998, opened 2018-06-20
Better handling of mimetype icons by guessing the type from the extension

goldoraf:
I used a different lib for mimetype lookup because the one used by cozy-client seems problematic. I've made the change in cozy-client too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cozy-drive",
-  "version": "1.5.7",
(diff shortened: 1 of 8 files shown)


goldoraf:
[(1) Certains fichiers n'ont pas la bonne icône](https://trello.com/c/w5neYxjT/995-1-certains-fichiers-nont-pas-la-bonne-ic%C3%B4ne)

PR #1001, opened 2018-06-20
Use SafariViewController to create account

kosssi:


src/authentication/withHasSafariPlugin.jsx
```diff
@@ -0,0 +1,48 @@
...
+      componentDidMount() {
+        this.checkSafariPlugin()
+      }
+
+      render() {
+        return <WrappedComponent {...this.props} {...this.state} />
+      }
+    }
+  }
+}
```
ptbrowne: I'm not sure this HOC is really useful :/

src/authentication/steps/Welcome.jsx
```diff
@@ -11,8 +12,54 @@ const getPlatformId = () =>
...
+        </a>
+      )
+    }
+
+    const url = `https://manager.cozycloud.cc/cozy/create?pk_campaign=drive-${getPlatformId() ||
+      'browser'}`
+
+    if (hasSafariPlugin) {
+      const openManager = () => {
+        SafariViewController.show(
```
ptbrowne: It is a bit disturbing that with the hasSafariPlugin, it's easier for us to know that SafariViewController exists and is usable but then we still have to know about it.

I wonder if the cozy-device-helper should not directly expose a way of calling this component.

PR #1002, opened 2018-06-22
Optimizing filtering performances

y-lohse:
This PR optimizes fetching and sorting files, especially on mobile. I still need to upgrade cozy-client-js after https://github.com/cozy/cozy-client-js/pull/270

Please ignore the list of commits, I will squash them later on.

- I made minor changes to the [`StackAdapter`](https://github.com/cozy/cozy-drive/pull/1002/files#diff-05794503f15d8f5deb3e90894c71798cR13)
- And more [important changes to the `PouchAdapter`](https://github.com/cozy/cozy-drive/pull/1002/files#diff-05794503f15d8f5deb3e90894c71798cR171)
- And I've adjusted the [index creation and warm-up](https://github.com/cozy/cozy-drive/pull/1002/files#diff-c29718694254b6d8e990fc33abd91492R25)


y-lohse:
[(5) [forum-5712] Le tri est lent](https://trello.com/c/UxXUjiar/965-5-forum-5712-le-tri-est-lent)

src/drive/actions/async.js
```diff
@@ -71,64 +75,81 @@ class Stack {
...
     sortOrder = 'asc',
     skip = 0,
-    limit = FILES_FETCH_LIMIT,
-    loadedFoldersCount = 0,
-    loadedFilesCount = 0
+    limit = FILES_FETCH_LIMIT
   ) => {
-    const shouldSeparateFilesAndFolders = sortAttribute === 'name'
-    const index = await cozy.client.data.defineIndex('io.cozy.files', [
+    const isFirstLoad = skip === 0
```
y-lohse: When sorting, we always want to separate files and folders now.

src/drive/actions/async.js
```diff
@@ -71,64 +75,81 @@ class Stack {
...
+      const folderSortingOrder = sortAttribute === 'name' ? sortOrder : 'asc'
+      const allFolders = await this.query({
+        folderId,
+        type: 'directory',
+        sortAttribute,
+        sortOrder: folderSortingOrder,
+        skip: 0,
+        limit: null
+      })
+      folders = allFolders.filter(folder => folder._id !== TRASH_DIR_ID)
```
y-lohse: The first time the view is loaded, we load *all* the folders — then when the user scrolls and we fetch more data, we can focus on the files. If there are tons of folders this might end up being slower, but that should be rare.

We also filter out the trash folder with javascript as it is faster than doing it through Pouch.

src/drive/actions/async.js
```diff
@@ -71,64 +75,81 @@ class Stack {
...
+      type: 'file',
+      sortAttribute,
+      sortOrder,
+      skip,
+      limit
+    })
+
+    return [...folders, ...files]
+  }
+
```
y-lohse: Whats follows is just a more generic `query` function, it hasn't changed much.

src/drive/actions/async.js
```diff
@@ -71,64 +75,81 @@ class Stack {
...
 
-  RECENT_FILES_INDEX_FIELDS = ['updated_at', 'trashed']
-  RECENT_FILES_QUERY_OPTIONS = {
-    selector: {
-      updated_at: { $gt: null },
-      trashed: false
-    },
-    sort: [{ updated_at: 'desc' }],
-    limit: FILES_FETCH_LIMIT
+  getIndex = async (name, fields) => {
```
y-lohse: Index are now cached so that we don't re-query them all the time.

src/drive/actions/async.js
```diff
@@ -148,101 +169,112 @@ class Stack {
...
     const parent = !!folder.dir_id && (await db.get(folder.dir_id))
-    const files = await this.getFolderContents(folderId)
+    const recentsFirst =
+      !!parent &&
+      shouldShowRecentsFirst(folder.path, parent.path, specialFolders)
+
+    const files = recentsFirst
+      ? await this.getSortedFolder(folderId, 'updated_at', 'desc')
+      : await this.getSortedFolder(folderId, 'name', 'asc')
+
```
y-lohse: We're now auto-sorting the special folders, same as with the Stack Adapter.

src/drive/actions/async.js
```diff
@@ -148,101 +169,112 @@ class Stack {
...
-          sort: [{ [sortAttribute]: sortOrder }],
-          skip: skipRows,
-          limit
-        })
-        .then(resp => resp.docs.map(f => normalizeFileFromPouchDB(f)))
-
-    if (shouldSeparateFilesAndFolders) {
-      const folders = await query({
-        type: 'directory'
+    const isFirstLoad = skip === 0
```
y-lohse: You'll notice that the content of `getSortedFolder` is literally the same between the StackAdapter and the PouchAdapter — only the underlying `query` method changes.

I'd like to avoid that duplication, but not sure how. They are both definitely class methods, but having one adapter `extend` the other doesn't seem right  either, as literally everything else changes. And having both `extend` a base class is maybe a bit overkill?

src/drive/actions/async.js
```diff
@@ -148,101 +169,112 @@ class Stack {
...
+
+    const files = await this.query({
+      folderId,
+      type: 'file',
+      sortAttribute,
+      sortOrder,
+      skip,
+      limit
+    })
+    return [...folders, ...files]
```
y-lohse: So same as above here — we load all the folders the first time, and then just the files.

src/drive/actions/async.js
```diff
@@ -263,10 +295,11 @@ class PouchDB {
   }
 }
 
+const StackAdapter = new Stack()
 export const getAdapter = state =>
   shouldWorkFromPouchDB(state)
     ? new PouchDB(state.settings.indexes)
-    : new Stack()
+    : StackAdapter
```
y-lohse: This helps preserving the caches.

src/drive/components/FileList.jsx
```diff
@@ -25,7 +25,7 @@ class FileList extends PureComponent {
 
   checkIntersectionsEntries = intersectionEntries => {
     if (intersectionEntries.filter(entry => entry.isIntersecting).length > 0) {
-      this.loadMoreRows(FILES_FETCH_LIMIT)
+      if (!this.state.isLoading) this.loadMoreRows(FILES_FETCH_LIMIT)
```
y-lohse: This call was triggered twice, so we got 2 load more fired every time.

src/drive/components/FileList.jsx
```diff
@@ -45,13 +45,11 @@ class FileList extends PureComponent {
...
-    // We're in /recent
     if (!this.props.displayedFolder) return false
-    if (isCordova()) {
-      if (this.props.files.length < FILES_FETCH_LIMIT) return false
+    else if (isCordova()) {
+      if (this.props.files.length < FILES_FETCH_LIMIT) return false // We're in /recent
       return !this.state.hasNoMoreRows
-    }
-    return this.props.files.length < this.props.fileCount
+    } else return this.props.files.length < this.props.fileCount
```
y-lohse: Just making the code pathes more explicit.

src/drive/actions/async.js
goldoraf: I'm not sure trying to avoid that duplication is worth it right now. Maybe we'll find a better way later if we intergate this work in cozy-client.

src/drive/actions/async.js
goldoraf: I think these 2 informations deserve a little comment in the code, don't they?

src/drive/components/FileList.jsx
goldoraf: I think it's not a good idea, it adds a level of indentation for nothing, and the "return fast" pattern is well-known.

PR #1003, opened 2018-06-22
fix: cozy-authentication would add global html styles

ptbrowne:
In Banks, the cozy-authentication page would be blank because it would add default element styles that would put the "[role=application]" element's height to 0. When importing cozy-authentication, we should not have to worry about global css styles.

I took this opportunity to move the files in a proper `src` directory and to remove the dependency on drive's webpack, the idea being to extract this package into cozy-libs.


src/authentication/webpack.config.js
```diff
@@ -1,31 +1,86 @@
...
+    alias: {
+      react: 'preact-compat',
+      'react-dom': 'preact-compat',
+      'create-react-class': 'preact-compat/lib/create-react-class'
+    }
+  }
+}
 
-  return merge(preactConfig, uiConfigNoExtract, {
+console.log(path.resolve(__dirname, '../..'))
```
y-lohse: Left-over from developing ?

PR #1006, opened 2018-06-25
fix: renamed index

y-lohse:
Renaming the index was unnecessary and caused the index to be recreated, leading to potential errors.


y-lohse:
[(5) [forum-5712] Le tri est lent](https://trello.com/c/UxXUjiar/965-5-forum-5712-le-tri-est-lent)

PR #1009, opened 2018-06-27
Fixed upload queue's FOUC and other issues

goldoraf:
In this PR, I also purged the queue when more files are added, and handled better the mimetypes' display in the queue.

### DIFF
--- a/src/drive/ducks/upload/UploadQueue.jsx
+++ b/src/drive/ducks/upload/UploadQueue.jsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react'
 import classNames from 'classnames'
 import { connect } from 'react-redux'
 import { translate } from 'cozy-ui/react/I18n'
+import { getFileMimetype } from 'drive/lib/getFileMimetype'
 
 import styles from './styles'
 import {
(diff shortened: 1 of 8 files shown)


src/drive/ducks/upload/UploadQueue.jsx
```diff
@@ -71,11 +78,18 @@ class UploadQueue extends Component {
     this.setState(state => ({ collapsed: !state.collapsed }))
   }
 
+  componentDidMount() {
+    this.wrapper.style.transition = '0.5s'
```
y-lohse: I think we should, as much as possible, keep the css in one consistent place, and right now this is in the stylus files. So I'm not very comfortable with setting the transition time in js while everything else is in the css file — it's going to be very hard to find.

Do you know *why* the uploader appears ? From what I understand, it should only start with the `upload-queue` class, so it should start translated and with the `visibility` set to hidden — so it shouldn't be visible at all on first render..?

If we have to set the transition after the first render, then I think it would be better to apply it via a `upload-queue--animated` class or whatever.

goldoraf: I'm not comfortable too with setting the transition like this, but the problem is: if I add a class on `componentDidMount`, this class will get lost on the first rerender (that will reapply classes). That's why I choose to use `style`.

And no, I don't really know why the uploader FOUCs, in theory it shouldn't be visible at all, but my guess is that the `transition` time mess this up.

goldoraf:
As discussed on Mattermost, I'm merging this hoping to see https://github.com/cozy/cozy-drive/pull/1009/commits/00efb3970fc17847c588ce3f081f34b752230a77 fix FOUC issues with high-end GPUs

PR #1013, opened 2018-06-28
fix: page title, fix #1012

### LINKED ISSUE #1012: Title of the page
cozy-drive instead of Cozy Drive.

Screenshot from fab.

image

If you use cozy-scripts the problem should be fixed by https://github.com/CPatchane/create-cozy-app/pull/627

y-lohse:
See #1012


enguerran:
Thanks for the amazing work, this is very valuable.
Nevertheless, could you update your commit message and the PR description to take into account the guidelines? https://github.com/cozy/cozy-guidelines#commit-messages

PR #1014, opened 2018-06-28
chore: temp remove ios openwith plugin

y-lohse:
I assumed we would soon continue work on this, but it's still postponed. In the meantime, just having the plugin present generates some maintenance, so I'd like to remove it. This commit can be reverted once we pick up the subject again.


y-lohse:

PR #1021, opened 2018-06-28
[WIP] Open files shared by link in the viewer

goldoraf:
Previously, when sharing a single file by link, we redirected to a download URL. We now want to open it in a light version of the viewer, with a white background and with a special version of the cozy-bar on top. I had to add a few props to `Viewer` to enable that.

### DIFF
--- a/src/components/Button/index.styl
+++ b/src/components/Button/index.styl
@@ -1,5 +1,6 @@
 @require 'settings/breakpoints.styl'
 @require 'utilities/display.styl'
+@require 'settings/z-index.styl'
 
 .u-visuallyhidden
     @extend $visuallyhidden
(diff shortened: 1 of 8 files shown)


src/components/Button/index.styl
```diff
@@ -20,3 +21,9 @@
         padding          .5rem .75rem
         opacity          .5
         background-color transparent
+
+.dri-cozyhomelink--in-bar
```
y-lohse: I suggest we rename this selector into something like `bar-homelink` because:

- styles are scoped, so we don't need a prefix
- `--in-bar` would be a BEM modifier, but there's no `.cozyhomelink` block or element

PR #1030, opened 2018-07-02
fix: table's last row bottom border

GoOz:
I introduced a regression a few days ago by moving the border of a table-row from the bottom to the top. But I forgot a usecase, so I added a bottom border to the last row of a table (fix in UI update) and because the border is now at the top we had a disgraceful double border on the mobile app with the `.coz-upload-status` element.
This PR  fixes that by removing this element's border.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "cozy-bar": "5.0.7",
     "cozy-client": "1.0.0-beta.14",
     "cozy-client-js": "0.10.0",
-    "cozy-ui": "10.1.0",
+    "cozy-ui": "^10.1.2",
     "create-react-context": "0.2.2",
     "date-fns": "1.29.0",
     "diacritics": "1.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -149,7 +149,7 @@
     "cozy-bar": "5.0.7",
     "cozy-client": "1.0.0-beta.14",
     "cozy-client-js": "0.10.0",
-    "cozy-ui": "10.1.0",
+    "cozy-ui": "^10.1.2",
```
y-lohse: Out of curiosity, what command do you use to upgrade `cozy-ui` ? It keeps adding the carret which we don't want (renovate will remove it, but still).

y-lohse: Next time, can you try to specify the version instead of `latest` please ? I think that will do the trick.

GoOz: Yeah sorry :)

PR #1033, opened 2018-07-02
Redux store cleaning after unlink

y-lohse:
There's a section of the store that didn't get cleaned unlinking from a Cozy. I fixed it, but I also added a technical debt card — as it turns out we have 2 `settings` sections in the store, and both are bout mobile stuff, but one is loaded all the time.

### DIFF
--- a/src/drive/actions/async.js
+++ b/src/drive/actions/async.js
@@ -1,5 +1,4 @@
-/* global cozy */
-import { isCordova } from '../mobile/lib/device'
+/* global cozy __TARGET__ */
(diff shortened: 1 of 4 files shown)


y-lohse:
[(1) 📱🐛 Après la révocation de l'app, les données ne se chargent plus](https://trello.com/c/PWELcLW4/776-1-%F0%9F%93%B1%F0%9F%90%9B-apr%C3%A8s-la-r%C3%A9vocation-de-lapp-les-donn%C3%A9es-ne-se-chargent-plus)

PR #1034, opened 2018-07-05
Lock sharing type

y-lohse:
Prevents the user from changing the sharing type after the first share, and adds a tooltip to explain this.


y-lohse:
[(1) Changer la modale pour prendre en compte le contrainte de partage seulement en lecture-écriture ou seulement en lecture-seule](https://trello.com/c/Kr1qBgvE/983-1-changer-la-modale-pour-prendre-en-compte-le-contrainte-de-partage-seulement-en-lecture-%C3%A9criture-ou-seulement-en-lecture-seule)

src/sharing/index.jsx
```diff
@@ -263,6 +264,7 @@ const OwnerSharingModal = ({ document, ...rest }) => (
             createContact={createContact}
             recipients={getRecipients(document.id)}
             link={getSharingLink(document)}
+            isShared={byDocId[document.id] !== undefined}
```
enguerran: Your commit message is `changed file extension`
But in the commit I see three changes
- the file has been renamed
- the function parameter has been changed
- a props in a third component has been added

Something is wrong with the commit:
1. if every changes are tighted together, then change the commit message (use the infinite form with a capital letter)
1. if changes are decorrelate (what I think), split your changes into several commits with good and informative commit messages

That is important because there is a lot of people working on the code and the git history (`git log`) is very useful to follow the code changes and understand the past, the present and be prepared for the future.

PR #1036, opened 2018-07-05
Fix offline boot

y-lohse:
There were é issues when starting the app offline:

- at least on iOS, the error we received was interpreted as meaning we had been disconnected. I inverted the logic — now we only consider ourselves revoked when we encounter specific error codes, in all other cases, we consider ourselves offline
- cozy-client-js was not restored at the application start, so PouchDB databases were unavailable


y-lohse:
Thanks! Habits are hard to change, but I'll get there.

PR #1040, opened 2018-07-10
Forbid some sharings

goldoraf:
Like described in the Trello card, we have to forbid sharing of a folder/file whose parent is shared or who have a shared child. We therefore need to get the paths of all shared folders/files. This is straightforward for folders, because they have a `path` property, but for files, we need to fetch their parent dir to get the path and append the file name.

I also fine-tuned some CSS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
   "dependencies": {
     "classnames": "2.2.6",
     "cozy-bar": "5.0.8",
-    "cozy-client": "1.0.0-beta.14",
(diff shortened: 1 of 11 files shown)


goldoraf:
[(3) Ecran de la modale d'un enfant ou un parent d'un élément partagé](https://trello.com/c/wZS5rpxu/992-3-ecran-de-la-modale-dun-enfant-ou-un-parent-dun-%C3%A9l%C3%A9ment-partag%C3%A9)

PR #1050, opened 2018-07-16
fix: Properly close action menu

y-lohse:
Until now, the action menu was closed when sharing due to a bug in cozy-ui. This also had some very strange side effect: the sharing modal would sometimes get `translateY` instead of the menu (presumably a bug in preact?).

This change closes it in a better way while also avoiding the translate bug.


y-lohse:
[(1) 🐛 Ouvrir deux fois la modale de partage est impossible sur mobile](https://trello.com/c/nxNLeOJM/1044-1-%F0%9F%90%9B-ouvrir-deux-fois-la-modale-de-partage-est-impossible-sur-mobile)

PR #1053, opened 2018-07-17
fix: Only init bar if there's a server url

y-lohse:
If you revoke the client, kill the app and restart it, the cozy bar would be initialized. This PR fixes it.


y-lohse:
[Release avec le partage activé et les correctifs mobiles](https://trello.com/c/cK3KGwEx/1052-release-avec-le-partage-activ%C3%A9-et-les-correctifs-mobiles)

PR #1056, opened 2018-07-18
Improve revocation and expired token handling

y-lohse:
There were a few issues when a token expired — the cozy-bar and cozy-client-js were keeping their old tokens, and their requests would fail.


y-lohse:
[🐛 Corriger le problème de non-rafraichissement du token OAuth](https://trello.com/c/QrphxR16/1042-%F0%9F%90%9B-corriger-le-probl%C3%A8me-de-non-rafraichissement-du-token-oauth)

PR #1058, opened 2018-07-18
fix: Scrolling in share modal

y-lohse:
The share modal didn't scroll when the content got bigger, now it does!


y-lohse:
[🐛 (1) Quand il y a beaucoup de participants, le scroll dépasse de la modale](https://trello.com/c/H1JxU1E4/1047-%F0%9F%90%9B-1-quand-il-y-a-beaucoup-de-participants-le-scroll-d%C3%A9passe-de-la-modale)

PR #1066, opened 2018-07-23
Added sharing recipients to the action menu

goldoraf:
Not much to add, it's a rather straightforward change

### DIFF
--- a/src/drive/styles/actionmenu.styl
+++ b/src/drive/styles/actionmenu.styl
@@ -46,3 +46,10 @@
     background embedurl("../assets/icons/icon-rename.svg") 1rem center no-repeat
 .fil-action-share
     background embedurl("../assets/icons/icon-share-grey.svg") 1rem center no-repeat
+    position   relative
+    
+    > div
+        position absolute
+        right    1rem
+        top      .1rem
+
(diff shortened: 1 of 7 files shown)


goldoraf:
[(2) Affiche l'indicateur de partage dans le file action menu sur mobile](https://trello.com/c/0WAmzJSP/1022-2-affiche-lindicateur-de-partage-dans-le-file-action-menu-sur-mobile)

src/sharing/components/Avatar.jsx
```diff
@@ -47,15 +54,21 @@ export const AvatarPlusX = ({ extraRecipients = [] }) => (
...
 }
 
-export const AvatarLink = () => (
-  <div className={styles['recipient-avatar']}>
+export const AvatarLink = ({ size = 'medium' }) => (
+  <div
+    className={cx(styles['recipient-avatar'], {
+      [styles['recipient-avatar--small']]: size === 'small'
+    })}
+  >
```
y-lohse: I wonder if we could avoid duplicating the class declarations. I was already wondering this the other day when I was working on this but didn't get around doing anything about it.

src/drive/styles/actionmenu.styl
```diff
@@ -46,3 +46,10 @@
     background embedurl("../assets/icons/icon-rename.svg") 1rem center no-repeat
 .fil-action-share
     background embedurl("../assets/icons/icon-share-grey.svg") 1rem center no-repeat
+    position   relative
+    
+    > div
```
y-lohse: I think `div` as a selector here is a bit confusing. Wouldn't we be better off giving it a class like `action-extra` or something like that?

PR #1068, opened 2018-07-25
feat: Add sharings view

y-lohse:
The commits go a bit back and forth, so I recommend looking at the final diff and this comment. I'll squash the commits before merging.

I've added a brand new which displays the list of shared documents. This was pretty challenging because:

a. It's not a regular folder, it's more a special view like Recent except you can navigate in it and stuff
b. Other views use cozy-client-js / pouch to fetch data, but we already lots of (better) code to handle sharings.

The result is, i think, a step towards where we will go in the future  as we migrate to cozy-client… but in the meantime it's a bit hacky.

1. [I added a new Container](https://github.com/cozy/cozy-drive/pull/1068/files#diff-a25ced079db66df431ae7ec70126fd20R1). It gets the list of shared documents from the `SharingProvider` and fetches the corresponding files and folders when needed
2. [Fetched data is dispatched and reduced](https://github.com/cozy/cozy-drive/pull/1068/files#diff-dba5935292b52df2da17bc1ea6222f12R13) so that it can be integrated into the redux store and be used by our usual components (`FileExplorer` and all)
3. From there on it's pretty much business as usual for `FolderView` and the subcomponents
4. [I had to modify the `breadcrumbs`](https://github.com/cozy/cozy-drive/pull/1068/files#diff-9e06191a586128b4bca10089eec2312dR24), because for the first time we have a clickable breadcrumb that is not a regular folder.

I think it comes down to this:

All the other containers rely on a `fetchMyData` function, which is sometimes called by `FileExplorer`, and sometimes by the `Nav`. The `SharingsContainer` handles all of that itself and injects the resulting data into the store.


y-lohse:
[(5) Je peux facilement retrouver les dossiers/fichiers "Mes partages"](https://trello.com/c/qv2MRAHI/1024-5-je-peux-facilement-retrouver-les-dossiers-fichiers-mes-partages)

PR #1069, opened 2018-07-30
Sharings view fixes

y-lohse:
The sorting fix is pretty straightforward.

For the breadcrumbs, when on the sharings view, we need to know if the folders are shared or not — because if they are, we don't want to show their (not shared) parent in the breadcrumb.


y-lohse:
[(5) Je peux facilement retrouver les dossiers/fichiers "Mes partages"](https://trello.com/c/qv2MRAHI/1024-5-je-peux-facilement-retrouver-les-dossiers-fichiers-mes-partages)

PR #1070, opened 2018-07-30
chore: Update file-opener plugin

y-lohse:
The file-opener plugin is out of date. Sadly there's no new official release, so in the meantime I've switched us to `master`. I've also opened an issue : https://github.com/pwlin/cordova-plugin-file-opener2/issues/219


y-lohse:
[🐛 Impossible d'accéder aux fichiers hors-ligne sur iOS](https://trello.com/c/uRxapjiz/1072-%F0%9F%90%9B-impossible-dacc%C3%A9der-aux-fichiers-hors-ligne-sur-ios)

enguerran:
By using the master branch instead of a fixed version, you may not have the same version between two builds (idempotence of builds).

y-lohse:
Yep, that's definitely the problem. But it's also the accepted solution in this case, and hopefully we can switch back to a fixed version soon.

PR #1074, opened 2018-08-02
Added sharing recipients to the more menu

goldoraf:
As the more menu on mobile is injected into the cozy-bar, it doesn't have access to the sharing context. That's why I had to use `RecipientsAvatars` [here](https://github.com/cozy/cozy-drive/pull/1074/files#diff-12918a0e2886ac9526cde1534cc4ca9fR78) instead of `SharedRecipients` and pass the recipients to the Toolbar.

### DIFF
--- a/src/drive/ducks/files/Toolbar.jsx
+++ b/src/drive/ducks/files/Toolbar.jsx
@@ -5,6 +5,7 @@ import React, { Component } from 'react'
 import { withRouter } from 'react-router'
 import { connect } from 'react-redux'
 import classNames from 'classnames'
+import { showModal } from 'react-cozy-helpers'
 
 import { ROOT_DIR_ID } from '../../constants/config'
 import { translate } from 'cozy-ui/react/I18n'
@@ -32,16 +33,7 @@ import {
 
 const { BarRight } = cozy.bar
 
-const toggleShowShareModal = state => ({
-  ...state,
-  showShareModal: !state.showShareModal
-})
-
 class Toolbar extends Component {
-  state = {
-    showShareModal: false
-  }
-
   render() {
     const cozyDev = cozy.client._url === 'http://cozy.tools:8080'
     const cozyRecette = cozy.client._url === 'https://recette.cozy.works'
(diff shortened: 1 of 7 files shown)


goldoraf:
[(3) Afficher l'indicateur de partage dans le menu "more" sur mobile](https://trello.com/c/H4Qhtx2E/1023-3-afficher-lindicateur-de-partage-dans-le-menu-more-sur-mobile)

src/drive/ducks/files/Toolbar.jsx
```diff
@@ -221,6 +215,19 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
...
+        <ShareModal
+          document={displayedFolder}
+          documentType="Files"
+          sharingDesc={displayedFolder.name}
+        />
+      ),
+      meta: {
+        hideActionMenu: true
+      }
+    }),
```
y-lohse: Nice improvement 👍

src/drive/styles/actionmenu.styl
```diff
@@ -48,7 +48,7 @@
     background embedurl("../assets/icons/icon-share-grey.svg") 1rem center no-repeat
     position   relative
     
-    > div
+    [class*='menu-recipients']
```
y-lohse: I think this sort of selector is dangerous. It's too easy to add a class that contains `menu-recipients` somewhere and accidentally have these rules applied.
Since we use it in 2 different places, I think it's worth having a dedicated class name.

src/sharing/components/Recipient.jsx
```diff
@@ -10,24 +10,28 @@ import { getDisplayName, getPrimaryEmail, getPrimaryCozy } from '..'
...
-    )}
-    {recipients
-      .slice(0, MAX_DISPLAYED_RECIPIENTS)
-      .map(recipient => (
-        <Avatar name={getDisplayName(recipient)} size={size} />
-      ))}
-  </div>
-)
+export const RecipientsAvatars = ({ recipients, link, size, className }) => {
+  // we slice first to clone the original array because reverse() mutates it
```
y-lohse: Aaargh good catch.
That being said, the comment mentioning `reveser()` only makes sense in the context of this PR. Maybe a better comment for the future would be to explain why we're reversing the array?

IIRC it's so that the recipients you've added last show up in the avatars.

PR #1076, opened 2018-08-03
Added a 'permissions required' popin to the sharing modal

goldoraf:
Some users that have installed the application a long time ago didn't ask for contacts' perm when they authorized the app (it wasn't in the OAuth scope yet). This PR detects 403s on contacts fetch and displays a popin for the user to renew authorization with the good scope.

### DIFF
--- a/src/sharing/components/ShareByEmail.jsx
+++ b/src/sharing/components/ShareByEmail.jsx
@@ -1,4 +1,5 @@
 import React, { Component } from 'react'
+import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
@@ -7,6 +8,7 @@ import Alerter from 'cozy-ui/react/Alerter'
 import SelectBox, { components } from 'cozy-ui/react/SelectBox'
 import ShareAutosuggest from './ShareAutosuggest'
 import { getPrimaryEmail } from '..'
+import { renewAuthorization } from 'drive/mobile/actions/authorization'
 
 import styles from '../share.styl'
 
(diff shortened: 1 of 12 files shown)


src/sharing/components/ShareByEmail.jsx
```diff
@@ -298,4 +366,6 @@ ShareByEmail.defaultProps = {
   contacts: []
 }
 
-export default ShareByEmail
+export default connect(null, dispatch => ({
+  renewAuthorization: client => dispatch(renewAuthorization(client))
+}))(ShareByEmail)
```
y-lohse: Bah, I see why it's there but i's annoying to link this part of the code with the general redux store :(

src/sharing/share.styl
```diff
@@ -172,6 +173,39 @@ input[type=text]
...
+        right .9375rem
+        margin 0
+        padding 0
+        background-size .875rem
+        background-color transparent
+        cursor pointer
+        display block
+        width .875rem
+        height .875rem
+        z-index 1
```
y-lohse: We probably want a variable for the z-index here.

src/sharing/components/ShareByEmail.jsx
```diff
@@ -38,6 +42,32 @@ ShareRecipientsInput.defaultProps = {
...
 
+const RequestPermissionPopin = ({ onClose, onAccept }, { t }) => (
+  <div className={styles['permission-required-popin']}>
+    <Button
+      theme="close"
+      className={styles['permission-required-popin-close']}
+      onClick={onClose}
+      extension="narrow"
+    >
+      <Icon icon="cross" width="14" height="14" color="#95999d" />
```
y-lohse: We should use a color variable here.

PR #1109, opened 2018-08-16
fix: Prevent clicks outside the menu to trigger actions

y-lohse:


y-lohse:
[🐛 Fermeture du menu d'action](https://trello.com/c/mBh8xzU8/1100-%F0%9F%90%9B-fermeture-du-menu-daction)

PR #1110, opened 2018-08-16
feat: Instantly switch views when navigating

y-lohse:
Switching back to regular routing instead of preloading the content on the previous view.


y-lohse:
[(2) Améliorer la perception de la réactivité du changement d'onglet](https://trello.com/c/iSagkdch/1029-2-am%C3%A9liorer-la-perception-de-la-r%C3%A9activit%C3%A9-du-changement-donglet)

PR #1117, opened 2018-08-22
Redux store and persisted state clean reset

goldoraf:
Here I used a [Dan Abramov strategy](https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992) to reset the redux store. There is no need to handle the `UNLINK` action in each duck anymore.


goldoraf:
[(2) 📱🐛 Vider totalement les données de l'app lors d'une déconnexion](https://trello.com/c/AkXIGtns/1035-2-%F0%9F%93%B1%F0%9F%90%9B-vider-totalement-les-donn%C3%A9es-de-lapp-lors-dune-d%C3%A9connexion)

PR #1120, opened 2018-08-22
Added sharing recipients list in the delete confirm

goldoraf:
It was just a matter of replacing the use of `confirm` by `showModal` so that I could use a `SharedDocument` component to provide the necessary sharing context to the confirm modal. I also added a specific `RecipientList` because we needed something with less margins and without status menu.


goldoraf:
[(2) Ajouter les destinataires du partage sur la modale de suppression](https://trello.com/c/W5q2ZmM5/978-2-ajouter-les-destinataires-du-partage-sur-la-modale-de-suppression)

src/lib/react-cozy-helpers/ModalManager.jsx
```diff
@@ -19,7 +19,10 @@ export default reducer
 
 export const showModal = component => ({
   type: SHOW_MODAL,
-  component
+  component,
+  meta: {
+    hideActionMenu: true
+  }
```
y-lohse: I don't think this is the right place for the `hideActionMenu ` thing because it's not correlated with the action of showing a modal. Can we find another place for it ? I didn't really mind where it was before, even though it wasn't very readable.

goldoraf: I think it's correlated, because I don't see a case where we would want to not hide the action menu when showing a modal. Am I wrong?

y-lohse: I think there might be cases where the modal is shown and there is no action menu at all, for example when using the main sharing button in the toolbar?
But even if we want to do it every time, the `meta` feels a bit like a hack anyway, because we're expressing 2 things with one action. If we really want to always close action menus when modals are shown, we might as well use the `action.type` in the reducers. But using 2 separate actions feels cleaner to me?
And yes this was also the case before, but at least the `meta` part was in a fairly high level part of the code, and I'm afraid burying it in a lower level makes the problem worse.

Do you see what I mean or did I miss something?

goldoraf: As said during our DS: we'll handle this pb later, it's probably up to the action menu to hide itself when an action is clicked.

PR #1130, opened 2018-08-27
fix: Prev state didn't always have authorization prop

y-lohse:
Just a small possible bug I found later on.

### DIFF
--- a/src/drive/store/persistedState.js
+++ b/src/drive/store/persistedState.js
@@ -13,7 +13,7 @@ const migrateSettings = async prevState => {
     analytics,
     wifiOnly
   } = prevState.mobile.settings
-  const { revoked } = prevState.mobile.authorization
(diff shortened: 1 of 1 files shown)


src/drive/store/persistedState.js
```diff
@@ -13,7 +13,7 @@ const migrateSettings = async prevState => {
     analytics,
     wifiOnly
   } = prevState.mobile.settings
-  const { revoked } = prevState.mobile.authorization
+  const { revoked } = prevState.mobile.authorization || {}
```
enguerran: It is a smell to not have consistent value in a state. It is better to initialize the state with an empty value or `undefined` (aka a value that means no value in the type you use). It will be easier to anticipate possible values in the state, even with a non-hard-typed language.

PR #1132, opened 2018-08-29
Photos: title and menus in the cozy bar

y-lohse:
Don't merge until we have https://github.com/cozy/cozy-client/pull/116 and a new version of cozy-client!

- I moved the title of the pages to the cozy-bar on mobile view, that was relatively easy.
- I also moved all the menus toolbar there
- The toolbar menus now use cozy-ui menu's
- Once renderer inside the cozy-bar, the components get the cozy-bar contact, so any HOC that relies on context won't work. To workaround this, I had to pass the Toolbars instance of `t`, `router`, and move some sharing code around

### DIFF
--- a/src/photos/ducks/albums/components/AlbumsView.jsx
+++ b/src/photos/ducks/albums/components/AlbumsView.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import styles from '../../../styles/layout'
+import { translate } from 'cozy-ui/react/I18n'
 
 import AlbumsToolbar from './AlbumsToolbar'
 import AlbumsList from './AlbumsList'
@@ -20,8 +21,9 @@ const Content = ({ list }) => {
   }
 }
 
-export default class AlbumsView extends Component {
+class AlbumsView extends Component {
   render() {
+    const { t } = this.props
     if (this.props.children) return this.props.children
     if (!this.props.albums) {
       return null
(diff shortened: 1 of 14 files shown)


y-lohse:
[(1) 🐛 Afficher le nom de la page et le menu "more" sur Photos (mobile)](https://trello.com/c/2qfkvO7E/1062-1-%F0%9F%90%9B-afficher-le-nom-de-la-page-et-le-menu-more-sur-photos-mobile)

src/photos/ducks/albums/components/AlbumsView.jsx
```diff
@@ -20,19 +21,21 @@ const Content = ({ list }) => {
...
+  render({ t }) {
     if (this.props.children) return this.props.children
     if (!this.props.albums) {
       return null
     }
     return (
       <div className={styles['pho-content-wrapper']}>
         <Topbar viewName="albums">
-          <AlbumsToolbar />
+          <AlbumsToolbar t={t} />
```
goldoraf: Why pass `t` here and not using `translate` HOC for the `AlbumToolbar` too? Am I missing something?

src/photos/ducks/albums/components/AlbumsView.jsx
```diff
@@ -20,19 +21,21 @@ const Content = ({ list }) => {
   }
 }
 
-export default class AlbumsView extends Component {
-  render() {
+class AlbumsView extends Component {
+  render({ t }) {
```
goldoraf: This way of getting props being specific to Preact, we shouldn't add more but preferably remove what we may find, should we?

src/photos/ducks/albums/components/AlbumsView.jsx
y-lohse: Yeah but it's easy to miss: The `Topbar` may wrap the `AlbumsToolbar` inside the cozy-bar. If that happens, `t` injected via the HOC will have the bar's translation context, so we need to pass it an instance of `t` that is bound to the correct context.

PR #1141, opened 2018-09-05
Update offline file copy when the file's content change

goldoraf:
The relevant changes are [this function](https://github.com/cozy/cozy-drive/compare/master...goldoraf:realtime-offline-files-updates?expand=1#diff-5be0a43208edead225fc91b8c0b76048R102) that is used by `RealtimeFiles` [here](https://github.com/cozy/cozy-drive/compare/master...goldoraf:realtime-offline-files-updates?expand=1#diff-9d445e7faef1808a83ef82ea6ba523efR57), but I also took the opportunity to move offline related code into its own duck and to connect things to the store differently.


goldoraf:
[(2) Mise à jour des fichiers marqués hors-ligne lorsqu’ils changent](https://trello.com/c/OwasFdQL/1055-2-mise-%C3%A0-jour-des-fichiers-marqu%C3%A9s-hors-ligne-lorsquils-changent)

src/drive/mobile/modules/offline/duck.js
```diff
@@ -0,0 +1,112 @@
...
+  _,
+  getState
+) => {
+  if (
+    isAvailableOffline(getState(), file.id) &&
+    file.md5sum !== prevFile.md5sum
+  ) {
+    await saveOfflineFileCopy(file)
+  }
+}
```
y-lohse: It would have been helpful to move the file in one commit and do the edits in a different one — as things are I can't tell what has changed and what has just been moved 🤷🏼‍♂️

`updateOfflineFileCopyIfNecessary ` is really nice and simple. I don't like it too much that we have functions like `openLocalFileCopy ` in a `duck` file even though they have no interaction with redux.

src/drive/web/modules/actionmenu/MenuItem.jsx
```diff
@@ -1,30 +1,22 @@
...
-  checkbox: {
-    value: isAvailableOffline(state)(ownProps.files[0].id),
-    onChange: () => {}
+export default class MenuItem extends Component {
+  onClick = e => {
+    // we need to stop propagation here so that the menu doesn't close itself
+    e.stopPropagation()
+    if (this.props.onClick) {
+      this.props.onClick()
+    }
```
y-lohse: Linking this for reference, a bit late : https://github.com/cozy/cozy-ui/pull/571

Rather than calling `e.stopPropagation()` all the time and `onClose` most of the time, I think it'd be better to let the menu close itself by default all the time, and when we don't want to close the menu — in our case, only when using the offline toggle — we can call `e.preventDefault()`, and cozy-ui can take care of stopping the propagation.

Not blocking because it's ok if we do it later, but it'd be a nice chance :)

src/drive/web/modules/drive/Container.jsx
```diff
@@ -40,6 +41,20 @@ const ShareMenuItem = ({ docId, ...rest }, { t }) => (
...
+  }),
+  (dispatch, ownProps) => ({
+    onToggle: () => dispatch(toggleAvailableOffline(ownProps.file))
+  })
+)(({ checked, onToggle, children, ...rest }) => (
+  <MenuItem {...rest}>
+    {children}
+    <Toggle id={children} checked={checked} onToggle={onToggle} />
+  </MenuItem>
+))
```
y-lohse: 👍

src/drive/web/modules/navigation/RealtimeFiles.jsx
```diff
@@ -82,7 +85,7 @@ class RealtimeFiles extends React.Component {
   normalizeId = doc => ({ ...doc, id: doc._id })
 
   componentWillUnmount() {
-    if (this.subscription) this.subscription.unsubscribe()
+    if (this.realtimeListener) this.realtimeListener.unsubscribe()
```
y-lohse: Good catch

src/drive/mobile/modules/offline/duck.js
goldoraf: > It would have been helpful to move the file in one commit and do the edits in a different one 

Yes, sorry for that, I didn't think I'd end up with so many changes, I should have been more methodic.

I tried to put every "actions" relative to local file copies because the reducer is very simple right now but we could add a lot of more state (like for instance the status of a file download, aka is it really available?) and make the offline feature better by enhancing these actions with redux state.
This particular function could go in `lib/filesystem.js`, but I think there's more refactoring than just that to do in this part of the code. No religion here, just a tentative arrangement of functions.

src/drive/web/modules/actionmenu/MenuItem.jsx
goldoraf: If you think your PR can be merged soon, why not?

y-lohse: Because it doesn't actually do what I describe in the last message yet :D

src/drive/mobile/modules/offline/duck.js
y-lohse: Yup, I figured it'd be something like that. It's still an improvement over what it was before :)

src/drive/web/modules/actionmenu/MenuItem.jsx
goldoraf: Oh I see... Too bad then :smile:

PR #1142, opened 2018-09-05
Replaced / by > as the breadcrumb sep

goldoraf:


y-lohse:
Although that will mean even more rebasing for me :(

PR #1143, opened 2018-09-06
Multiple fixes on photos

y-lohse:
The commits are explicit about what they do. 

The last one is a bit strange, there's more context [here](https://github.com/cozy/cozy-client/issues/144).

### DIFF
--- a/src/photos/components/Topbar.jsx
+++ b/src/photos/components/Topbar.jsx
@@ -69,13 +69,25 @@ const TopbarTitle = ({ children }) => (
   <h2 className={styles['pho-content-title']}>{children}</h2>
 )
 
-class Topbar extends Component {
-  backToAlbums = () => {
-    // go to parent
-    let url = this.props.router.location.pathname
-    this.props.router.push(url.substring(0, url.lastIndexOf('/')))
-  }
+const BackToAlbumsButton = withRouter(({ router }) => {
+  const url = router.location.pathname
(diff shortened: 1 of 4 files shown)


src/photos/components/Topbar.jsx
```diff
@@ -87,22 +93,27 @@ class Topbar extends Component {
...
   render() {
-    const { children, viewName, breakpoints: { isMobile } } = this.props
+    const { children, viewName, breakpoints: { isMobile }, router } = this.props
     const isAlbumContent = viewName === 'albumContent'
     const title = <TopbarTitle>{this.renderTitle()}</TopbarTitle>
     const responsiveTitle = isMobile ? <BarCenter>{title}</BarCenter> : title
 
     const responsiveMenu = isMobile ? <BarRight>{children}</BarRight> : children
 
+    const url = router.location.pathname
```
goldoraf: This is not ideal to process `parentUrl` on each render (it's a very small optim I know). I guess we can't connect to the router from `BackToAlbumsButton` when we render the button in the mobile bar and thus in cozy-bar, am I right? But we may store `parentUrl` in an instance variable and process it only once at least. What do you think?

y-lohse: Yep, you're right — that's why it wasn't done, but we can definitely use an instance variable to store it. i'll amend the PR.

PR #1144, opened 2018-09-06
Added fetchMore on albums view

goldoraf:
Straightforward change...

### DIFF
--- a/src/photos/ducks/albums/components/AlbumsView.jsx
+++ b/src/photos/ducks/albums/components/AlbumsView.jsx
@@ -9,7 +9,7 @@ import ErrorComponent from 'components/Error/ErrorComponent'
 import Topbar from '../../../components/Topbar'
 
 const Content = ({ list }) => {
-  const { fetchStatus, data } = list
+  const { fetchStatus } = list
   switch (fetchStatus) {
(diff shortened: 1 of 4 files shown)


src/photos/ducks/albums/components/AlbumsView.jsx
```diff
@@ -9,15 +9,15 @@ import ErrorComponent from 'components/Error/ErrorComponent'
 import Topbar from '../../../components/Topbar'
 
 const Content = ({ list }) => {
-  const { fetchStatus, data } = list
+  const { fetchStatus } = list
```
y-lohse: What is the `fetchStatus` when we're `fetchingMore`? Does it change at all?

I'm asking because it'd be cool if we could avoid the coupling introduced [here](https://github.com/cozy/cozy-drive/pull/1144/files#diff-41bc1b57031edefa1a5a326588bccdbcR19), but without `fetchStatus` I'm not sure it's possible.

goldoraf: No, right now the `fetchStatus` doesn't change when loading more rows... I guess it'd be nice to have it.

PR #1148, opened 2018-09-10
docs: fix the command to run the app in watch mode

Crash--:


y-lohse:
Thanks! The change is good but would you mind changing your commit message so it's comity-zen compliant? You can fin our guidelines on this [here](https://github.com/cozy/cozy-guidelines#commit-messages) 😉

PR #1149, opened 2018-09-10
Sharing preview

y-lohse:
This PR is ready for review but will need a new version of cozy-client including [these changes](https://github.com/cozy/cozy-client/pull/150) before merging.

With this PR, sharings are created with a *preview* URL, from which the recipient can accept the sharing.
I've repurposed the page we use to display sharings by links because it's essentially the same UI. The main change is a new button in the header that can be used to accept the sharing.

I've also switched a menu to the cozy-ui version — but it's packed together with another commit, sorry about that :(


y-lohse:
[(3) En tant que destinataire d'un partage, je peux visualiser le contenu avant d'accepter le partage](https://trello.com/c/9vomUaDf/1026-3-en-tant-que-destinataire-dun-partage-je-peux-visualiser-le-contenu-avant-daccepter-le-partage)

src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -21,65 +27,140 @@ const DownloadFilesButton = ({ t, onDownload, size = 'normal' }) => (
...
+  </div>
+)
+
+class PublicToolbar extends React.Component {
+  state = {
+    discoveryLink: null
+  }
+
+  componentDidMount() {
+    if (window.location.pathname === '/preview') this.loadSharingDiscoveryLink()
```
Crash--: `this.props.router.location.pathname` ?

src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -21,65 +27,140 @@ const DownloadFilesButton = ({ t, onDownload, size = 'normal' }) => (
...
+      const { sharecode } = getQueryParameter()
+
+      const link = client
+        .collection('io.cozy.sharings')
+        .getDiscoveryLink(sharingId, sharecode)
+      this.setState({ discoveryLink: link })
+    } catch (err) {
+      console.warn('Failed to load sharing discovery link', err)
+    }
+  }
```
Crash--: Not a big fan of having these requests and manipulations inside a component. Is there an easy way to move it outside?

y-lohse: As we discussed, one way to do it might be to use redux for this. It's a bit of a pain though, as we'd have to add dedicated actions, reducers and a store section for this view, which isn't even in the main app. Given the extra code bloat and fragmentation that would introduce, I'm honestly not sure it's worth it.

We could also transform the function into a standalone function, it would take `client` as a parameter and return the link / `null`. The call would still be in the component, but at least the code itself would be a simple js function ?

PR #1157, opened 2018-09-11
fix: 🐛 Updating UI to v11 and fixing button bugs accordingly

GoOz:
Updating UI to v11, fixing the missing required `label` props everywhere and also fixing the rendering of action buttons in viewer selection bar (but the download action is still not working)

ping @y-lohse ^

~~⚠️ Needs cozy-bar to be [merged](https://github.com/cozy/cozy-bar/pull/229) & deployed to avoid some inconvenience.~~ ✅

### DIFF
--- a/src/viewer/ViewerControls.jsx
+++ b/src/viewer/ViewerControls.jsx
@@ -3,6 +3,7 @@ import classNames from 'classnames'
 import Hammer from 'hammerjs'
 
 import { translate } from 'cozy-ui/react/I18n'
+import { Button } from 'cozy-ui/react/Button'
 import { downloadFile } from 'cozy-client'
 
 import styles from './styles'
@@ -96,14 +97,16 @@ class ViewerControls extends Component {
                 styles['pho-viewer-toolbar-actions']
               )}
             >
-              <button
+              <Button
+                theme="secondary"
                 className={styles['coz-action-download']}
                 onClick={() => {
                   downloadFile(currentFile)
                 }}
-              >
-                {t('Viewer.actions.download')}
-              </button>
+                icon="download"
+                label={t('Viewer.actions.download')}
+                subtle
+              />
             </div>
             {onClose && (
               <div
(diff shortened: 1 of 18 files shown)


y-lohse:
So on my end, the close button in the viewer is now hidden, see [here](
https://www.dropbox.com/s/v2256fdm881wiri/Capture%20d%27%C3%A9cran%202018-09-12%2016.44.45.png?dl=0).

For the download button, it can't be clicked right now because the `div.pho-viewer-imageviewer` is above `.pho-viewer-toolbar`, which contains the button. But I don't understand why, since the toolbar has a fixed position and a z-index..?

GoOz:
The hidden close button is due to the unmerged PR on Cozy-bar I talked about

As for the download button, I'll take a look.

GoOz:
Ok I figured it out. 

There's a CSS property `pointer-events: none;` on the wrapper of the download button. Its purpose is to make sure you can click on other toolbar, mainly the one from the PDF reader.
I could just add `pointer-events: initial;` to the buttons but every button from the PDF toolbar behind a button from our selection bar wouldn't be clickable anyway.

I'll see if there could be an UX answer to that issue.

src/viewer/ViewerControls.jsx
```diff
@@ -72,6 +73,9 @@ class ViewerControls extends Component {
       children
     } = this.props
     const { hidden } = this.state
+
+    const isPDF = currentFile.class === 'pdf' ? true : false
```
GoOz: Is there a better way to do this?

GoOz:
So we decided, for now, to hide the download button when there's a PDF displayed since the PDF toolbar already has its own download button.

GoOz:
It should be good to go now. Bar is updated and released.

y-lohse:
The UI is good, but the download button is still broken. It used a previous version of cozy-client and needs updating, posssibly using our redux actions.

GoOz:
Ok then I let you handle it, I don't think I would be able to do it myself.

y-lohse:
[Téléchargement depuis le viewer](https://trello.com/c/0nEmon4r/1242-t%C3%A9l%C3%A9chargement-depuis-le-viewer)

y-lohse:
Waiting for https://github.com/cozy/cozy-client/pull/196 now…

y-lohse:
cozy-client has been updated but we found more bugs, still a work in progress.

@GoOz we also need the latest version of the cozy-bar for this release, but it's still not clear to me why? I know it's something about `aria-hidden` and the icons, but could you point me to what exactly is going on please?

y-lohse:
Ok thanks, I understand. I guess the other problem is that it's a global declaration — does cozy-ui have many of those? I'm not sure it should add anything to the global css scope unless we explicitly say so, and even then probably just classes?

Crash--:
Close this one. See #1200

PR #1158, opened 2018-09-12
Pick a folder from another app

y-lohse:
This PR adds a new intent that allows another app to open a folder picker.

The main component is [here](https://github.com/cozy/cozy-drive/pull/1158/files#diff-e41dd17299fbb5c1542671b38f659ef8R1), but we wanted to make this work without using the URL for navigation, as a preparation for the "move to" feature. This required a bunch of other changes I've highlighted bellow.


src/drive/store/rootReducer.js
```diff
@@ -19,25 +19,25 @@ import {
...
+    availableOffline
+  }
+
+  const mobileReducer = combineReducers({
+    authorization,
+    settings,
+    replication,
+    mediaBackup
+  })
+
```
y-lohse: I had to move these — strangely, inside an intent, the `view` reducer was undefined at the time of import, but was perfectly fine inside the function.

src/drive/styles/topbar.styl
```diff
@@ -13,5 +13,9 @@ $coz-bar-size = 3rem
     flex 0 0 auto
 
 +small-screen()
-    .fil-topbar
+  .fil-topbar
+    margin-right .5rem
+    margin-left .5rem
+
+  .fil-topbar.hidden-mobile
```
y-lohse: Usually we hide the topbar on mobile, but inside this interapp modal we want to keep it displayed.

src/drive/web/modules/filelist/File.jsx
```diff
@@ -278,9 +277,7 @@ class File extends Component {
   open(e, attributes) {
     e.stopPropagation()
     if (isDirectory(attributes)) {
-      this.props.onFolderOpen(attributes.id).then(() => {
-        this.props.router.push(getFolderUrl(attributes.id, this.props.location))
-      })
+      this.props.onFolderOpen(attributes.id)
```
y-lohse: Since we don't want to always use the URL to navigate, the calls to router have been moved to a component higher up the chain.

src/drive/web/modules/filelist/FileList.jsx
```diff
@@ -7,11 +7,20 @@ import FileListBody from 'drive/web/modules/filelist/FileListBody'
 
 import styles from 'drive/styles/filelist'
 
-const FileList = ({ canSort, fileActions, ...rest }) => (
+const FileList = ({
+  canSort,
+  fileActions,
+  withSelectionCheckbox = true,
```
y-lohse: During a previous refactor, we stopped respecting the `withSelectionCheckbox ` prop — but not all view should have it.

src/drive/web/modules/filelist/FileListBody.jsx
```diff
@@ -53,7 +54,13 @@ export const FileListBody = ({
         else if (files.length === 0 && !isTypingNewFolderName)
           return <EmptyContent {...props} />
         else
-          return <FileListRows files={files} withSelectionCheckbox {...props} />
+          return (
+            <FileListRows
+              files={files}
+              withSelectionCheckbox={withSelectionCheckbox}
```
y-lohse: Same as before, re-adding support for the `withSelectionCheckbox ` prop.

src/drive/web/modules/navigation/Breadcrumb.jsx
```diff
@@ -1,6 +1,6 @@
 /* global __TARGET__ cozy */
 import React, { Component } from 'react'
-import { Link, withRouter } from 'react-router'
+import { withRouter } from 'react-router'
```
y-lohse: Same deal with the `Breadcrumb` as with the `File` component — we don't want to be tied to a URL navigation at all cost. This file now exposes "dumb" components that can be used without router, and components with a router attached, that work just like before.

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -54,16 +57,20 @@ class FileExplorer extends Component {
...
   render() {
-    const { children, ...props } = this.props
+    const { children, ...otherProps } = this.props
     return (
       <RealtimeFiles>
         <SharingProvider doctype="io.cozy.files" documentType="Files">
-          {React.cloneElement(React.Children.only(children), props)}
+          {React.cloneElement(React.Children.only(children), {
+            ...otherProps,
+            onFolderOpen: this.navigateToFolder
```
y-lohse: When we click on a folder, this is where the navigation is happening now.

src/drive/web/modules/public/LightFolderView.jsx
```diff
@@ -75,6 +80,7 @@ class DumbFolderView extends React.Component {
         <Content>
           <FileList
             onFileOpen={this.showInViewer}
+            onFolderOpen={this.navigateToFolder}
```
y-lohse: Similar change to `FileExplorer`, the routing is handled here.

src/drive/web/modules/services/components/Picker/Picker.jsx
```diff
@@ -0,0 +1,129 @@
...
+  getVisibleFiles
+} from 'drive/web/modules/navigation/duck'
+import styles from './picker.styl'
+
+class Picker extends Component {
+  state = {
+    path: []
+  }
+
+  updateBreadcrumb = folder => {
```
y-lohse: We're handling the breadcrumbs without a router here. I'm not super happy with the implentation, mostly because if we start the navigation somewhere else than in the root folder, it wouldn't really work. I think this is something we can refine when we work on the folder-moving.

src/drive/web/modules/services/components/Picker/Picker.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+  goBack = () => {
+    const { path } = this.state
+    this.navigateTo(path[path.length - 2])
+  }
+
+  componentDidMount() {
+    const root = document.getElementById('main')
+    const data = root.dataset
+    this.headerIcon = data.cozyIconPath
```
y-lohse: Another thing I'm not very happy about — but not sure where would be a good place to get this icon. Maybe passing it down as a prop from the `main.jsx` file?

y-lohse:
[(5) Sélectionner un dossier en inter-app](https://trello.com/c/HGIRn2ii/870-5-s%C3%A9lectionner-un-dossier-en-inter-app)

PR #1169, opened 2018-09-17
feat: Add debug build to android command

Crash--:
To build debug APK


Crash--:
just the extra command :)

PR #1175, opened 2018-09-19
Get app slug from stack

CPatchane:
Necessary to use the new cozy-bar navigation
and the home app detection


CPatchane:
@y-lohse it should be fixed 👍

Crash--:
👍

PR #1176, opened 2018-09-19
Websql migration

y-lohse:
Migrates our databases from websql to idb !

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
     "classnames": "2.2.6",
     "cozy-bar": "5.0.8",
     "cozy-client": "2.19.2",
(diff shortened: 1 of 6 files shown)


y-lohse:
[(3) 📱🐛 Erreur WebSQL sous iOS](https://trello.com/c/fZhVxxLK/954-3-%F0%9F%93%B1%F0%9F%90%9B-erreur-websql-sous-ios)

y-lohse:
We have access to the version number, but not if the user comes from a previous installation. I think checking the adapter that is actually used is our best option right now. We'll can revert this commit in a few releases once everyone has been updated.

PR #1179, opened 2018-09-24
Moving files and folder

y-lohse:
Super early PR, I'll comment things here as I go along. For now it's just a placeholder modal.

Requires this PR in cozy-client : https://github.com/cozy/cozy-client/pull/192
Require this PR in cozy-ui : https://github.com/cozy/cozy-ui/pull/621


src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+    const { entries, onClose } = this.props
+    const { client } = this.context
+    const { folderId } = this.state
+
+    const entry = entries[0]
+
+    try {
+      await client
+        .collection('io.cozy.files')
+        .updateFileMetadata(entry._id, { dir_id: folderId })
```
Crash--: Is there a way to use Cozy client (withMutation) instead of a direct call? https://github.com/cozy/cozy-client/blob/master/docs/guide.md#mutating-data

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+    const { client, t } = this.context
+    const { folderId } = this.state
+
+    const contentQuery = client =>
+      client
+        .find('io.cozy.files')
+        .where({ dir_id: folderId })
+        .sortBy({ name: 'asc' })
+
+    const breadcrumbQuery = client => client.get('io.cozy.files', folderId)
```
Crash--: Can we create the request constant somewhere else ? We don't need to recreate these request everytime we render the component, right ?

src/drive/web/modules/navigation/getFolderPath.js
```diff
@@ -0,0 +1,57 @@
...
+      // if we don't have one, we add it manually
+      path.unshift({
+        id: isBrowsingTrash ? TRASH_DIR_ID : ROOT_DIR_ID
+      })
+    }
+  }
+  return path
+}
+
+export default getFolderPath
```
Crash--: 👍 moving thing outside of `duck/reducer` is always a good idea :)

src/drive/web/modules/navigation/getFolderPath.spec.js
```diff
@@ -0,0 +1,216 @@
...
+          dir_id: '789'
+        }
+      }
+      const path = getFolderPath(curentFolder, location, true, ['456'])
+      expect(path.length).toEqual(2)
+      expect(path[0]).toEqual(curentFolder.parent)
+      expect(path[1]).toEqual(curentFolder)
+    })
+  })
+})
```
Crash--: yaeh baby! `expect('team_creates_tests').toEqual(true)`

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+                </div>
+              </div>
+            )
+          }}
+        </Query>
+        <div>
+          <Button theme="secondary" onClick={onClose}>
+            Annuler
+          </Button>
+          <Button onClick={this.moveEntries}>Déplacer</Button>
```
Crash--: Everytime we'll render the `MoveModal`, we'll trigger an update of this button because everytime we create an instance of `moveEntries`

If we had, in the constructor : `this.moveEntries = this.moveEntries.bind(this)`, we'll create an instance of this method only once.  And if the `Button` is smart enough, it will not be rendered several times :)

src/drive/web/modules/filelist/File.jsx
```diff
@@ -351,14 +351,15 @@ class File extends Component {
...
         )}
-        {actionMenuVisible && (
-          <ActionMenu
-            file={attributes}
-            reference={this.filerowMenuToggle}
-            actions={actions}
-            onClose={this.hideActionMenu}
-          />
-        )}
+        {actions &&
```
Crash--: We can have the ActionMenu displayed when we don't have any action? Or is it because sometimes this component is rendered before we have data ?

src/drive/web/modules/move/MoveModal.jsx
y-lohse: Yeah good point, I have to investigate that.

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+    <Topbar hideOnMobile={false}>
+      {path.length > 1 &&
+        isMobile && (
+          <PreviousButton onClick={() => navigateTo(path[path.length - 2])} />
+        )}
+      <Breadcrumb path={path} onBreadcrumbClick={navigateTo} opening={false} />
+      <div role="toolbar">add folder</div>
+    </Topbar>
+  )
+)
```
Crash--: Can this component have its specific file ? ;)

src/drive/web/modules/move/MoveModal.jsx
Crash--: Never mind, didn't see the arrow function :(

src/drive/web/modules/move/MoveModal.jsx
Crash--: Never mind, didn't see the arrow function :(

src/drive/web/modules/move/MoveModal.jsx
y-lohse: Definitely something I want to look into when cleaning! But they need to be recreated when `folderId` changes, which may or may not be the only time the component re-renders.

src/drive/web/modules/move/MoveModal.jsx
y-lohse: So our guidelines currently [recommend the opposite](https://github.com/cozy/cozy-guidelines#binding-event-handlers), but it's debatable. We can open an issue there, you're not the first one to raise this issue.

That being said, are you sure it will re-render the button? `this.moveEntries` is a reference to our function, and the reference doesn't change, even when doing a new render pass.

src/drive/web/modules/move/MoveModal.jsx
y-lohse: Absolutely yes. I'm planning on splitting things up further and into more files when I'm closer to finishing the PR.

src/drive/web/modules/filelist/File.jsx
y-lohse: There are views where files don't have any actions (moving view, interapp folder picking…), and in that case we don't want to show the button. It's the same for the [`ActionMenu`](https://github.com/cozy/cozy-drive/blob/b17e10a181c744eb7166018842b131853ae5cbcd/src/drive/web/modules/filelist/File.jsx#L354).

y-lohse:
I'm more or less done with this PR, I think a review can already be done.

What's missing:

- activating the move feature on all screens (recents, sharings, etc)
- putting it behind a flag in the URL
- instead of starting in the root folder, start wherever we are at the moment
- upgrade cozy-ui and cozy-client

src/drive/locales/en.json
```diff
@@ -534,6 +535,15 @@
       "noNetwork": "You're currently offline."
     }
   },
+  "Move": {
+    "to": "Move to:",
+    "action": "Move",
+    "cancel": "Cancel",
+    "success": "%{subject} has been moved to %{target}.",
+    "error": "Something went wrong while moving this item, please try again later.",
+    "cancelled": "%{subject} has been moved back to it's original location.",
```
Crash--: Can we display the original path?

src/drive/web/modules/drive/Container.jsx
```diff
@@ -26,6 +27,15 @@ import {
...
 import styles from 'drive/styles/actionmenu'
 
+const isMoveToActive = () => {
+  try {
+    const url = new URL(window.location)
+    return url.search.includes('logistics')
+  } catch (e) {
+    return false
+  }
+}
```
Crash--: Is it temporary? :)

src/drive/web/modules/filelist/File.jsx
```diff
@@ -103,6 +103,7 @@ const SelectBox = ({ withSelectionCheckbox, selected, onClick }) => (
 const FileName = ({
   attributes,
   isRenaming,
+  disabled,
```
Crash--: Hmm. A `FileName` can be `disabled` ? What does that mean? Can we find a better name?

src/drive/web/modules/move/MoveExplorer.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import fileListStyles from 'drive/styles/filelist'
+import FileListHeader, {
+  MobileFileListHeader
+} from 'drive/web/modules/filelist/FileListHeader'
+
+const MoveExplorer = ({ children }) => (
+  <div className={fileListStyles['fil-content-table']} role="table">
```
Crash--: Accessibility 👍

Crash--:
Do we need cozy-ui@12 ?

src/drive/web/modules/drive/Container.jsx
y-lohse: Yes!

src/drive/web/modules/filelist/File.jsx
y-lohse: Yeah you're right, that's not very clear, i'll rename it.

src/drive/locales/en.json
y-lohse: We could, but it requires another query to get the name of the original folder, since we only have the `id`. That said the original directory should already be loaded in the cozy-client redux store.

I thought it was better to avoid the extra query and just display a message like that. When we start supporting multiple files, we probably won't be able to show the original path either, because there might be more than one.
That said we can change the behavior if you think it's important to have the original path, it's not a big change to make!

PR #1181, opened 2018-09-25
fix: Disabling sharing by links UI

y-lohse:
We weren't always setting the `loading` state to false. While at it, I switched from `then/catch` to `async/await`.


y-lohse:
[(0.5) 🐛 Impossible de désactiver certains partages par lien](https://trello.com/c/UeEtnyGa/990-05-%F0%9F%90%9B-impossible-de-d%C3%A9sactiver-certains-partages-par-lien)

PR #1182, opened 2018-09-27
Fix the docker command in README

Gara64:


y-lohse:
Argh sorry, we're going to need a commit message that conforms to semantic commit messages. Maybe `fix: Docker command in the Readme` should do the trick!

Gara64:
Looks good now!

y-lohse:
Perfect!

PR #1183, opened 2018-09-27
fix: Provide full context to bar components

y-lohse:
This fixes our problem of components not showing up in the cozy-bar — more context can be found here : https://github.com/cozy/cozy-bar/issues/144#issuecomment-425138507

### DIFF
--- a/src/drive/web/modules/drive/Toolbar.jsx
+++ b/src/drive/web/modules/drive/Toolbar.jsx
@@ -1,15 +1,16 @@
 /* global cozy */
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
+import SharingProvider, { SharedDocument } from 'sharing'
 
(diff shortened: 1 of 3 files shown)


y-lohse:
[(2j) Gestion des menus actions (more et selections)](https://trello.com/c/t1DQaY22/980-2j-gestion-des-menus-actions-more-et-selections)

src/drive/web/modules/drive/Toolbar.jsx
```diff
@@ -1,9 +1,11 @@
 /* global cozy */
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
+import SharingProvider from 'sharing'
```
Crash--: we already ` import { SharedDocument } from 'sharing' ` (l13)

PR #1188, opened 2018-10-01
Use name_prefix Cozy

CPatchane:
* Changes in manifest
* Changes in all .ejs files


CPatchane:
Yep, you need it to publish correctly to the registry

PR #1189, opened 2018-10-02
fix: We can now cancel the OAuth registration process. 

Crash--:
And in this case, cozy-client throws an error, we have to catch it and treat it as not an error ;)


Crash--:
[Erreur login - aller retour](https://trello.com/c/NZogwnmj/1269-erreur-login-aller-retour)

PR #1192, opened 2018-10-03
fix: ShareAutoSuggest only display 20 contacts

Crash--:


Crash--:
[(2) PARTAGE - dropdown de contacts](https://trello.com/c/DQjdjnXJ/1205-2-partage-dropdown-de-contacts)

PR #1193, opened 2018-10-04
fix: Force websql on safari in services

y-lohse:


targets/drive/web/services.jsx
```diff
@@ -34,13 +34,16 @@ document.addEventListener('DOMContentLoaded', () => {
     token: data.cozyToken
   })
 
+  const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
```
Crash--: const iOS =  /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;

(isSafari || iOS) ?

PR #1194, opened 2018-10-04
Force Index migration on mobile

y-lohse:
All indexes were not properly migrated from websql to idb, this PR fixe the problem.


y-lohse:
Aw, well let me know if you want explanations about some of that stuff!

PR #1199, opened 2018-10-05
Update u iv11

Crash--:
based on the work done in #1157

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,7 +162,7 @@
     "cozy-client": "2.19.2",
     "cozy-client-js": "0.13.0",
     "cozy-realtime": "1.1.0",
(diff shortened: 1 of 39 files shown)


package.json
```diff
@@ -198,7 +190,8 @@
     "redux-raven-middleware": "1.2.0",
     "redux-thunk": "2.3.0",
     "url-polyfill": "1.1.0",
-    "webpack-node-externals": "1.7.2"
+    "webpack-node-externals": "1.7.2",
+    "whatwg-fetch": "^3.0.0"
```
Crash--: Why on Earth this line is here? We already have it in master :o

PR #1206, opened 2018-10-08
fix: display initial on toolbar->button->svg

Crash--:


src/drive/web/modules/viewer/FilesViewer.jsx
```diff
@@ -15,10 +15,12 @@ const getParentPath = router => {
 
 class FilesViewer extends Component {
   state = {
-    currentFile: null
+    currentFile: null,
+    shoudLoadMore: true
```
y-lohse: This variable appears to be unused, do we really need it?

src/drive/web/modules/viewer/FilesViewer.jsx
```diff
@@ -49,15 +51,24 @@ class FilesViewer extends Component {
...
+  // If we get close of the last file fetched (-5), but we know there are more in the folder
+  // (it shouldn't happen in /recent), we fetch FILES_FETCH_LIMIT more files
+  async fetchMoreIfNecessary() {
+    const { files, params, location, fetchMoreFiles } = this.props
     const currentIndex = files.findIndex(f => f.id === params.fileId)
-    if (files.length !== fileCount && files.length - currentIndex <= 5) {
-      const folderId = getFolderIdFromRoute(location, params)
-      fetchMoreFiles(folderId, files.length, FILES_FETCH_LIMIT)
+    const folderId = getFolderIdFromRoute(location, params)
+    // As fetchMoreIfNecessary is called every time, we had to check if we need to
```
y-lohse: `every time` is not very clear here, Is it at every render, or every mount, or something else?

src/drive/web/modules/viewer/FilesViewer.jsx
```diff
@@ -49,15 +51,24 @@ class FilesViewer extends Component {
...
+    // fetch more files. To do so, we simply store the latest length of files loaded
+    if (
+      currentIndex - files.length - 5 > 0 &&
+      this.lastLengthFetched !== files.length
```
y-lohse: I don't understand how this works. Since we're setting `this.lastLengthFetched` to be `files.length`, `this.lastLengthFetched !== files.length` will always be false the second time, won't it?

Crash--:
Close this one. See #1207 for style issues. Will open a new PR for the file viewer bug

PR #1214, opened 2018-10-10
fix: Sharing album by link

Crash--:


Crash--:
[[Regression] Photos - Impossible de partager un album](https://trello.com/c/6mi3gwhH/1304-regression-photos-impossible-de-partager-un-album)

PR #1218, opened 2018-10-12
feat: upload corrupted files

Crash--:


src/drive/mobile/lib/media.js
```diff
@@ -39,7 +39,48 @@ export const requestAuthorization = async () => {
...
+          else if (result.progress !== undefined)
+            progressCallback(result.progress)
+          else if (result.thumbnail) thumbnailCallback(result.thumbnail)
+          else {
+            resolve(result)
+          }
+        },
+        reject
+      )
+    })
```
Crash--: needs some mutualisation here ;)

src/drive/mobile/modules/mediaBackup/duck/tmp.js
```diff
@@ -0,0 +1,62 @@
...
+            )
+          )
+          return memo.push(photo.id)
+        }
+      })
+    }
+    return memo
+  }, [])
+  try {
+    await Promise.all(promises)
```
Crash--: @y-lohse I make uploading parallel. Do you think it can be an issue ?

y-lohse: Mmmh no I think it should be ok!

targets/drive/mobile/config.xml
```diff
@@ -101,8 +101,10 @@
     </plugin>
     <plugin name="cordova-plugin-x-socialsharing" spec="~5.4.1" />
     <plugin name="cordova-plugin-safariviewcontroller" spec="~1.5.4" />
-    <plugin name="cordova-ios-plugin-no-export-compliance" spec="git+https://github.com/Cordobo/cordova-plugin-ios-no-export-compliance.git" />
-    <plugin name="io.cozy.jsbackgroundservice" spec="https://github.com/cozy/cordova-jsbackgroundservice.git#1.3.0" />
-    <plugin name="io.cozy.plugins.listlibraryitems" spec="https://github.com/cozy/cordova-plugin-list-library-items.git#1.5.0" />
+    <plugin name="cordova-ios-plugin-no-export-compliance" spec="git+https://github.com/Cordobo/cordova-plugin-ios-no-export-compliance.git" />    <plugin name="io.cozy.jsbackgroundservice" spec="https://github.com/cozy/cordova-jsbackgroundservice.git#1.3.0" />
     <plugin name="com-darryncampbell-cordova-plugin-intent" spec="^1.1.1" />
+    <plugin name="cordova-plugin-browsersync" spec="~0.1.7" />
```
y-lohse: I don't think this is supposed to be part of this PR?

src/drive/mobile/modules/mediaBackup/duck/tmp.js
```diff
@@ -0,0 +1,63 @@
+/* global cozy */
+/* 
+This file is TEMPORARY. It covers a stack bug and we need to synchronize again some files
```
y-lohse: Maybe we can give the file or more explicit name than `tmp`? `checkCorruptedFiles` is actually pretty good!

src/drive/mobile/modules/mediaBackup/duck/tmp.js
```diff
@@ -0,0 +1,63 @@
...
+export const checkCorruptedFiles = async (
+  photosOnDevice,
+  alreadyUploaded,
+  client,
+  dispatch
+) => {
+  const onProgressUpdate = progress => {
+    dispatch({ type: CURRENT_UPLOAD_PROGRESS, progress })
+  }
+  const onThumbnailGenerated = thumbnailUrl => {}
```
y-lohse: We could declare these functions outside of `checkCorruptedFiles ` so they aren't recreated with every call.

src/drive/mobile/modules/mediaBackup/duck/tmp.js
```diff
@@ -0,0 +1,63 @@
...
+          try {
+            await updateLibraryItem(
+              value.file_doc,
+              photo,
+              throttle(onProgressUpdate, 500),
+              onThumbnailGenerated
+            )
+            const newObj = files.reduce((acc, v) => {
+              if (v.file_doc._id === value.file_doc._id) {
+                console.log('ici')
```
y-lohse: I don't think we need this one :)

src/drive/mobile/modules/mediaBackup/duck/tmp.js
```diff
@@ -0,0 +1,63 @@
...
+            await localforage.setItem('CORRUPTED_FILES', newObj)
+            return memo.push(photo.id)
+          } catch (error) {
+            console.log('error on CORRUPTED FILES', error)
+          }
+        }
+      })
+    }
+    return memo
+  }, [])
```
y-lohse: Maybe we can rewrite the code block above in a more functional style to make things clearer? There's a `reduce` inside a `some` inside a `reduce`, it makes thingsa bit hard to follow along.

Maybe something like:

```
files.filter(onlyFilesWithNames).map(getPhotoOnDevice).forEach(reuploadFile)
```

Do you see what I mean? If not I think at least we should rename some of the variables like `ids`, `value` and `v` so that they express their purpose :)

Crash--: Come on! Nothing more functional than `reduce`. I really don't like chaining, so I'll juste rename some vars as advised :)

src/drive/mobile/modules/mediaBackup/duck/tmp.js
Crash--: `checkCorruptedFiles` is only call when there is a sync. Declaring `onProgressUpdate` outside means passing dispatch outside... Can do it but does it worth it? :D

targets/drive/mobile/config.xml
Crash--: So painful that Cordova 7 reads the package.json to update this one

src/drive/mobile/modules/mediaBackup/duck/tmp.js
y-lohse: Oh right, I missed that `dispatch` is in the context. Probably not worth the bother then.

src/drive/mobile/modules/mediaBackup/duck/tmp.js
y-lohse: Well it's a `reduce` call but there are definitely side effects :p. I also think we can use `forEach` instead of `some` if we leave the overall structure as it is, since we're not doing anything with the return value of `some`.

Crash--: Why using `forEach`? We can't break the loop with `forEach`. With `some`, when we return something, it breaks the loop, so I don't iterate over all files again and again.

PR #1220, opened 2018-10-12
Updated test config

y-lohse:
Creating this PR so we don't forgot about this. Adding tests is currently a pain in the ass, because importing almost any component imports a lot of things and jest can't handle everything. This PR reworks the config until it works, but it's not finished yet.

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -2,9 +2,14 @@
   "roots": ["<rootDir>/src"],
   "moduleFileExtensions": ["js", "jsx", "json", "styl"],
   "moduleNameMapper": {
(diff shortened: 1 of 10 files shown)


nono:
@y-lohse just a friendly reminder that you have this WIP pull request

y-lohse:
Yeah thanks! I'm afraid it will stay like this for a while though, not sure when I (or anyone) will get to work on this again.

y-lohse:
Ok here it is — this new test config should allow us to import pretty much any file in cozy-drive in the test-context without everything crashing horribly.

Crash--:
Cool !

PR #1222, opened 2018-10-16
Lots of UI fixes

y-lohse:
Probably easiest to review commit by commit, they are very atomic.

We should wait a bit before merging this, as the recently introduced `body-scroll-lock` in the modals seem to be faulty on some browser (chrome responsive, but still investigating).

### DIFF
--- a/src/sharing/components/Recipient.jsx
+++ b/src/sharing/components/Recipient.jsx
@@ -14,12 +14,27 @@ import { getDisplayName, getPrimaryEmail, getPrimaryCozy } from '..'
 
 const MAX_DISPLAYED_RECIPIENTS = 3
 
-export const RecipientsAvatars = ({ recipients, link, size, className }) => {
+export const RecipientsAvatars = ({
+  recipients,
+  link,
+  size,
+  className,
+  onClick
+}) => {
   // we reverse the recipients array because we use `flex-direction: row-reverse` to display them correctly
   // we slice first to clone the original array because reverse() mutates it
   const reversedRecipients = recipients.slice().reverse()
   return (
-    <div className={classNames(styles['recipients-avatars'], className)}>
+    <div
+      className={classNames(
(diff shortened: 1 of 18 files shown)


src/sharing/components/ShareByEmail.jsx
```diff
@@ -110,7 +112,7 @@ const ShareTypeSelect = ({ options, onChange }) => (
       classNamePrefix="needsclick react-select"
       components={{ DropdownIndicator, Option }}
       styles={customStyles}
-      defaultValue={options[1]}
+      defaultValue={options[0]}
```
Crash--: Should we use the 'value' of the option? Instead of the index ?

src/sharing/components/Recipient.jsx
```diff
@@ -27,12 +33,17 @@ export const RecipientsAvatars = ({ recipients, link, size, className }) => {
             .slice(MAX_DISPLAYED_RECIPIENTS)
             .map(recipient => getDisplayName(recipient))}
           size={size}
+          onClick={onClick}
```
Crash--: Really don't like having to pass down this props everywhere

y-lohse:
[(0.5) [Partage] UI marouflage 2](https://trello.com/c/bh2mPBxZ/1212-05-partage-ui-marouflage-2)

src/sharing/components/ShareByEmail.jsx
y-lohse: Mmh no I don't think so, `react-select` can handle an object as a value!

src/sharing/components/Recipient.jsx
y-lohse: The `onClick` no longer goes down to the avatars, it stays one level above. It's still a bit of drilling, but I'm not sure it makes sense to keep it anywhere above that?

PR #1224, opened 2018-10-16
Websql migration corrections

y-lohse:
When we're telling PouchDB to use the `sqlite` adapter on iOS, it starts using it right away. The documents are eventually replicated, but the indexes are missing and that's why we're seeing an error.

We now check for missing indexes, and recreate them if they are missing. The good news is it fixes the problem in all cases, even people having only done a partial migration. the problem is that the app will still show an error message the first time it is loaded, and will need a view change or restart to get the new indexes. But I don't see a better way to handle the situation :/


y-lohse:
[iOS - 1.10.2 erreur fichiers - sqlite vs. idb](https://trello.com/c/26RrQxP8/1314-ios-1102-erreur-fichiers-sqlite-vs-idb)

Crash--:
@y-lohse lint issues

PR #1225, opened 2018-10-17
fix(ModalFileList): Firefox overflow bug in modal

y-lohse:
Tagging @GoOz just in case it's of interest to you: in firefox, the file list in a modal was taking up all it's height instead of being scrollable. This change fixes the problem, and I haven't seen any other consequences on desktop or mobile, in Chrome, Firefox or Safari.

### DIFF
--- a/src/drive/styles/filelist.styl
+++ b/src/drive/styles/filelist.styl
@@ -3,6 +3,7 @@
 
 .fil-content-table
     @extend $table
+    overflow auto
 
 .fil-content-head
     @extend $table-head


GoOz:
Funny, this looks pretty similar from https://trello.com/c/PxOedeMo/1316-folder-infinite-scroll-gooz?menu=filter&filter=*

The thing is `overflow: auto` doesn't fix the bug of this card but `overflow: hidden` does. Those two solutions won't work together.

I wonder why all of a sudden File list gets broken on every browser.

y-lohse:
[(8) Déplacer un fichier ou un dossier](https://trello.com/c/OSfXpP1q/847-8-d%C3%A9placer-un-fichier-ou-un-dossier)

PR #1227, opened 2018-10-18
fix(PublicLink): Translate button label

y-lohse:
The button is rendered in the cozy bar, so it still doesn't have the right context.
See https://github.com/cozy/cozy-bar/issues/144 for discussions on a long term solution.


y-lohse:
[Partage - label manquant](https://trello.com/c/nMF4CeHN/1311-partage-label-manquant)

PR #1228, opened 2018-10-18
iOS hotfix release

y-lohse:


Crash--:
@y-lohse Do we still need it?

PR #1231, opened 2018-10-19
Activate sentry on the web

y-lohse:


Crash--:
It seems I've some troubles with this PR : 

{"error":"Two different projects were specified"}
capture d ecran 2018-11-02 a 10 56 27

y-lohse:
When does this happen?

Crash--:
When an error happens locally

PR #1233, opened 2018-10-19
fix(ActionMenu): Re-align recipients on mobile

y-lohse:


y-lohse:
[(3) PARTAGE - Problème affichage dropdown](https://trello.com/c/G8O8m1fY/1194-3-partage-probl%C3%A8me-affichage-dropdown)

PR #1236, opened 2018-10-23
Move multiple files and folders

y-lohse:
Still requires a new version of cozy-ui before we can merge it, but the rest should be good to go!


src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -33,52 +35,65 @@ class MoveModal extends React.Component {
...
     const { folderId } = this.state
 
-    const entry = entries[0]
-    const currentDirId = entry.dir_id
-
     try {
+      this.setState({ isMoveInProgress: true })
+      await Promise.all(
+        entries.map(entry => this.moveEntry(entry._id, folderId))
+      )
```
Crash--: Just to say something : I would have created a `moveEntries` function.

PR #1239, opened 2018-10-29
feat: Use references to track upload dir

y-lohse:
The main change [is here](https://github.com/cozy/cozy-drive/pull/1239/files#diff-f4a3d99ad5cf87a040b080caf7bb1d0eR175) — the function that gets the dir id has been overhauled and can be found [here](https://github.com/cozy/cozy-drive/pull/1239/files#diff-f4a3d99ad5cf87a040b080caf7bb1d0eR94). All the other changes are helper functions and signature changes.


y-lohse:
[(2) DOSSIER - Dossiers magiques "Backuped from my mobile" / "Backed up from my mobile"](https://trello.com/c/gn5xZ726/1063-2-dossier-dossiers-magiques-backuped-from-my-mobile-backed-up-from-my-mobile)

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -21,7 +21,10 @@ import {
   CURRENT_UPLOAD_PROGRESS
 } from './reducer'
 import { checkCorruptedFiles } from './checkCorruptedFiles'
+
 const ERROR_CODE_TOO_LARGE = 413
+const REFERENCE_PHOTOS_DIR = 'io.cozy.apps/photos'
+const REFERENCE_UPLOAD_DIR = 'io.cozy.apps/photos/mobile'
```
Crash--: We now have `doctypes.js`

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+      sort: [{ created_at: 'desc' }]
+    }
+  )
+
+  return referencedFolders.filter(
+    folder => /^\/\.cozy_trash/.test(folder.attributes.path) === false
+  )
+}
+
+const addFolderReference = (folderId, reference) =>
```
Crash--: should be `async` and fetchJSON awaited ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+
+const ensureFolderWithReferenceExists = async (path, reference) => {
+  const referencedFolders = await getFoldersWithReference(reference)
+
+  if (referencedFolders.length) {
+    return referencedFolders[0]
+  } else {
+    const dir = await cozy.client.files.createDirectoryByPath(path, false)
+    await addFolderReference(dir._id, reference)
+    return dir
```
Crash--: try / catch ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+
+const ensureHasCreatedAtIndex = () =>
+  cozy.client.fetchJSON('POST', '/data/io.cozy.files/_index', {
+    index: {
+      fields: ['created_at']
+    }
+  })
+
+const getUploadDir = async () => {
+  await ensureHasCreatedAtIndex()
```
Crash--: Why do we need to check if the `created_at` index is created ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+
+  const uploadedFolders = await getFoldersWithReference(REFERENCE_UPLOAD_DIR)
+
+  if (uploadedFolders.length >= 1) {
+    return uploadedFolders[0]
```
Crash--: How can we have more than one upload folder? Taking the first one is a bit arbitrary ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+      data: [
+        {
+          type: reference,
+          id: reference
+        }
+      ]
+    }
+  )
+
+const ensureFolderWithReferenceExists = async (path, reference) => {
```
Crash--: `getOrCreateFolder` ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+    const mediaFolderName = t('mobile.settings.media_backup.media_folder')
+    const uploadFolderName = t('mobile.settings.media_backup.backup_folder')
+    const legacyUploadFolderName = t(
+      'mobile.settings.media_backup.legacy_backup_folder'
+    )
+
+    await ensureFolderWithReferenceExists(
+      `/${mediaFolderName}`,
+      REFERENCE_PHOTOS_DIR
+    )
```
Crash--: Why this one is not in the `try` block ?

src/drive/mobile/modules/mediaBackup/duck/index.js
```diff
@@ -34,9 +37,95 @@ const currentMediaUpload = (media, uploadCounter, totalUpload) => ({
...
+
+    try {
+      const legacyFolder = await cozy.client.files.statByPath(
+        `/${mediaFolderName}/${legacyUploadFolderName}`
+      )
+      await addFolderReference(legacyFolder._id, REFERENCE_UPLOAD_DIR)
+      return legacyFolder
+    } catch (err) {
+      // the legacy folder doesn't exist, so we create the new one
+      const uploadFolder = await ensureFolderWithReferenceExists(
```
Crash--: can we check the error type before creating the folder ?

src/drive/mobile/modules/mediaBackup/duck/index.js
y-lohse: What would we do with a caught exception here? I don't think we can recover from an error here, so I feel like the best place to catch the error is in the main backup function where there [already is a try/catch block](https://github.com/y-lohse/cozy-drive/blob/magic-folders/src/drive/mobile/modules/mediaBackup/duck/index.js#L160).

src/drive/mobile/modules/mediaBackup/duck/index.js
y-lohse: Because in `getFoldersWithReference`, we sort the files by `created_at`, and if there is no index beforehand Couch can't resolve the request.

src/drive/mobile/modules/mediaBackup/duck/index.js
y-lohse: It can happen if you delete a folder that is referenced, and then restore it from the trash.
Since the results are ordered by `created_at`, the first one is the folder that we used most recently.

src/drive/mobile/modules/mediaBackup/duck/index.js
y-lohse: It is, [in the function above](https://github.com/y-lohse/cozy-drive/blob/magic-folders/src/drive/mobile/modules/mediaBackup/duck/index.js#L160) ;)

Unless you see something useful we could do here in case we can't create the `Photos` folder?

src/drive/mobile/modules/mediaBackup/duck/index.js
Crash--: Can you put that in comments ? :) This is interesting !

PR #1240, opened 2018-10-29
chore: Update matomo URL

y-lohse:
The variables are still called `piwik`, but I don't think renaming them matters.


y-lohse:
Nope, we're not shipping any.

PR #1244, opened 2018-10-31
fix: Stop synchro params

Crash--:
When switching to `False`, we `cancelMediaBackup` and force settings to be set to `false` by adding a new param to the function. 

This is because of https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/modules/mediaBackup/duck/index.js#L266


Crash--:
[(0.5j) APP synchro - backup non souhaité](https://trello.com/c/dbyD6Tk3/1317-05j-app-synchro-backup-non-souhait%C3%A9)

PR #1245, opened 2018-11-02
refacto: Use same configs as cozy-scripts. 

Crash--:
- Use same eslint config as cozy-scripts 
- Use same babel config as cozy-scripts

- update our codebase to follow our new config. Had to add some `eslint-disable-next-line` :( 
- fix one bug (now visible in the console) 

=> First step to migrate to cozy-scripts and react 

Next actions: 
 - [ ] migrate our webpack config


Crash--:
[(1) Utiliser les configs communes](https://trello.com/c/nJeC267v/1299-1-utiliser-les-configs-communes)

package.json
```diff
@@ -149,15 +123,20 @@
     "xml2js": "0.4.19"
   },
   "dependencies": {
+    "babel-loader": "^8.0.4",
```
Crash--: need to check if it's still needed

Crash--: Toujours besoin tant qu'on n'utilise pas les config webpack de cozy-scripts

package.json
```diff
@@ -149,15 +125,20 @@
...
     "cozy-client": "3.0.2",
     "cozy-client-js": "0.13.0",
     "cozy-realtime": "1.1.0",
+    "cozy-scripts": "^1.1.1",
     "cozy-ui": "14.2.0",
     "create-react-context": "0.2.2",
     "date-fns": "1.29.0",
     "diacritics": "1.3.0",
+    "eslint-config-cozy-app": "1.1.4",
+    "eslint-config-prettier": "3.1.0",
```
y-lohse: I think these should be devDependencies?

src/drive/mobile/lib/tracker.js
```diff
@@ -51,7 +51,9 @@ export const startHeartBeat = () => {
   if (trackerInstance) {
     try {
       trackerInstance.push(['enableHeartBeatTimer', mobileHeartBeatDelay])
-    } catch (err) {}
+    } catch (err) {
+      //console.log(err)
```
y-lohse: Why the commented out logs? Is it so the variable gets used?

Crash--: No empty blocks...

I added these lines for several reasons : 
- I think we should really take care of our errors and how we manage them. 
- I have to add several `disable case assignation` because I don't know really what to do with these cases and if we need to remove it or not

PR #1256, opened 2018-11-06
feat: Active moving files and folders

y-lohse:


y-lohse:
[(0.5) Publications - Drive/Photos](https://trello.com/c/Ff3vICFp/1355-05-publications-drive-photos)

PR #1260, opened 2018-11-07
Add empty service for photo

Gara64:
This allows to build a service for photos that does nothing yet. 
The build will generate a onPhotoUpload.js file that will be run separately as a node process


package.json
```diff
@@ -6,8 +6,9 @@
     "build:drive": "npm run build:drive:browser",
     "build:drive:browser": "npm run internal:build -- --env.app=drive --env.target=browser --env.production",
     "build:drive:mobile": "npm run internal:build -- --env.app=drive --env.target=mobile --env.production",
-    "build:photos": "npm run build:photos:browser",
+    "build:photos": "npm run build:photos:browser && npm run build:photos:services",
```
Crash--: `npm-run-all build:photos:browser build:photos:services`

and if we need/want , you can use `--parallel`

Gara64: Ok, I actually copy-pasted from cozy-banks :)
Is there a reason we wouldn't want `--parallel`?

Crash--: For your use case, nope

config/webpack.target.services.js
```diff
@@ -0,0 +1,30 @@
...
+
+const TARGET_DIR = path.resolve(__dirname, '../targets/')
+
+module.exports = function(production, app) {
+  var entry = {
+    onPhotoUpload: path.resolve(TARGET_DIR, './photos/services/onPhotoUpload')
+  }
+  var plugins = [
+    new DefinePlugin({
+      __TARGET__: JSON.stringify('services')
```
y-lohse: Will we really need this? We use the `__TARGET__` var to differentiate between web and mobile, but for a service script I'm not sure it will be needed :)

Gara64: Right, I removed it

PR #1262, opened 2018-11-07
Public album view on mobile

y-lohse:
On mobile, on this view, we display a custom header, but it was broken.


y-lohse:
[PARTAGE - album photo par lien - spinner infini](https://trello.com/c/LB6m638y/1391-partage-album-photo-par-lien-spinner-infini)

PR #1265, opened 2018-11-08
fix(PublicAlbum): Better desktop view

y-lohse:
After fixing the mobile view I accidentally made the desktop view slightly worse :/


targets/photos/web/public/index.styl
```diff
@@ -6,7 +6,7 @@
     display flex
     flex-direction column
     width 100%
-    padding-top 1.875rem
+    padding-top 3rem
```
Crash--: u-pt ?

PR #1267, opened 2018-11-09
 fix: Don't use color var in stylus func

GoOz:
Some cleaning in preparation for a future [release of Cozy-UI](https://github.com/cozy/cozy-ui/pull/641) with color variables as CSS custom properties.

~I couldn't test it because I could make Photos works.~ It works fine :)

~⚠️ Meanwhile if your Photos apps works well, you should consider testing it locally before merging.~

### DIFF
--- a/src/viewer/styles.styl
+++ b/src/viewer/styles.styl
@@ -43,7 +43,7 @@
             width      40%
 
         .pho-viewer-toolbar
-            background linear-gradient(to bottom, alpha(charcoalGrey, .50), alpha(charcoalGrey, .33), alpha(charcoalGrey, 0))
+            background linear-gradient(to bottom, rgba(50, 54, 63, .5), rgba(50, 54, 63, .33), rgba(50, 54, 63, 0))
 
         .pho-viewer-canceled
             height      100%


GoOz:
[Utiliser custom properties sur UI pour la personnalisation](https://trello.com/c/oHPjhtmr/336-utiliser-custom-properties-sur-ui-pour-la-personnalisation)

y-lohse:
Why shouldn't we use the variable in this context?

GoOz:
@y-lohse Because it won't work soon. Future Stylus colors vars will look like this `dodgerBlue: var(--dodgerBlue)`.
Stylus don't treat CSS custom properties as variable so it won't get the hexa color attached to it and so it won't be able to use any buitin color function such as `alpha()`, `darken()` or even `rgba()`.
And we will need those custom properties for brand customisation pruposes.

Apart from that, in this case, the color used in the viewer toolbar background doesn't matter much, it just need to be a dark grey. It doesn't have to be linked to the palette afaik.
I guess you can create your own Stylus var on Photos to replace what I did but so far it would only be used there, so i don't see any benefit.

But if you have some other solution, I'm open to discussion :)

GoOz:
Can it be merged?

y-lohse:
Yes, normally the one who creates the PR merges it :)

GoOz:
@y-lohse oh yeah sorry… every team works differently and I'm having a hard time adjusting myself 😕

edas:
Shouldn't we immediatly give them the 3 CSS variable they need instead of converting to color code now and dd CSS variables later ?

PR #1268, opened 2018-11-09
fix: Force bar to be in public mode

y-lohse:
It works on some views because the bar automatically goes into public mode when the URL contains `/public`… but not on all the views.


y-lohse:
[(0.5) PARTAGE - vue publique en vue mobile - menu principal](https://trello.com/c/GJEJdBgJ/1353-05-partage-vue-publique-en-vue-mobile-menu-principal)

PR #1270, opened 2018-11-12
feat: Expose preact as react

y-lohse:
The cozy-bar is now using react by default, but we want it to use preact — for that we have to expose preact under the "react" name in the global scope.


y-lohse:
[(0.5) PARTAGE - vue publique en vue mobile - menu principal](https://trello.com/c/GJEJdBgJ/1353-05-partage-vue-publique-en-vue-mobile-menu-principal)

PR #1271, opened 2018-11-13
WIP Serve chunks correctly before the bar

CPatchane:
Manually include the chunks instead of keeping HtmlWebpackPlugin put them at the end of the `<head>` tag


CPatchane:
No need this changes anymore with https://github.com/cozy/cozy-bar/pull/344

PR #1273, opened 2018-11-14
feat: inApp login with safariviewcontroller

Crash--:


src/drive/mobile/lib/cozy-helper.js
```diff
@@ -19,7 +19,7 @@ export const initClient = url => {
       'io.cozy.jobs:POST:sendmail:worker'
     ],
     oauth: {
-      redirectURI: 'http://localhost',
+      redirectURI: 'cozydrive://auth',
```
y-lohse: Don't we need some new code to handle this deeplink?

Crash--: No need :
- to handle deeplink. It’s already done somewhere ;) 
- migrating to cozy-auth will come soon. But as it needs more work and as is only a technical task I prefer to do it in a new PR.

PR #1274, opened 2018-11-14
Initialize clustering

Gara64:
Note that nothing will be active, as I haven't defined the trigger in the manifest yet

### DIFF
--- a/src/photos/ducks/clustering/settings.js
+++ b/src/photos/ducks/clustering/settings.js
@@ -1,0 +1,35 @@
+import { cozyClient, log } from 'cozy-konnector-libs'
+import { DOCTYPE_PHOTOS_SETTINGS } from '../../../drive/lib/doctypes'
+import {
+  SETTING_TYPE,
+  DEFAULT_SETTING,
+  DEFAULT_PERIOD,
+  DEFAULT_MODE
+} from './consts'
+
+export const createDefaultSetting = async () => {
+  log('info', 'Create default setting')
+  return cozyClient.data.create(DOCTYPE_PHOTOS_SETTINGS, DEFAULT_SETTING)
+}
+
+export const readSetting = async () => {
+  const settings = await cozyClient.data.findAll(DOCTYPE_PHOTOS_SETTINGS)
+  return settings.find(doc => doc.type === SETTING_TYPE)
+}
+
+export const defaultParameters = setting => {
+  const param = setting.parameters.filter(param => {
+    return param.period === DEFAULT_PERIOD
+  })
+  const dParam =
+    param.length > 0
+      ? param[0].modes.find(mode => mode.name === DEFAULT_MODE)
+      : null
+  return dParam
+    ? {
+        eps: dParam.eps,
+        epsTemporal: dParam.eps_temporal,
+        epsSpatial: dParam.eps_spatial
+      }
+    : dParam
+}
--- a/src/photos/ducks/clustering/settings.js
+++ b/src/photos/ducks/clustering/settings.js
@@ -1,5 +1,5 @@
 import { cozyClient, log } from 'cozy-konnector-libs'
-import { DOCTYPE_PHOTOS_SETTINGS } from '../../../drive/lib/doctypes'
+import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
 import {
   SETTING_TYPE,
   DEFAULT_SETTING,
@@ -7,7 +7,7 @@ import {
   DEFAULT_MODE
 } from './consts'
 
-export const createDefaultSetting = async () => {
+export const createDefaultSetting = () => {
   log('info', 'Create default setting')
   return cozyClient.data.create(DOCTYPE_PHOTOS_SETTINGS, DEFAULT_SETTING)
 }
@@ -21,15 +21,15 @@ export const defaultParameters = setting => {
   const param = setting.parameters.filter(param => {
     return param.period === DEFAULT_PERIOD
   })
-  const dParam =
+  const defaultParam =
     param.length > 0
       ? param[0].modes.find(mode => mode.name === DEFAULT_MODE)
       : null
-  return dParam
+  return defaultParam
     ? {
-        eps: dParam.eps,
-        epsTemporal: dParam.eps_temporal,
-        epsSpatial: dParam.eps_spatial
+        eps: defaultParam.eps,
+        epsTemporal: defaultParam.eps_temporal,
+        epsSpatial: defaultParam.eps_spatial
       }
-    : dParam
+    : defaultParam
 }
--- a/targets/photos/services/onPhotoUpload.js
+++ b/targets/photos/services/onPhotoUpload.js
@@ -1,5 +1,76 @@
+import { log, cozyClient } from 'cozy-konnector-libs'
+
+import { DOCTYPE_FILES } from '../../../src/drive/lib/doctypes'
+import {
+  readSetting,
+  defaultParameters,
+  createDefaultSetting
+} from '../../../src/photos/ducks/clustering/settings'
+
+// Returns the photos metadata sorted by date
+const extractInfo = photos => {
+  let info = photos
+    .map(file => {
+      const photo = {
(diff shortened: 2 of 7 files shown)


src/photos/ducks/clustering/settings.js
```diff
@@ -0,0 +1,35 @@
+import { cozyClient, log } from 'cozy-konnector-libs'
```
Crash--: (this question is certainly addressed to @y-lohse :D) 

Why `cozy-konnector-libs` is exposing `cozyClient` ? Is it cozy-client or cozy-client-js ? We need it from there because `services` is executed in the same way as a connector ?

targets/photos/services/onPhotoUpload.js
```diff
@@ -1,5 +1,76 @@
...
+
+  log('info', `Get changes on files since ${lastSeq}`)
+  const result = await cozyClient.fetchJSON(
+    'GET',
+    `/data/${DOCTYPE_FILES}/_changes?include_docs=true&since=${lastSeq}`
+  )
+
+  return result.results.map(res => res.doc).filter(doc => {
+    return (
+      doc.class === 'image' && doc._id.indexOf('_design') !== 0 && !doc.trashed
```
Crash--: FYI: now we can use a beautiful and long awaited [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Syntax) method instead of `indexOf`

Gara64: Ah! Glad to get rid of this ugly `indexOf`. Thanks!

src/photos/ducks/clustering/settings.js
Gara64: Yes it is, the services are run in the same way as the connectors, so we use `cozy-konnector-libs` to have a `cozy-client-js` exposed and initialized.

src/photos/ducks/clustering/settings.js
```diff
@@ -0,0 +1,35 @@
...
+import { DOCTYPE_PHOTOS_SETTINGS } from '../../../drive/lib/doctypes'
```
y-lohse:
```suggestion
import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
```

src/photos/ducks/clustering/settings.js
```diff
@@ -0,0 +1,35 @@
...
+import {
+  SETTING_TYPE,
+  DEFAULT_SETTING,
+  DEFAULT_PERIOD,
+  DEFAULT_MODE
+} from './consts'
+
+export const createDefaultSetting = async () => {
```
y-lohse: There's no `await` here so we can ditch the `async` ;)

```suggestion
export const createDefaultSetting = () => {
```

src/photos/ducks/clustering/settings.js
```diff
@@ -0,0 +1,35 @@
...
+export const readSetting = async () => {
+  const settings = await cozyClient.data.findAll(DOCTYPE_PHOTOS_SETTINGS)
+  return settings.find(doc => doc.type === SETTING_TYPE)
+}
+
+export const defaultParameters = setting => {
+  const param = setting.parameters.filter(param => {
+    return param.period === DEFAULT_PERIOD
+  })
+  const dParam =
```
y-lohse: What does `dparam` stand for? Can we give it a more explicit name?

targets/photos/services/onPhotoUpload.js
```diff
@@ -1,5 +1,76 @@
+import { log, cozyClient } from 'cozy-konnector-libs'
+
+import { DOCTYPE_FILES } from '../../../src/drive/lib/doctypes'
```
y-lohse:
```suggestion
import { DOCTYPE_FILES } from 'drive/lib/doctypes'
```

targets/photos/services/onPhotoUpload.js
```diff
@@ -1,5 +1,76 @@
...
+import {
+  readSetting,
+  defaultParameters,
+  createDefaultSetting
+} from '../../../src/photos/ducks/clustering/settings'
```
y-lohse:
```suggestion
} from 'photos/ducks/clustering/settings'
```

src/photos/ducks/clustering/settings.js
y-lohse: That being said, the only thing happening here with cozy-client-js is a fetch, which is definitely supported by `cozy-client` too. If we can find a good way to initialize cozy-client inside a service…

targets/photos/services/onPhotoUpload.js
```diff
@@ -1,5 +1,76 @@
...
+
+// Returns the photos metadata sorted by date
+const extractInfo = photos => {
+  let info = photos
```
y-lohse: You've been using `const` elsewhere, is there a reason to use `let` here?

targets/photos/services/onPhotoUpload.js
```diff
@@ -1,5 +1,76 @@
...
+        id: file._id,
+        name: file.name
+      }
+      if (file.metadata) {
+        photo.date = file.metadata.datetime
+        photo.gps = file.metadata.gps
+      } else {
+        photo.date = file.created_at
+      }
+      const hours = new Date(photo.date.slice(0, 19)).getTime() / 1000 / 3600
```
y-lohse: So we want a timestamp expressed in hours?

Also, why do we have to extract a part of the `photo.date` string? What else is there in it?

targets/photos/services/onPhotoUpload.js
Gara64: The clustering will reorganize the elements, therefore it is not a `const`

targets/photos/services/onPhotoUpload.js
Gara64: Yes we use the hours as the unit for our temporal metric, as it easier to understand use than milliseconds in our case. 
For the `photo.date`, it is useful after the clustering, we need to save the period of a newly created album (dates of first and last photos), so we can use the ones already saved in the file.

src/photos/ducks/clustering/settings.js
Gara64: Agreed,  if there is an easy way to use `cozy-client` on the service side, I'm up to use it

targets/photos/services/onPhotoUpload.js
y-lohse: Ok, thanks for confirming the hour-timestamp, it's not usual but it makes sense!

For the slice, my main concern is that `0, 19` looks very arbitrary and without looking at one of these dates, it's hard to figure out what is happening. What happens if we don't slice the string at all? Can we still use it in `new Date()`?

targets/photos/services/onPhotoUpload.js
y-lohse: Ah ok, I see what you mean.

So it's ok for me, but note that `const` in javascript doesn't *actually* mean that the variable will be immutable (I know, I know 🤷🏼‍♂️). `const` only means you can't re-assign a new value to that variable, but changing attributes of an object will still work. Since `info` is never re-assigned in this context, I think we should use `const` here, to stay consistent with the rest of the codebase.

targets/photos/services/onPhotoUpload.js
Gara64: The date format generated by the stack is like `2018-11-15T00:07:19.519852762+01:00`. So the `0,19` slice was to keep the precision to the second. However, I admit this is not very useful, and more importantly, I realize the `+01:00` might be important in case of different timezone. So I'll remove the slice :)

targets/photos/services/onPhotoUpload.js
Gara64: Ah yes the infamous `const` being not const :D 
Ok for using `const` here then.

src/photos/ducks/clustering/settings.js
Gara64: It stands for `default param`, I'll make it clearer

PR #1276, opened 2018-11-14
feat: Add context definition on component

kosssi:


Crash--:
Let me close this one @kosssi ;). We have something else on track

PR #1277, opened 2018-11-14
fix(Photos): Configurable load more button label

y-lohse:


y-lohse:
[(1) Pagination des albums](https://trello.com/c/Ah0aKivw/1138-1-pagination-des-albums)

PR #1278, opened 2018-11-14
fix: Selection scroll mobile

Crash--:
Tested on Android & iOS seems to work as expected.


y-lohse:
Uuh ok :D And on the web? Has this line been made completely useless in the latest cozy-ui version @GoOz ?

Crash--:
Works also on small screen

PR #1280, opened 2018-11-19
refactor(Move): Rename components

y-lohse:
I renamed all the `MoveXXX` components to just `XXX` — they are all in a folder called `Move`, I think that's explicit enough. Meant to do it earlier but oh well.


Crash--:
Si ça ne tenait qu'à moi j'aurais des noms de fichiers super long et parlant 
 
`MoveModalHeader` and so on... I think I've do too much Java during my studies... (react team also :D https://github.com/facebook/react/tree/master/packages/react-dom/src/client) 

So I really don't know what to think about this one. If you prefer, then let's go.

y-lohse:
I mean generally I also prefer long names, but these components are exclusively used in the `Move` modal so I feel like the extra `Move` in the name is quite useless.

PR #1284, opened 2018-11-20
Add knn computation on temporal and spatial metrics

paultranvan:
The goal of this PR is to compute the `eps` parameter that will be used for the actual clustering. 

It computes the [KNN](https://en.wikipedia.org/wiki/K-nearest_neighbors_algorithm) algorithm on the dataset, giving a metric using k dimensions (1 for temporal and 2 for spatial). Then, we find the `eps` by extracting a significative slope on the sorted data.

### DIFF
--- a/src/photos/ducks/clustering/knn.js
+++ b/src/photos/ducks/clustering/knn.js
@@ -54,5 +54,6 @@ export default class KNN {
         return distances[indexSignSlope]
       }
     }
+    return null
   }
 }


Gara64:
@y-lohse About the short name variables: I agree this is to avoid most of the time, but I really don't know how to be explicit for mathematical operations. Like, how would you name the result of `Math.pow(Math.sin(x / 2), 2)`? I commited some changes to make it more explicit when it's possible, though.

For the maths libs, there are some, but I didn't want to import a full size one just to compute a mean or a standard deviation :)

PR #1286, opened 2018-11-21
fix: Sort albums by creation date

y-lohse:


y-lohse:
[Liste albums - tri par date](https://trello.com/c/sYJra0HD/1444-liste-albums-tri-par-date)

PR #1297, opened 2018-11-26
refacto: Use cozy-device-helper

Crash--:
2 things in this PR: 
- Fix Standalone mode with the redirection to localhost when we're on the Web Platform
- Remove `device` helper and use `cozy-device-helper`


src/viewer/Viewer.jsx
```diff
@@ -143,7 +144,7 @@ export default class Viewer extends Component {
       case 'video':
         return isMobile() ? NoViewer : VideoViewer
```
Crash--: I don't know if `isMobile()` is the same as `isMobileApp()` since we only test the `window.navigator`. 

Maybe we should rename it to `isMobileBrowser()` ?

y-lohse: Yeah no I think it's the same and this one should be replaced as well :)

y-lohse:
```suggestion
        return isMobileApp() ? NoViewer : VideoViewer
```

PR #1299, opened 2018-11-27
fix: Load dev config only for dev

Crash--:


config/webpack.config.dev.js
```diff
@@ -8,6 +8,14 @@ const pkg = require(path.resolve(__dirname, '../package.json'))
 module.exports = {
   devtool: 'cheap-module-source-map',
   externals: ['cozy'],
+  module: {
+    rules: [
+      {
+        test: require.resolve('cozy-bar/dist/cozy-bar.js'),
```
CPatchane: With that, when you will build cozy-bar it will need `cozy-bar/dist/cozy-bar.js` which is not available when you link the cozy-bar and run `yarn watch:mobile`

config/webpack.target.browser.js
```diff
@@ -32,7 +35,10 @@ module.exports = function(production, app) {
   if (
     fs.existsSync(path.resolve(__dirname, `../targets/${app}/web/services.jsx`))
   ) {
-    entry.services = [path.resolve(__dirname, 'expose-react.js'), path.resolve(__dirname, `../targets/${app}/web/services`)]
+    entry.services = [
+      path.resolve(__dirname, 'expose-react.js'),
```
CPatchane: I think that we don't need React for services

y-lohse: We shouldn't, but we do — somehow components that use `BarLeft` etc are imported on the services view.

Crash--:
It's clearly a revert of #1296. Just kept some lines and mainly https://github.com/cozy/cozy-drive/pull/1296/files#diff-1f2d88e45e2d2a2de34056266bb8b318R10

PR #1300, opened 2018-11-27
feat(authentication): Add contextTypes

drazik:
React requires `contextTypes` to be defined in order to use `context` in a component. That's not the case with Preact, so we didn't define it on every components. But now we encounter some `this.context.something is undefined` errors when we migrate our apps to React.


Crash--:
We already added `contextType` https://github.com/cozy/cozy-drive/pull/1287/files 

Did you installed this lib from the `@next` tag ?

drazik:
Didn't see that there is a `next` tag. It works with it. Thanks.

PR #1303, opened 2018-11-27
Feat/new app login

Crash--:
- Upgrade cozy-* (UI & Bar) 
- Add some native plugins 
 - keyboard
 - startApp
- New login template for onBoarding


src/authentication/src/steps/SelectServer.jsx
```diff
@@ -23,6 +27,15 @@ export class SelectServer extends Component {
   }
 
   componentDidMount() {
+    //if the cordova plugin is here, then shrink the view on iOS
+    if (window.Keyboard && window.Keyboard.shrinkView) {
+      window.Keyboard.shrinkView(true)
+    } else {
+      console.warn(
+        'Cozy-Authentication needs a cozy/cordova-plugin-keyboard plugin to better works.'
```
y-lohse:
```suggestion
        'Cozy-Authentication needs a cozy/cordova-plugin-keyboard plugin to work better.'
```

src/authentication/src/steps/SelectServer.jsx
```diff
@@ -60,6 +61,22 @@ export class SelectServer extends Component {
   }
 
   onChange(value) {
+    if (value.includes('www.') && (value.match(/\./g) || []).length === 1) {
```
y-lohse: This line is hard to decipher, I'm still not sure what we're testing for?

Crash--: If an user puts a dot, it's mainly because this user has a custom domain (since we can't have any dot in our cozy url). But we also accept the user writes `www.ylohse` and in this case we want to use our `mycozy.cloud` domain.

Crash--:
Reste à faire : 

- Attendre le merge de https://github.com/cozy/cozy-ui/pull/737 
- Gérer l'input quand on sélectionne "custom domain"

PR #1305, opened 2018-11-27
Save clustering results

Gara64:
Note there is no test yet as we discovered some bugs impacting jest that need to be fixed first through their own PR.

### DIFF
--- a/src/photos/ducks/clustering/albums.js
+++ b/src/photos/ducks/clustering/albums.js
@@ -25,19 +25,16 @@ const createReferences = async (album, photos) => {
 }
 
 const createAutoAlbum = async (albums, photos) => {
-  if (!photos || photos.length < 1) {
-    return null
-  }
   // Check if an album already exists for these photos. If not, create it
   const name = albumName(photos)
   const album = albums ? albums.find(album => album.name === name) : null
-  if (!album) {
+  if (album) return album
+  else {
     const created_at = new Date()
     const period = albumPeriod(photos)
     const album = { name, created_at, auto: true, period }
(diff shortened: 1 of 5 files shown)


src/photos/ducks/clustering/albums.js
```diff
@@ -0,0 +1,63 @@
...
+export const findAllAutoAlbums = async () => {
+  const albums = await cozyClient.data.findAll(DOCTYPE_ALBUMS)
+  return albums.filter(album => album.auto)
+}
+
+// TODO: deal with updates
+export const saveClustering = async clusters => {
+  const albums = await findAllAutoAlbums()
+  for (const photos of clusters) {
+    const album = await createAutoAlbum(albums, photos)
```
y-lohse: How about not calling `createAutoAlbum` at all if there are no photos? This would save us a check there and the `if (album)` check here.

src/photos/ducks/clustering/albums.js
```diff
@@ -0,0 +1,63 @@
...
+  // Check if an album already exists for these photos. If not, create it
+  const name = albumName(photos)
+  const album = albums ? albums.find(album => album.name === name) : null
+  if (!album) {
+    const created_at = new Date()
+    const period = albumPeriod(photos)
+    const album = { name, created_at, auto: true, period }
+    return await cozyClient.data.create(DOCTYPE_ALBUMS, album)
+  }
+  return album
```
y-lohse: I think this would be more readable with something like:

```
 const album = albums ? albums.find(album => album.name === name) : null
if (album) return album
else {
  //...create the new album
}
```

PR #1308, opened 2018-11-28
fix: Defensive code for sharing instance checking

y-lohse:


y-lohse:
[(0.5) PARTAGE - non affichés dans la vue "SHARINGS" - cas spécifique Claire](https://trello.com/c/IOXPAsov/1220-05-partage-non-affich%C3%A9s-dans-la-vue-sharings-cas-sp%C3%A9cifique-claire)

src/sharing/state.js
```diff
@@ -65,7 +65,8 @@ export const revokeSharingLink = permissions => ({
 })
 export const receivePaths = paths => ({ type: RECEIVE_PATHS, paths })
 export const matchingInstanceName = (instanceUri = '') => shareMember =>
-  shareMember.instance.toLowerCase() === instanceUri.toLowerCase()
+  shareMember.instance &&
```
Crash--: So a shareMember can not have an instance? What does that really mean?

y-lohse: Yeah, when tehy haven't accepted the sharing yet, ie. `pending` status.

PR #1313, opened 2018-12-03
Update clustering for new photos

Gara64:
Each time a set of new photos are uploaded, we have to check theirs dates to determine how the clusters must be updated.
We have 4 cases for each photo : 

* The photo is more recent than the newest cluster.
* The photo is older than the oldest cluster.
* The photo is inside a current cluster.
* The photo is between 2 clusters.

 For each case, we take the relevant cluster(s) and recompute it by retrieving all its photos.

### DIFF
--- a/src/photos/ducks/clustering/albums.js
+++ b/src/photos/ducks/clustering/albums.js
@@ -49,12 +49,11 @@ export const findAutoAlbums = async () => {
   ])
   const results = await cozyClient.data.query(autoAlbums, {
     selector: { auto: true },
-    sort: [{ name: 'desc' }]
+    sort: [{ name: 'asc' }]
   })
   return results
 }
 
-// TODO: deal with updates
 export const saveClustering = async (clusters, albums) => {
   for (const photos of clusters) {
     if (photos && photos.length > 0) {
@@ -67,7 +66,7 @@ export const saveClustering = async (clusters, albums) => {
 const findPhotosByAlbum = async album => {
   album._type = DOCTYPE_ALBUMS
   const files = await cozyClient.data.fetchReferencedFiles(album, {})
-  if (files && files.included.length > 0) {
+  if (files && files.included) {
     const attributes = files.included.map(file => {
       const attributes = file.attributes
       attributes.id = file.id
@@ -99,6 +98,7 @@ export const albumsToClusterize = async (photos, albums) => {
   const clusterize = {}
 
   for (const photo of photos) {
+    // Find clusters matching this photo
     const matchingAlbums = matchingClusters(photo, albums)
     if (matchingAlbums.length > 0) {
       const key = matchingAlbums[1]
@@ -107,6 +107,7 @@ export const albumsToClusterize = async (photos, albums) => {
       if (clusterize[key]) {
         clusterize[key].push(photo)
       } else {
+        // Save the photos referenced by the matchig albums
         const photosToClusterize = await findPhotosToClusterize(matchingAlbums)
         photosToClusterize.push(photo)
         clusterize[key] = photosToClusterize
--- a/src/photos/ducks/clustering/albums.js
+++ b/src/photos/ducks/clustering/albums.js
@@ -1,5 +1,7 @@
 import { cozyClient, log } from 'cozy-konnector-libs'
 import { DOCTYPE_ALBUMS } from 'drive/lib/doctypes'
+import { matchingClusters } from './matching'
+import flatten from 'lodash/flatten'
 
 // An auto album name is the date of the first photo
 const albumName = photos => {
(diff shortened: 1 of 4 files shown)


Gara64:

src/photos/ducks/clustering/albums.js
```diff
@@ -37,24 +42,113 @@ const createAutoAlbum = async (albums, photos) => {
...
-          `${photos.length} photos clustered into: ${JSON.stringify(album)}`
-        )
+      const album = await createAutoAlbum(photos, albums)
+      await createReferences(photos, album)
+    }
+  }
+}
+
+const findPhotosByAlbum = async album => {
+  album._type = DOCTYPE_ALBUMS
```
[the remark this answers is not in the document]
Gara64: The documents returned [here](https://github.com/cozy/cozy-drive/blob/e8452b3175079760342bb447d225bf10e09161dc/src/photos/ducks/clustering/albums.js#L50) do not include the `type`, I assumed it's a behaviour from `cozy-client-js`, although I don't understand why.

src/photos/ducks/clustering/albums.js
```diff
@@ -37,24 +42,113 @@ const createAutoAlbum = async (albums, photos) => {
...
+      const attributes = file.attributes
+      attributes.id = file.id
+      return attributes
+    })
+    return extractPhotosInfo(attributes)
+  }
+  return []
+}
+
+const findPhotosToClusterize = async albums => {
```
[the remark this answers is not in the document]
Gara64: Yeah I had a hard time for the naming :/
In this case, we are trying to find the photos already clusterized for these albums, because at least one new photo (not clusterized) match this cluster by date ; so we need to recompute the whole cluster.
Maybe `findPhotosToReclusterize` would be better?

src/photos/ducks/clustering/albums.js
```diff
@@ -37,24 +42,113 @@ const createAutoAlbum = async (albums, photos) => {
...
+        photo.datetime = file.metadata.datetime
+        photo.gps = file.metadata.gps
+      } else {
+        photo.datetime = file.created_at
+      }
+      const hours = new Date(photo.datetime).getTime() / 1000 / 3600
+      photo.timestamp = hours
+      return photo
+    })
+    .sort((pa, pb) => pa.timestamp < pb.timestamp)
```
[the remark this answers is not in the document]
Gara64: The sort is indeed important. Maybe a name like `prepareDataset` including the sort would be better?

src/photos/ducks/clustering/matching.js
```diff
@@ -0,0 +1,49 @@
...
+  const olderEnd = olderAlbum ? new Date(olderAlbum.period.end) : null
+  // Photo inside cluster
+  if (photoDate <= newerEnd && photoDate >= newerStart) {
+    return [newerAlbum]
+  }
+  // Photo between clusters
+  if (olderEnd && photoDate < newerStart && photoDate > olderEnd) {
+    return [newerAlbum, olderAlbum]
+  }
+  return []
```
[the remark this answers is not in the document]
Gara64: I see what you mean, ok for splitting in 2 functions

src/photos/ducks/clustering/matching.js
```diff
@@ -40,9 +48,10 @@ export const matchingClusters = (photo, albums) => {
     return [edge]
   }
   for (let i = 0; i < albums.length; i++) {
-    const matches = adjacentToClusters(photo, albums[i], albums[i + 1])
-    if (matches.length > 0) {
-      return matches
+    if (photoInsideCluster(photo, albums[i])) {
+      return [albums[i]]
+    } else if (photoBetweenClusters(photo, albums[i], albums[i + 1])) {
```
y-lohse: Yep, I think this is better! Now we could even *not* call `photoBetweenClusters ` when there is no `albums[i + 1]` — that way the `photoBetweenClusters` function doesn't have to check it's parameter.

src/photos/ducks/clustering/albums.js
```diff
@@ -37,24 +43,97 @@ const createAutoAlbum = async (albums, photos) => {
...
+export const saveClustering = async (clusters, albums) => {
   for (const photos of clusters) {
     if (photos && photos.length > 0) {
-      const album = await createAutoAlbum(albums, photos)
-      const refs = await createReferences(album, photos)
-      if (refs) {
-        log(
-          'info',
-          `${photos.length} photos clustered into: ${JSON.stringify(album)}`
+      const album = await createAutoAlbum(photos, albums)
```
Crash--: `createAutoAlbum` alors que par moment ça ne créé pas l'album, c'est pas top. 

getOrCreateAlbum ?

src/photos/ducks/clustering/albums.js
```diff
@@ -37,24 +43,97 @@ const createAutoAlbum = async (albums, photos) => {
...
+      const key = matchingAlbums[1]
+        ? matchingAlbums[0]._id + ':' + matchingAlbums[1]._id
+        : matchingAlbums[0]._id
+      if (clusterize[key]) {
+        clusterize[key].push(newPhoto)
+      } else {
+        // Save the photos referenced by the matchig albums
+        const photosToClusterize = await findPhotosToReclusterize(
+          matchingAlbums
         )
```
Crash--: Il semblerait qu'on ne fasse jamais (dans sa pile d'appel) de try / catch sur le `findPhotosToRelusterize`?

Gara64: Non en effet, tu penses que c'est nécessaire ?

Crash--: J'ai l'impression que cette fonction est quand même assez centrale. Maintenant le catch ne sera utile que si on sait quoi faire dans ce cas : 
- ajouter des logs pour avoir de la lecture si jamais ça pète un jour ? 
- ?

Gara64: Ok, j'ai ajouté un try/catch avec un log pour avoir l'info

PR #1318, opened 2018-12-05
fix: Disable offline for services

y-lohse:
Offline for services was originally enabled for offline-search. However the search doesn't use the offline database anymore. We have no immediate plans to migrate back to it, and in the meantime the offline part is problematic on safari mobile browser where it tends to fail.


y-lohse:
[Modale de PDF depuis banks sur safari mobile](https://trello.com/c/bId1Yvd4/1472-modale-de-pdf-depuis-banks-sur-safari-mobile)

PR #1319, opened 2018-12-05
feat: Cluster-based timeline

Gara64:


src/photos/ducks/timeline-clusters/index.js
```diff
@@ -18,6 +19,7 @@ const TIMELINE_QUERY = client =>
     .sortBy({
       'metadata.datetime': 'desc'
     })
+    .include(['albums'])
```
Crash--: If I remember correctly, we have a bug in `CozyClient` with the `include` keywords :  CozyClient generates too much requests instead of doing its optimizations to make only one or two requests. 

Is it something you saw?

Gara64:

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
+  photosNotClustered.forEach(photo => {
     const datetime =
-      p.metadata && p.metadata.datetime ? p.metadata.datetime : Date.now()
-    // here we want to get an object whose keys are months in a l10able format
-    // so we only keep the year and month part of the date
-    const month = datetime.slice(0, 7) + '-01T00:00'
-    /* istanbul ignore else */
-    if (!sections.hasOwnProperty(month)) {
-      sections[month] = []
+      photo.metadata && photo.metadata.datetime
```
y-lohse: To avoid this kind of conditions, we can use `get(photo, 'metadata.datetime', photo.created_at)` from lodash ;)

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
 })
 
-const getPhotosByMonth = photos => {
-  let sections = {}
-  photos.forEach(p => {
+const formatDate = date => {
+  return format(date, 'DD MMMM YYYY')
+}
+
+const sectionTitle = album => {
```
y-lohse: Can we prefix this function with a verb to be consistent with the other ones (`getPhotosByClusters`, `handle…` etc)? Like `getSectionTitle` maybe?

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
+  }
+  return formatDate(new Date(album.name))
+}
+
+/**
+ * A section matches if:
+ * - The datetime is inside the period.
+ * - The section has no period (not a cluster) and is the datetime's day.
+ */
+const matchingSection = (sections, datetime) => {
```
y-lohse: Same here, maybe `getMatchingSection`?

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
+      photo.metadata && photo.metadata.datetime
+        ? photo.metadata.datetime
+        : photo.created_at
+
+    const sectionDate = matchingSection(sections, new Date(datetime))
+    if (sectionDate) {
+      sections[sectionDate].photos.push(photo)
+    } else {
+      // Create a new section for this day, without a period, to differentiate from clusters' sections
+      const day = datetime.slice(0, 10)
```
y-lohse: This still looks odd to me, isn't there a `Date` method that formats the datetime as we want it?

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
+ * inside existing clusters, or group them per day otherwise.
+ */
+const getPhotosByClusters = photos => {
+  const sections = {}
+  const photosNotClustered = []
+
+  photos.forEach(p => {
+    const refAlbums = p.albums ? p.albums.data : null
+    // TODO Ensure unicity on the service side
+    const album = refAlbums ? refAlbums.find(ref => ref.auto) : null
```
y-lohse: We can do things like this as well:

```

const refAlbums = p.albums ? p.albums.data : []
const album = refAlbums.find(ref => ref.auto)
```

And while we're at it we can probably simplify it some more:

```

const clusterAlbum = p.albums ? p.albums.data.find(ref => ref.auto) : null
```

What do you think?

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
...
-  // are inserted on top of the list, and months can become unordered
-  const sortedMonths = Object.keys(sections)
-    .sort((a, b) => new Date(a).getTime() - new Date(b).getTime())
-    .reverse()
 
-  return sortedMonths.map(month => {
+  // We deal with the not-clustered photos after the loop to make sure all
+  // the sections have been processed.
+  if (photosNotClustered.length > 0) {
+    handlePhotosNotClustered(sections, photosNotClustered)
```
y-lohse: I think it's a bit dangerous to pass around the `sections` object, as it's a critical variable. 
For example in `handlePhotosNotClustered `, we add new sections, but those section have a `date` property that other sections don't have and they don't have a period. Having these slight differences scattered across multiple functions makes them hard to find, and changing the structure tricky. There's also the fact that `period` seems to be only used for the non-custered part of the code whereas `title` and `photos` are important afterwards.

 I'm not quite sure what the best approach is here — inlining `handlePhotosNotClustered` might be a good start though.

src/photos/ducks/timeline-clusters/index.js
Gara64: We could use the [`toLocaleDateString`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/toLocaleDateString) but we have to pass a locale. Or we could use the date-fns `format`, but I'm not sure how this would be better as we know the date format here is YYYY-MM-DD

src/photos/ducks/timeline-clusters/index.js
Gara64: I'm not a big fan of using this kind of syntaxic help from external libs, but I might be too old fashion ;)

src/photos/ducks/timeline-clusters/index.js
```diff
@@ -36,39 +38,122 @@ const TIMELINE_MUTATIONS = query => ({
     })
 })
 
-const getPhotosByMonth = photos => {
-  let sections = {}
-  photos.forEach(p => {
+const formatDate = date => {
+  return format(date, 'DD MMMM YYYY')
```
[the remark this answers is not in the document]
Gara64: Hm true, we could adapt to english, but I don't see how to make it generic for all language, though...

src/photos/ducks/timeline-clusters/index.js
Gara64: What is `p.albums.data` is `undefined`? Won't this fail with a `Cannot read property 'find' of undefined`?

src/photos/ducks/timeline-clusters/index.js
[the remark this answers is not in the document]
y-lohse: I think we should use the translation system for this :) That way we can specify `DD MMMM YYYY` in french, something else in english, etc.
I also think we need to use the `f` function provided by cozy-ui, and not directly `format` from date-fns, so the name of months and days are properly translated.

src/photos/ducks/timeline-clusters/index.js
y-lohse: Yeah no big deal, I just find it gets more readable when you have more than one test like that. But the code is fine as it is :)

src/photos/ducks/timeline-clusters/index.js
y-lohse: Do we really know this though? The data comes from `photo.metadata.datetime` or `photo.created_at` and hasn't been parsed. `created_at` is almost certainly a standard time string, but can `datetime` be something we don't expect?
I also think that with `const day = f(new Date(datetime), 'YYYY-MM-DD')`, it's really obvious what we're getting. With `.slice(0, 10)` the reader first has to figure out what the variable contains.
Another option would be to put this in a named function, — so it becomes something like `const day = getSectionDate(datetime)`?

src/photos/ducks/timeline-clusters/index.js
y-lohse: Ah shit yes, you're right. So maybe not the third version 😝

PR #1321, opened 2018-12-06
fix: Guard against missing indexes

y-lohse:


y-lohse:
[(0.5j) App mobile - offline](https://trello.com/c/IgZEyMLq/1459-05j-app-mobile-offline)

PR #1329, opened 2018-12-11
 feat: Handle clustering parameters 

Gara64:


src/photos/ducks/clustering/matching.js
```diff
@@ -58,3 +60,27 @@ export const matchingClusters = (photo, albums) => {
...
+ */
+export const matchingParameters = (parameters, photos) => {
+  // Take the average date in the photos to compare with parameters periods.
+  const datetime = averageDate(photos)
+
+  const lastParams = parameters[parameters.length - 1]
+  if (new Date(lastParams.period.end) <= datetime) {
+    // The date is newer than the last parameters
+    return lastParams
+  } else if (new Date(parameters[0].period.start) >= datetime) {
```
y-lohse: Why not a `firstParam` variable here? :)

src/photos/ducks/clustering/matching.js
```diff
@@ -58,3 +60,27 @@ export const matchingClusters = (photo, albums) => {
...
+    // The date is older than the first parameters
+    return parameters[0]
+  } else {
+    // The date is inside the parameters periods
+    return parameters.find(param => {
+      return new Date(param.period.end) >= new Date(datetime)
```
y-lohse: `datetime ` is already a `Date` object ;)

```suggestion
      return new Date(param.period.end) >= datetime
```

src/photos/ducks/clustering/settings.js
```diff
@@ -3,33 +3,64 @@ import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
...
-        epsSpatial: defaultParam.eps_spatial
+const getDefaultParameters = params => {
+  const mode = params.modes.find(mode => mode.name === DEFAULT_MODE)
+  if (!mode || !mode.epsTemporal || !mode.epsSpatial) {
+    return null
+  }
+  mode.epsMax = Math.max(mode.epsTemporal, mode.epsSpatial)
+  mode.maxBound =
+    mode.epsMax * 2 < DEFAULT_MAX_BOUND ? mode.epsMax * 2 : DEFAULT_MAX_BOUND
+  mode.cosAngle = gradientAngle(mode.epsMax, COARSE_COEFFICIENT)
```
y-lohse: Is it safe to directly modify an object inside `params.modes`?

src/photos/ducks/clustering/settings.js
```diff
@@ -3,33 +3,64 @@ import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
...
+  return mode
+}
+
+export const findPhotosDefaultParameters = (setting, photos) => {
+  const matchingParams = matchingParameters(setting.parameters, photos)
+  return matchingParams ? getDefaultParameters(matchingParams) : null
```
y-lohse: Without knowledge of the algorithm, this seems odd — if we found a matching parameter, why are we calling `getDefaultParameters `?

src/photos/ducks/clustering/settings.js
```diff
@@ -3,33 +3,64 @@ import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
...
+  }
+  mode.epsMax = Math.max(mode.epsTemporal, mode.epsSpatial)
+  mode.maxBound =
+    mode.epsMax * 2 < DEFAULT_MAX_BOUND ? mode.epsMax * 2 : DEFAULT_MAX_BOUND
+  mode.cosAngle = gradientAngle(mode.epsMax, COARSE_COEFFICIENT)
+  return mode
+}
+
+export const findPhotosDefaultParameters = (setting, photos) => {
+  const matchingParams = matchingParameters(setting.parameters, photos)
```
y-lohse: Can we rename `matchingParameters ` with a verb prefix, like `find` or `get` please?

src/photos/ducks/clustering/settings.js
```diff
@@ -3,33 +3,64 @@ import { DOCTYPE_PHOTOS_SETTINGS } from 'drive/lib/doctypes'
...
+  return matchingParams ? getDefaultParameters(matchingParams) : null
+}
+
+export const findLastDefaultParameters = setting => {
+  const lastParams = setting.parameters[setting.parameters.length - 1]
+  return lastParams ? getDefaultParameters(lastParams) : null
+}
+
+export const defaultSetting = photos => {
```
y-lohse: Same here, probably `getDefaultSetting`?

targets/photos/services/onPhotoUpload.js
```diff
@@ -29,93 +28,112 @@ import {
...
-          saveClustering(clusters, albumsToSave)
-        }
+const createNewClusters = async (setting, dataset, albums) => {
+  const clusterize = await albumsToClusterize(dataset, albums)
+  if (clusterize) {
+    for (const [id, photos] of Object.entries(clusterize)) {
+      const params = findPhotosDefaultParameters(setting, photos)
+      if (!params) {
+        log('warn', 'No parameters for clustering found')
+        break
```
y-lohse: Just to be sure, if we can't find the params for one cluster, we stop doing *all* the clusters? Because otherwise we'd want `continue` instead of `break` here :)

targets/photos/services/onPhotoUpload.js
```diff
@@ -29,93 +28,112 @@ import {
...
+      // Use default
+      setting = defaultSetting(dataset)
+    }
+    await clusterizePhotos(setting, dataset)
+  } else {
+    const photos = await getNewPhotos(setting.lastSeq)
+    if (photos.length < 1) {
+      log('warn', 'Service called but no photos found to clusterize')
+      return
+    }
```
y-lohse: Not a big deal, but we could probably factorize this part so we only have to do the check and warning once.

src/photos/ducks/clustering/settings.js
Gara64: The naming might be confusing indeed. Actually, this is an anticipation of the future: we could have several kind of parameters (the modes), to customize the granularity of the clustering, e.g. `macro`, `micro`, 'super micro`, etc. Today, there is only one mode, thus the `default`.
So I should rename this `getDefaultParameterMode`

targets/photos/services/onPhotoUpload.js
Gara64: Right, well seen!

src/photos/ducks/clustering/settings.js
Gara64: We wouldn't be? We need the params written in the db and computes some other ones based on them.

y-lohse: The risk is that another function somewhere in the flow will use the same object, and behaves differently because we modified it here. I don't think it's the case here, but it's generally something to be mindful of.

Gara64:

Gara64:

PR #1330, opened 2018-12-12
WIP: switching to cozy scripts build

y-lohse:


y-lohse:
[(2) Mettre à jour webpack](https://trello.com/c/LZbJ0fsF/1300-2-mettre-%C3%A0-jour-webpack)

CPatchane:
🎉 🎉 🎉

CPatchane:
> the mobile build also builds the intents and public chunks, and they may get included in the final build, although they will never be called
We could change that to not build them if we want the mobile target, the mobile app won't app intents services and public page in all cases

src/drive/targets/mobile/fastlane/report.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Crash--: I think you can edit the gitignore to add this fastlane folder

src/drive/targets/browser/index.ejs
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
```
Crash--: 👍 for `browser` instead of `web`

webpack/mobile.config.js
```diff
@@ -0,0 +1,11 @@
+const { ProvidePlugin } = require('webpack')
+
+module.exports = {
+  plugins: [
+    new ProvidePlugin({
+      PouchDB: 'pouchdb',
+      pouchdbFind: 'pouchdb-find',
+      pouchdbAdapterCordovaSqlite: 'pouchdb-adapter-cordova-sqlite'
```
Crash--: Is it used somewhere?

y-lohse: The adapter, yes, it's used in cozy-client-js. The 2 other variables I'm not sure…

PR #1334, opened 2018-12-14
fix: Use more color variables

y-lohse:
Replaces some of our hard-coded color codes by variables. The simpler ones.

### DIFF
--- a/src/drive/styles/filelist.styl
+++ b/src/drive/styles/filelist.styl
@@ -18,9 +18,9 @@
     @extend $table-row-head
 
 .fil-content-row
-    @extend $table-row 
+    @extend $table-row
     &:hover
-        .fil-content-file-select	
+        .fil-content-file-select
             opacity 1
     &--center
         justify-content center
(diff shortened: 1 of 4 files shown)


Crash--:
Is it just a first pass?

y-lohse:
Yep, these are the "easy" replacements. The remaining ones are new colors, colors with transparency, or generally things that require some thought before being replaced.

Crash--:
Can we change also here : https://github.com/cozy/cozy-drive/blob/master/src/sharing/components/ShareByEmail.jsx#L96 and here https://github.com/cozy/cozy-drive/blob/master/src/sharing/components/ShareByEmail.jsx#L100

y-lohse:
[(2) Codes couleurs en dur](https://trello.com/c/0KjHoPp7/1468-2-codes-couleurs-en-dur)

y-lohse:
I've added the message in our corresponding trello card, we can tackle these colors when we really work on it. i'll merge this in the meantime, even if it's incomplete.

PR #1337, opened 2018-12-17
 feat: Auto albums unicity 

Gara64:


Gara64:

src/photos/ducks/clustering/albums.js
```diff
@@ -50,17 +68,47 @@ export const findAutoAlbums = async () => {
...
+    sort: [{ name: 'desc' }]
   })
   return results
 }
 
-export const saveClustering = async (clusters, albums) => {
+const addToClusters = async (clusters, albumsToSave) => {
+  let i = 0
+  for (const photos of clusters) {
+    await addAutoAlbumReferences(photos, albumsToSave[i++])
```
y-lohse: Actually, I would inline this function. It really really needs to have the same number of clusters and albums, si it would be good to keep the function close to the condition. This also lessens the risk of someone calling this function from somewhere else, unaware of he requirement.

Note that you can avoid having to carry around an `i` variable by doing `await Promise.all(clusters.map((photos, index) => addAutoAlbumReferences(photos, albumsToSave[index]))`

src/photos/ducks/clustering/albums.js
```diff
@@ -50,17 +68,47 @@ export const findAutoAlbums = async () => {
...
+      // The clustering structure has not changed: only new photos to add
+      await addToClusters(clusters, albumsToSave)
+    } else {
+      // The clustering structure has changed: remove the impacted clusters
+      // and create the new structure
+      await recreateClusters(clusters, albumsToSave)
     }
+  } else {
+    // No cluster exist yet: create them
+    await createClusters(clusters)
```
y-lohse: Instead of having `createClusters ` and `recreateClusters `, I suggest having `createClusters` and `removeClusters` to avoid code duplication.

src/photos/ducks/clustering/albums.js
```diff
@@ -100,6 +149,10 @@ const findPhotosToReclusterize = async albums => {
   return photos
 }
 
+const photoExists = (photos, photo) => {
```
y-lohse: How about naming this function `photoExistsInCluster` or `photoExistsInCollection` or something like that? Because like… the photo exists, it's right there. I'm not even totally sure we need this function since it's just a wrapper around a one liner?

src/photos/ducks/clustering/albums.js
```diff
@@ -50,17 +68,47 @@ export const findAutoAlbums = async () => {
   ])
   const results = await cozyClient.data.query(autoAlbums, {
     selector: { auto: true },
-    sort: [{ name: 'asc' }]
+    sort: [{ name: 'desc' }]
```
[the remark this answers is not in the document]
Gara64: Not quite, it is a small optimization. 
This is based on the assumption a new uploaded photo will probably be the one with the most recent date, so it is likely to cluster with the most recent auto album. This avoids to iterate over the whole collection to find the newest album.

src/photos/ducks/clustering/albums.js
Gara64: Ok for the renaming. I prefer to keep the function, as the semantic of the name gives a quicker insight than doing `!photos.find(p => p.id === photo.id)`

PR #1347, opened 2018-12-18
fix: Use new mutate signature

y-lohse:


src/photos/ducks/timeline/index.jsx
```diff
@@ -19,12 +19,12 @@ const TIMELINE_QUERY = client =>
...
   uploadPhoto: async (file, dirId) => {
-    return query.client.mutate({
+    return client.mutate({
       mutationType: 'UPLOAD_PHOTO',
       execute: () =>
-        query.client.collection('io.cozy.files').createFile(file, { dirId })
+        client.collection('io.cozy.files').createFile(file, { dirId })
     })
   },
   deletePhoto: photo => query.client.destroy(photo)
```
Crash--: deletePhoto methods will still work? Or do we need to use `deletePhoto: photo => client.destroy(photo)`

y-lohse:
[Album photo - pagination](https://trello.com/c/0Z6W0sEN/1511-album-photo-pagination)

src/photos/ducks/timeline/index.jsx
y-lohse: Still worked, but I'm changing it anyway to stay consistent.

PR #1348, opened 2018-12-18
fix: Update intents view URL

y-lohse:


y-lohse:
[(2) Mettre à jour webpack](https://trello.com/c/LZbJ0fsF/1300-2-mettre-%C3%A0-jour-webpack)

PR #1366, opened 2019-01-04
fix: path in Recent view

cedricmessiant:
- display path of the parent folder under the filename
- click on the path redirects to parent folder

https://trello.com/c/J9GzS2Ao

### DIFF
--- a/src/drive/web/modules/filelist/File.spec.jsx
+++ b/src/drive/web/modules/filelist/File.spec.jsx
@@ -1,0 +1,26 @@
+'use strict'
+
+import { getParentLink } from './File'
+
+describe('getParentLink function', () => {
+  it('should return the first link in the element ancestors', () => {
+    const div = document.createElement('div')
+    const link = document.createElement('a')
+    link.className = 'my-link'
+    const span = document.createElement('span')
+    const span2 = document.createElement('span')
+    div.appendChild(link)
+    link.appendChild(span)
+    span.appendChild(span2)
+    const result = getParentLink(span2)
+    expect(result).toEqual(link)
+  })
+
+  it('should return null if there is no link in the element ancestors', () => {
+    const div = document.createElement('div')
+    const span = document.createElement('span')
+    div.appendChild(span)
+    const result = getParentLink(span)
+    expect(result).toBeNull()
+  })
+})
--- a/src/drive/web/modules/navigation/duck/reducer.js
+++ b/src/drive/web/modules/navigation/duck/reducer.js
@@ -360,18 +360,32 @@ export const getFileById = ({ view }, id) => {
 
 const isRootFolder = folder => folder.id === ROOT_DIR_ID
 
-const ensureFileHavePath = ({ view }, file) => ({
+export const ensureFileHavePath = ({ view }, file) => ({
   ...file,
+  displayedPath: getDisplayedFilePath({ view }, file),
   path: getFilePath({ view }, file)
 })
 
+/*
+  TODO: deprecate and remove getFilePath
+*/
 export const getFilePath = ({ view }, file) => {
   const { displayedFolder } = view
-  return isDirectory(file)
-    ? file.path
-    : displayedFolder && !isRootFolder(displayedFolder)
-      ? `${displayedFolder.path}/${file.name}`
-      : `/${file.name}`
+  if (isDirectory(file)) {
+    return file.path
+  } else {
+    const folderPath = displayedFolder
+      ? !isRootFolder(displayedFolder)
+        ? displayedFolder.path
+        : ''
+      : file.path // file.path is the parent folder path
+    return `${folderPath}/${file.name}`
+  }
+}
+
+export const getDisplayedFilePath = ({ view }, file) => {
+  const { displayedFolder } = view
+  return file.path || (displayedFolder ? displayedFolder.path : '')
 }
 
 export const getOpenedFolderId = state => state.view.openedFolderId
--- a/jest.config.json
+++ b/jest.config.json
@@ -11,9 +11,17 @@
     "^components(.*)": "<rootDir>/src/components$1",
     "^folder-references(.*)": "<rootDir>/src/folder-references$1"
   },
+  "snapshotSerializers": ["enzyme-to-json/serializer"],
   "transform": {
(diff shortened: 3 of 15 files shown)


src/drive/web/modules/drive/Container.jsx
```diff
@@ -64,6 +64,7 @@ const isAnyFileReferencedByAlbum = files => {
 }
 
 const mapStateToProps = state => ({
+  files: state.view.files,
```
y-lohse: I'm afraid don't understand this change.

This prop is passed to the `Container` component, which is then rendered by the FileExplorer. But the [FIleExplorer already injects a `files` prop](https://github.com/y-lohse/cozy-drive/blob/photos-relations/src/drive/web/modules/navigation/FileExplorer.jsx#L84).

So now we're injecting the prop twice, but one time using directly the state value, and one time using the `getVisibleFiles` selector.

Do we really want to inject the same prop twice? And why does injecting it directly with   `state.view.files` work, but now with `getVisibleFiles`?

Crash--: Oh nice catch! 
In fact, we display only the filename if we don't have a `displayedFolder`https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/navigation/duck/reducer.js#L372 

Since we're in the `recent view`, we don't have any displayedFolder. 

We should probably edit this `getVisibilesFiles` selector ! 👍

y-lohse: Aaah yes, I didn't go all the way down to the `getFilePath ` selector, this does explain the problem. So yeah I also think it would be best to fix this problem in the selectors.

cedricmessiant: oh, just saw your comments (I was already on it). In the end, I did this: https://github.com/cozy/cozy-drive/pull/1366/commits/cf88e57c149943b77d32870226cca69f7e4b8228

It feels strange and misleading that we have two different information in `path` attribute. In case of a file Recent view, we have the path to the parent dir and for other files, we have the full path. I am afraid to break stuff so I don't want to go too far here but I don't really like this :-)

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -358,20 +358,17 @@ export const getFileById = ({ view }, id) => {
...
 export const getFilePath = ({ view }, file) => {
   const { displayedFolder } = view
-  return isDirectory(file)
+  const isRecentView = displayedFolder === null
+  return isDirectory(file) || isRecentView
     ? file.path
-    : displayedFolder && !isRootFolder(displayedFolder)
-      ? `${displayedFolder.path}/${file.name}`
-      : `/${file.name}`
+    : `${displayedFolder.path}/${file.name}`.replace('//', '/')
```
Crash--: Good to know, we can't create a folder or a file with an `/` in its name. And if we create a folder / file with an `/`, it'll be replaced by `:`.  So this regex seems to work even if I would prefer to check with `isRootFolder` since it more robust

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -358,20 +358,17 @@ export const getFileById = ({ view }, id) => {
...
-
 const ensureFileHavePath = ({ view }, file) => ({
   ...file,
   path: getFilePath({ view }, file)
 })
 
 export const getFilePath = ({ view }, file) => {
   const { displayedFolder } = view
-  return isDirectory(file)
+  const isRecentView = displayedFolder === null
```
y-lohse: Continuing the discussion  started here https://github.com/cozy/cozy-drive/commit/cf88e57c149943b77d32870226cca69f7e4b8228#r31858308

The `displayedFolder ` being `null` is indeed not enough to mean we're in the recent view. I think the sharing view for example also has no `displayedFolder` set.

But generally speaking, could we do something like this?

```
return file.path || `${displayedFolder ? displayedFolder.path : ''}/${file.name}`.replace('//', '/')
```

cedricmessiant: In case of the sharing view, we want the same behaviour (even though it's not done yet:  https://trello.com/c/7bV17E2g )

As I said above I thought about the solution you suggest but I'm afraid that the `file.path` condition may create regressions. If you think it's ok, I'm ok with this solution. Also, I still don't understand why in some cases we want the file name in the file path and in other cases we don't.

Crash--: Since we will do it for the sharing view, we can rename the proposed `isRecentView` by something more explicit ? 
```js 
const viewsThatNeedToDisplayPath = ['recent', 'sharings'];
if(viewsThatNeedToDisplayPath.includes(currentView)){
...
}
```

Also, I'm not a big fan of the `.replace('//', '/')` regex :(

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -360,18 +360,32 @@ export const getFileById = ({ view }, id) => {
...
   ...file,
+  displayedPath: getDisplayedFilePath({ view }, file),
   path: getFilePath({ view }, file)
 })
 
+/*
+  TODO: deprecate and remove getFilePath
+*/
 export const getFilePath = ({ view }, file) => {
   const { displayedFolder } = view
```
Crash--: What do you think about adding a console.warn here with the trace of the call and a link to this discussion? Like that, you'll see where and when this function is called?

src/drive/web/modules/filelist/File.spec.jsx
```diff
@@ -0,0 +1,26 @@
+'use strict'
```
y-lohse: Why the `use strict` here?

src/drive/web/modules/navigation/duck/reducer.js
y-lohse: I think this function is called by redux `connect`, so it will be called all the time :/

src/drive/web/modules/filelist/File.spec.jsx
cedricmessiant: For no particular reason. I think I saw it in another file and thought that it is better to use it when it's possible. Do you want me to remove it?

PR #1372, opened 2019-01-10
fix(Sharings): Remove underline on path on hover

cedricmessiant:
I use `+small-screen()` selector, it seems to match mobile resolutions


cedricmessiant:
in fact we don't have the link on mobile but since it shares the same class it was underlined : https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/filelist/File.jsx#L170-L182

PR #1374, opened 2019-01-10
test: Testcafe tests update and travis configuration

CozyKourai:
New PR for Testcafé (Replacing https://github.com/cozy/cozy-drive/pull/1369)

Include all tests and config file for travis


CozyKourai:
[1ers TestCafe](https://trello.com/c/gZiKlk6a/1518-1ers-testcafe)

.travis.yml
```diff
@@ -18,13 +18,16 @@ env:
...
+  firefox: latest #testcafe
   apt:
     sources:
       - ubuntu-toolchain-r-test
       - deadsnakes
+      - google-chrome
     packages:
       - g++-4.8
       - python3.5
+      - google-chrome-stable fluxbox
```
Crash--: Since we're currently only testing with firefox, can we remove this step?

.travis.yml
```diff
@@ -62,3 +67,10 @@ deploy:
     script: yarn run deploy:$COZY_APP_SLUG
     on:
       tags: true
+# dist: trusty
+sudo: required
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3
```
Crash--: We should use something way more robust than just a sleep 3. 

Something like this : https://github.com/elgalu/docker-selenium/blob/master/xvfb/bin/wait-xvfb.sh

testcafe/runner-photos.js
```diff
@@ -0,0 +1,30 @@
+const createTestCafe = require("testcafe");
+let testcafe = null;
+
+createTestCafe("localhost", 1337, 1338)
+  .then(tc => {
```
Crash--: can we use a newer syntax and not deal with Promise ? 
```js
try {
const tc = await createTestCafe() 
}catch(e){
 console.log("Tests failed: " + failedCount);
    tc.close();
}
```

testcafe/runner-photos.js
```diff
@@ -0,0 +1,30 @@
...
+        .browsers(["firefox:headless"])
+        //.browsers(["firefox"])
+
+        .screenshots(
+          "reports/screenshots/",
+          true,
+          "${DATE}_${TIME}/test-${TEST_INDEX}-${FILE_INDEX}.png"
+        )
+        .run(
+          { selectorTimeout: 200000 },
```
Crash--: I don't want to wait 200 seconds to see if a selector has failed. When the test will fail, they will take ages to run... 

60s is already more than enough, no?

testcafe/tests/helpers/roles.js
```diff
@@ -0,0 +1,19 @@
...
+const page = new Page();
+
+export const regularUser = Role(
+  `${config.photosUrl}`,
+  async t => {
+    await t
+      .typeText(page.password, `${config.password}`)
+      .click(page.loginButton)
+      .expect(getPageUrl())
+      .contains("#/photos"); //Checks if the current page URL contains the '#/photos' string
```
Crash--: Ici c'est clairement un cas qu'on pourrait améliorer. Éviter d'avoir des strings en dur par ci par là et qui serait dissociée du code. 

Typiquement dans le code de photos on pourrait exposer une variable avec le chemin et l'utiliser ici pour le checker.

Ainsi, le jour où on change l'adresse de redirection après login dans le code, le test ne sera pas en échec

testcafe/tests/pages/photos-model.js
```diff
@@ -0,0 +1,52 @@
...
+  constructor() {
+    this.nameInput = Selector("#developer-name");
+
+    this.btnUpload = Selector('[class*="pho-toolbar"]')
+      .find("span")
+      .find("input");
+    this.divUpload = Selector('[class*="upload-queue"]');
+
+    this.modalUpload = Selector('[class*="c-alert-wrapper"]', {
+      visibilityCheck: true
```
Crash--: Par défaut, testcafé ne check pas si c'est visible, c'est ça ? 

Est-ce que ce visibilityCheck, fonctionne aussi sur les images pour savoir si elles ont bien été chargées par exemple ?

testcafe/tests/pages/photos-model.js
```diff
@@ -0,0 +1,52 @@
...
+    //Top Option bar & Confirmation Modal
+    this.barPhoto = Selector('[class*="coz-selectionbar"]');
+    this.barPhotoBtnAddtoalbum = this.barPhoto.find("button").nth(0); //ADD TO ALBUM
+    this.barPhotoBtnDl = this.barPhoto.find("button").nth(1); //DOWNLOAD
+    this.barPhotoBtnDelete = this.barPhoto.find("button").nth(2); //DELETE
+    this.modalDelete = Selector('[class*="c-modal"]').find("div");
+    this.modalDeleteBtnDelete = this.modalDelete.find("button").nth(2); //REMOVE
+    this.allPhotos = Selector('[class^="pho-photo"]').find(
+      '[class^="pho-photo-item"]'
+    );
```
Crash--: can you add a //!FIXME tag? And create a new trello card to use IDs ?

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
+import { Selector, Role } from "testcafe"; //import testcafe function
+import { regularUser } from "./helpers/roles"; //import roles for login
+import config from "../config"; //import url & psswd config
+//import { generateRandomInteger } from "./helpers/utils.js";
+const _ = require("lodash");
```
Crash--: Can we use only one system to use external code? ie use import instead of require. And to make the thing right, we onyl want to import random not all the lib. 

Something like 
```
import random from 'lodash/random';
```

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
...
+
+test("Uploading 1 pic from Photos view", async t => {
+  //new pic shows up
+
+  await t
+    .setFilesToUpload(page.btnUpload, ["../data/IMG0.jpg"])
+    .expect(page.divUpload.visible)
+    .ok()
+    .expect(page.modalUpload.exists)
+    .ok({ timeout: 50000 })
```
Crash--: Why do we have to write a `timeout` ?

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
...
+      "../data/IMG-JPG.jpg",
+      "../data/IMG-PNG.png",
+      "../data/IMG-GIF.gif"
+    ])
+    .expect(page.divUpload.visible)
+    .ok()
+    .expect(page.modalUpload.exists)
+    .ok({ timeout: 50000 })
+    .expect(page.divUpload.child("h4").innerText)
+    .contains("Uploaded 3 out of 3 successfully");
```
Crash--: I think this is something we do not want neither. We can have minor change on our strings. 

What do you think about using our translation system here to check the string ?

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
...
+    .ok()
+
+    .expect(page.barPhotoBtnAddtoalbum.visible)
+    .ok()
+    .expect(page.barPhotoBtnDl.visible)
+    .ok()
+    .expect(page.barPhotoBtnDelete.visible)
+    .ok();
+
+  // TODO - Add check on label text ??
```
Crash--: yes plz

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
...
+});
+
+test("Open 1st pic", async t => {
+  //Right arrow shows up. Navigatio to other pics is OK, Closing pic (X or 'esc') is Ok
+
+  await t
+    .click(page.photoThumb(0))
+    .expect(page.photoFull.visible)
+    .ok()
+    .takeScreenshot();
```
Crash--: Add an url check?

testcafe/tests/photos_crud.js
```diff
@@ -0,0 +1,255 @@
...
+    .notOk() //1st photo, so previous button does not exist
+    .hover(page.photoNavNext) //1st photo, so next button should exists
+    .expect(page.photoNavNextBtn.visible)
+    .ok() //Next arrow is shown
+    .click(page.photoNavNextBtn);
+
+  const photo2src = await page.photoFull.getAttribute("src");
+  await t
+    .expect(photo1src)
+    .notEql(photo2src) //Photo has change, so src is different
```
Crash--: and check if url has changed?

edas:
General comment : 

* Do not use manual timeouts. Try to avoid them, testcafe should help you in that by its promise system. If you are sure you can't do without, please discuss the problem with the team, and if you all confirm there is no better solution, put a comment explaining why in your source code

* Try to select react components directly and not their html rendering. If you have to select an html element, try to select by id, by a data-* attribute, or by a specific classname (like `photo-item` which is not used anywhere for anything else than what you search for) and not generic ones (like `button`) unless you really want to be generic. Using id is tricky because you may generate conflicts so you have to use very specific ids; selecting React components is probably the way to go.

* Add comments! `ok` expects a parameter with a string that could explain some things

testcafe/runner-photos.js
CozyKourai: The default is 100s.
I changed the timeout because I was having issues on .works server which are much slower, which cause my tests to fall because not all pictures on the page were loaded.
I can probably refine and improve this point anyway.

testcafe/tests/photos_crud.js
CozyKourai: I think this need to change too! 
Actually the part I really need is the "3" in "3 ou of 3 succesfully") because even if the upload fail, we still have the same message but with "0 out of 3"
So maybe a regular expression checking we have "3" twice in the string will be enough

CozyKourai:
[Tests E2E - Amélioration des tests existants et retours de PR](https://trello.com/c/lBMoXXux/1551-tests-e2e-am%C3%A9lioration-des-tests-existants-et-retours-de-pr)

CozyKourai:
[Tests E2E - Amélioration des tests existants et retours de PR](https://trello.com/c/lBMoXXux/1551-tests-e2e-am%C3%A9lioration-des-tests-existants-et-retours-de-pr)

CozyKourai:
Close pour tests Travis sur mon repo directement

testcafe/tests/photos_crud.js
CozyKourai: I'll add label check when we'll add keyword in testcafé, otherwise those checks will have string issues too!

PR #1376, opened 2019-01-11
feat(Sharings): Add path to parent folders

cedricmessiant:


cedricmessiant:
[(1) Vue 'SHARINGS' - path](https://trello.com/c/7bV17E2g/1210-1-vue-sharings-path)

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -25,13 +25,23 @@ class SharingFetcher extends React.Component {
       const resp = await client
         .collection('io.cozy.files')
         .all({ keys: sharedDocuments })
+
+      const parent_ids = resp.data.map(f => f.dir_id)
+      const parentsResp = await client.collection('io.cozy.files').all({
+        keys: parent_ids
+      })
+      const parents = parentsResp.data
```
Crash--: We only fetch the first parent, right? 

So if my file is in : 
/Admin/foo/bar/toto.PNG

we'll only display : 
/bar/

right?

Crash--: OK just read the test... Sorry.

cedricmessiant: The path we get from the stack is the full path of the directory

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -25,13 +25,23 @@ class SharingFetcher extends React.Component {
       const resp = await client
         .collection('io.cozy.files')
         .all({ keys: sharedDocuments })
+
+      const parent_ids = resp.data.map(f => f.dir_id)
```
y-lohse: Nitpick:

```suggestion
      const parentIds = resp.data.map(f => f.dir_id)
```

To stay consistent with the casing :)

PR #1382, opened 2019-01-14
Quick fix services module from cozy-scripts config

CPatchane:
~~__PLEASE WAIT FOR THE v1.10.0 of cozy-scripts BEFORE MERGING THIS PR__~~

Quick fix the modules path when building app services.

The change need to be well tested on cozy-scripts side and for other applications building process, I think it will be added into the next major version of cozy-scripts since it could be a breaking changes.


app.config.js
```diff
@@ -13,16 +13,9 @@ configurationFiles.push(
 )
 
 const isDrive = process.env.COZY_APP_SLUG === 'drive'
-const isPhotosServices = process.env.COZY_APP_SLUG === 'photos:services'
 
 if (isDrive) configurationFiles.push(require('./webpack/drive.config.js'))
-else if (isPhotosServices) {
-  configurationFiles.push(
-    require('cozy-scripts/config/webpack.config.services')
```
CPatchane: The v1.10.0 of cozy-scripts will add automatically the services config into bundles

app.config.js
```diff
@@ -13,16 +13,10 @@ configurationFiles.push(
 )
 
 const isDrive = process.env.COZY_APP_SLUG === 'drive'
-const isPhotosServices = process.env.COZY_APP_SLUG === 'photos:services'
+const isPhotos = process.env.COZY_APP_SLUG === 'photos'
```
Gara64:
```suggestion
const isPhotos = process.env.COZY_APP_SLUG === 'photos' || process.env.COZY_APP_SLUG === 'photos:services'
```

app.config.js
```diff
@@ -48,4 +42,20 @@ const extraConfig = {
...
+  configurationFiles[0].multiple.services
+) {
+  // FIXME: Will be handled correctly by next major version of cozy-scripts
+  configurationFiles[0].multiple.services.__mergeStrategy.strategy[
+    'resolve.modules'
+  ] = 'replace'
+  configurationFiles[0].multiple.services.resolve.modules = [
+    SRC_DIR,
+    'node_modules'
+  ]
```
Gara64: This fails if resolve is not defined.
```suggestion
      configurationFiles[0].multiple.services.resolve = {
    ...configurationFiles[0].multiple.services.resolve,
    modules: [SRC_DIR, 'node_modules']
  }
```

Gara64: It fails if resolve is not defined.
This could do the trick:
```
  configurationFiles[0].multiple.services.resolve = {
    ...configurationFiles[0].multiple.services.resolve,
    modules: [SRC_DIR, 'node_modules']
  }
```

Crash--: Can you explain what this code do and why?

app.config.js
CPatchane: I think we don't need the script `photos:services` anymore since we build services with the photos browser application

Gara64: Ok, fair enough

package.json
```diff
@@ -10,12 +10,10 @@
     "watch:drive:browser": "COZY_APP_SLUG=drive cozy-scripts watch --manifest src/drive/targets/manifest.webapp --build-dir build/drive --src-dir src/drive",
     "watch:drive:mobile": "COZY_APP_SLUG=drive NODE_ENV=mobile:development cozy-scripts watch --manifest src/drive/targets/manifest.webapp --src-dir src/drive",
     "watch:drive:standalone": "COZY_APP_SLUG=drive NODE_ENV=mobile:development cozy-scripts start --no-stack --manifest src/drive/targets/manifest.webapp --src-dir src/drive",
-    "build:photos": "npm-run-all --parallel build:photos:browser build:photos:services",
+    "build:photos": "npm-run-all --parallel build:photos:browser",
     "build:photos:browser": "COZY_APP_SLUG=photos:browser cozy-scripts build --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
```
Gara64: `COZY_APP_SLUG=photos:browser` should be replaced by `COZY_APP_SLUG=photos`, so `isPhotos` has the expected value

package.json
```diff
@@ -10,12 +10,10 @@
...
-    "build:photos:services": "COZY_APP_SLUG=photos:services cozy-scripts build --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
     "watch:photos": "npm run watch:photos:browser",
     "watch:photos:browser": "COZY_APP_SLUG=photos:browser cozy-scripts watch --debug --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
```
Gara64: Same remark here

Gara64:
@y-lohse Yes, it's ok. It's a bit less convenient when we just want to test a service, but it's not a big deal.

CPatchane:
@Gara64 It was already building the whole application with the `photos:services`

y-lohse:
I know, but the plan was to move towards building them separately rather then the opposite direction :p

PR #1385, opened 2019-01-15
feat(Drive): Require DevTools during DEV

Crash--:
This is just a POC to have access to the React Dev Tools panel. 

But I want to discuss about a more global subject.

We've several entry points with Drive / Photos and we need to do the same thing several times. 

How can we have some kind pf mutualisation here? Should we create a DriveWrapper to wrap IconSprite / devtools and so on?

PS : If you want to crash your browser, do not hesitate to enable React dev tools highlights :D


cedricmessiant:
what is the alternative? either we have only one entry point but I guess it's not trivial or we can add a wrapper component (or an hoc maybe) to add the features that are shared between the different entry points?

y-lohse:
👍  to the shared component / file / HOC between entry points.This might even end up in cozy-libs because i think the problem is the same across all our apps.

PR #1388, opened 2019-01-17
Fix/icon

Crash--:
- Use PublicLayout for LightFileViewer
- Add `isFile` attribute to know if we are rendering a file or a folder.
- Added some proptypes 
- Changed Photos public view (buttons order / padding)

### DIFF
--- a/src/drive/targets/public/index.jsx
+++ b/src/drive/targets/public/index.jsx
@@ -3,7 +3,6 @@
 import 'whatwg-fetch'
 import React from 'react'
 import { render } from 'react-dom'
-import { IconSprite } from 'cozy-ui/transpiled/react'
 
 import { Router, Route, Redirect, hashHistory } from 'react-router'
 import CozyClient, { CozyProvider } from 'cozy-client'
@@ -97,7 +96,9 @@ const init = async () => {
       <I18n lang={lang} polyglot={polyglot}>
         <CozyProvider store={store} client={client}>
           {isFile ? (
-            <LightFileViewer files={[data]} />
+            <PublicLayout>
+              <LightFileViewer files={[data]} isFile={true} />
+            </PublicLayout>
           ) : (
             <Router history={hashHistory}>
               <Route component={PublicLayout}>
@@ -106,7 +107,6 @@ const init = async () => {
               <Redirect from="/*" to={`files/${id}`} />
             </Router>
           )}
-          <IconSprite />
         </CozyProvider>
       </I18n>,
       root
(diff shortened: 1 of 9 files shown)


src/drive/web/modules/public/LightFileViewer.jsx
```diff
@@ -15,4 +17,8 @@ const LightFileViewer = ({ files }) => (
   </div>
 )
 
+LightFileViewer.propTypes = {
+  files: PropTypes.array.isRequired,
```
cedricmessiant: could you use `arrayOf` so that we know exactly what do the files look like here?

src/drive/targets/public/index.jsx
```diff
@@ -106,7 +107,6 @@ const init = async () => {
               <Redirect from="/*" to={`files/${id}`} />
             </Router>
           )}
-          <IconSprite />
```
cedricmessiant: why did you remove `IconSprite`? `PublicLayout` provides the icons?

src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -147,26 +165,34 @@ class PublicToolbar extends React.Component {
...
       )
     }
   }
 }
-
+PublicToolbar.propTypes = {
+  isFile: PropTypes.bool.isRequired,
+  renderInBar: PropTypes.bool,
+  breakpoints: PropTypes.object.isRequired,
+  files: PropTypes.array.isRequired
```
cedricmessiant: same here and same for the shape of the `breakpoints` object above

src/drive/targets/public/index.jsx
Crash--: Exactly! https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/public/PublicLayout.jsx#L10

src/drive/web/modules/public/PublicToolbar.jsx
Crash--: When cozy-ui will expose its own propTypes. Currently it's not the case and I don't want to put that kind of work in this PR.

src/drive/web/modules/public/LightFileViewer.jsx
Crash--: Since `files` is something coming from the API, it can have a lot of different shapes. 

Is it possible to only check for the presence of one or two keys ?

Crash--:
[PARTAGE - icones vues publiques (suite)](https://trello.com/c/Rs1koHfT/1560-partage-icones-vues-publiques-suite)

PR #1355, opened 2018-12-21
feat: Fetch photos through relationships

y-lohse:
This PR reverts some of the previous changes we made, where we load photos as part of their parent album. We still keep the spinners we added last time though!

Needs https://github.com/cozy/cozy-client/pull/306 or the photo pagination will be broken.

PR #1393, opened 2019-01-18
fix: Better adding and removing photos from albums

y-lohse:
Follow-up of https://github.com/cozy/cozy-drive/pull/1355/files#diff-b99860e2348055cdd5e05388380200ddR49

### DIFF
--- a/src/photos/ducks/albums/index.jsx
+++ b/src/photos/ducks/albums/index.jsx
@@ -43,7 +43,17 @@ const addPhotos = async (album, photos) => {
 }
 
 const ALBUM_MUTATIONS = client => ({
-  updateAlbum: album => client.save(album),
+  updateAlbum: async album => {
+    const unique = await client
+      .collection(DOCTYPE_ALBUMS)
+      .checkUniquenessOf('name', album.name)
+    if (unique !== true) {
+      Alerter.error('Albums.create.error.already_exists', { name })
+      return
+    } else {
+      return client.save(album)
+    }
+  },
   deleteAlbum: album => client.destroy(album),
   addPhotos,
   removePhotos: async (album, photos, clearSelection) => {
@@ -68,7 +78,9 @@ const ALBUMS_MUTATIONS = client => ({
       }
       const album = { _type: DOCTYPE_ALBUMS, name, created_at }
 
-      const unique = await client.validate(album)
+      const unique = await client
+        .collection(DOCTYPE_ALBUMS)
+        .checkUniquenessOf('name', album.name)
       if (unique !== true) {
         Alerter.error('Albums.create.error.already_exists', { name })
         return
--- a/src/photos/ducks/albums/index.jsx
+++ b/src/photos/ducks/albums/index.jsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import { withRouter } from 'react-router'
 import { Query, withMutations } from 'cozy-client'
-import omit from 'lodash/omit'
 import SharingProvider from 'sharing'
 import AlbumsView from './components/AlbumsView'
 import AlbumPhotos from './components/AlbumPhotos'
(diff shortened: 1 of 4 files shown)


src/photos/targets/browser/doctypes.js
```diff
@@ -36,8 +36,7 @@ export default {
     doctype: DOCTYPE_ALBUMS,
     attributes: {
       name: {
-        type: 'string',
-        unique: true
```
y-lohse: I had to remove this because the `validate` function of cozy-client is buggy — when we update the album, it will find that one album already has the same name, but fail to realize that it's the same one.

y-lohse:
[(1) Relations de fichiers dans cozy-client](https://trello.com/c/7IkEB6u4/1545-1-relations-de-fichiers-dans-cozy-client)

src/photos/ducks/albums/index.jsx
```diff
@@ -25,13 +24,31 @@ const ALBUMS_QUERY = client =>
...
+      Alerter.info('Alerter.photos.already_added_photo')
+    } else {
+      Alerter.success('Albums.add_photos.success', {
+        name: album.name,
+        smart_count: photos.length
+      })
+    }
+  } catch (error) {
+    Alerter.error('Albums.add_photos.error.reference')
+  }
```
Crash--: I know this is not part of this PR, but I think that our `duck` functions, should do nothing related to the UI. Even if it's only an alert. WDYT?

src/photos/ducks/albums/index.jsx
```diff
@@ -25,13 +24,31 @@ const ALBUMS_QUERY = client =>
 export const ALBUM_QUERY = (client, ownProps) =>
   client.get(DOCTYPE_ALBUMS, ownProps.router.params.albumId).include(['photos'])
 
+const addPhotos = async (album, photos) => {
+  try {
+    const photoCountBefore = album.photos.data.length
+    await album.photos.addById(photos.map(({ _id }) => _id))
```
Crash--: Oh. Is it really `addById` and not `addByIds` ? :(

src/photos/targets/browser/doctypes.js
Crash--: Why are we removing this one?

Crash--: Oh! Didn't see this comment before posting mine... 

Do you created an issue?

src/photos/ducks/albums/index.jsx
y-lohse: It's `addById` for backwards compatibility reasons. But `addById (_id)` also works of course.

src/photos/ducks/albums/index.jsx
y-lohse: I'm not too sure. Yes, they shouldn't do major UI work, but alerts..?

This code is more or less like a redux action. So yeah, you wouldn't have much UI code there, but you could have things like `dispatch(alertError('message'))`. And that's pretty much what `Alerter.error()` does. But if you have an idea of a more suitable place to trigger the alerts, I'm happy to consider it !

src/photos/targets/browser/doctypes.js
y-lohse: I have now : https://github.com/cozy/cozy-client/issues/335

src/photos/ducks/albums/index.jsx
```diff
@@ -25,13 +24,31 @@ const ALBUMS_QUERY = client =>
...
+}
+
 const ALBUM_MUTATIONS = client => ({
   updateAlbum: album => client.save(album),
   deleteAlbum: album => client.destroy(album),
   addPhotos,
   removePhotos: async (album, photos, clearSelection) => {
     try {
-      await album.photos.remove(photos)
+      await album.photos.removeById(photos.map(({ _id }) => _id))
```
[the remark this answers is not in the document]
y-lohse: It is! Passing a single id as a parameter is the more "normal" use case though, but both are supported.

y-lohse:
@Crash-- @cedricmessiant sorryyyyyy I added another commit. It re-implements name checks so that we don't have a regression here. It's related to https://github.com/cozy/cozy-client/issues/335.

PR #1394, opened 2019-01-18
chore: Only run testcafe on non-master branches

y-lohse:


Crash--:
c'est bizarre on n'a que 2 checks maintenant alors qu'avant on en avait 3.

y-lohse:
Yeeep, not sure why. Maybe because the PR comes from a fork? I'll check.

y-lohse:
Yeah, same thing in #1395, only 2 checks for PRs hat come from a fork.

PR #1402, opened 2019-01-22
fix: Pagination on public albums

y-lohse:
The issue was pretty dumb…

Not sure what could be tested on this though. Any ideas?


y-lohse:
[Album partagé - pagination](https://trello.com/c/DSVUBzdY/1544-album-partag%C3%A9-pagination)

PR #1405, opened 2019-01-24
fix(Photos): Sharing view error when link is revoked

Crash--:


src/photos/components/ErrorUnshared.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import ErrorShare from 'components/Error/ErrorShare'
+import { Main } from 'cozy-ui/react/Layout'
+import styles from '../targets/public/index.styl'
+import classNames from 'classnames'
+
+const ErrorUnsharedComponent = () => (
```
cedricmessiant: maybe add a snapshot test for this component? ;-)

src/photos/components/ErrorUnshared.jsx
```diff
@@ -0,0 +1,21 @@
...
+  <div
+    className={classNames(
+      styles['pho-public-layout'],
+      styles['pho-public-layout--full'],
+      'u-pt-3'
+    )}
+  >
+    <Main className="u-pt-1-half">
+      <ErrorShare errorType={`public_album_unshared`} />
```
cedricmessiant: why not `errorType="public_album_unshared"` ?

src/photos/targets/public/App.jsx
```diff
@@ -134,6 +122,9 @@ class App extends Component {
 const ConnectedApp = props => (
   <Query query={ALBUM_QUERY} {...props}>
     {({ data: album, fetchStatus }) => {
+      if (fetchStatus === 'failed') {
```
cedricmessiant: maybe I missed something but I think you can remove `state` in the dumb component since its `fetchStatus` can't be `failed`

src/photos/components/ErrorUnshared.jsx
Crash--: As already said, I really like the `={}` syntax because I find it more consistent with the rest of the code : 
boolean={true/false}
object={object}
string={'string'} 

;)

Crash--:
@y-lohse @cedricmessiant Hop, new version with : 
- removed state 
- propTypes 
- Snapshot
- removed `{...props}`

💪

PR #1409, opened 2019-01-24
feat(Onboarding): We can use COZY_FLAGS=beta-onboarding yarn build:dr…

Crash--:
…ive:mobile to build with the beta url


Crash--:
> Cool trick! Is it documented somewhere? ALlI see [is this](https://github.com/cozy/cozy-libs/tree/master/packages/cozy-flags#flags-enabled-at-build-time) but your commit message mentions a different variable?

You right, we should document this variable in cozy-scripts. https://github.com/CPatchane/create-cozy-app/pull/1028

PR #1412, opened 2019-01-28
fix: Use permissions to get photo shared id

y-lohse:
We're now using the permissions route to get the shared document id on photos as well.


src/sharing/getSharedDocument.js
```diff
@@ -0,0 +1,14 @@
+import get from 'lodash/get'
+
+const getSharedDocument = async client => {
+  const { data: permissionsData } = await client
+    .collection('io.cozy.permissions')
+    .getOwnPermissions()
+
+  const permissions = permissionsData.attributes.permissions
+  const sharedDocumentId = get(Object.values(permissions), '0.values.0')
```
cedricmessiant: are you sure that the first value will always be the right one? (`collection` I guess according to the unit test)

y-lohse: In fact, I am not. In the case of photo sharings we get `collection` and `files` and want `collection`. In the case of file sharing, we only receive `files`, and that's what we want.

So so far, yes, we always want the first one. But I'm not sure where the names come from and if they are reliable or not, so using the order seems like the best bet. Do you another approach in mind?

cedricmessiant: I guess that here you don't know if you are in file sharing or photo? So no, I don't have a better approach. But maybe add a comment in this case, it may be helpful in the future!

PR #1414, opened 2019-01-28
fix: Redirect on public pages

y-lohse:
We were using the react router to open URLs on the public page, but:

- [we don't always have a router in the public page](https://github.com/y-lohse/cozy-drive/blob/share-redirect/src/drive/targets/public/index.jsx#L93)
- `router.push` is meant for routes that are part of the app, but here we redirect to completely different pages, so `window.location` is more appropriate

### DIFF
--- a/src/drive/web/modules/public/PublicToolbar.jsx
+++ b/src/drive/web/modules/public/PublicToolbar.jsx
@@ -19,7 +19,7 @@ import DownloadIcon from 'drive/assets/icons/icon-download-16.svg'
 
 const { BarRight } = cozy.bar
 
-const DownloadFilesButton = ({ t, onDownload, size = 'normal', isFile }) => (
+const DownloadFilesButton = ({ t, onDownload, size, isFile }) => (
   <DownloadButton
     label={
       isFile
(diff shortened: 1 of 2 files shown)


y-lohse:
[(1) Menu d'action - vue mobile - item non cliquable - vue publique de partage](https://trello.com/c/JIuXLcBf/1573-1-menu-daction-vue-mobile-item-non-cliquable-vue-publique-de-partage)

src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -63,20 +62,21 @@ const MoreMenu = ({ t, onDownload, onOpenInCozy, onCreateCozy, isFile }) => (
...
       onDownload={onDownload}
-      onOpenInCozy={discoveryLink ? () => redirectTo(discoveryLink) : false}
+      onOpenInCozy={
+        discoveryLink ? () => openExternalLink(discoveryLink) : false
+      }
       onCreateCozy={
         discoveryLink
           ? false
-          : () => redirectTo(getHomeLinkHref('sharing-drive'))
+          : () => openExternalLink(getHomeLinkHref('sharing-drive'))
```
Crash--: Can you add propTypes to this component ?  :D

PR #1420, opened 2019-01-31
fix: Don't make a request if we are public

Crash--:


Crash--:
et vu que tout est pété : 

Test Suites: 17 passed, 17 total
Tests:       90 passed, 90 total
Snapshots:   5 passed, 5 total
Time:        95.667s
Ran all test suites.
✨  Done in 105.79s.

PR #1426, opened 2019-02-01
fix: Change clipboard library

y-lohse:


src/sharing/components/ShareByLink.jsx
```diff
@@ -19,8 +19,8 @@ class ShareByLink extends React.Component {
     }
   }
 
-  copyLinkToClipboard = async () => {
-    await copy(this.props.link)
+  copyLinkToClipboard = () => {
+    copy(this.props.link)
```
Crash--: if(copy(this.props.link)) 
  Alerter.success(`${this.props.documentType}.share.shareByLink.copied`)
else
  Alerter.error ?

cedricmessiant: :+1:

PR #1427, opened 2019-02-01
Publication hotfix

y-lohse:
Not a real PR — just a branch of which we can run a new android release with a hotfix.


Crash--:
@y-lohse can we close ?

y-lohse:
Yup.

PR #1428, opened 2019-02-04
chore(travis): Add script to choose testing instance

CozyKourai:
all instances are in the script for now, should we put them in travis encypted env vars ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,17 +22,17 @@ stages:
   - build
 jobs:
   include:
-    - name: "Lint"
-      stage: "prebuild"
+    - name: 'Lint'
+      stage: 'prebuild'
       <<: *node
       script: yarn lint
-    - name: "Unit tests"
-      stage: "prebuild"
+    - name: 'Unit tests'
+      stage: 'prebuild'
       <<: *node
       script: yarn test
     - &build-web
-      name: "Drive web"
-      stage: "build"
+      name: 'Drive web'
+      stage: 'build'
       <<: *node
       env:
         - COZY_APP_SLUG=drive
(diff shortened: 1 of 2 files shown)


CozyKourai:
[(1) Choisir une instance de test dans travis](https://trello.com/c/Q1tm2Lut/1615-1-choisir-une-instance-de-test-dans-travis)

CozyKourai:
Ok on travis! 
- **continuous-integration/travis-ci/pr** show script returning testcafe0
- **continuous-integration/travis-ci/push** show script returning "not a PR"

scripts/deploy-tests
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
Crash--: can we add an extension? `deploy-tests.sh` ?

.travis.yml
```diff
@@ -50,6 +50,7 @@ script:
   - yarn tx
   - yarn test
   - yarn build:$COZY_APP_SLUG:browser
+  - ./scripts/deploy-tests
```
Crash--: No need to `chmod a+x` ?  :o

CozyKourai: I did `chmod` on the deploy-tests file locally, but it seems to run without it on travis

cozy-bank doesn't seem to do `chmod` either https://github.com/cozy/cozy-banks/blob/master/.travis.yml#L43

scripts/deploy-tests
CozyKourai: yes, we can.
I didn't do it because cozy-bank didn't do it ! (https://github.com/cozy/cozy-banks/blob/master/scripts/deploy-on-cozy)

.travis.yml
y-lohse: Yeah if the file is marked as executable before pushing it it works on travis too 👍

CozyKourai:
failed due to photos e2e tests

PR #1430, opened 2019-02-05
chore(testcafe): use testing instance from travis script

CozyKourai:


CozyKourai:
[(1) Choisir une instance de test dans travis](https://trello.com/c/Q1tm2Lut/1615-1-choisir-une-instance-de-test-dans-travis)

.travis.yml
```diff
@@ -44,8 +42,9 @@ jobs:
       script:
         - yarn tx
         - yarn build:$COZY_APP_SLUG:browser
-        - ./scripts/deploy-tests.sh
-        - test TRAVIS_PULL_REQUEST && yarn testcafe:$COZY_APP_SLUG
+        - ./scripts/choose-test-instance.sh
+        - source test.sh
```
y-lohse: I don't think this line is supposed to be here?

testcafe/tests/helpers/utils.js
```diff
@@ -1,7 +1,25 @@
...
+
+if (INSTANCE_TESTCAFE.includes('tools')) {
+  //Local server
+  TESTCAFE_PHOTOS_URL = `http://photos.${process.env.INSTANCE_TESTCAFE}`
+  TESTCAFE_DRIVE_URL = `http://drive.${process.env.INSTANCE_TESTCAFE}`
+} else {
+  //not local server!
+  TESTCAFE_PHOTOS_URL = `https://${process.env.INSTANCE_TESTCAFE.slice(
+    0,
+    9
```
y-lohse: I'm not super comfortable with slicing strings using magic number like that. It's not very easy to read and If we rename the test instances this will break.
Something [like this](https://repl.it/@y_lohse/DistinctRecentBinary) would be better I think.

.travis.yml
```diff
@@ -44,8 +42,9 @@ jobs:
...
+        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then yarn testcafe:$COZY_APP_SLUG; fi
```
y-lohse: 👍

.travis.yml
CozyKourai: Actually, I need to source test.sh to get my var, as travis doesn't seems to retain var from script : https://github.com/travis-ci/travis-ci/issues/7472#issuecomment-291363061 and https://stackoverflow.com/questions/34021558/travis-ci-using-environment-variables-in-the-deploy-section-of-the-travis-yml

scripts/choose-test-instance.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
nono: Do you know [shellcheck](https://github.com/koalaman/shellcheck)? It's a tool to check common errors in shell scripts and make them more robust.

I also think that using `set -eu` for this script could be safer. See https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/ for an explanation.

scripts/choose-test-instance.sh
```diff
@@ -0,0 +1,9 @@
...
+
+INSTANCE_TESTCAFE="None"
+if [ "$TRAVIS_PULL_REQUEST" != "false" ];
+then
+ let INSTANCE_NUM=$TRAVIS_PULL_REQUEST%4;
+ INSTANCE_TESTCAFE="testcafe"$INSTANCE_NUM".cozy.rocks";
+fi
+echo "export INSTANCE_TESTCAFE=$INSTANCE_TESTCAFE" > setinstance.sh
```
nono: I think you can avoid using a temporary script by using `export INSTANCE_TESTCAFE="$INSTANCE_TESTCAFE"` here, and calling this script from `.travis.yml` like this `- . ./scripts/choose-test-instance.sh`. The `.` tells the shell to run the script inside the current process, instead of creating a sub-shell.

PR #1438, opened 2019-02-11
Feat: New cozy creation

Crash--:
- New Onboarding Process


Crash--:
@y-lohse @cedricmessiant I think you can start to review this PR. 

There is : 
- some console.log() but I'll remove them later after a final final final test because it's pretty boring to remove them test and rewrite them later to debug... 
- I'll write some tests, specially for the new `onboarding` file 
- I have used arrow function in a class, and our guideline say to not do so because it's harder to test. The alternative is to bind all the functions to this in the constructor ?

y-lohse:
> * I have used arrow function in a class, and our guideline say to not do so because it's harder to test. The alternative is to bind all the functions to this in the constructor ?

Haven't started reviewing yet, but yes.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,119 @@
...
+  getAccessToken,
+  readSecret,
+  clearSecret,
+  clearState,
+  checkExchangedInformations,
+  checkIfOnboardingLogin
+} from './src/utils/onboarding'
+
+class MobileRouter extends Component {
+  async checkState(receivedState, code, cozy_url, history) {
```
y-lohse: What is `receivedState `? We're not using it at all as far as I can tell.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,119 @@
...
+    console.log({ localSecret })
+    console.log({ localState })
+    try {
+      const clientInfo = await secretExchange(
+        localSecret,
+        cozy_url,
+        this.props.client
+      )
+
+      const { onboarding_secret, onboarding_state } = clientInfo
```
y-lohse: I understand what we are doing, but not why. We send a secret and receive the same secret back? What is the point of that? What is the difference between `state` and `secret`?

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,119 @@
...
+          localState,
+          onboarding_state
+        )
+      )
+        throw new Error('ERROR', 'exchanged informations are not good')
+
+      const getTokenRequest = await getAccessToken(
+        clientInfo,
+        cozy_url,
+        code,
```
y-lohse: i also thought we would get the access token in exchange for an information that only the app and the stack knows, for example the `secret` or the `state`. But here we seem to be using the `code` which can be found in the email? Is this correct? Can we call this route without exchanging secrets first?

src/drive/targets/mobile/index.jsx
```diff
@@ -246,6 +183,86 @@ var app = {
...
+      /*Since we can update our OauthConfig sometimes, we need to
+        override the cached one */
+      realOauthOptions =
+        clientInfos !== null ? { ...clientInfos, ...getOauthOptions() } : null
+      const token = getToken(store.getState())
+      const stackClient = client.getStackClient()
+
+      stackClient.setOAuthOptions(realOauthOptions)
+      stackClient.setCredentials(token)
+      await restoreCozyClientJs(client.options.uri, realOauthOptions, token)
```
y-lohse: I think the `await` has become useless here.

package.json
```diff
@@ -9,6 +9,7 @@
     "watch:drive": "npm run watch:drive:browser",
     "watch:drive:browser": "COZY_APP_SLUG=drive cozy-scripts watch --manifest src/drive/targets/manifest.webapp --build-dir build/drive --src-dir src/drive",
     "watch:drive:mobile": "COZY_APP_SLUG=drive NODE_ENV=mobile:development cozy-scripts watch --manifest src/drive/targets/manifest.webapp --src-dir src/drive",
+    "watch:drive:mobile:hot": "COZY_APP_SLUG=drive NODE_ENV=mobile:development  cozy-scripts start --hot --mobile --manifest src/drive/targets/manifest.webapp --src-dir src/drive ",
```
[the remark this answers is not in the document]
Crash--: Not yet :(

src/authentication/MobileRouter.jsx
Crash--: Oh, good catch! We need to check the receivedState with our localState before calling the `secretExchange`

src/authentication/MobileRouter.jsx
Crash--: The idea is : 
- The app creates a secret 
- The app gives this secret to the onboarding page 
- Our cloudery creates the oauth client and give us back the secret
- We check if the secret is the same than before 
- if yes, we exchange the secret to get the token

src/authentication/MobileRouter.jsx
Crash--: We are using the `clientInfo` and the clientInfo contains : `onboarding_secret` &  `onboarding_state`

We can call this route without exchanging secret first, but you'll get an error

y-lohse: Ok, this is not super obvious from the code but it's the behaviour I expected 👍

src/authentication/MobileRouter.jsx
y-lohse: Mmmh ok. And what is the difference between state and secret ?

Crash--: Once is used in our side, the stack checks the other

package.json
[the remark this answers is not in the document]
y-lohse: I think we can remove it in that case, no?

src/authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,125 @@
...
+  return encodeURIComponent(JSON.stringify(oauthStuff))
+}
+
+const getCozyUrl = cozy_url => {
+  return `https://${cozy_url}`
+}
+export const secretExchange = (secret, cozy_url, client) => {
+  const response = client.stackClient.fetchJSON(
+    'POST',
+    getCozyUrl(cozy_url) + '/auth/secret_exchange',
```
y-lohse: I'd rename the function `addProtocolToURL` or something like that, `getCozyUrl(cozy_url)` just looks weird 😛

src/drive/targets/mobile/index.jsx
```diff
@@ -205,37 +153,39 @@ var app = {
       persistedState
     )
     return this.store
-  },
+  }
 
-  onDeviceReady: async function() {
+  onDeviceReady = async () => {
+    console.log('onDeviceReady')
+    if (this.isStarting === true) {
```
y-lohse: We probably also want to return if `stardedApp ` is `true` ?

src/drive/targets/mobile/index.jsx
```diff
@@ -246,6 +196,89 @@ var app = {
       })
     }
   }
-}
 
-app.initialize()
+  startApplication = async () => {
+    console.log('StartApplication', console.trace())
+    if (this.stardedApp) return
```
y-lohse: I think we could use `appReady` instead of `stardedApp`, by resolving the promise with `true` for example. Then `if (this.appReady === true)` would work and not require an extra state-tracking variable.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,128 @@
...
+        clientInfo,
+        router: history
+      })
+      console.log('afterAuth', afterAuth)
+      clearState()
+      clearSecret()
+    } catch (error) {
+      console.log('error', error)
+      clearState()
+      clearSecret()
```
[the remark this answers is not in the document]
Crash--: I don't know if I can use `finally` here since I can dispatch `LOGOUT`. And I don't know, if in this case, the `finally` will be executed. If you are sure about it, then yes I can use the finally. 

No need to `await` since I don't need their return value or something else

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,128 @@
...
+      onboarding,
+      onboardingInformations
+    } = this.props
+
+    if (!isAuthenticated) {
+      if (checkIfOnboardingLogin(onboardingInformations)) {
+        /* We need to hide() the ViewController since the ViewController is still active 
+        when the application cames from background (specialy on iOS)
+        */
+        window.SafariViewController.hide()
```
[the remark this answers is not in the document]
Crash--: MobileRouter is only called in a cordova context.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,128 @@
...
+        const { code, state, cozy_url } = onboardingInformations
+        this.checkState(state, code, cozy_url, history)
```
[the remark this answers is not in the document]
Crash--: It's not pretty clear, but if checkState failed, I dispatch a `logout` action. A kind of complet `reset` of the app. So the return will not be called, right?

src/authentication/src/Authentication.jsx
```diff
@@ -50,9 +51,16 @@ class Authentication extends Component {
...
       const cozyClient = this.context.client
       const { client, token } = await register(cozyClient, url)
-      onComplete({
+
+      const destructuredToken = {
+        accessToken: token.accessToken,
+        refreshToken: token.refreshToken,
+        scope: token.scope,
+        tokenType: token.tokenType
+      }
```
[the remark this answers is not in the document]
Crash--: This is a bit awkward, but we have 2 situations : 
- Conventionnel login 
- Autologin with deeplink 

Once return us an object type : `AccessToken` (something like : https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/OAuthClient.js#L327) 

But the second only give us the same information but without the object type. 

So in order to not have to deal with this difference in the saveCredential() method (https://github.com/cozy/cozy-drive/pull/1438/files/9d919ccef65ab555eca7796e88eadf5c5597060f#diff-faee20f502e0c8af9ec0f733df03cd7bR31) I've done it here after discussion with ya.

src/authentication/src/steps/ButtonLinkRegistration.jsx
```diff
@@ -1,32 +1,77 @@
...
+      clientName,
+      redirectURI,
+      softwareID,
+      softwareVersion,
+      clientURI,
+      logoURI,
+      policyURI,
+      scope
+    })
+    const url = `https://staging-manager.cozycloud.cc/cozy/create?domain=cozy.works&pk_campaign=drive-${getPlatform() ||
```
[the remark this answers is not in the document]
Crash--: Good catch! I need to be sure that this new onboarding is already deployed on our production env. Before changing this URL

src/authentication/src/steps/Welcome.jsx
```diff
@@ -31,6 +32,7 @@ export class Welcome extends Component {
       <ButtonLinkRegistration
         label={t('mobile.onboarding.welcome.create_my_cozy')}
         size={isMobile ? 'normal' : 'large'}
+        onboarding={onboarding}
```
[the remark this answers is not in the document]
Crash--: All you see in src/authentication/ will be soon a cozy-lib

src/authentication/src/steps/test.json
```diff
@@ -0,0 +1,25 @@
...
+      "permissions": [
+        "io.cozy.contacts",
+        "io.cozy.files:GET:io.cozy.files.music-dir",
+        "io.cozy.jobs:POST:sendmail:worker"
+      ],
+      "secret": "test",
+      "state": "test"
+    }
+  }
+}
```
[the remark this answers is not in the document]
Crash--: i'll remove this file. No need it anymore 👍

src/authentication/MobileRouter.jsx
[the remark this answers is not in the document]
cedricmessiant: ok, I see!

src/drive/mobile/modules/authorization/duck/selectors.js
```diff
@@ -0,0 +1,27 @@
...
+
+const getProp = (state, key) => {
+  if (
+    state.mobile &&
+    state.mobile.authorization &&
+    state.mobile.authorization[key] !== undefined
+  ) {
+    return state.mobile.authorization[key]
+  }
+  console.warn(`Authorization prop not found: ${key}`)
```
cedricmessiant: if there are `console.{log,warn}` that will stay in the code, please add `// eslint-disable no-console` to avoid useless warnings

src/drive/targets/mobile/index.jsx
```diff
@@ -205,37 +153,39 @@ var app = {
...
+      return
+    }
+    this.isStarting = true
+    console.log('afeter isStarting')
     const store = await this.getStore()
-    const client = await this.getClient()
-    const polyglot = await this.getPolyglot()
-
+    this.startApplication()
+    await this.appReady
```
cedricmessiant: I don't understand this. What is this for?

src/drive/targets/mobile/index.jsx
```diff
@@ -246,6 +196,89 @@ var app = {
...
+        <CozyProvider store={store} client={client}>
+          <DriveMobileRouter history={hashHistory} onboarding={onboarding} />
+        </CozyProvider>
+      </I18n>,
+      root,
+      () => {
+        console.log('should Resolve')
+        this.stardedApp = true
+        this.isStarting = false
+        this.resolvePromise()
```
cedricmessiant: ok, I think I understand `this.appReady` now, sorry!

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,128 @@
...
+      console.log({ onboarding_state })
+      if (
+        !checkExchangedInformations(
+          localSecret,
+          onboarding_secret,
+          localState,
+          onboarding_state
+        )
+      )
+        throw new Error('ERROR', 'exchanged informations are not good')
```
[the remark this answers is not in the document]
cedricmessiant: (il y en avait un ou deux autres que je n'ai pas relevé il me semble)

src/authentication/src/steps/Welcome.jsx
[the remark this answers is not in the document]
cedricmessiant: cool !

src/authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,125 @@
...
+}
+const generateState = () => {
+  return generateRandomString()
+}
+const generateSecret = () => {
+  return generateRandomString()
+}
+export const checkIfOnboardingLogin = onboardingInformations => {
+  console.log('checkfif', onboardingInformations)
+  if (onboardingInformations.code !== null) return true
```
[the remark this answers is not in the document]
Crash--: It should not be possible: https://github.com/cozy/cozy-drive/pull/1438/files#diff-df4befc41afaffe2333725bb24e8cdc4R14

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,131 @@
...
-      <Revoked
-        router={history}
-        onLogBackIn={onAuthenticated}
-        onLogout={onLogout}
-      />
-    )
-  } else {
-    return <Router history={history}>{appRoutes}</Router>
+      const getTokenRequest = await getAccessToken(
+        clientInfo,
```
y-lohse: I think here we could pass the required infos directly, rather than the opaque `clientInfo ` object.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,131 @@
...
+        cozy_url,
+        code,
+        this.props.client
+      )
+      const token = await getTokenRequest.json()
```
y-lohse: Shouldn't we do the `.json()` inside `getAccessToken` ?

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,131 @@
...
+      if (getTokenRequest.status !== 200) {
+        throw new Error('token.error')
+      }
+
+      await this.props.onAuthenticated({
+        url: `https://${cozy_url}`,
```
y-lohse: Nit, but also kind of not nit: I think we should rename this variable. Not only is it snake cased and the linter should be yelling at it, but it's also not a fully formed URL and we have to add the protocol here and there. I think `instanceName` or `instanceDomain` would be better.

src/authentication/MobileRouter.jsx
```diff
@@ -1,39 +1,131 @@
...
+      )
+    } else {
+      return <Router history={history}>{appRoutes}</Router>
+    }
   }
 }
 
+MobileRouter.propTypes = {
+  onboarding: onboardingPropTypes.isRequired,
+  onboardingInformations: onboardingInformationsPropTypes.isRequired,
```
y-lohse: Now that most of the review is behind us, I have some more nits — I think `onboarding ` and `onboardingInformations ` are not so great names.
`onboarding` should be something like `oauthInfos` I think, and `onboardingInformations ` maybe `autloginInformations`. But none of them really do much onboarding.

src/authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,124 @@
...
+  return generateRandomString()
+}
+const generateSecret = () => {
+  return generateRandomString()
+}
+export const checkIfOnboardingLogin = onboardingInformations => {
+  return onboardingInformations.code !== null
+}
+
+export const checkExchangedInformations = (
```
y-lohse: Nit: I would personally inline this function, it does nothing complicated and putting it behind a function makes the thing a bit more opaque.

src/authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,124 @@
...
+    secret = generateSecret()
+    await writeSecret(secret)
+  }
+  let state = await readState()
+
+  if (!state) {
+    state = generateState()
+    await writeState(state)
+  }
+  const oauthStuff = {
```
y-lohse: Nit: `oauthData` instead of `oauthStuff `? 😛

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -92,7 +91,9 @@ export function resetClient(client, clientInfo = null) {
     cozy.client.offline.destroyAllDatabase()
   }
   // reset cozy-client
-  client.getClient().resetClientId()
+  client.getStackClient().resetClientId()
+  client.getStackClient().setUri('')
+  client.getStackClient().setCredentials(null)
```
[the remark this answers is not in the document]
Crash--: See : https://github.com/cozy/cozy-client/pull/376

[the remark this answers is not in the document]
cedricmessiant: :+1:

PR #1441, opened 2019-02-12
chore: Use chrome for photo integration tests

y-lohse:
Should prevent [this kind](https://travis-ci.org/cozy/cozy-drive/jobs/492082141#L1417) of error. But we should probably use the same configuration for drive and photos?


y-lohse:
Ok, in that case, should be disable integration tests on the CI? We still have several cases where they fail although everything is fine, and it looks like resolving these different issues will still take a while?

PR #1443, opened 2019-02-12
fix: Catch exception on public page

Crash--:
I know this is not the best solution to this problem but let's create a new card for that. 

Here, we just want to catch the exception if the call is made on a public page.

### DIFF
--- a/src/viewer/CallToAction.jsx
+++ b/src/viewer/CallToAction.jsx
@@ -1,0 +1,65 @@
+import React, { Component } from 'react'
+import Icon from 'cozy-ui/react/Icon'
+
+import localforage from 'localforage'
+import {
+  isClientAlreadyInstalled,
+  isLinux,
+  NOVIEWER_DESKTOP_CTA
+} from 'components/pushClient'
+import styles from './styles'
+
+export default class CallToAction extends Component {
+  state = {
+    mustShow: false
+  }
+
+  async componentDidMount() {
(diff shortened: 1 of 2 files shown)


src/viewer/CallToAction.jsx
```diff
@@ -0,0 +1,65 @@
...
+  NOVIEWER_DESKTOP_CTA
+} from 'components/pushClient'
+import styles from './styles'
+
+export default class CallToAction extends Component {
+  state = {
+    mustShow: false
+  }
+
+  async componentWillMount() {
```
cedricmessiant: `componentWillMount` is now unsafe/deprecated, we will need to change this when we move to React (https://reactjs.org/docs/react-component.html#unsafe_componentwillmount ). In this case though, I don't really see how we can do it nicely. Maybe we can use `componentDidMount` but since we already have performance issues due to multiple rendering, it may be a bad idea.

Crash--: I edited the code, but I don't think it was the responsibility of this PR to do it. I already made an effort to separate the component in order to be able to edit it later so... ;)

Crash--:
[(S) SCN - partage par lien d'un fichier](https://trello.com/c/mQzLJmjK/1597-s-scn-partage-par-lien-dun-fichier)

Crash--:
[Vue publique de partage - exceptions](https://trello.com/c/ppQtpCRs/1642-vue-publique-de-partage-exceptions)

PR #1444, opened 2019-02-13
fix: Upgrade react-mesure to fix resizeObserver loop

Crash--:
This commit is exactly what we wanted : https://github.com/souporserious/react-measure/pull/118/files ;) 

We should be able to run our photos tests on chrome now


Crash--:
[Réussir à faire passer les tests de Photos sur Chrome](https://trello.com/c/ptscPVyK/1650-r%C3%A9ussir-%C3%A0-faire-passer-les-tests-de-photos-sur-chrome)

PR #1447, opened 2019-02-14
fix: Make the textviewer handle less files

y-lohse:
The `class` attribute will report `text` even for complex text formats like docx etc. I added mime type detection to narrow the file formats we display using the viewer, even though it can probably handle more than that.


y-lohse:
[Viewer - loader infini, onglet plante (Chrome)](https://trello.com/c/jrKCu7xT/1654-viewer-loader-infini-onglet-plante-chrome)

PR #1454, opened 2019-02-19
fix: Upgrade cozy-ui

y-lohse:
No breaking changes for us because:

- We still use a custom avatar component, so the size changes don't affect us yet
- We don't use the stylesheet from cozy-ui (because I don't think we need it? We import the stuff we really want.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "cozy-konnector-libs": "4.11.4",
     "cozy-realtime": "1.1.0",
     "cozy-scripts": "1.11.0",
-    "cozy-ui": "17.9.0",
+    "cozy-ui": "19.0.1",
     "create-react-context": "0.2.2",
     "date-fns": "1.29.0",
     "diacritics": "1.3.0",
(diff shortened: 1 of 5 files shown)


y-lohse:
[update cozy-UI](https://trello.com/c/Hz8NcJpL/1674-update-cozy-ui)

y-lohse:
> Does't it also break the Auth screen ? (I upgraded to v18 in my branch and the screens are not centered anymore).

You're right, good catch, it broke the auth screen. @GoOz in https://github.com/cozy/cozy-ui/pull/829#issue-250403112 we changed the wizard alignment on mobile — does this mean some wizard screens must be top-aligned, and others must be centered (like our login screen) ?

GoOz:
Most wizard screens on mobile are top aligned, but some are centered.
Basically I think the rule is: if there's an element with a user interaction such as form elements, it should be top aligned (because of the keyboard), if not then it should be centered.

y-lohse:
Ok thanks! And if we want to center the screen, should we just override the necessary rules? Or are there other classes to use?

GoOz:
Oh yeah sorry I meant to fix that but… well I forgot. You should add the class `.wizard-wrapper--center` to the element with the class `wizard-wrapper`. I think that's pretty much it.

PR #1455, opened 2019-02-19
 chore(travis): wait for rundeck to finish deploy before tests

CozyKourai:
WIP (as I want to check the travis execution), but already a question for review : 

For now, we have 2 files and 3 commands : 
- `./scripts/choose-test-instance.sh` (executed in script, it set an envirronement variable for testcafe)
- `downcloud.js` (in after success)
- `./scripts/deploy-on-instance.sh` (in after success)

`./scripts/deploy-on-instance.sh` does : 
- deploy app on instance using rundeck
- check for rundeck status
- when status is "successful', launch testcafé

Tell me if this is ok (I'll rename deploy-on-instance to deploy-and-test or something), or I need to merge/split some files.


y-lohse:
I think the best would be to have one script for downcloud, and one for the tests — the one for the tests would choose the instance, deploy, check for the status and launch the tests.

CozyKourai:
[(𝐒) Détecter la fin de la tâche rundeck pour savoir quand l'installation est finie](https://trello.com/c/X1fLdEHC/1618-%F0%9D%90%92-d%C3%A9tecter-la-fin-de-la-t%C3%A2che-rundeck-pour-savoir-quand-linstallation-est-finie)

.travis.yml
```diff
@@ -51,8 +51,11 @@ jobs:
         - yarn build:$COZY_APP_SLUG:browser
         - ./scripts/choose-test-instance.sh
         - source setinstance.sh
-      before_deploy:
         - &downcloud-cert if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then eval "$(ssh-agent -s)" && chmod 600 /tmp/id_rsa_travis_downcloud && ssh-add /tmp/id_rsa_travis_downcloud; fi # Needs to be inlined and not in a script to work.
+        - *downcloud-cert
+        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then APP_URL=$(node scripts/downcloud.js build/$COZY_APP_SLUG); fi
+        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then ./scripts/deploy-on-instance.sh $APP_URL; fi
```
Crash--: - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then 
APP_URL=$(node scripts/downcloud.js build/$COZY_APP_SLUG)
./scripts/deploy-on-instance.sh $APP_URL;
fi

curl.out
```diff
@@ -0,0 +1,23 @@
...
+        <option name='slug' value='photos' />
+      </options>
+    </job>
+    <description>cd /root/gozy-cluster-controller/ &amp;&amp; /root/.virtualenvs/cozy/bin/python ./gozy-cli install_application "${option.slug}" "${option.source}" -i "${option.instance}" -f</description>
+    <argstring>-instance testcafe.cozy.rocks -slug photos -source https://downcloud.cozycloud.cc/upload/photos/mobile/1.15.3-e19ee9cd80ed1fcc040f06d793552467d7068a5b/photos.tar.gz</argstring>
+    <successfulNodes>
+      <node name='gozy-adm-stg' />
+    </successfulNodes>
+  </execution>
+</executions>
```
Crash--: À supprimer ? On peut pas écrire sur le `stdout` et lire à partir de là ?

CozyKourai: oups! that was not supposed to be in the commit!

Anyway, I used stdout, and there are no more temp file!

CozyKourai:
**continuous-integration/travis-ci/pr** is failed because photo tests failed!
We can check the passing tests (drive) and the failing tests (photos)

**continuous-integration/travis-ci/push** is unaffected by change from this PR

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# return status for an executions
+function getRundeckStatus {
+  EXEC_ID=$1
+  local __resultvar=$2
+
+STATUS=$(curl -s -X GET \
+     -H "X-Rundeck-Auth-Token: $RUNDECK_TOKEN" \
+     -H "Content-Type: application/json" https://rundeck.cozycloud.cc/api/27/execution/$EXEC_ID | grep -o -P "(?<=status=').*(?=' )")
+eval $__resultvar="$STATUS"
```
y-lohse: Maybe it's because of my limited bash-scripting knowledge, but I think we need to decide if we want to keep using functions in this script, and if so, how we manage our variables.

I don't think we need the functions — `executeRundeck ` is only called in one place and could be inlined. `getRundeckStatus ` is called twice, but as mentioned in another comment, I think we could call it only once. So my preference would be to not use functions at all and inline the curl calls as much as possible.

If for some reason we want to keep the functions around, the variables usage needs to be cleared up:
- the function gets some variables as parameters, but takes others from the global scope (`RUNDECK_TOKEN `) — we should do one or the other, but not both
- we are creating new global variables (`EXEC_ID` and `STATUS`), but we are updating `JOB_STATUS` with the `eval` trick — again, either this function creates global variables, or it doesn't, but doing both is inconsistent.

So if we keep the function, I would suggest we rely only on global variables and try to not create too many.

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "↳ ℹ️  Deploying $COZY_APP_SLUG ($1) on $INSTANCE_TESTCAFE:"
+EXECUTION_ID=$(executeRundeck $INSTANCE_TESTCAFE $JOB_UPDATE_STAGING $COZY_APP_SLUG $1 | grep -o -P "(?<=execution id=').*(?=' href)")
+echo $EXECUTION_ID
+
+# wait for execution
+sleep 1 && getRundeckStatus $EXECUTION_ID JOB_STATUS && echo "↳ ℹ️  Job $JOB_STATUS"
+
+# wait a little more if need for exec to end
+while [  $JOB_STATUS == 'running' ]; do
+  sleep 1 && getRundeckStatus $EXECUTION_ID JOB_STATUS && echo "↳ ℹ️  Job $JOB_STATUS"
```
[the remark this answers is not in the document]
CozyKourai: I remove the line before the loop, but then I add to initialize JOB_STATUS  to 'running' before, otherwise it doesn't enter the loop

scripts/deploy-and-test-on-instance.sh
CozyKourai: I remove both functions, hence removing a lot of variables. 
I agree it's "better looking' that way.

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -eu
+
+JOB_UPDATE_STAGING="8d452c26-5a39-4162-b3b8-4da5efdf0b76"
+
+RUNDECK_TOKEN='2hE42p5phY1qSJ0YJUch2AqnYbXrhDec'
+INSTANCE_TESTCAFE='testcafe.cozy.rocks'
```
y-lohse: I think this line and the next 2 need to me removed, no?

CozyKourai: Indeed....

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -eu
+
+JOB_UPDATE_STAGING="8d452c26-5a39-4162-b3b8-4da5efdf0b76"
```
Crash--: is it always the same? This is a global token for the .rock env?

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,37 @@
...
+     -d "{\"argString\":\"-instance $INSTANCE_TESTCAFE -slug $COZY_APP_SLUG -source $1 \"}"  \
+     https://rundeck.cozycloud.cc/api/27/job/$JOB_UPDATE_STAGING/run | grep -o -P "(?<=execution id=').*(?=' href)")
+echo $EXECUTION_ID
+
+#init JOB_STATUS
+JOB_STATUS='running'
+
+# wait for exec to end
+while [  $JOB_STATUS == 'running' ]; do
+  sleep 1
```
Crash--: 1 second is pretty agressive. 10 ? 20 ? 30 ? 60 ?

scripts/deploy-and-test-on-instance.sh
CozyKourai: Always the same for the `[cozy.rocks - Staging] Install app on instance` job. 
We'll need another key if we want to run other job.
Do you want me to put an envirronement var in travis ?

y-lohse: Yep, it's the job's ID, never changes.

scripts/deploy-and-test-on-instance.sh
CozyKourai: well for what I've seen for now, it take about 4 to 6 seconds to run. So I can change to 5, but I think 10 or more will be too long!

y-lohse: I think the rundeck UI can tell us what the average runtime for this job is, and we could pick a value that's about the same :)

y-lohse: Agree that 10 would be too long, 5 sounds good if that's the average job runtime!

CozyKourai: Yep, I just checked, It's more between 5 and 8 seconds (I was a little optimistic) for the many runs I did yesterday while doing the dev

Crash--: Go for 5 or 10 ;)

scripts/deploy-and-test-on-instance.sh
```diff
@@ -0,0 +1,37 @@
...
+       -H "X-Rundeck-Auth-Token: $RUNDECK_TOKEN" \
+       -H "Content-Type: application/json" https://rundeck.cozycloud.cc/api/27/execution/$EXECUTION_ID | grep -o -P "(?<=status=').*(?=' )")
+  echo "↳ ℹ️  Job $JOB_STATUS"
+done
+
+
+if [  $JOB_STATUS == 'succeeded' ] ; then
+   echo "↳ ✅ Execution $EXECUTION_ID succeeded"
+   #Launch tests on newly install server
+   yarn testcafe:$COZY_APP_SLUG;
```
Crash--: I'm wondering if we have to add a sleep here. 

We deploy a new version, and we know that we have some cache on the registry and the installation take some time, no ? ping @y-lohse

scripts/deploy-and-test-on-instance.sh
Crash--: Does staging also contains `.red` domain for instance? 

This is just a question to know if we will need to work on it later to be able to deploy on other "env"

CozyKourai: I'm not sure, let me check with infra.
Anyway if we want to deploy to INT ie, we just need to change (or add) this job, like they did in the old cozy-bank script (https://github.com/cozy/cozy-banks/commit/aa36466d5fc870a8cd659cb32160546bc7ad9e04)

scripts/deploy-and-test-on-instance.sh
y-lohse: Yeah but here we aren't using the registry to do the installation in this case, we're forcing it through rundeck. So no cache, and probably no sleep necessary :)

scripts/deploy-and-test-on-instance.sh
CozyKourai: `.red` is on INT, not stagging, so we'll need another token 
rundeck jobs should run on `.red` but Seb told me nobody uses it yet!

Crash--: Thanks!

scripts/deploy-and-test-on-instance.sh
Crash--: 🤞

PR #1459, opened 2019-02-21
 feat: Add a limit to the clustering 

Gara64:
This is the first PR of an incoming serie about the clustering execution by batch.

### DIFF
--- a/src/photos/targets/services/onPhotoUpload.js
+++ b/src/photos/targets/services/onPhotoUpload.js
@@ -1,4 +1,4 @@
-import { log } from 'cozy-konnector-libs'
+import { cozyClient, log } from 'cozy-konnector-libs'
 
 import {
   getChanges,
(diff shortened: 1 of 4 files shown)


src/photos/targets/services/onPhotoUpload.js
```diff
@@ -137,27 +142,40 @@ const recomputeParameters = async setting => {
   }
   log('info', `Compute clustering parameters on ${files.length} photos`)
 
-  const dataset = prepareDataset(files)
+  const dataset = await prepareDataset(files)
```
y-lohse: The function wasn't async before?

src/photos/targets/services/onPhotoUpload.js
```diff
@@ -168,27 +142,40 @@ const recomputeParameters = async setting => {
...
-const onPhotoUpload = async () => {
-  log('info', `Service called with COZY_URL: ${process.env.COZY_URL}`)
-
-  let setting = await readSetting()
-  const lastSeq = setting ? setting.lastSeq : 0
-
-  const changes = await getChanges(lastSeq)
+const runClustering = async (setting, since, filteredIds) => {
+  const changes = await getChanges(since, CHANGES_RUN_LIMIT, filteredIds)
   if (!changes || changes.photos.length < 1) {
```
y-lohse: I don't think `changes ` can return nothing, so the condition could be simplified, no?

src/photos/targets/services/onPhotoUpload.js
Gara64: Oh right, this function turns async in my modifications, but it is not part of this PR. I change this.

PR #1465, opened 2019-02-25
feat: Link the mobile and web Drive applications

nono:
See https://docs.cozy.io/en/cozy-stack/auth/#linked-applications


Crash--:
does it break already registered OAuth client?

nono:
> does it break already registered OAuth client?

Nope, the already registered OAuth clients will still work. They just won't be linked to the web applications: we will fix that later via a fixer in the stack (https://docs.cozy.io/en/cozy-stack/cli/cozy-stack_fixer_link-app/).

Crash--:
And I don't know if this is still used, but we need this SOFTWARE_ID to be equal to the bundle name for play store rating: https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/modules/settings/RatingModal.jsx#L125

PR #1466, opened 2019-02-25
fix: Don't crash when album is not found

y-lohse:
When the request album is not found (because it has been deleted, or the id is wrong, or some other reason), we don't want to render it.

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -6,6 +6,7 @@
     ".svg$": "<rootDir>/jestHelpers/mocks/iconMock.js",
     "styles": "identity-obj-proxy",
     "^drive/(.*)": "<rootDir>/src/drive/$1",
+    "^photos/(.*)": "<rootDir>/src/photos/$1",
     "^sharing(.*)": "<rootDir>/src/sharing$1",
     "^react-cozy-helpers(.*)": "<rootDir>/src/lib/react-cozy-helpers$1",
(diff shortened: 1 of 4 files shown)


y-lohse:
[Photo : promise rejection lors de la suppression d'un album](https://trello.com/c/U1I4sL1Z/1666-photo-promise-rejection-lors-de-la-suppression-dun-album)

y-lohse:
The only halfway relevant test I can think of would be to snapshot test `ConnectedAlbumPhotos`, but it's a `Query` component so it's quite hard to test and involves lots of changes compared to the change. Plus this fix is already here to make a test work. Is there a specific way of testing this that you had in mind?

cedricmessiant:
With this change, you will render a loader if the album is not here. Is that what we want?
About the test, you could separate the function that is in the brackets and test it. Not very pretty. It would better to have mock `withRouter` and `Query` but would require some work.

NB: In another project, for this pattern (render a loader until we have data), I created a HOC named `withLoadingIndicator`

y-lohse:
> With this change, you will render a loader if the album is not here. Is that what we want?

For now, yes.

I added some tests and had to do some tiny refactoring to make them work, probably bets to replay the PR commit by commit and see if you still ✅it.

Crash--:
We can create a new card to deal with this loader. 

The main idea was to unlock @CozyKourai at the beginning. Since now we have tests, I still   ✅

PR #1470, opened 2019-02-27
fix: Show album photos

y-lohse:
I previously [removed a spreaded `props`](https://github.com/cozy/cozy-drive/pull/1466/commits/c799b9cb0c817ed9a2d3c90fcdfa6d42cc978097#diff-b99860e2348055cdd5e05388380200ddL149) — most of it is react-router stuff that is unnecessary, but the `children` prop is important as it is used to render the photo viewer / photo picker.

### DIFF
--- a/src/photos/ducks/albums/index.jsx
+++ b/src/photos/ducks/albums/index.jsx
@@ -130,7 +130,7 @@ const ConnectedAddToAlbumModal = props => (
   </Query>
 )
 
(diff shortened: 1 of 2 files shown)


y-lohse:
[Photo : promise rejection lors de la suppression d'un album](https://trello.com/c/U1I4sL1Z/1666-photo-promise-rejection-lors-de-la-suppression-dun-album)

PR #1472, opened 2019-02-28
tests: Automatic login in MobileRouter

y-lohse:
I think we could improve these tests, especially the ones that check mismatching secrets and states, by returning something testable from the function in MobileRouter. Maybe the error ?

### DIFF
--- a/src/authentication/MobileRouter.spec.jsx
+++ b/src/authentication/MobileRouter.spec.jsx
@@ -1,0 +1,257 @@
+import { MobileRouter } from './MobileRouter'
+import React from 'react'
+import { shallow } from 'enzyme'
+import { writeState, writeSecret } from './src/utils/onboarding'
+
+describe('MobileRouter', () => {
+  it('should render the appRoutes when all is well', () => {
+    const app = shallow(
+      <MobileRouter
+        appRoutes={<div />}
+        isAuthenticated={true}
+        isRevoked={false}
+        onboarding={{}}
+        onboardingInformations={{}}
+        history={{}}
+        onAuthenticated={jest.fn()}
+        onLogout={jest.fn()}
+        appIcon={''}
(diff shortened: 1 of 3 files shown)


src/authentication/MobileRouter.spec.jsx
```diff
@@ -0,0 +1,227 @@
...
+    )
+    expect(app).toMatchSnapshot()
+  })
+
+  it('should render nothing when onboarding', () => {
+    const onboardingInformations = {
+      code: '123',
+      state: 'abc',
+      cozy_url: 'http://lol.com'
+    }
```
cedricmessiant: maybe check that `doOnboardingLogin` is called with the right params here?

src/authentication/MobileRouter.spec.jsx
```diff
@@ -0,0 +1,227 @@
...
+      client = {
+        stackClient: {
+          fetchJSON: jest.fn(),
+          fetch: jest.fn()
+        }
+      }
+      state = '123'
+      code = '111222333'
+      url = 'instance.cozy.test'
+      history = jest.fn()
```
cedricmessiant: you could define all the spies in the `describe` function and just call `jest.clearAllMocks()` or `jest.resetAllMocks()` in `afterEach`?

src/authentication/MobileRouter.spec.jsx
```diff
@@ -0,0 +1,227 @@
...
+        onboarding_state: '123'
+      })
+      client.stackClient.fetch.mockResolvedValue({
+        json: async () => {
+          return {
+            error: 'test error'
+          }
+        },
+        status: 500
+      })
```
cedricmessiant: maybe also add a test to test the case where the promise is rejected?

PR #1473, opened 2019-03-01
fix: Viewer on public album

y-lohse:
Any idea for relevant tests? Shallow-rendering snapshot gives:

```
<div>
  <Main
    className="u-pt-1-half"
  >
    <Selection />
    <IconSprite />
  </Main>
</div>
```

…which is pretty useless. And I don't think deep-rendered snapshot are very useful to catch regressions.

### DIFF
--- a/src/photos/targets/public/App.jsx
+++ b/src/photos/targets/public/App.jsx
@@ -14,7 +14,7 @@ import styles from './index.styl'
 import { ALBUM_QUERY } from '../../../../src/photos/ducks/albums/index'
 import ErrorUnsharedComponent from 'photos/components/ErrorUnshared'
 
-class App extends Component {
+export class App extends Component {
   onDownload = selected => {
     const photos = selected.length !== 0 ? selected : null
     this.downloadPhotos(photos)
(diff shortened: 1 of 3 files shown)


y-lohse:
[Photos - visionneuse album photos vue publique](https://trello.com/c/u5Uyeu31/1689-photos-visionneuse-album-photos-vue-publique)

src/photos/targets/public/App.jsx
```diff
@@ -44,8 +44,17 @@ class App extends Component {
       .downloadArchive(allPhotos.map(({ _id }) => _id), album.name)
   }
 
+  renderViewer(children) {
```
Crash--: Can we add a comment to explain why and when we can have `children` ?

cedricmessiant:
About the test, I'm not sure you can do something relevant as the code stays as it is. 
Maybe a deep rendered snapshot is better than nothing though (it's our responsibility to update it carefully).

y-lohse:
Here we go, added some tests. The snapshots are still pretty big but it's not a full render either.

src/photos/targets/public/App.jsx
```diff
@@ -114,6 +126,9 @@ App.propTypes = {
   photos: PropTypes.array.isRequired,
   fetchMore: PropTypes.func.isRequired
 }
+App.contextTypes = {
+  t: PropTypes.func
```
Crash--: 👍

PR #1477, opened 2019-03-01
Use existing photos in clustering

Gara64:
If new photos impact existing clusters, we need to be able to re-compute them efficiently. 
This PR deals with files having a referenced cluster (auto-album) and the building of a Map to associate clusters with existing and new photos.

### DIFF
--- a/src/photos/ducks/clustering/albums.js
+++ b/src/photos/ducks/clustering/albums.js
@@ -2,7 +2,7 @@ import { cozyClient, log } from 'cozy-konnector-libs'
 import { DOCTYPE_ALBUMS } from 'drive/lib/doctypes'
 import { getMatchingClusters } from './matching'
 import { prepareDataset } from './utils'
-import flatten from 'lodash/flatten'
+import { flatten, union, intersection, difference } from 'lodash'
 
 // An auto album name is the date of the first photo
 const albumName = photos => {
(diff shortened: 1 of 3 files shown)


src/photos/ducks/clustering/albums.js
```diff
@@ -173,34 +180,73 @@ const photoExistsInCollection = (photos, photo) => {
...
+                const photosToClusterize = await findPhotosToReclusterize([
+                  album
+                ])
+                mergedValues = mergedValues.concat(photosToClusterize)
+              }
+            }
+            mergedValues.push(newPhoto)
+            // resort values
+            const sortedMergedValues = mergedValues.sort(
+              (a, b) => a.timestamp - b.timestamp
```
y-lohse: There are 3 different places where we re-sort the album, do we really need to do it that often?

src/photos/ducks/clustering/albums.js
```diff
@@ -2,7 +2,7 @@ import { cozyClient, log } from 'cozy-konnector-libs'
 import { DOCTYPE_ALBUMS } from 'drive/lib/doctypes'
 import { getMatchingClusters } from './matching'
 import { prepareDataset } from './utils'
-import flatten from 'lodash/flatten'
+import { flatten, union, intersection, difference } from 'lodash'
```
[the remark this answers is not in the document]
Gara64: Ah yes, thanks, I forgot it loads the whole package with this syntax.

src/photos/ducks/clustering/albums.js
Gara64: Well, there are indeed 3 cases where we sort: 
1. When we add a new photo to a sorted set of photos. [We resort the whole array](https://github.com/cozy/cozy-drive/blob/829e957643d245a8195573c0db8a425d959aff07/src/photos/ducks/clustering/albums.js#L170), as I assumed it was smart enough to not recompute the whole array (I might be wrong)
2. When we query photos referenced by albums. [We sort all the photos](https://github.com/cozy/cozy-drive/blob/829e957643d245a8195573c0db8a425d959aff07/src/photos/ducks/clustering/albums.js#L159) after querying each album separately as we cannot group it into a single sorted couchdb query.
3. When we "merge" albums. We need to [re-sort the set of photos](https://github.com/cozy/cozy-drive/blob/829e957643d245a8195573c0db8a425d959aff07/src/photos/ducks/clustering/albums.js#L224) into a single sorted set. 

The case 1. could be replaced by a function if the sort is not "smart". The case 2. should be expressed by a single query, but it is not possible from what I know (maybe in cozy-client?). The case 3. seems mandatory.

PR #1481, opened 2019-03-04
fix(OfflineFile): Upgrade fileOpener2 and remove encodeURIComponent

Crash--:
It should fix the problem with have on iOS with our offline files. 

We don't need to `encodeURIComponent` anymore. 

See https://github.com/pwlin/cordova-plugin-file-opener2/issues/14


Crash--:
[App mobile iOS - erreur si mis dispo offline](https://trello.com/c/2j5aviMm/1721-app-mobile-ios-erreur-si-mis-dispo-offline)

y-lohse:
Me neither, but let's go with that.

PR #1484, opened 2019-03-04
fix(travis): sometimes rundeck doesn't launch

CozyKourai:
This PR is for investigation : 
add echo on curl to investigate when rundeck is not running


CozyKourai:
[Install des apps via rundeck ne se lance pas](https://trello.com/c/tT9UtCca/1718-install-des-apps-via-rundeck-ne-se-lance-pas)

y-lohse:
Ok for me, but are we suspecting the CURL command to fail? If not, I think we might as well add a lot of logs everywhere.

PR #1488, opened 2019-03-06
refactor: Save clustering results

Gara64:


src/photos/ducks/clustering/albums.js
```diff
@@ -204,7 +261,7 @@ export const albumsToClusterize = async (newPhotos, albums) => {
             )
             clusterAlbumsExist = true
             break
-          } else if (intersection(matchingAlbums, clusterAlbums).length > 1) {
+          } else if (intersection(matchingAlbums, clusterAlbums).length > 0) {
```
y-lohse: Hm, I'm a bit worried that we missed this. Could we add a test to cover this?

Gara64: Hum yes, it would be better. However, I cannot test this function directly because of the problem we had with cozy-client-js (We couldn't import it with jest for obscure reason). So any idea how to test it?

src/photos/ducks/clustering/albums.js
```diff
@@ -91,54 +123,80 @@ const createClusters = async clusters => {
...
+        photos.filter(p => p.clusterId).map(p => p.clusterId)
+      )
+      if (clusterIds.length === 0) {
+        // No clusterId : create the new cluster
+        const album = await createAutoAlbum(photos)
+        refsCount += await addAutoAlbumReferences(photos, album)
+      } else if (clusterIds.length === 1) {
+        const album = clusterAlbums.find(album => album._id === clusterIds[0])
+        if (processedAlbumsIds.includes(album._id)) {
+          // Album already processed for another cluster: remove the refs and create a new album
```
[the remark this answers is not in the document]
Gara64: Consider this scenario: we have 2 clusters: c1 = [p1, p2], c2 = [p3]. p1 and p3 used to be in the same cluster, so referenced by the same album, say a1. p2 is a newly uploaded photo.
We first process c1. a1 is found because of p1,so we add it as a `referenced-by` in p2.
Then we process c2. a1 is found again, but it has already been used in c1, another cluster. So we cannot use it anymore: we remove the reference and create a3, the new `referenced-by` of p3.

PR #1493, opened 2019-03-07
fix: Safer way to check if auth should start

y-lohse:
I'm not sure why, but in some cases the redux store does not have the default `onboardingInformations` set, and when that happens the auth module crashed.


tpetitberghien:
[App Android - bloqué splash screen (device One Plus?)](https://trello.com/c/YyEf8x3v/1731-app-android-bloqu%C3%A9-splash-screen-device-one-plus)

PR #1498, opened 2019-03-11
Fix clustering gradient

Gara64:


Gara64:
@y-lohse I think it could be useful to keep the debug logs even after we are confident on the execution. It could be useful to be able to monitor the service on a specific instance.

y-lohse:
In that case maybe the `debug` level isn't the correct one? Not sure what the exact effects of the levels are though.

Gara64:
I assumed that `info` was always displayed, while `debug` was only for debug instances. However, this is not the case, as said here: https://github.com/cozy/cozy-stack/issues/1546
So yeah, as they are semantically more 'info' than 'debug' we better use info. I'll revert the last commit

PR #1500, opened 2019-03-12
fix: Public page link as a button

GoOz:
Changed the link to create a Cozy from the public sharing page to a button style (but it's still a link).

I hope I didn't miss any other link but according to the code it seems like this is it.

That was a very hard task to do for a morning task 😫 *`</sarcasm>`*

### DIFF
--- a/src/components/Button/CozyHomeLink.jsx
+++ b/src/components/Button/CozyHomeLink.jsx
@@ -5,7 +5,6 @@ import styles from './index.styl'
 
 const CozyHomeLink = ({ from, embedInCozyBar = false, t }) => (
   <ButtonLink
-    subtle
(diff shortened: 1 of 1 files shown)


GoOz:
[(𝐗𝐒) Partage par lien vue publique - boutons vs. liens](https://trello.com/c/qwbIUoRk/1638-%F0%9D%90%97%F0%9D%90%92-partage-par-lien-vue-publique-boutons-vs-liens)

y-lohse:
C'est à cause de ca oui. Je te recommande de ne plus passer par un fork pour drive :-)

y-lohse:
(Mais tu peux merger quand même)

PR #1501, opened 2019-03-12
Visual Review : Upload screenshots, and get testcafe to fail if not accepted

CozyKourai:
For each fixture in testcafe : I create a new VisualReview Object, so https://visualreview.cozycloud.cc/#/ will have 2 projects (Photos and Drive), and one suite for each fixture.
Each time we launch the test, a new run will be created in the suite.
I only check each run once, when all screenshots are uploaded.

----

Also, I'm using the fixture/tests name quite a lot : 
- to declare the fixture/test
- in console.log to provide infos
- to create the visual review object
- to name my screenshots

testcafe doesn't have a prop like test.name , so I think I'll use const like
`const FIXTURE_UPLOAD= 'Uploads Photos'`
`const TEST_UP_1 = 'Uploads 1 Photos'`
etc... and use those const to name my fixtures/test/screenshots/etc....

What do you think is best : 
declare each const in the scenario file (like `photos_start_upload_photos.js`), or create a `data-name` or something file and put everything here ?

### DIFF
--- a/testcafe/runner-photos.js
+++ b/testcafe/runner-photos.js
@@ -1,6 +1,10 @@
 const createTestCafe = require('testcafe')
+const postCommentToGithub = require('./tests/helpers/comment-on-github.js')
 
 async function runRunner() {
+  //init vrErrorMsg
+  process.env.vrErrorMsg = ''
+
   const tc = await createTestCafe()
   const runner = await tc.createRunner()
   const response = await runner
@@ -8,10 +12,10 @@ async function runRunner() {
       //Scenario that just upload photos, so we don't need to do it in every test.
       'testcafe/tests/photos/photos_start_upload_photos.js',
 
-      'testcafe/tests/photos/photos_crud.js',
-      'testcafe/tests/photos/create_full_album_scenario.js',
-      'testcafe/tests/photos/create_empty_album_scenario.js',
-
+      // 'testcafe/tests/photos/photos_crud.js',
+      // 'testcafe/tests/photos/create_full_album_scenario.js',
+      // 'testcafe/tests/photos/create_empty_album_scenario.js',
+      //
       //Scenario that just delete photos, so we don't need to do it in every test.
       'testcafe/tests/photos/photos_end_delete_all_data.js'
     ])
@@ -24,6 +28,8 @@ async function runRunner() {
     )
     .run({ assertionTimeout: 6000 }, { pageLoadTimeout: 6000 })
   tc.close()
+
+  await postCommentToGithub()
 
   if (response > 0) throw Error(response)
 }
--- a/testcafe/tests/helpers/utils.js
+++ b/testcafe/tests/helpers/utils.js
@@ -29,6 +29,21 @@ export const getPageUrl = ClientFunction(() => window.location.href)
 
 export const goBack = ClientFunction(() => window.history.back())
 
+export const getNavigatorOs = ClientFunction(() => navigator.platform)
+
+//User Agent is needed for VisualReview, but we don't need the all string
+export const getNavigatorName = ClientFunction(() => {
+  const navigatorKey = ['MSIE', 'Firefox', 'Safari', 'Chrome', 'Opera'],
+    userAgent = navigator.userAgent
+  return navigatorKey.find(
+    navigatorName => userAgent.indexOf(navigatorName) !== -1
+  )
+})
+
+export const getResolution = ClientFunction(
+  () => `${window.screen.width} x ${window.screen.height}`
+)
+
 export const getElementWithTestId = Selector(
   id => document.querySelectorAll(`[data-test-id='${id}']`)
   //getElementsByAttribute is not part of W3C DOM, while querySelectorAll is.
--- a/testcafe/tests/helpers/visualreview-utils.js
+++ b/testcafe/tests/helpers/visualreview-utils.js
@@ -1,0 +1,20 @@
+import VisualReview from 'visualreview-client'
+import { t } from 'testcafe'
+
+//Put this const in travis after POC
+const VISUALREVIEW_INSTANCE = 'visualreview.cozycloud.cc'
+
+export class VisualReviewTestcafe extends VisualReview {
+  constructor(options) {
+    super(options)
+    this.options.protocol = 'https'
+    this.options.hostname = VISUALREVIEW_INSTANCE
+  }
+
+  async takeScreenshotAndReview(imageName) {
+    await t.takeScreenshot(imageName)
+
+    //the path needs to be in const but i need to define the screenshots tree 1st
+    this.uploadScreenshot('./reports/screenshots/' + imageName)
+  }
+}
--- a/testcafe/tests/helpers/visualreview-utils.js
+++ b/testcafe/tests/helpers/visualreview-utils.js
@@ -1,5 +1,6 @@
 import VisualReview from 'visualreview-client'
 import { t } from 'testcafe'
+import { getNavigatorOs, getNavigatorName, getResolution } from './utils'
 
 //Put this const in travis after POC
 const VISUALREVIEW_INSTANCE = 'visualreview.cozycloud.cc'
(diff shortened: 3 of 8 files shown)


CozyKourai:
[Intégration de VisualReview-node-client dans drive et 1ers uploads de screens automatiques sur https://visualreview.cozycloud.cc](https://trello.com/c/7HBNAZ1U/1738-int%C3%A9gration-de-visualreview-node-client-dans-drive-et-1ers-uploads-de-screens-automatiques-sur-https-visualreviewcozycloudcc)

CozyKourai:
[(𝐒) Intégration des retours de visualReview dans les tests > Failed si des screens demandent une validaton manuelle](https://trello.com/c/LIsCcvoI/1740-%F0%9D%90%92-int%C3%A9gration-des-retours-de-visualreview-dans-les-tests-failed-si-des-screens-demandent-une-validaton-manuelle)

testcafe/tests/helpers/utils.js
```diff
@@ -29,6 +29,23 @@ export const getPageUrl = ClientFunction(() => window.location.href)
 
 export const goBack = ClientFunction(() => window.history.back())
 
+export const getNavigatorOs = ClientFunction(() => navigator.platform)
+
+//User Agent is needed for VisualReview, but we don't need the all string
+export const getNavigatorName = ClientFunction(() => {
+  var aKeys = ['MSIE', 'Firefox', 'Safari', 'Chrome', 'Opera'],
```
y-lohse: Please don't use `var` to declare variables, and we can use complete variable names, no need to shorten them.

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -0,0 +1,46 @@
...
+const VISUALREVIEW_INSTANCE = 'visualreview.cozycloud.cc'
+
+export class VisualReviewTestcafe extends VisualReview {
+  constructor(options) {
+    super(options)
+    this.options.protocol = 'https'
+    this.options.hostname = VISUALREVIEW_INSTANCE
+  }
+
+  async takeScreenshotAndReview(imageName) {
```
y-lohse: The function is called `takeScreenshotAndReview ` but there's no review happening, so I think we should rename it. `saveScreenshot` or something like that?

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -0,0 +1,46 @@
...
+    await t.takeScreenshot(imageName)
+
+    this.options.properties.os = await getNavigatorOs()
+    this.options.properties.browser = await getNavigatorName()
```
y-lohse: We could set the os and the browser in the constructor, since they won't change over time?

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -0,0 +1,46 @@
...
+    this.options.properties.resolution = await getResolution()
+
+    //the path needs to be in const but i need to define the screenshots tree 1st
+    this.uploadScreenshot('./reports/screenshots/' + imageName)
+  }
+
+  async checkVr() {
```
y-lohse: Maybe we can find a better name, like `checkStatus`? We're in the VisualReview class so we know it's something about `vr`.

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -0,0 +1,46 @@
...
+    let runAnalysis = await this.getJsonStatusForCurrentRun()
+    let runStatus = 'accepted'
+
+    for (let imgDiff in runAnalysis.diffs) {
+      if (runAnalysis.diffs[imgDiff].status != 'accepted') {
+        runStatus = runAnalysis.diffs[imgDiff].status
+        throw new Error(
```
y-lohse: Since we're throwing an error, the execution will stop right there and we won't know if more than one image ha a regression. Is this what we want? I would think we'd want to report all the failures, no?

CozyKourai: Each time testcafe is launched, VisualReview will create one run for each fixture.
Ie : One run for `upload photos` and one for `delete photos`
in the GUI, one run contains several images : ie : https://visualreview.cozycloud.cc/#/19/27/271 

If several screenshots are failing in this fixture, there will be only one URL to open to review.
Also the GUI shows how many images are failed/pending for this fixture.
As long as there is at least one failed/pending images, we'll have to do the manual review, hence throwing an error just for the 1st one.
(the real analysis for picture is done on the server as soon as they are uploaded, we are just getting the status here)

I can enhance this function to check all status and throw the error at the end, but there will still be only one url to review.

(and as again, we have a different run for each fixture : 
if `upload photos` fails, we'll get an url to review, but testcafe will continue running and tests the other fixtures. so `delete photos` will get a different url to review if it fails too! (or nothing if all screens in this fixtures are ok!)

testcafe/tests/helpers/visualreview-utils.js
CozyKourai: `takeScreenshotAndUpload` ?

testcafe/tests/helpers/visualreview-utils.js
y-lohse: Ah yes ok! In that case we don't even need the `for` loop. We can use `runAnalysis.diffs.filter()` to check if any of the runs failed. This is cleaner than manually looping and then throwing in the middle of the loop.

testcafe/tests/helpers/visualreview-utils.js
CozyKourai: After a bit of research, no I cannot set the `os` and `browser` in the constructor.

I cannot even create an `initStart` function to call right after creating the vr object, as I'm doing it in a `.before` (So I have one run for one fixture!)

The before/after fixture hooks are executed before the first test/after the last test on the server side so you cannot use interactions with the `ClientFunction` in them. (And `getNavigatorOs`and `getNavigatorName` are both using `ClientFunction`

testcafe/tests/helpers/visualreview-utils.js
CozyKourai: Done and thanks to that change I realize I was not always getting the status for the whole run.
As for now, the server doesn't send back a global run status (it always stays in "running"), so I had to introduce a delay.

I can change the function and add a parameter to tell it how many pictures we are expecting in the response, but I think it's more relevant to wait for this afternoon meeting to define next steps as some changes on server sides will be required to go on production, so we'll be able to get a proper status for the whole run.

CozyKourai:
[⚡Tests Visuels](https://trello.com/c/plCmXXJA/1716-%E2%9A%A1tests-visuels)

testcafe/tests/helpers/utils.js
```diff
@@ -29,6 +29,25 @@ export const getPageUrl = ClientFunction(() => window.location.href)
...
+export const getNavigatorName = ClientFunction(() => {
+  let navigatorKey = ['MSIE', 'Firefox', 'Safari', 'Chrome', 'Opera'],
+    userAgent = navigator.userAgent,
+    index = navigatorKey.length - 1
+  for (
+    index;
+    index > -1 && userAgent.indexOf(navigatorKey[index]) === -1;
+    index--
+  );
+  return navigatorKey[index]
```
cedricmessiant: you can just do:

`return navigatorKey.find(navigatorName => userAgent.indexOf(navigatorName) !== -1)`

CozyKourai:
[VisualReview > Message Git quand des screensshot fail](https://trello.com/c/hmkpFblj/1758-visualreview-message-git-quand-des-screensshot-fail)

testcafe/runner-photos.js
```diff
@@ -1,17 +1,22 @@
...
       //Scenario that just upload photos, so we don't need to do it in every test.
       'testcafe/tests/photos/photos_start_upload_photos.js',
 
-      'testcafe/tests/photos/photos_crud.js',
-      'testcafe/tests/photos/create_full_album_scenario.js',
-      'testcafe/tests/photos/create_empty_album_scenario.js',
-
+      // 'testcafe/tests/photos/photos_crud.js',
+      // 'testcafe/tests/photos/create_full_album_scenario.js',
+      // 'testcafe/tests/photos/create_empty_album_scenario.js',
```
y-lohse: Why are these commented out?

testcafe/runner-photos.js
```diff
@@ -1,17 +1,22 @@
 const createTestCafe = require('testcafe')
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
```
y-lohse: Nice, TIL.

testcafe/tests/helpers/utils.js
```diff
@@ -29,6 +29,21 @@ export const getPageUrl = ClientFunction(() => window.location.href)
 
 export const goBack = ClientFunction(() => window.history.back())
 
+export const getNavigatorOs = ClientFunction(() => navigator.platform)
+
+//User Agent is needed for VisualReview, but we don't need the all string
+export const getNavigatorName = ClientFunction(() => {
+  let navigatorKey = ['MSIE', 'Firefox', 'Safari', 'Chrome', 'Opera'],
```
y-lohse:
```suggestion
  const navigatorKey = ['MSIE', 'Firefox', 'Safari', 'Chrome', 'Opera'],
```

testcafe/runner-photos.js
```diff
@@ -25,6 +30,17 @@ async function runRunner() {
...
+  if (
+    typeof process.env.TRAVIS_PULL_REQUEST !== 'undefined' &&
+    process.env.TRAVIS_PULL_REQUEST &&
+    process.env.vrErrorMsg != ''
+  ) {
+    const message = `Visual Review - Please review screenshots, then restart build. ${
+      process.env.vrErrorMsg
+    }`
+    await exec(`yarn run cozy-ci-github "${message}"`)
+  }
```
y-lohse: I wonder if it's the right place for this — we'll want to do the same for drive, so wouldn't it make sense to have this in a separate script?

testcafe/runner-photos.js
CozyKourai: I remove them for quicker local testing (as those tests don't have screenshots for now), and forgot to un-comment before commit!

testcafe/tests/photos/photos_end_delete_all_data.js
```diff
@@ -1,24 +1,40 @@
...
-fixture`Delete all photos`.page`${TESTCAFE_PHOTOS_URL}/`.beforeEach(async t => {
-  await t.useRole(photosUser)
-  await page.waitForLoading()
-  await page.initPhotosCount()
-})
+fixture`Delete all photos`.page`${TESTCAFE_PHOTOS_URL}/`
+  .before(async ctx => {
+    ctx.vr = new VisualReviewTestcafe({
+      projectName: 'PHOTOS',
+      suiteName: `fixture : delete photos`
```
Crash--: We should also init diff to something like 4%

PR #1502, opened 2019-03-12
fix: move creation button to menu on mobile

GoOz:
I should have seen it… That was too easy.

Here, now the button is hidden on mobile and a new `MenuItem` with the same action is added on Album sharing and the button is properly displayed when in cozy-bar on Drive as well.

NB: `getHomeLinkHref` function has been extracted to be properly called in any other context.

### DIFF
--- a/src/components/Button/CozyHomeLink.jsx
+++ b/src/components/Button/CozyHomeLink.jsx
@@ -2,29 +2,28 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { ButtonLink } from 'cozy-ui/react'
 import styles from './index.styl'
+import getHomeLinkHref from './getHomeLinkHref'
 
-const CozyHomeLink = ({ from, embedInCozyBar = false, t }) => (
+const CozyHomeLink = ({ from, embedInCozyBar = false, t, size, className }) => (
   <ButtonLink
     label={t('Share.create-cozy')}
     icon="cozy-negative"
-    className={embedInCozyBar ? styles['bar-homelink'] : ''}
+    className={embedInCozyBar ? styles['bar-homelink'] : className}
     href={getHomeLinkHref(from)}
+    size={size}
   />
 )
 
 CozyHomeLink.propTypes = {
   from: PropTypes.string,
   embedInCozyBar: PropTypes.bool,
-  t: PropTypes.func.isRequired
+  t: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['tiny', 'small', 'large', 'normal'])
 }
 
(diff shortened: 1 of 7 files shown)


GoOz:
[(𝐗𝐒) Partage par lien vue publique - boutons vs. liens](https://trello.com/c/qwbIUoRk/1638-%F0%9D%90%97%F0%9D%90%92-partage-par-lien-vue-publique-boutons-vs-liens)

src/components/Button/getHomeLinkHref.jsx
```diff
@@ -0,0 +1,6 @@
+const getHomeLinkHref = from =>
```
cedricmessiant: - shouldn't we move `https://manager.cozycloud.cc/cozy/create` in constants?
- could you please add a simple test for this function?

src/components/Button/getHomeLinkHref.spec.js
```diff
@@ -0,0 +1,14 @@
+import getHomeLinkHref from './getHomeLinkHref'
+
+describe('getHomeLinkHref', () => {
```
cedricmessiant: great, thanks!

GoOz:
[(𝐗𝐒) Partage par lien vue publique - boutons vs. liens](https://trello.com/c/cPeMYuqF/1113-%F0%9D%90%97%F0%9D%90%92-partage-par-lien-vue-publique-boutons-vs-liens)

src/components/Button/CozyHomeLink.jsx
```diff
@@ -2,29 +2,27 @@ import React from 'react'
...
+    size={size}
   />
 )
 
 CozyHomeLink.propTypes = {
   from: PropTypes.string,
   embedInCozyBar: PropTypes.bool,
-  t: PropTypes.func.isRequired
+  t: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(['tiny', 'small', 'large'])
```
y-lohse: `size` has no default value and in one of the cases, we don't provide it as a prop, so it will be undefined and probably produce proptypes warning. Could you make sure there are no warnings coming from this code please?

src/photos/targets/public/App.jsx
```diff
@@ -89,6 +94,14 @@ export class App extends Component {
                       position="right"
                       className="u-hide--desk"
                     >
+                      <MenuItem
+                        onSelect={() =>
+                          (window.location = getHomeLinkHref('sharing-drive'))
```
y-lohse:
```suggestion
                          (window.location = getHomeLinkHref('sharing-photos'))
```

PR #1504, opened 2019-03-14
Create instance on travis for testing, and delete it after

CozyKourai:
- [x] Create instance, deploy app, run tests, delete instance
- [x] Case : test failed. We still need to delete instance !
- [x] Remove comments/unnecessary logs
- [x] Squah


CozyKourai:
[(𝐌) Déploiement auto : creation d'une instance à la volée (vraie instance, pas testcafe1...)](https://trello.com/c/4d79e6gw/1696-%F0%9D%90%8C-d%C3%A9ploiement-auto-creation-dune-instance-%C3%A0-la-vol%C3%A9e-vraie-instance-pas-testcafe1)

scripts/rundeck.sh
```diff
@@ -0,0 +1,43 @@
...
+    LOG_OUTPUT=$(curl -s -X GET \
+         -H "X-Rundeck-Auth-Token: $RUNDECK_TOKEN" \
+         -H "Content-Type: application/json" https://rundeck.cozycloud.cc/api/27/execution/$1/output)
+
+    JOB_STATUS=$(echo $LOG_OUTPUT | grep -o -P "(?<=<execState>).*(?=</execState> )")
+    echo "↳ Job $JOB_STATUS"
+  done
+  if [  $JOB_STATUS == 'succeeded' ] ; then
+    echo "↳ ✅ Execution ${1} succeeded"
+    #The job can be succesfull. but return an error, so we need to check!
```
cedricmessiant: I know this is on rundeck side but did you checked with infra if it is normal or if it's a bug?

CozyKourai: Yep! https://mattermost.cozycloud.cc/test-team/pl/qmoahembdfrnpfszgazrtm1pma

scripts/rundeck.sh
```diff
@@ -0,0 +1,43 @@
...
+    if [[ "${LOG_OUTPUT,,}" == *"error"* ]]; then
+      echo "❌ Execution ${1} return an error :" && echo $LOG_OUTPUT && exit 1;
```
y-lohse:
```suggestion
      echo "❌ Execution ${1} returned an error :" && echo $LOG_OUTPUT && exit 1;
```

scripts/create-instance-and-test.sh
```diff
@@ -0,0 +1,33 @@
...
+INSTANCE_ID="${COZY_APP_SLUG,,}"$TRAVIS_PULL_REQUEST
+
+
+echo "↳ ℹ️  Creating instance..."
+INSTANCE_CREATE_EXECUTION_ID=$(runRundeckJob $JOB_CREATE_TEST_INSTANCE "-commitid $INSTANCE_ID -environment $INSTANCE_ENV -locale $INSTANCE_LOCALE")
+getRundeckStatus $INSTANCE_CREATE_EXECUTION_ID
+INSTANCE_CREATE_STATUS=$?
+echo $INSTANCE_CREATE_STATUS
+
+if [ $INSTANCE_CREATE_STATUS == 0 ] ; then
```
y-lohse: If the creation failed, will the execution stop and an error be reported?

scripts/create-instance-and-test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -eu
+source ./scripts/rundeck.sh
+
+JOB_CREATE_TEST_INSTANCE="7170b8bc-10d2-4099-b424-8e194a5d5884"
+
+INSTANCE_ENV='stg'
+INSTANCE_LOCALE='en'
+#We need a different name for photos and drive build otherwise we'll try to create the same instance twice in travis.
+INSTANCE_ID="${COZY_APP_SLUG,,}"$TRAVIS_PULL_REQUEST
```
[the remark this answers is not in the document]
CozyKourai: It convert to lowecase  (Rundeck won't accept an id with uppercase), but I just realize that now our `COZY_APP_SLUG` is now already in lower case! so I can remove it if you want

.travis.yml
```diff
@@ -42,10 +42,9 @@ jobs:
       script:
         - yarn tx
         - yarn build:$COZY_APP_SLUG:browser
-        - ./scripts/choose-test-instance.sh
-        - source setinstance.sh
         - &downcloud-cert if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then eval "$(ssh-agent -s)" && chmod 600 /tmp/id_rsa_travis_downcloud && ssh-add /tmp/id_rsa_travis_downcloud; fi # Needs to be inlined and not in a script to work.
-        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then APP_URL=$(node scripts/downcloud.js build/$COZY_APP_SLUG) && ./scripts/deploy-and-test-on-instance.sh $APP_URL; fi
+        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then ./scripts/create-instance-and-test.sh; fi
+        - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then ./scripts/delete-instance.sh; fi
```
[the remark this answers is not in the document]
CozyKourai: I can merge `create instance` , `deploy instance` and `test` in one file as the execution of `tests` depends on `deploy instance`, which depend on `create instance`, but I think we should keep the `delete instance` apart.
If the `create-deploy-test` script fail at some point, we still want to try to `delete the instance` .
Otherwise, if we restart the buid (or commit more stuff to the PR), there is a risk that the instance already exist, so the build might fail again.

scripts/create-instance-and-test.sh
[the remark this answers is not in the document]
y-lohse: TIL! But yeah if it's not useful, might as well get rid of it :)

.travis.yml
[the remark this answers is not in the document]
y-lohse: > If the create-deploy-test script fail at some point, we still want to try to delete the instance .

Yep ok, that makes sense. I think it's still worth splitting or merging the other ones 👍

> Otherwise, if we restart the buid (or commit more stuff to the PR), there is a risk that the instance already exist, so the build might fail again.

I guess that's a risk anyway, I think travis jobs can be interrupted at pretty much any point, either manually or by other commits. But you're right, lets keep that risk as low as possible.

scripts/create-instance-and-test.sh
CozyKourai: if the execution status return `fail` -> error & message
if the execution status body contains `error` -> error & message (except for the `deploy job`which always return the log with `level=error`)

I just add a check on whether the POST return something or not which display the request if nothing is return

PR #1506, opened 2019-03-15
fix: Dates timezone

Gara64:
**Context**: the stack stores photos' datetime with its own timezone, UTC+00 for our infra.
However, this creates an offset if the user's locale is not the same. Typically with a UTC+02 timezone, 2 hours will be "added" to all photos on the front side when we create the date with `new Date(date)`.
This is particularly obvious when there are several cluster for the same day, as we show the hour. 

This PR fixes this behavior by using the constructor `new dates(year, month ...)`, as the arguments are interpreted as the local time, as stated [here](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date#Param%C3%A8tres)

### DIFF
--- a/src/photos/ducks/timeline-clusters/index.jsx
+++ b/src/photos/ducks/timeline-clusters/index.jsx
@@ -3,11 +3,16 @@ import { Query } from 'cozy-client'
 import Timeline from '../timeline/components/Timeline'
 import PhotoBoard from '../../components/PhotoBoard'
 import {
-  differenceInCalendarDays,
-  differenceInCalendarMonths,
-  differenceInHours
-} from 'date-fns'
-import { formatDMY, formatD, formatH } from './dates'
+  formatDMY,
+  formatD,
+  formatH,
+  formatYMD,
+  isSameDay,
+  isSameMonth,
+  isSameHour,
+  isEqualOrOlder,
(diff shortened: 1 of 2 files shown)


src/photos/ducks/timeline-clusters/index.jsx
```diff
@@ -95,22 +98,21 @@ const getSectionTitleHours = (dates, index, section, f) => {
...
-      return (
-        (datetime >= startPeriod && datetime <= endPeriod) ||
-        differenceInCalendarDays(datetime, endPeriod) === 0
-      )
+      const startPeriod = sections[date].album.period.start
+      const endPeriod = sections[date].album.period.end
+      const isInsidePeriod =
+        isEqualOrNewer(datetime, startPeriod) &&
+        isEqualOrOlder(datetime, endPeriod)
+      return isInsidePeriod || isSameDay(f, datetime, endPeriod)
```
y-lohse: If `datetime` and `endPeriod` are the same day, then doens't that mean `isInsidePeriod ` is true as well?

Gara64:
All the formats used are defined in the `dates.js` file. I'm not sure how you would avoid passing the `f` variable as it is retrieved from the props?

PR #1508, opened 2019-03-20
New PDF viewer

y-lohse:
Using pdf.js for viewing pdf files, with a minimal home-gown UI around it!


y-lohse:

src/viewer/PdfJsViewer.jsx
```diff
@@ -0,0 +1,167 @@
...
+import throttle from 'lodash/throttle'
+import { Button, Spinner } from 'cozy-ui/react'
+import withFileUrl from './withFileUrl'
+import NoViewer from './NoViewer'
+import styles from './styles'
+
+export const MIN_SCALE = 0.25
+export const MAX_SCALE = 3
+
+const ToolbarButton = ({ icon, onClick, disabled }) => (
```
Crash--: Move it to its own file? :-°

src/viewer/Viewer.jsx
```diff
@@ -143,7 +142,7 @@ export default class Viewer extends Component {
       case 'video':
         return isMobile() ? NoViewer : VideoViewer
       case 'pdf':
-        return isMobileApp() ? NativePdfViewer : PdfViewer
```
Crash--: So we can remove `NativePdfViewer` file? And remove the native plugin in config.xml ?

src/viewer/__snapshots__/PdfJsViewer.spec.jsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PDFViewer with a pdf that does not load should show a fallback 1`] = `
```
cedricmessiant: maybe shallow rendering is not the best here?

src/viewer/PdfJsViewer.jsx
```diff
@@ -0,0 +1,167 @@
...
+    this.wrapper = null
+    this.resizeListener = null
+
+    this.onLoadSuccess = this.onLoadSuccess.bind(this)
+    this.onLoadError = this.onLoadError.bind(this)
+    this.nextPage = this.nextPage.bind(this)
+    this.previousPage = this.previousPage.bind(this)
+    this.scaleUp = this.scaleUp.bind(this)
+    this.scaleDown = this.scaleDown.bind(this)
+    this.setWrapperSize = this.setWrapperSize.bind(this)
```
cedricmessiant: you could remove all these binds if you used class properties instead of methods like you already do for `setWrapperSize`

cedricmessiant: you can even remove the constructor if you define `wrapper` and `resizeListener` as properties also

src/viewer/PdfJsViewer.jsx
```diff
@@ -0,0 +1,167 @@
...
+    this.resizeListener = throttle(this.setWrapperSize, 500)
+    window.addEventListener('resize', this.resizeListener)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.resizeListener)
+  }
+
+  setWrapperSize = () => {
+    this.setState({ width: this.wrapper.getBoundingClientRect().width })
```
cedricmessiant: don't we have to check that `this.wrapper` is not `null` here?

src/viewer/PdfJsViewer.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+  async onLoadSuccess({ numPages }) {
+    this.setState({
+      totalPages: numPages,
+      loaded: true
+    })
+  }
+
+  async onLoadError(error) {
+    console.warn(error)
```
cedricmessiant: add `// eslint-disable-next-line no-console` before?

cedricmessiant: (and do we really want to keep this `console.warn`?)

src/viewer/Viewer.jsx
y-lohse: The file is removed, but yes, I forgot about the plugin!

src/viewer/__snapshots__/PdfJsViewer.spec.jsx.snap
y-lohse: Mmmh I'll check this, I really don't want to render the pdf in a test, but maybe just of the error it would work.

cedricmessiant: I guess you could mock `Document` component but whatever, it will be covered by e2e tests I guess !

PR #1516, opened 2019-03-22
chore: Remove eslint warning by removing console or disable-next-line

Crash--:
I'm tired of seeing so much warnings when linting ;).

I removed some, for others I just disable the check since I think they can be useful on Sentry.

What do you think?


src/drive/mobile/lib/background.js
```diff
@@ -1,3 +1,4 @@
+/* eslint no-console: 0 */
```
cedricmessiant: I don't really like global ignores since it means we may forgot `console.*` in this file but if it's a temporary solution until we remove all these console?

Crash--: This is !

PR #1521, opened 2019-03-26
chore: add more log and check for rundeck script

CozyKourai:
I added some checks and logs

I also add a `sleep 10` before retrying in case of conflicts.
Regarding conflicts, it seems rundeck cannot run the same job at the same time (not depending on the parameters)
there are years-old issues open on that point like https://github.com/rundeck/rundeck/issues/1426 which are still active.


CozyKourai:
[Mise en place de log sur chaque commande exécutée](https://trello.com/c/YXoieb7N/1805-mise-en-place-de-log-sur-chaque-commande-ex%C3%A9cut%C3%A9e)

y-lohse:
Is this ready for review? If not do you mind putting a WIP label in the PR title until it's ready?

Crash--:
Why  we are not using the stderror to log error trace?

```
myerror=`/mycommand 2>&1 `
echo $myerror 
```

CozyKourai:
[Bug de déploiement ?](https://trello.com/c/6a7tWxBh/1782-bug-de-d%C3%A9ploiement)

PR #1523, opened 2019-03-27
Viewer CSS refactor

y-lohse:
I refactored the CSS for the viewer. Everything should look almost the same after this PR, but with simpler code.

I strongly suggest to review this PR commit by commit, otherwise the changes will be very hard to follow. I will squash everything once it's ✅.


y-lohse:

src/viewer/NoViewer.jsx
```diff
@@ -1,16 +1,22 @@
 import React from 'react'
 import { connect } from 'react-redux'
-import classNames from 'classnames'
 import { translate } from 'cozy-ui/react/I18n'
-import { Button } from 'cozy-ui/react/Button'
-import Alerter from 'cozy-ui/react/Alerter'
+import { Icon, Button, Alerter } from 'cozy-ui/transpiled/react'
```
Crash--: Hmm now we are importing from the transpiled and the non transpiled one. Can we just add an alias in our webpack config ? ;)

src/viewer/styles.styl
```diff
@@ -41,11 +41,25 @@
     align-items      center
     flex-direction   column
 
+    h2
+        max-width   90%
+
+.pho-viewer-imageviewer
+.pho-viewer-nav + .pho-viewer-pdfviewer
```
GoOz: I know this is legacy code but I guess there is something smarter to do here, like giving all those element a same class. Not blocking comment though.

src/viewer/styles.styl
```diff
@@ -48,6 +48,22 @@
       margin-left      0
       width            100%
 
+.pho-viewer-canceled
+.pho-viewer-noviewer
+.pho-viewer-audioviewer
+.pho-viewer-textviewer
```
GoOz: Not sure about how the JSX looks like but I wonder if it wouldn't be better to have a common class `.pho-viewer` (or something else generic if it's already in use for another element) to do this.
I am not comfortable having several classes doing the same thing (my personal limit being 2)

src/viewer/controls.styl
```diff
@@ -24,6 +26,9 @@
             margin-left 20%
 
 .pho-viewer-nav
+    position   fixed
+    top        5.625rem // because of the toolbar
```
GoOz: Is it possible to get this height via a variable? Using it to set this position and the height of the toolbar?

src/viewer/AudioViewer.jsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react'
+import { Icon } from 'cozy-ui/transpiled/react'
 
 import withFileUrl from './withFileUrl'
 import styles from './styles'
+import AudioIcon from './icons/icon-type-audio.svg'
 
 const AudioViewer = ({ file, url }) => (
   <div data-test-id="viewer-audio" className={styles['pho-viewer-audioviewer']}>
+    <Icon icon={AudioIcon} width={160} height={140} />
```
GoOz: Genuine interrogation: since `width` & `height` props expects only numbers what's the main difference between:
```jsx
<Icon icon={AudioIcon} width={160} height={140} />
```
and
```jsx
<Icon icon={AudioIcon} width="160" height="140" />
```

src/viewer/NoViewer.jsx
```diff
@@ -78,6 +84,39 @@ const NoViewerButton = ({ file, fallbackUrl, t, onError }) => {
...
+    case 'pdf':
+      icon = IconPdf
+      break
+    default:
+      icon = IconFiles
+      break
+  }
+
+  return <Icon icon={icon} width={160} height={140} />
+}
```
Crash--: FileIcon.jsx ? :D

src/viewer/controls.styl
```diff
@@ -0,0 +1,121 @@
...
+@require 'settings/breakpoints.styl'
+@require 'settings/palette.styl'
+
+.pho-viewer-controls
+    display         flex
+    flex-direction  row
+    justify-content center
+    align-items center
+    width           100%
+    height          100%
```
Crash--: u-flex u-flex-row u-flex-justify-center ... ? :D

src/viewer/AudioViewer.jsx
y-lohse: I think the doc says number or strings? But otherwise, `160` has a type of `Number` where as `"160"` has a type of `String`. So the 2 will behave differently depending on what you do with them in js. So for example `"160" + 3 = "1603"` whereas `160 + 3 = 163`.

AFAIK we don't actually manipulate these variables though, they simply end up in the HTML markup, so `160` (the number) ends up converted into `"160"` (the string).

src/viewer/controls.styl
y-lohse: I'd rather not convert everything to utility classes in this PR. I did it for the css I touched, but there's lots more that could be improved.

src/viewer/NoViewer.jsx
y-lohse: Talked through elsewhere: there shouldn't be a problem migrating to the transpiled version progressively, no need for an alias just yet.

src/viewer/styles.styl
y-lohse: I agree, I really do, but I'm not looking for more enhancements on this PR. I know there's still a lot that can be improved in this code, but the PR has to stop somewhere :)

y-lohse:
Thanks for the review! I moved the `FileIcon` into a standalone file :). 
I really appreciate the other suggestions for improvements, but I'm specifically looking for feedback on the new code that I added or the changes I made. There's a lot left to improve but I've already spent a fair amount of time on this :)

src/viewer/styles.styl
GoOz: Yes, sorry, I didn't mention it but this is cleary a non-blocking comment. :)

PR #1526, opened 2019-03-27
Viewer improvements

y-lohse:
Needs #1523 + rebase to be merged, but can already be reviewed.

- Enable horizontal and vertical scrolling inside a zoomed pdf
- Makes sure the viewer is correctly displayed in all situations

### DIFF
--- a/src/drive/web/modules/drive/FileOpenerExternal.jsx
+++ b/src/drive/web/modules/drive/FileOpenerExternal.jsx
@@ -61,7 +61,7 @@ class FileOpener extends Component {
         {!loading &&
           !fileNotFound && (
             <Viewer
-              style={{ top: '3rem' }}
+              fullscreen={false}
               files={[file]}
               currentIndex={0}
               onChange={doNothing}
--- a/src/drive/web/modules/drive/FileOpenerExternal.jsx
+++ b/src/drive/web/modules/drive/FileOpenerExternal.jsx
@@ -8,6 +8,8 @@
 
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+import { withRouter } from 'react-router'
+import get from 'lodash/get'
 
 import { Spinner, Alerter, translate } from 'cozy-ui/react'
 import styles from './styles'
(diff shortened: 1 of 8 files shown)


y-lohse:

src/drive/web/modules/drive/FileOpenerExternal.jsx
```diff
@@ -43,8 +44,14 @@ class FileOpener extends Component {
     }
   }
 
+  navigateToDrive = () => {
+    const parentDir = this.state.file.dir_id || ''
```
Crash--: can we be more defensive here? this.state.file && this.state.file.dir_id ?

src/drive/web/modules/drive/FileOpenerExternal.jsx
```diff
@@ -70,11 +78,13 @@ const getFileId = ownProps => {
...
       fileId: PropTypes.string.isRequired
     }).isRequired
   }).isRequired,
-  fileId: PropTypes.number
+  fileId: PropTypes.number,
+  withCloseButtton: PropTypes.bool
 }
 
-export default translate()(FileOpener)
+export default translate()(withRouter(FileOpener))
```
Crash--: This file is only used when we open a document from a deeplink, right?

y-lohse: No, it's also used in the OPEN file intent view.

y-lohse: Oh hang on that's a good point, there might not be a router in that case. Not that it would be used, but I still need to check this.

y-lohse:
> What does it brings to wrap the viewer in an Overlay?

The viewer doesn't use `position fixed` anymore, so it's easier to control when we have other things to display around it. But in the places where we want the viewer to take up all the screen, wrapping it in `Overlay` essentially gives it

```
position fixed
top 0
bottom 0
left 0
right 0
```

src/drive/web/modules/drive/FileOpenerExternal.jsx
y-lohse: All good, I can confirm it works in both cases!

PR #1540, opened 2019-04-01
Replacing embedurl with icons

y-lohse:
- [x] In the file list
- [x] In the action menu on mobile
- [x] In the upload panel


y-lohse:
[(𝐋) Custo Drive pour Orange 🍊](https://trello.com/c/ML4nmtCW/1762-%F0%9D%90%8B-custo-drive-pour-orange-%F0%9F%8D%8A)

src/components/Image/ImageLoader.jsx
```diff
@@ -127,16 +127,20 @@ class ImageLoader extends React.Component {
 
   render() {
     const { src } = this.state
-    const { render } = this.props
-    return src ? render(src) : false
+    const { render, renderFallback } = this.props
+
+    if (src) return render(src)
+    else if (renderFallback) return renderFallback()
+    else return false
```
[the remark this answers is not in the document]
y-lohse: I can never remember this. Hold on, let me look it up.

Edit: Yep, you're right : https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

src/drive/styles/filelist.styl
```diff
@@ -115,11 +114,10 @@
     padding-left 0
 
 .fil-content-cell.fil-content-file
-    position       relative
+    flex 0 0 35%
```
Crash--: I don't like magic value, 35%, 5%. Why?

src/drive/web/modules/filelist/FileThumbnail.jsx
```diff
@@ -0,0 +1,62 @@
...
+const FileThumbnail = ({
+  file,
+  withSharedBadge,
+  breakpoints: { isMobile }
+}) => {
+  const isImage = file.class === 'image'
+  return (
+    <div
+      className={cx(styles['fil-content-cell'], styles['fil-file-thumbnail'], {
+        'u-pl-0': !isMobile
```
Crash--: can't we use a css cozy-ui breakpoint here?

src/drive/web/modules/filelist/FileThumbnail.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import { withBreakpoints } from 'cozy-ui/react'
+import { ImageLoader } from 'components/Image'
+import { SharedBadge } from 'sharing'
+import FileIcon from './FileIcon'
```
Crash--: we should avoid using ./ . Can you use `drive/...` (yeah I know, alias...)

src/drive/web/modules/filelist/File.jsx
```diff
@@ -10,10 +10,9 @@ import RenameInput from 'drive/web/modules/drive/RenameInput'
 import { default as DesktopActionMenu } from 'drive/web/modules/actionmenu/ActionMenu'
 import MobileActionMenu from 'drive/web/modules/actionmenu/MobileActionMenu'
 import { isDirectory } from 'drive/web/modules/drive/files'
-import { ImageLoader } from 'components/Image'
 import { Button, Icon, withBreakpoints, MidEllipsis } from 'cozy-ui/react'
-import { SharedBadge, SharedStatus } from 'sharing'
-import { getFileMimetype } from 'drive/lib/getFileMimetype'
+import { SharedStatus } from 'sharing'
+import FileThumbnail from './FileThumbnail'
```
Crash--: alias

src/drive/web/modules/filelist/FileThumbnail.jsx
y-lohse: What do you mean? `isMobile` comes from cozy-ui and is based on the same breakpoints.

src/drive/styles/filelist.styl
y-lohse: Every column in the file list gets a % of the total width. The file part was 40% before, now it's 5% for the icon and 35 for the file name. Would you be more comfortable with it the % values had names like `column-width-icon`?

src/drive/web/modules/filelist/FileThumbnail.jsx
Crash--: We don't have new utilities classes to manage this kind of thing? u-pl-0-XXX ?

src/drive/styles/filelist.styl
Crash--: Yep having a centralized place with variables should be enough. We can do it latter

src/drive/web/modules/filelist/FileThumbnail.jsx
y-lohse: The margin/padding classes are not responsive no. There aren't many cases where we ned this, AFAIK.

PR #1546, opened 2019-04-02
fix: Fetch file when props have changed

Crash--:
- Component is not unmounted when switching back to other app 
- it's better to use `routeParams` than `router.params` since it's an abstraction and is up to date even if something went wrong 
- react-router v3 doesn't expose render props, right? I wanted to have only props.fileId but... 

- Is there a way to mock   `await cozy.client.files.statById(id, false)` in a test?

### DIFF
--- a/src/drive/web/modules/drive/FileOpenerExternal.jsx
+++ b/src/drive/web/modules/drive/FileOpenerExternal.jsx
@@ -27,16 +27,25 @@ class FileOpener extends Component {
     file: null
   }
   componentWillMount() {
-    this.loadFileInfo()
+    if (this.props.fileId) {
+      this.loadFileInfo(this.props.fileId)
+    } else if (this.props.routeParams.fileId) {
+      this.loadFileInfo(this.props.routeParams.fileId)
+    }
   }
 
-  async loadFileInfo() {
+  componentDidUpdate(prevProps) {
+    if (prevProps.fileId !== this.props.fileId) {
+      return this.loadFileInfo(this.props.fileId)
+    }
+    if (prevProps.routeParams.fileId !== this.props.routeParams.fileId) {
+      return this.loadFileInfo(this.props.routeParams.fileId)
+    }
+  }
+  async loadFileInfo(id) {
     try {
       this.setState({ fileNotFound: false })
-      const resp = await cozy.client.files.statById(
-        getFileId(this.props),
-        false
-      )
+      const resp = await cozy.client.files.statById(id, false)
       const file = { ...resp, ...resp.attributes, id: resp._id }
       this.setState({ file, loading: false })
     } catch (e) {
@@ -71,10 +80,6 @@ class FileOpener extends Component {
       </div>
     )
   }
-}
-
-const getFileId = ownProps => {
-  return ownProps.fileId || ownProps.router.params.fileId
 }
 
 FileOpener.PropTypes = {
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/drive/FileOpenerExternal.jsx
```diff
@@ -27,16 +27,27 @@ class FileOpener extends Component {
...
+      this.loadFileInfo(this.props.fileId)
+    } else {
+      if (this.props.routeParams.fileId) {
+        this.loadFileInfo(this.props.routeParams.fileId)
+      }
+    }
   }
 
-  async loadFileInfo() {
+  componentWillReceiveProps(nextProps) {
```
cedricmessiant: I would prefer that we don't use `componentWillReceiveProps` ( https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops ). We should use `getDerivedStateFromProps` for this (or maybe `componentDidUpdate` as we are performing a side effect).

By the way, why do you need to do this in both `componentWillMount` and `componentWillReceiveProps` ?

src/drive/web/modules/drive/FileOpenerExternal.jsx
```diff
@@ -27,16 +27,27 @@ class FileOpener extends Component {
     file: null
   }
   componentWillMount() {
-    this.loadFileInfo()
+    if (this.props.fileId) {
+      this.loadFileInfo(this.props.fileId)
+    } else {
+      if (this.props.routeParams.fileId) {
```
y-lohse: Out of curiosity, why use `else { if {` instead of an `else if` statement? I find nested `if`'s to be a code smell, but that might just be me.

src/drive/web/modules/drive/FileOpenerExternal.jsx
y-lohse: Doing it in `componentWillReceiveProps ` solves the bug that we had, because the component is not unmounted in some cases, but the document changes.

But as you say it should be possible to use something else than `componentWillReceiveProps` here :)

Crash--: > I would prefer that we don't use componentWillReceiveProps ( https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops ). We should use getDerivedStateFromProps for this (or maybe componentDidUpdate as we are performing a side effect).

Because we're not using React yet... 

> By the way, why do you need to do this in both componentWillMount and componentWillReceiveProps ?

Because `componentWillReceiveProps` is not call during the initial render()

Crash--: OK, `getDerivatedStateFromProps` was added to Preact 8.3.0, the version we use. 

But DidUdapte will do the job :)

src/drive/web/modules/drive/FileOpenerExternal.jsx
Crash--: Because `else if` doesn't exist in JS and I don't like to rely on indentation ? :D 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else#Description

src/drive/web/modules/drive/FileOpenerExternal.jsx
Crash--: Resolved. I use Didupdate, but again, I need to call it in the constructor and in the didupdate method ;)

src/drive/web/modules/drive/FileOpenerExternal.jsx
y-lohse: Well we're using blacks with `{ }` here so it wouldn't rely on indentation :)

Crash--:
[App Banks -> Drive PDF - 2x (iOS only)](https://trello.com/c/ROHdp6DR/1673-app-banks-drive-pdf-2x-ios-only)

PR #1547, opened 2019-04-02
fix: Download link is now blank

Crash--:


Crash--:
[Regression Viewer - style roll du lien "Télécharger"](https://trello.com/c/bXd5R0uR/1837-regression-viewer-style-roll-du-lien-t%C3%A9l%C3%A9charger)

PR #1548, opened 2019-04-02
fix: Fix file list column width

y-lohse:


y-lohse:
[(𝐋) Custo Drive pour Orange 🍊](https://trello.com/c/ML4nmtCW/1762-%F0%9D%90%8B-custo-drive-pour-orange-%F0%9F%8D%8A)

PR #1549, opened 2019-04-03
Drive classification scenario

CozyKourai:
OK for Review
(CI might still be running as a 2nd run is needed to validate new screenshots)

### DIFF
--- a/testcafe/tests/drive/classification_scenario.js
+++ b/testcafe/tests/drive/classification_scenario.js
@@ -1,0 +1,379 @@
+import { driveUser } from '../helpers/roles'
+import {
+  TESTCAFE_DRIVE_URL,
+  SLUG,
+  isExistingAndVisibile
+} from '../helpers/utils'
+import { initVR } from '../helpers/visualreview-utils'
+import {
+  FILE_FROM_ZIP_PATH,
+  FILE_PDF,
+  FILE_TXT,
+  maskDriveFolderWithDate,
+  maskDeleteFolder,
+  maskMoveMoadal
+} from '../helpers/data'
+import PrivateDrivePage from '../pages/drive/drive-model-private'
+import Modal from '../pages/drive/drive-modal-model'
+
+const moveMoadal = new Modal()
+const privateDrivePage = new PrivateDrivePage()
+
+//Scenario const
+const FEATURE_PREFIX = 'ClassicationScenario'
+
+const FIXTURE_INIT = `${FEATURE_PREFIX} 1- Prepare Data`
+const TEST_CREATE_FOLDER = `1-1 Create Folders`
+const TEST_UPLOAD = `1-2 Upload file`
+
+const FIXTURE_CLASSIFICATION = `${FEATURE_PREFIX} 2- Classification actions on file`
+const TEST_RENAME_FILE = `2-1 Rename file`
+const TEST_MOVE_FILE_CANCEL = `2-2 Move Modal and Cancel`
+const TEST_MOVE_FILE = `2-3 Move file`
+const TEST_MOVE_FOLDER = `2-4 Move folder`
+
+const FIXTURE_TRASH = `${FEATURE_PREFIX} 3- Trash and Restore`
+const TEST_DELETE_FOLDER = `3-1 Delete Folder`
+const TEST_RESTORE_FOLDER = `3-2 Restore Folder`
+const TEST_DELETE_FOLDER_FROM_DRIVE = `3-3 Delete Folder from Drive`
+const TEST_NAVIGATE_IN_TRASH = `3-4 Navigate in Trash`
+const TEST_EMPTY_TRASH = `3-5 Empty Trash`
+
+//************************
+//Tests when authentified
+//************************
+fixture`${FIXTURE_INIT}`.page`${TESTCAFE_DRIVE_URL}/`
+  .before(async ctx => {
+    await initVR(ctx, SLUG, FIXTURE_INIT)
+  })
+  .beforeEach(async t => {
+    console.group(`\n↳ ℹ️  Loggin & Initialization`)
+    await t.useRole(driveUser)
+    await privateDrivePage.waitForLoading()
+    console.groupEnd()
+  })
+  .after(async ctx => {
+    await ctx.vr.checkRunStatus()
+  })
+
+test(`${TEST_CREATE_FOLDER}`, async t => {
+  await t.maximizeWindow() //Real fullscren for VR
+  console.group(`↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_CREATE_FOLDER}`)
+
+  await privateDrivePage.addNewFolder(
+    `${FEATURE_PREFIX}-Folder1`,
+    `${FEATURE_PREFIX}/${TEST_CREATE_FOLDER}-Folder1-1`
+  )
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_CREATE_FOLDER}-Folder1-2`
+  )
+
+  //Create 2nd Folder
+  await privateDrivePage.addNewFolder(
+    `${FEATURE_PREFIX}-Folder2`,
+    `${FEATURE_PREFIX}/${TEST_CREATE_FOLDER}-Folder2-1`
+  )
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_CREATE_FOLDER}-Folder2-2`
+  )
+  console.groupEnd()
+})
+
+test(`${TEST_UPLOAD}`, async t => {
+  console.group(
+    `↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_UPLOAD} (in "${FEATURE_PREFIX}-Folder2" folder)`
+  )
+  await privateDrivePage.goToFolder(`${FEATURE_PREFIX}-Folder2`)
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_UPLOAD}-1`
+  )
+  await privateDrivePage.uploadFiles([`${FILE_FROM_ZIP_PATH}/${FILE_PDF}`])
+  await t.fixtureCtx.vr.setMaksCoordonnates(maskDriveFolderWithDate)
+  await privateDrivePage.takeScreenshotsForUpload(
+    `${FEATURE_PREFIX}/${TEST_UPLOAD}-2`,
+    true
+  )
+  await privateDrivePage.uploadFiles([`${FILE_FROM_ZIP_PATH}/${FILE_TXT}`])
+  await privateDrivePage.takeScreenshotsForUpload(
+    `${FEATURE_PREFIX}/${TEST_UPLOAD}-3`,
+    true
+  )
+  console.groupEnd()
+})
+
+//************************
+//Tests when authentified
+//************************
+fixture`${FIXTURE_CLASSIFICATION}`.page`${TESTCAFE_DRIVE_URL}/`
+  .before(async ctx => {
+    await initVR(ctx, SLUG, FIXTURE_CLASSIFICATION)
+  })
+  .beforeEach(async t => {
+    console.group(`\n↳ ℹ️  Loggin & Initialization`)
+    await t.useRole(driveUser)
+    await privateDrivePage.waitForLoading()
+    console.groupEnd()
+  })
+  .after(async ctx => {
+    await ctx.vr.checkRunStatus()
+  })
+
+test(`${TEST_RENAME_FILE}`, async t => {
+  console.group(`↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_RENAME_FILE}`)
+  const [fileName, ext] = FILE_PDF.split('.')
+  //set mask only once, all screen in this test use the same mask
+  await t.fixtureCtx.vr.setMaksCoordonnates(maskDriveFolderWithDate)
+
+  await privateDrivePage.goToFolder(`${FEATURE_PREFIX}-Folder2`)
+  await privateDrivePage.renameElementByName(
+    FILE_PDF,
+    `${fileName}2.${ext}`,
+    `${FEATURE_PREFIX}/${TEST_RENAME_FILE}-1`,
+    { exitWithEnter: true }
+  )
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_RENAME_FILE}-1-rename2`,
+    true
+  )
+
+  await privateDrivePage.renameElementByName(
+    `${fileName}2.${ext}`,
+    FILE_PDF,
+    `${FEATURE_PREFIX}/${TEST_RENAME_FILE}-2`,
+    { exitWithEnter: false }
+  )
+
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_RENAME_FILE}-2-rename2`,
+    true
+  )
+  console.groupEnd()
+})
+
+test(`${TEST_MOVE_FILE_CANCEL}`, async t => {
+  console.group(`↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_MOVE_FILE_CANCEL}`)
+  //set mask only once, all screen in this test use the same mask
+  await privateDrivePage.goToFolder(`${FEATURE_PREFIX}-Folder2`)
+
+  console.log('Show Action Menu & Cancel')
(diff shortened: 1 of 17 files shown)


CozyKourai:
[(𝐌) SCN - parcours Drive "rangement"](https://trello.com/c/fsWPwkDD/1694-%F0%9D%90%8C-scn-parcours-drive-rangement)

testcafe/tests/pages/drive/drive-model-private.js
```diff
@@ -91,17 +119,21 @@ export default class PrivateDrivePage extends DrivePage {
...
     const numOfFiles = files.length
     let rowCountStart
     if (!t.fixtureCtx.isVR)
       rowCountStart = await this.getContentRowCount('Before Upload')
 
     console.log(`Uploading ${numOfFiles} file(s)`)
-
-    await isExistingAndVisibile(this.btnUpload, 'Upload Button')
-    await t.setFilesToUpload(this.btnUpload, files)
+    if (isDragAndDrop) {
```
CozyKourai: Not "real" drag&Drop upload, but I'm not using the `<input type="file">` from the upload button, but the hidden `<input type="file">` whitch is use for drag&droping

CozyKourai:
last commit ` fix trash error by playing scenario first ` to avoid screenshots problem while in "trash" : As other tests were not design to use screenshots, folders name are using date and time, and so will change every run. 
Because of it I end up with error like 
image

I just move this test so it's the 1st to pass, so there are no other folders in trash!

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+  FILE_PDF,
+  FILE_TXT,
+  maskDriveFolderWithDate,
+  maskDeleteFolder,
+  maskMoveModale
+} from '../helpers/data'
+
+import PrivateDrivePage from '../pages/drive/drive-model-private'
+const privateDrivePage = new PrivateDrivePage()
+import Modal from '../pages/drive/drive-modal-model'
```
y-lohse: Nit: Imports should be grouped at the top of the file :)

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+
+const FIXTURE_TRASH = `${FEATURE_PREFIX} 3- Trash and Restore`
+const TEST_DELETE_FOLDER = `3-1 Delete Folder`
+const TEST_RESTORE_FOLDER = `3-2 Restore Folder`
+const TEST_DELETE_FOLDER_FROM_DRIVE = `3-3 Delete Folder from Drive`
+const TEST_NAVIGATE_IN_TRASH = `3-4 Navigate in Trash`
+const TEST_EMPTY_TRASH = `3-5 Empty Trash`
+
+// const FIXTURE_CLEANUP = `${FEATURE_PREFIX} 3- Cleanup Data`
+// const TEST_DELETE_FOLDER = `3-1 Delete Folder`
```
y-lohse: If these are not used we might as well delete the lines.

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+test(`${TEST_UPLOAD}`, async t => {
+  console.group(
+    `↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_UPLOAD} (in "${FEATURE_PREFIX}-Folder2" folder)`
+  )
+  await privateDrivePage.goToFolder(`${FEATURE_PREFIX}-Folder2`)
+  await t.fixtureCtx.vr.takeScreenshotAndUpload(
+    `${FEATURE_PREFIX}/${TEST_UPLOAD}-1`
+  )
+  await privateDrivePage.uploadFiles([`${FILE_FROM_ZIP_PATH}/${FILE_PDF}`])
+  await t.wait(5000)
```
y-lohse: This is to wait for the thumbnails?

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+  )
+  console.groupEnd()
+})
+
+test(`${TEST_UPLOAD_DD}`, async t => {
+  console.group(
+    `↳ ℹ️  ${FEATURE_PREFIX} : ${TEST_UPLOAD_DD} (in "${FEATURE_PREFIX}-Folder2" folder)`
+  )
+  await privateDrivePage.goToFolder(`${FEATURE_PREFIX}-Folder2`)
+  await privateDrivePage.uploadFiles([`${FILE_FROM_ZIP_PATH}/${FILE_TXT}`])
```
y-lohse: I don't see the difference between this text and the previous one, except the file that is uploaded — but how is it done via drag and drop ?

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+  )
+
+  await privateDrivePage.renameElementByName(
+    `${fileName}2.${ext}`,
+    FILE_PDF,
+    `${FEATURE_PREFIX}/${TEST_RENAME_FILE}-2`,
+    { exitWithEnter: false }
+  )
+
+  await t.fixtureCtx.vr.setMaksCoordonnates(maskDriveFolderWithDate)
```
y-lohse: A few lines earlier a comment says we're only setting the mask once, and here we're setting it again?

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+  FILE_TXT,
+  maskDriveFolderWithDate,
+  maskDeleteFolder,
+  maskMoveModale
+} from '../helpers/data'
+
+import PrivateDrivePage from '../pages/drive/drive-model-private'
+const privateDrivePage = new PrivateDrivePage()
+import Modal from '../pages/drive/drive-modal-model'
+const moveModale = new Modal()
```
y-lohse: Nit, but it's all over the file: `modal` instead of `modale`

testcafe/tests/drive/classification_scenario.js
```diff
@@ -0,0 +1,395 @@
...
+  )
+  //Remove comment below when https://trello.com/c/RoNPTPZV/1692-modale-d%C3%A9placement-touch-echap-ne-ferme-pas-la-modale is fixed
+  // console.log('Show Move Modale, and Cancel (esc)')
+  // await privateDrivePage.showMoveModalForElement(FILE_PDF)
+  // // no need to screenshot again the modal
+  // await t.pressKey('esc')
+  // await t.fixtureCtx.vr.takeScreenshotAndUpload(
+  //   `${FEATURE_PREFIX}/${TEST_MOVE_FILE_CANCEL}-4`,
+  //   true
+  // )
```
y-lohse: Why is this section commented out?

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -29,6 +29,7 @@ export class VisualReviewTestcafe extends VisualReview {
 
   //@param { jsonObject } masK { x, y, width, height } : coordonate for mask. order is the same in gimp rectangle selection
   async setMaksCoordonnates(maskCoordonates) {
+    this.resetMask
```
y-lohse: Why is this lin here? It's also not calling the function, so I think we can remove it :)

testcafe/tests/pages/drive/drive-model-private.js
```diff
@@ -1,13 +1,40 @@
...
 import DrivePage from './drive-model'
 
-export default class PrivateDrivePage extends DrivePage {
+import Modal from './drive-modal-model'
+const moveModale = new Modal()
+
+export default class privateDrivePage extends DrivePage {
+  constructor() {
+    super()
+    this.elementActionMenuByRowIndex = value => {
```
y-lohse: Shouldn't this be a class method?

testcafe/tests/pages/drive/drive-model-private.js
```diff
@@ -232,17 +265,21 @@ export default class PrivateDrivePage extends DrivePage {
...
+      .expect(this.modalFooter.visible)
       .ok('Delete button does not show up')
-      .click(this.modalDeleteBtnDelete)
-    await t.takeScreenshot()
-    const rowCountEnd = await this.getContentRowCount('After')
-    await t.expect(rowCountEnd).eql(rowCountStart - filesIndexArray.length)
+      .click(this.modalSecondButton)
+    if (!t.fixtureCtx.isVR) {
+      const rowCountEnd = await this.getContentRowCount('After')
+      await t.expect(rowCountEnd).eql(rowCountStart - filesIndexArray.length)
```
y-lohse: We're using `rowCountStart` here but it can be `undefined`, is that normal?

testcafe/tests/pages/drive/drive-model-private.js
```diff
@@ -264,21 +301,110 @@ export default class PrivateDrivePage extends DrivePage {
...
+  //@param {String} screenshotPath : path for screenshots taken in this test
+  //@param { integer } index : file/folder index
+  //@param {String} newName : new name for file/folder
+  //@param {bool} : exitWithEnter
+  async renameElementsByIndex(index, newName, screenshotPath, exitWithEnter) {
+    await this.selectElements([index])
+
+    await t.click(this.cozySelectionbarBtnRename)
+    await isExistingAndVisibile(this.foldersNamesInputs, 'Folder Name input')
+    //dates showsw up here, so there is a mask for screenshots
```
y-lohse:
```suggestion
    //dates shows up here, so there is a mask for screenshots
```

testcafe/tests/pages/drive/drive-model-private.js
```diff
@@ -1,13 +1,40 @@
...
   getPageUrl,
   isExistingAndVisibile,
   overwriteCopyCommand,
   getLastExecutedCommand
 } from '../../helpers/utils'
 import DrivePage from './drive-model'
 
-export default class PrivateDrivePage extends DrivePage {
+import Modal from './drive-modal-model'
+const moveModale = new Modal()
```
y-lohse: Is this the best place to initialize this? Wouldn't it make more sense to do it in the constructor for example?

testcafe/tests/pages/drive/drive-model.js
```diff
@@ -98,8 +100,14 @@ export default class DrivePage {
     this.cozySelectionbarBtnShare = this.cozySelectionbar
       .find('button')
       .withText('SHARE') //!FIX ME : do not use text! Do not use .nth(x) because the buttons count changes if one or several files/folders are selected
-    this.modalDelete = Selector('[class*="c-modal"]').find('div')
-    this.modalDeleteBtnDelete = this.modalDelete.find('button').nth(2) //REMOVE
+    this.cozySelectionbarBtnRename = this.cozySelectionbar
+      .find('button')
+      .withText('RENAME') //!FIX ME : do not use text! Do not use .nth(x) because the buttons count changes if one or several files/folders are selected
```
y-lohse: What's the hold-up for fixing this?

testcafe/tests/pages/drive/drive-modal-model.js
```diff
@@ -0,0 +1,31 @@
...
+      .find('h2')
+      .withAttribute('data-test-id', 'path-title')
+    this.folderOrFileName = this.modaleContent
+      .find('div')
+      .withAttribute('data-test-id', 'fil-file-filename-and-ext')
+  }
+
+  async moveElementTo(destinationFolder) {
+    await this.goToBaseFolder()
+    await this.goToFolder(destinationFolder)
```
y-lohse: I'm confused: `goToFolder ` is used to navigate to another folder on the main page, not on the modal. How does it know to navigate inside the modal?

testcafe/tests/drive/classification_scenario.js
CozyKourai: https://github.com/cozy/cozy-drive/pull/1549#discussion_r272105270

There are 2 `<input type="file">` on the page.
It's not a "real" drag'n'drop, but we don't want to spend time developping a real one.
I can but a regular upload if you think it's better

testcafe/tests/drive/classification_scenario.js
CozyKourai: yes, but only before merging https://github.com/cozy/cozy-drive/pull/1556

testcafe/tests/drive/classification_scenario.js
CozyKourai: Because the bug https://trello.com/c/RoNPTPZV/1692-modale-d%C3%A9placement-touch-echap-ne-ferme-pas-la-modale is not fixed, so if I uncomment it, the test will failed
(I wrote to remove this comment in the card)

testcafe/tests/pages/drive/drive-modal-model.js
CozyKourai: Because the selector use to `goToFolder` , `this.folderOrFileName` is redefined here, so I can use the same function. But this will likely change while working on https://trello.com/c/22GedOwK/1850-%F0%9D%90%92-check-de-tous-les-selecteurs-sortir-des-classes-et-faire-un-fichier-communs-avec-tous-les-selecteurs-en-exports

testcafe/tests/pages/drive/drive-model-private.js
CozyKourai: well `this.elementActionMenuByRowIndex` is a selector (with a parameter) and I already have a few of them written that way (ie: https://github.com/cozy/cozy-drive/blob/master/testcafe/tests/pages/photos/photos-model.js#L39)

testcafe/tests/drive/classification_scenario.js
y-lohse: But we're not using the `isDragAndDrop ` parameter here?

testcafe/tests/pages/drive/drive-model-private.js
y-lohse: Ok, hadn't noticed it before. Maybe an improvement for another PR.

testcafe/tests/pages/drive/drive-model-private.js
CozyKourai: `rowCountStart`and `rowCountEnd` are only defined and used when we don't use the screenshots as it was the way to check the expected files where removed, but with the screenshots we don't need those checks as the screenshots will show the files were removed

testcafe/tests/drive/classification_scenario.js
y-lohse: Ok! It doesnt have to be in this PR, but it would be good to use a variable like `THUMBNAIL_DELAY` instead of a hard-coded value like that. Not only is it easier to change, it also explains why the timeout is there :)

testcafe/tests/pages/drive/drive-model-private.js
y-lohse: Oh sorry my bad, I missed the second `if (!t.fixtureCtx.isVR) ` 👍

testcafe/tests/pages/drive/drive-modal-model.js
y-lohse: Mmh ok. This is definitely a bit of a tricky behavior, so we'll see how we can improve it with the other card then :)

testcafe/tests/drive/classification_scenario.js
CozyKourai: that's true. and It's not supposed to be that way! Let me add it again!

testcafe/tests/pages/drive/drive-model.js
CozyKourai: https://trello.com/c/22GedOwK/1850-%F0%9D%90%92-check-de-tous-les-selecteurs-sortir-des-classes-et-faire-un-fichier-communs-avec-tous-les-selecteurs-en-exports : I will have a better idea of which cozyBar buttons shows up, in order to use relevant (and easy to read) selector

testcafe/tests/pages/drive/drive-modal-model.js
```diff
@@ -0,0 +1,31 @@
...
+} from '../../helpers/utils'
+import { t } from 'testcafe'
+import DrivePage from './drive-model'
+
+export default class MoveModal extends DrivePage {
+  constructor() {
+    super()
+
+    //Overwrite some selector to use our regular drive method in modal context
+    this.modaleContent = getElementWithTestId('fil-content-modal')
```
y-lohse:
```suggestion
    this.modalContent = getElementWithTestId('fil-content-modal')
```

testcafe/tests/pages/drive/drive-modal-model.js
```diff
@@ -0,0 +1,31 @@
...
+
+    this.breadcrumb = this.modal
+      .find('h2')
+      .withAttribute('data-test-id', 'path-title')
+    this.folderOrFileName = this.modaleContent
```
y-lohse:
```suggestion
    this.folderOrFileName = this.modalContent
```

PR #1550, opened 2019-04-03
Panning the pdf viewer

y-lohse:
Most notably on iOS, it was not possible to pan inside a zoomed pdf. This is because somehow, the swipe and pan handlers defined on the ViewerControls kept intercepting regular touch events even though they shouldn't.

The fix I propose is to disable these event handlers when we are zoomed into the pdf.


y-lohse:
[Swipe sur pdf viewer sur iOS](https://trello.com/c/7cqTnbhx/1845-swipe-sur-pdf-viewer-sur-ios)

PR #1552, opened 2019-04-04
Fix permissions

gregorylegarec:
* Explicit default to `["ALL"]`
* Fix methods prop and use verbs instead. This change could imply regressions as some permissions are passing from `["ALL"]` to a smaller verbset ⚠️


gregorylegarec:
@Crash-- @y-lohse I will let you merge, I think it's pretty important to get this update into today's releases :)

Crash--:
https://github.com/cozy/cozy-drive/pull/1554

PR #1555, opened 2019-04-04
Fix/personalization

Crash--:
- Sharing > Select > Check 
- ErrorSharing 
- DropZone

### DIFF
--- a/src/components/Button/getHomeLinkHref.jsx
+++ b/src/components/Button/getHomeLinkHref.jsx
@@ -1,6 +1,6 @@
 const getHomeLinkHref = from =>
   `https://manager.cozycloud.cc/cozy/create${
-    from ? `?pk_campaign=${encodeURIComponent(from)}` : ''
-  }`
+    from ? `?pk_campaign=${encodeURIComponent(from)}` : '?'
+  }&pk_kwd=cozy`
 
 export default getHomeLinkHref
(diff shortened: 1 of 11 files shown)


src/components/Button/getHomeLinkHref.jsx
```diff
@@ -1,6 +1,6 @@
 const getHomeLinkHref = from =>
   `https://manager.cozycloud.cc/cozy/create${
     from ? `?pk_campaign=${encodeURIComponent(from)}` : ''
-  }`
+  }&pk_kwd=cozy`
```
cedricmessiant: What is `pk_kwd` ?

src/drive/web/modules/upload/DropzoneTeaser.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { translate, Icon } from 'cozy-ui/react'
+
+import styles from 'drive/styles/dropzone'
+
+const DropzoneTeaser = translate()(({ t, currentFolder }) => (
+  <div className={styles['fil-dropzone-teaser']}>
+    <div className={styles['fil-dropzone-teaser-claudy']}>
+      <Icon icon="cozy-negative" size={'2.5em'} color="white" />
```
y-lohse: Can we use a pixel value here instead of em please?

src/drive/web/modules/upload/DropzoneTeaser.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { translate, Icon } from 'cozy-ui/react'
+
+import styles from 'drive/styles/dropzone'
+
+const DropzoneTeaser = translate()(({ t, currentFolder }) => (
```
cedricmessiant: a little snapshot test for this component would be nice ;-)

PR #1556, opened 2019-04-04
fix: change screenshots when uploading files (thumb problems)

CozyKourai:
With this PR the screenshots will be a bit different from before so CI will failed until they are set as new baseline, and we have to wait to merge to do that, other wise, other PR will fail because of screenshots! 

New screenshots : 
- When uploading files, now, we wait 10s, then reload to avoid thumbnail problem : https://visualreview.cozycloud.cc/api/image/21707
- Because of this reload, the "upload div" can't be seen, So I added a screenshot on this component only : https://visualreview.cozycloud.cc/api/image/21704

### DIFF
--- a/testcafe/tests/helpers/visualreview-utils.js
+++ b/testcafe/tests/helpers/visualreview-utils.js
@@ -29,6 +29,7 @@ export class VisualReviewTestcafe extends VisualReview {
 
   //@param { jsonObject } masK { x, y, width, height } : coordonate for mask. order is the same in gimp rectangle selection
   async setMaksCoordonnates(maskCoordonates) {
+    this.resetMask()
     this.options.mask = {
       excludeZones: [
         {
@@ -50,6 +51,7 @@ export class VisualReviewTestcafe extends VisualReview {
       //re-init mask
       this.resetMask()
     }
+    //always wait for 1s before taking screenshots
     await t.takeScreenshot(`${screenshotsPath}.png`)
 
     this.options.properties.os = await getNavigatorOs()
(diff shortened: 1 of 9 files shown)


CozyKourai:
[(𝐒) Corriger quand la méthode d'appel du screenshot est appelée après l'upload sur photo](https://trello.com/c/9dBkqsfc/1795-%F0%9D%90%92-corriger-quand-la-m%C3%A9thode-dappel-du-screenshot-est-appel%C3%A9e-apr%C3%A8s-lupload-sur-photo)

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -29,6 +29,7 @@ export class VisualReviewTestcafe extends VisualReview {
 
   //@param { jsonObject } masK { x, y, width, height } : coordonate for mask. order is the same in gimp rectangle selection
   async setMaksCoordonnates(maskCoordonates) {
+    this.resetMask
```
Crash--: resetMask()

testcafe/tests/helpers/visualreview-utils.js
```diff
@@ -50,14 +51,34 @@ export class VisualReviewTestcafe extends VisualReview {
...
+  async takeElementScreenshotAndUpload(
+    selector,
+    screenshotsPath,
+    hasMask = false
+  ) {
+    if (!hasMask) {
+      //re-init mask
+      this.resetMask()
+    }
+    //always wait for 1s before taking screenshots
```
y-lohse: How are we waiting 1s before taking the screenshot?

PR #1568, opened 2019-04-11
fix: Reduce text viewer padding

y-lohse:
After the viewer refactor, the text viewer was a bit too low on the page.


y-lohse:
[Affichage viewer text/Md](https://trello.com/c/gwiKjPqS/1875-affichage-viewer-text-md)

PR #1573, opened 2019-04-11
fix(Photo): Upload string

Crash--:
- [ ] tx push -s after meger


Crash--:
[Photos - labels FR et ES](https://trello.com/c/8omOajZr/1790-photos-labels-fr-et-es)

PR #1574, opened 2019-04-12
Use next instead of hasMore (undefined)

cedricmessiant:
`hasMore` is `undefined`, we have to use `next` instead

https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination


Crash--:
https://github.com/cozy/cozy-client/issues/345

cedricmessiant:
Oh, thanks. Took a quick look at cozy-client, it is not obvious how `hasMore` could be `undefined`

PR #1580, opened 2019-04-15
fix: Update RealTime when context token change

Crash--:
If the context has a change, we unsuscribe the socket and registrer it again with the new information

### DIFF
--- a/src/drive/web/modules/navigation/RealtimeFiles.spec.jsx
+++ b/src/drive/web/modules/navigation/RealtimeFiles.spec.jsx
@@ -1,0 +1,113 @@
+import { mount } from 'enzyme'
+import React from 'react'
+import realtime from 'cozy-realtime'
+import { RealtimeFiles } from './RealtimeFiles'
+global.cozy = {
+  client: {
+    offline: {
+      getDatabase: jest.fn()
+    }
+  }
+}
+
+jest.mock('cozy-realtime', () => {
+  const mockReturn = {
+    onCreate: jest.fn(() => mockReturn),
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/navigation/RealtimeFiles.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+}
+
+jest.mock('cozy-realtime', () => {
+  const mockReturn = {
+    onCreate: jest.fn(() => mockReturn),
+    onUpdate: jest.fn(() => mockReturn),
+    onDelete: jest.fn(() => mockReturn),
+    unsubscribe: jest.fn(() => mockReturn)
```
cedricmessiant: seems strange

Crash--: This is a hacky solution, because realtime let you chain your call : 

realtime.subscribe().onCreate().onDelete()

with that hacky solution, we can mock it. If you have something better ... :)

Crash--:
> There is some commented code that can be removed, but more generally speaking: does that even happen? When the client's token changes, does it trigger an update of all components???

https://reactjs.org/docs/legacy-context.html
```
If contextTypes is defined within a component, the following lifecycle methods will receive an additional parameter, the context object:

constructor(props, context)
componentWillReceiveProps(nextProps, nextContext)
shouldComponentUpdate(nextProps, nextState, nextContext)
componentWillUpdate(nextProps, nextState, nextContext)
```

So I think so

PR #1584, opened 2019-04-16
Fix: some bugs

Crash--:
- Upgrade cozy-ui to fix bug scroll after creating a new folder 
- Add ScrollToTop component in order to scrollTo when we mount <Settings />


src/drive/mobile/modules/settings/Settings.jsx
```diff
@@ -37,8 +37,13 @@ class Settings extends Component {
   render() {
     const { t } = this.props
     const isDebug = this.state.tapCount >= 3
+    /*
+    See https://reacttraining.com/react-router/web/guides/scroll-restoration
+    for the ScrollToTop explanation
```
cedricmessiant: :+1:

PR #1587, opened 2019-04-16
fix: add hover on arrow in viewer navigation

CozyKourai:
For now, there a no specific hover when navigating in the viewer.
Now, it happens like this : 
- choosing a starting file/photos
- open the viewer, and check there is a next file
- navigate to next file
- hover the **previous** button (which brings us back to the file we opened)
- take screenshots

Same thing when navigating to previous image, we will hover the **next** button.
So 1st screenshots will fail as for now, there were no arrow

### DIFF
--- a/testcafe/tests/pages/viewer/viewer-model.js
+++ b/testcafe/tests/pages/viewer/viewer-model.js
@@ -62,6 +62,11 @@ export default class Viewer {
     )
     for (let i = startIndex; i < startIndex + numberOfNavigation; i++) {
       await this.navigateToNextFile(i)
+      //navigateToNextFile brings you on the next file, so we are sure there is a Previous Button on this image
+      await t.hover(selectors.btnViewerNavPrevious, {
+        offsetX: 0,
+        offsetY: 0
+      })
       if (t.fixtureCtx.isVR)
         await t.fixtureCtx.vr.takeScreenshotAndUpload(
           `${screenshotPath}-${i}-next`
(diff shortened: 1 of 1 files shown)


CozyKourai:
[Viewer > Navigation > Ajouter un hover sur les fleches pour rendre les screenshots plus explicit](https://trello.com/c/OlYZWpmT/1870-viewer-navigation-ajouter-un-hover-sur-les-fleches-pour-rendre-les-screenshots-plus-explicit)

testcafe/tests/pages/viewer/viewer-model.js
```diff
@@ -62,6 +62,11 @@ export default class Viewer {
     )
     for (let i = startIndex; i < startIndex + numberOfNavigation; i++) {
       await this.navigateToNextFile(i)
+      //navigateToNextFile brings you on the next file, so we are sure there is a Previous Button on this image
+      await t.hover(selectors.btnViewerNavPrevious, {
+        offsetX: 0,
```
y-lohse: I don't think we need to specify an offset in that case?

CozyKourai: I think it's best to add an offset when hovering for screenshots, so we are 100% sure the hover will be on exact same spot test after test.
I agree here we don't really need it, as `btnViewerNavPrevious/Next` is not that big so there is not so many chance to trigger another div on page.

So do we keep the offset as best practice or do we remove it here?

Crash--: I think we can keep it. I prefer to set the position manually to be sure where the cursor will be. We already had some issues when not doing that.

PR #1591, opened 2019-04-16
fix: Use transpiled version of cozy-ui

y-lohse:
There are some differences in css orders when we build the app for production, but using the transpiled version of cozy-ui solves the problem.

See https://github.com/cozy/cozy-ui/issues/335 for the original issue.


y-lohse:
[App - problèmes d'affichage](https://trello.com/c/vED6SbG3/1883-app-probl%C3%A8mes-daffichage)

Crash--:
👍

PR #1593, opened 2019-04-17
chore(testcafe): Remove fix error js

CozyKourai:
No more cozy-bar js error, so we can remove 
```
skipJsErrors: true,
skipUncaughtErrors: true
```


Crash--:
In fact there was no discussion about that. I just wanted to be sure that new version of the bar fixed our issue. It is the case. 

Since we don't know which version is served by the stack, I think we can close this PR and let the ignore in our production test suite 👍

PR #1606, opened 2019-04-18
Fix/menutoolbar

Crash--:
- No need to renderItems() if the menu is not opened (better perf and less css) 
- I don't like to cloneElement. So I just wrapped the item in a div only when needed
- No need to wrap the item if we don't add the onclick 
- Sometimes Item is wrapped in an HOC so we can't check the type. The idea is that case is to provide a "type" props. If we don't do that, we'll be able to close the Menu by clicking on the `<hr />`. Good luck with that because hr is pretty small ;)  

edit: formatting by @y-lohse

### DIFF
--- a/src/components/Menu/index.jsx
+++ b/src/components/Menu/index.jsx
@@ -66,20 +66,20 @@ export default class Menu extends Component {
           </button>
         )}
         {opened && (
-          <Overlay
-            onClick={this.close}
-            onEscape={this.close}
-            className={styles['cozy-menu-overlay']}
-          />
+          <>
+            <Overlay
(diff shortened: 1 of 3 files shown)


Crash--:
[Menu d'action courant - reste ouvert après choix](https://trello.com/c/iUQQXZc1/1898-menu-daction-courant-reste-ouvert-apr%C3%A8s-choix)

Crash--:
> I don't understand what was causing the initlal bug 🤔

Because `nodeName` was a... Preact property.

Crash--:
updated screenshots baseline since the menu is now hidden ;)

PR #1611, opened 2019-04-18
Fix/quick view

Crash--:
Several points in this PR : 

- We now have a concept of "view". View is based on our navigation system. Currently we have 4 views : `folder / trash / recent / sharings`

- For each of this view we have a special "slice" of the redux store to store the concerned files : 
 `files : { shared: [], recent: [], trashed: [], folder: [] }`

- We have a selector to get the `currentView` based the latest dispatched action. It is not the `displayedView` but the desired one 

- Tried to use the redux store instead of the `location` when I could without rewriting a few big part of the app. 

- Added lot of documentation about internal stuff


Crash--:
[Vue Drive/Récents - clic rapide](https://trello.com/c/7vO1ECD7/1695-vue-drive-r%C3%A9cents-clic-rapide)

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -75,6 +75,11 @@ const displayedFolder = (state = null, action) => {
   }
 }
 
+const latestAction = (state, action) => {
+  return action.type
```
y-lohse: Do we need a default value or state for this reducer?

Crash--:
I think adding tests to SharingsContainer should be doable since some are already there. For the "actions" part, it seems like a huge effort. 

@y-lohse be prepared, tomorrow I'll have to work on a new task so... :D

cedricmessiant:
I don't think you can be sure that another action has not been be dispatched in the meantime. If this action has nothing to do with files fetching you still want to update the store with recent files. 

`files` state may contain recent files, shared files, etc (and fetch the files "on demand") but the route should select only the files it is interested in. 

For example, the store could look like this:

```
files: {
  shared: [],
  recent: [],
},
currentDisplayedFolder: 'recent'
```

Of course it is a huge change and we probably can't do this to fix this bug but maybe we could create a technical card.

Crash--:
> I don't think you can be sure that another action has not been be dispatched in the meantime. If this action has nothing to do with files fetching you still want to update the store with recent files.

You right. After 2 minutes of thinking, we can receive UPLOAD_SUCCESS or other related sync stuff. 

At first, I wanted to do something similar at what you proposed. It was a bit more complicated although. Your solution should be doable.

Crash--:
Some news: 

We now have: 
```jsx
 files : { shared: [], recent: [], trashed: [], folder: [] },
currentView: 'recent' // https://github.com/cozy/cozy-drive/pull/1611/commits/4eeea76b527218031fafffc93b997c8a7b56a8d8#diff-47554ec30330a7ef6eb836864a43b9f4R78
```

Based on the `currentView` we return the right reducer path: https://github.com/cozy/cozy-drive/pull/1611/commits/4eeea76b527218031fafffc93b997c8a7b56a8d8#diff-47554ec30330a7ef6eb836864a43b9f4R425 

I found out some "issues" doing this stuff: 
- I had to hack this realtime line (https://github.com/cozy/cozy-drive/pull/1611/commits/4eeea76b527218031fafffc93b997c8a7b56a8d8#diff-9d445e7faef1808a83ef82ea6ba523efR98) because we're dispatching the exact same action for a deletion and to empty the trash. 
```
Quand dans un nfolder, on supprimer des fichiers : 
- TRash_files(ids)
 + delete_file(id) (x X)


Quand on supprime depuis la corbeille des fichiers définitivement : 
- Destroy_files(ids)
    + delete_file(id) (x X)
```

But the delete_file is not dispatched by the click, but by the real time (https://github.com/cozy/cozy-drive/pull/1611/commits/4eeea76b527218031fafffc93b997c8a7b56a8d8#diff-9d445e7faef1808a83ef82ea6ba523efR99). Since we're removing files here, we can have a race between this and this line https://github.com/cozy/cozy-drive/blob/4eeea76b527218031fafffc93b997c8a7b56a8d8/src/drive/web/modules/trash/actions.js#L79. Sometimes, the map doesn't have the right value. 


There is still a lot of work to be done: 
- I think I broke the pagination
- We can refactor the `Container` stuff. Since we now have a specific `path` for each view, we can optimise them to ne re-render if they are not concerned. 
- We can remove lot of stuff (no need to store the latest action). 
- We can change some check. Currently we are checking if the displayFolder is null to know if we're displaying trash or not.

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -20,7 +21,11 @@ export class SharingFetcher extends React.Component {
     client: PropTypes.object.isRequired,
     t: PropTypes.func.isRequired
   }
-
+  constructor(props) {
+    super(props)
+    this.fetchSharedFiles = null
+    this.fetSharedParents = null
```
cedricmessiant: typo: `fetchSharedParents`

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -50,15 +59,25 @@ export class SharingFetcher extends React.Component {
       }))
       this.props.fetchSuccess(filesWithPath)
     } catch (e) {
-      this.setState({ error: e })
-      this.props.fetchFailure(e)
+      /* 
+      If the error is a cancelable promise, don't use setState sinnce the
```
cedricmessiant: typo: `since`

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -50,15 +59,25 @@ export class SharingFetcher extends React.Component {
       }))
       this.props.fetchSuccess(filesWithPath)
     } catch (e) {
-      this.setState({ error: e })
-      this.props.fetchFailure(e)
+      /* 
+      If the error is a cancelable promise, don't use setState sinnce the 
+      component is not mounted anymore
+      */
+      if (e.isCanceled !== true) {
```
cedricmessiant: why not `if (!e.isCanceled)` ?

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -50,15 +59,25 @@ export class SharingFetcher extends React.Component {
...
+      }
     }
   }
 
   componentDidMount() {
     this.fetchSharedDocuments()
   }
-
+  componentWillUnmount() {
+    //In order to not dispatch fetch sucess, we cancel the promises
```
cedricmessiant: typo: `success`

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,95 @@ import {
...
     ) {
       this.navigateToFolder(
         getFolderIdFromRoute(newProps.location, newProps.params)
       )
-    } else if (
-      urlHasChanged(this.props, newProps) &&
-      isRecentFilesView(newProps)
-    ) {
-      this.props.fetchRecentFiles()
     }
```
cedricmessiant: what if the url is sharings/trash ? did I miss something?

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,95 @@ import {
...
   }
 
+  /*
+   @performance 
+
+    Thanks to this, we're updating twice the redux store (url and fetchfolder) 
+    and we cause a lof of re-render issues
+
+    Very deep in the three (File.jsx), we use navigateToForlder renamed as
```
cedricmessiant: typos: tree / navigateToFolder

src/drive/web/modules/navigation/RealtimeFiles.jsx
```diff
@@ -93,7 +93,9 @@ export class RealtimeFiles extends React.Component {
   }
 
   onDocumentDeletion = doc => {
-    if (this.isInCurrentView(doc)) this.props.deleteFile(this.normalizeId(doc))
+    //Ne pas faire un deleteFile quand on vide la corbeille car on a déjà une action pour cela
```
cedricmessiant: :fr:

src/lib/promise.js
```diff
@@ -0,0 +1,18 @@
+//https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html
+export const makeCancelable = promise => {
```
cedricmessiant: maybe add a test for this helper?

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -55,10 +55,15 @@ export const META_DEFAULTS = {
   hideActionMenu: true
 }
 
-export const openFolder = folderId => {
+export const openFolder = (
+  folderId,
+  openFolderAction = OPEN_FOLDER,
```
cedricmessiant: why not use another param for the action (something like `context` for example) with a default value instead of using different action types? If you use action type you have to add the type in every `switch()` on type whereas using a distinct param, you could treat special cases only when it's necessary

Crash--: In fact this is precisely the goal.  I want to be able to do what I want for a special action without worrying about the consequence because the scoped is controlled. 

The goal is to have in a few weeks / months a clean reducer, cleaned Controllers and cleaned Action. 

I thought about this solution, but I didn't want to add complexity to the reducer and do not want having to do something like : 

```jsx
case OPEN_FOLDER:
if(action.payload.params === 'trash'){
return {
...state,
view: trash
}
}else {
return {
...state,
view: folder
}
}
```

Do you see any value to do that ? Except the `you have to switch on the new action in the reducer`

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -50,15 +59,25 @@ export class SharingFetcher extends React.Component {
...
+    if (this.fetchSharedFiles) this.fetchSharedFiles.cancel()
+    if (this.fetSharedParents) this.fetSharedParents.cancel()
```
ptbrowne: typo: fetchSharedParents

src/drive/web/modules/filelist/File.jsx
```diff
@@ -366,6 +369,27 @@ class File extends Component {
...
+  withFilePath: PropTypes.bool,
+  withSharedBadge: PropTypes.bool.isRequired,
+  isAvailableOffline: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool,
+  breakpoints: PropTypes.object.isRequired,
+  selectionModeActive: PropTypes.func,
+  //When a user click on a Folder
+  onFolderOpen: PropTypes.func.isRequired,
+  //onFileOpen : When a user click on a File
+  onFileOpen: PropTypes.func.isRequired,
```
ptbrowne: `/** */` are better suited for JSDOC

src/lib/promise.js
Crash--: will use the one from cozy-client ;)

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,90 @@ import {
...
     } else {
-      this.props.fetchFolder(
-        getFolderIdFromRoute(this.props.location, this.props.params)
-      )
+      this.props.fetchFolder(getFolderIdFromRoute(location, params))
     }
   }
 
+  /*
+    We need this method since <Nav is using href and not onClick
```
ptbrowne: typo: Nav

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,90 @@ import {
...
+
+    This is only needed for this case, since the navigation in a Folder 
+    is managed by navigateToFolder function.
+
+    We should perhaps have : 
+    - FileFolderExplorer 
+    - RecentFileExplorer 
+    - TrashFileExplorer ?
+
+    We have to use newProps and newProps.location, because has we use both :
```
ptbrowne: d/has/

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,90 @@ import {
...
+    - props location 
+    and 
+    - redux store 
+    props.location is the first to be updated 
+
+    Also we check the matching between the openedFilderId (and not the displayFolderId)
+    and the new props location (folder id in the url)
+
+    I can't use isRecentFilesView for instance, since when we click on the "Recent" links
+    we don't dispatch any action. We're just listeninng the URL, and dispatch the actio
```
ptbrowne:
```suggestion
    we don't dispatch any action. We're just listening the URL, and dispatch the action
```

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -81,17 +73,28 @@ export const openFolder = folderId => {
...
+        and display. To do that, we get the folder id from the latest OPEN_FOLDER call
+      */
+      const currentFolderId = getOpenedFolderId(getState())
+      if (folder._id === currentFolderId) {
+        return dispatch({
+          type: openFolderActionSucess,
+          folder,
+          fileCount: folder.contents.meta.count || 0,
+          files: folder.contents.data
+        })
```
ptbrowne: IMHO this kind of thing should be done in the reducer. It should ignore the success if it is for the wrong id.

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -62,22 +68,70 @@ const isOpening = (state = false, action) => {
...
 
+/*
+  It seems that we have a distinction between 
+  `displayedFolder` which is set after a success (generaly async)
+  and 
+  `openedFolderId` which is set just after the dispatch of OPEN_FOLDER
+  without waiting for any success / faillure 
+
+  Also `openedFilderId` is just the uuid of the folder. 
+  `displayFolder` is a io.cozy.file object
```
ptbrowne:
```suggestion
  `displayedFolder` is a io.cozy.file object
```

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -62,22 +68,70 @@ const isOpening = (state = false, action) => {
...
+    case FETCH_SHARINGS:
+    case FETCH_RECENT:
       return null
     default:
       return state
   }
 }
 
+/*
+  CurrentView is not async. It is called at the begining
```
ptbrowne: beginning

src/drive/web/modules/navigation/duck/actions.jsx
Crash--: You absolutely right! But we have a "special" reducer in drive. We splited / sliced our reducer and state. So for instance here https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/navigation/duck/reducer.js#L197 I can't have a direct access to the "global" state since in this method I'm scoped. So no acess `state.openedFolderId` :(

Maybe I missed something to get access to it. I'm very open to suggestion here !

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -62,8 +62,8 @@ export class SharingFetcher extends React.Component {
       }))
       this.props.fetchSuccess(filesWithPath)
     } catch (e) {
-      /* 
-      If the error is a cancelable promise, don't use setState sinnce the 
+      /*
+      If the error is a cancelable promise, don't use setState sinnce the
```
ptbrowne: since

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -94,31 +94,23 @@ export class SharingFetcher extends React.Component {
...
-      }
-      if (
-        !this.props.params.folderId &&
-        this.props.params.folderId !== prevProps.params.folderId
-      ) {
-        this.fetchSharedDocuments()
-      }
+    }
+    else if (movedToSharedRoot) {
+      // if we start the navigation inside a folder in the saring view, and navigate back to the root, we need to load the root content again
```
ptbrowne: sharing

If (majuscule)

src/drive/web/modules/drive/SharingsContainer.jsx
```diff
@@ -94,31 +94,23 @@ export class SharingFetcher extends React.Component {
...
+      this.fetchSharedDocuments()
+    }
+    else if (movedAwayFromSharedRoot) {
+      // in case we open a folder from the root fo the sharing view
```
ptbrowne: In case (majuscule

to the sharing view

src/lib/promise.js
Crash--: @y-lohse : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/utils.js#L9 ;)

src/drive/web/modules/trash/Container.jsx
```diff
@@ -36,9 +42,30 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
...
 
+/*
+
+navigateToFolder = async folderId => {
+    await this.props.fetchFolder(folderId)
+    this.props.router.push(getFolderUrl(folderId, this.props.location))
+  }
+  fetchFolder: folderId => dispatch(openFolder(folderId)),
+
+*/
```
Crash--: we can remove

src/drive/web/modules/navigation/getFolderPath.js
```diff
@@ -3,25 +3,59 @@ import { ROOT_DIR_ID, TRASH_DIR_ID } from 'drive/constants/config'
...
+  /*   switch (openedFolderId && !displayedFolder.parent) {
+    case ROOT_DIR_ID:
+    case TRASH_DIR_ID:
+      path.push({ id: openedFolderId })
+      return path
+    default:
+      console.log('brefore break')
+      break
+  } */
+  //console.log('after break')
```
Crash--: we can remove all theses comments too.

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -188,45 +256,112 @@ const indexFor = (file, array, compareFn, start = 0, end = array.length) => {
...
+        Special case for the Trash dir since `Trash` has its own 
+        reducer to store its files and that Trash is a "regular"
+        folder
+        */
+      if (action.folder.id === TRASH_DIR_ID) {
+        return {
+          ...state,
+          trashed: action.files
+        }
+      }
```
Crash--: I think we can remove this special case since now we use `OPEN_FOLDER_FROM_TRASH` right?

src/drive/web/modules/navigation/RealtimeFiles.jsx
Crash--: See : 
 `
I had to hack this realtime line (4eeea76#diff-9d445e7faef1808a83ef82ea6ba523efR98) because we're dispatching the exact same action for a deletion and to empty the trash.
Quand dans un nfolder, on supprimer des fichiers : 
- TRash_files(ids)
 + delete_file(id) (x X)


Quand on supprime depuis la corbeille des fichiers définitivement : 
- Destroy_files(ids)
    + delete_file(id) (x X)
But the delete_file is not dispatched by the click, but by the real time (4eeea76#diff-9d445e7faef1808a83ef82ea6ba523efR99). Since we're removing files here, we can have a race between this and this line

cozy-drive/src/drive/web/modules/trash/actions.js

Line 79 in 4eeea76

 ids: files.map(f => f.id), 
. Sometimes, the map doesn't have the right value.

`
https://github.com/cozy/cozy-drive/pull/1611#issuecomment-485674260

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -15,48 +15,119 @@ import {
...
+        )
+      }
     }
+    /**
+     *  Peut-être qu'ici on doit dispatcher le FETCH_SHARINGS histoire d'éviter un didUpdate dans le SharingContainer
+     */
+
+    /* if(isSharingsFilesViewByPath(newProps)){
+      
+    } */
```
Crash--: we can remove previous lines

src/drive/web/modules/navigation/Breadcrumb.jsx
```diff
@@ -159,7 +159,7 @@ class RouterBreadCrumb extends Component {
 
     goToFolder(folderId).then(() => {
       this.toggleOpening()
-      router.push(getFolderUrl(folderId, location))
+      //router.push(getFolderUrl(folderId, location))
```
Crash--: we can remove

src/drive/web/modules/navigation/duck/reducer.js
y-lohse: Yep.

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -89,6 +153,15 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
   fetchMoreFiles: (folderId, skip, limit) =>
     dispatch(fetchMoreFiles(folderId, skip, limit)),
   fetchFolder: folderId => dispatch(openFolder(folderId)),
+  fetchFolderFromTrash: folderId =>
```
y-lohse: @Crash-- Could we use `props.onFolderOpen` instead, as defined [in the TrashContainer](https://github.com/cozy/cozy-drive/blob/fix/QuickView/src/drive/web/modules/trash/Container.jsx#L46) ?

y-lohse: No, because I'm silly and the TrashContainer is further down the tree, so the prop doesn't exist there.

nono:
Ping

PR #1613, opened 2019-04-19
feat: Set statusbar to black when fullscreen viewer

Crash--:
We already had cordova-status-bar installed 

Didn't test on android yet

before : 
IMG_1D3714DA0F16-1

after
IMG_40F282EAA03A-1

### DIFF
--- a/src/viewer/Viewer.jsx
+++ b/src/viewer/Viewer.jsx
@@ -1,3 +1,4 @@
+/* global StatusBar */
 import React, { Component } from 'react'
 import cx from 'classnames'
 
@@ -44,6 +45,10 @@ export default class Viewer extends Component {
 
   componentWillUnmount() {
     document.removeEventListener('keyup', this.onKeyUp, false)
+    if (isMobileApp) {
+      StatusBar.styleDefault()
+      StatusBar.backgroundColorByName('white')
+    }
   }
 
   onKeyUp = e => {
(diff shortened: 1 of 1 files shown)


src/viewer/Viewer.jsx
```diff
@@ -101,6 +106,10 @@ export default class Viewer extends Component {
     const hasNext = currentIndex < fileCount - 1
     // this `expanded` property makes the next/previous controls cover the displayed image
     const expanded = currentFile && currentFile.class === 'image'
+    if (isMobileApp) {
+      StatusBar.styleBlackOpaque()
+      StatusBar.backgroundColorByName('black')
```
y-lohse: That's… very black. Can we use the same grey as the viewer ?

Crash--: In fact, I tried to put the logic to cozy-ui in order to have this behavior enabled by default but it seems that the plugin doesn't work very well with css var :/. 

Also : 
> The Android 5+ guidelines specify using a different color for the statusbar than your main app color (unlike the uniform statusbar color of many iOS apps)

So not sure having the same gray is the way to go 

@joel-costa wanted the same a you: https://gyazo.com/81c4f349c8032384a5fb87a95829a1e0

y-lohse: Mmh ok, if the guidelines forces us to have a different color, so be it. Still not sure about a pure black though but I'll let @joel-costa decide about that. Otherwise the code LGTM.

joel-costa: Ideally, we are supposed to have :
- On iOS : the exact same dark gray
- on Android : a gray darker than the background (I think we had chosen a darker: + 24% with @kosssi )

@Crash-- : in cozy-ui, is it asked for iOS to also have a color darker than the background ?

nono:
Ping @Crash--

Crash--:
I'm waiting for the use of our new Viewer lib to work on this subject

nono:
Ping @Crash--

Crash--:
Done, check the  1.27.1 release

PR #1617, opened 2019-04-23
fix: Import transpiled css on interapp view

y-lohse:
Missed that one.


y-lohse:
[Régressions drive](https://trello.com/c/a0hem9dF/1911-r%C3%A9gressions-drive)

PR #1618, opened 2019-04-23
Photos spinners

y-lohse:
- Sets the color of spinners on the `Loading` component — the prop was never used as far as I can tell
- The spinner on the public album wasn't showing up because the `<IconSprite />` was included only on a specific code path, but we need it all the time


y-lohse:
[Regressions affichage Photos - spinner](https://trello.com/c/blt71RKQ/1902-regressions-affichage-photos-spinner)

PR #1619, opened 2019-04-24
fix: Use cozy-script alias for cozy-ui transpiled

y-lohse:
The new version of cozy-scripts has some bug fixes and it's own  alias for cozy-ui, which seem to work even in production mode.

This is the plan B PR.


y-lohse:
Found a better fix.

PR #1620, opened 2019-04-24
Fix transpiled imports

y-lohse:
See explanations in the code.


y-lohse:
[Modale PDF Banks - styles boutons](https://trello.com/c/hRfZvpSn/1917-modale-pdf-styles-boutons)

Crash--:
Let's merge this one but can we find a "cleaner" fix? In cozy-scripts in UI ?

PR #1623, opened 2019-04-24
Update RatingModal.jsx

Crash--:
- Fix crash because contextType was not there


tpetitberghien:
[[Drive] Page blanche au lieu du contenu de l'app](https://trello.com/c/ItOa4s0N/1916-drive-page-blanche-au-lieu-du-contenu-de-lapp)

PR #1628, opened 2019-04-25
Revert "refactor: Don't display items if menu is not opened"

y-lohse:
This reverts commit da6707c8000e31c7e6c03adf591c493aff2efdaa.

We need to keep the menu in the DOM so the upload event can fire.

### DIFF
--- a/src/components/Menu/index.jsx
+++ b/src/components/Menu/index.jsx
@@ -64,20 +64,20 @@ export default class Menu extends Component {
           </button>
         )}
         {opened && (
-          <>
-            <Overlay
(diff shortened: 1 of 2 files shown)


y-lohse:
[App mobile 1.18.1 - upload impossible (bloquant release)](https://trello.com/c/jRKVZD8v/1922-app-mobile-1181-upload-impossible-bloquant-release)

Crash--:
We should write a comment in the code to explain that stuff. I'm sure in 6 months, I'll do the same mistake again. :D

PR #1631, opened 2019-04-25
Menu display issues

y-lohse:
- The upload button in the main toolbar did not take the whole width
- The "toggle offline" menu item was mis-aligned


y-lohse:
[Regression visuelle - menu action offline décalé](https://trello.com/c/NsaM75rL/1921-regression-visuelle-menu-action-offline-d%C3%A9cal%C3%A9)

PR #1634, opened 2019-04-29
fix: open 2nd photo, not random

CozyKourai:
I need to validate the screenshot for `photos : PhotosCrud 1- Photos Navigation` as it will be the good picture this time


CozyKourai:
[Testcafe : Photo crud open a random photos instead of opening the second photo](https://trello.com/c/uuhooVMD/1925-testcafe-photo-crud-open-a-random-photos-instead-of-opening-the-second-photo)

CozyKourai:
Well, the scenario was "check for viewer for a photo, nor 1st, nor last", and at the time (way before using screenshots, as photos crud is the 1st test I wrote) it seems relevant to use a random photos instead of always using the 2nd one.

PR #1639, opened 2019-04-29
fix: Don't allow user to share to a trashed contact

cedricmessiant:
Takes only contacts that don't have `trashed` attribute or for which `trashed` is `false`


cedricmessiant:
[Drive - Contacts supprimés dans la liste des contacts proposés pour les partages](https://trello.com/c/F2u345xb/1841-drive-contacts-supprim%C3%A9s-dans-la-liste-des-contacts-propos%C3%A9s-pour-les-partages)

PR #1640, opened 2019-04-29
Display fullname in suggestions if possible

cedricmessiant:
Use fullname if we have it. Else, use primary email.

### DIFF
--- a/src/models/Contact.js
+++ b/src/models/Contact.js
@@ -1,0 +1,22 @@
+import { Contact as DoctypeContact } from 'cozy-doctypes'
+
+class Contact extends DoctypeContact {
+  static getInitials(contactOrRecipient) {
+    if (Contact.isContact(contactOrRecipient)) {
+      return DoctypeContact.getInitials(contactOrRecipient)
+    } else {
+      const s = contactOrRecipient.public_name || contactOrRecipient.name
+      return (s && s[0].toUpperCase()) || ''
+    }
+  }
+
+  static getDisplayName(contact) {
+    if (Contact.isContact(contact)) {
+      return DoctypeContact.getDisplayName(contact)
+    } else {
+      return contact.name || ''
+    }
+  }
+}
+
+export default Contact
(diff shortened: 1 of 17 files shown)


src/sharing/components/ContactSuggestion.jsx
```diff
@@ -32,12 +31,17 @@ export const ContactSuggestion = ({ contactOrGroup, contacts, t }) => {
...
-    const email = Contact.getPrimaryEmail(contactOrGroup)
-    name = email || getDisplayName(contactOrGroup)
-    avatarText =
-      Contact.getInitials(email) || Contact.getInitials(contactOrGroup)
+    if (contactOrGroup.fullname) {
+      name = contactOrGroup.fullname
+      avatarText = Contact.getInitials(contactOrGroup)
+    } else {
+      name = Contact.getPrimaryEmail(contactOrGroup)
+      avatarText = name[0].toUpperCase()
```
Crash--: 'Passing a string to Contact.getInitials will be deprecated soon.' :D

cedricmessiant: yeah, that's why I use this :-)

Crash--: Is it possible to have a contactOrGroup without email? Should we test if name[0] exists ?

src/models/Contact.js
```diff
@@ -0,0 +1,26 @@
+import { Contact as OriginalContact } from 'cozy-doctypes'
```
Crash--: DoctypeContact ?

src/models/Contact.js
```diff
@@ -0,0 +1,26 @@
...
+  static isContact(candidate) {
+    return candidate._type === OriginalContact.doctype
+  }
+
+  static getInitials(contactOrRecipient) {
+    if (Contact.isContact(contactOrRecipient)) {
+      return OriginalContact.getInitials(contactOrRecipient)
+    } else {
+      const s = contactOrRecipient.public_name || contactOrRecipient.name
+      return s[0].toUpperCase() || ''
```
Crash--: Can we have `s === undefined` or ` === null` ? Should we test if s[0] exists?

src/models/Contact.js
cedricmessiant: :+1:

cedricmessiant:
Made the changes and added some tests. I wonder if we should move `isContact` to `cozy-doctypes`

Crash--:
i think we should

cedricmessiant:
waiting for https://github.com/cozy/cozy-libs/pull/418 to be able to merge

PR #1641, opened 2019-04-30
fix: Properly display pdf fallback

y-lohse:
We were passing the wrong prop to the fallback component.

### DIFF
--- a/src/viewer/PdfJsViewer.jsx
+++ b/src/viewer/PdfJsViewer.jsx
@@ -98,7 +98,7 @@ export class PdfJsViewer extends Component {
   }
 
   render() {
-    const { url } = this.props
+    const { url, file } = this.props
     const {
       loaded,
       errored,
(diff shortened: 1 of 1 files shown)


y-lohse:
[Viewer PDF Android 9 🍊](https://trello.com/c/oFtLVTqN/1929-viewer-pdf-android-9-%F0%9F%8D%8A)

src/viewer/PdfJsViewer.jsx
```diff
@@ -109,7 +109,7 @@ export class PdfJsViewer extends Component {
       renderAllPages
     } = this.state
 
-    if (errored) return <NoViewer file={url} />
+    if (errored) return <NoViewer file={file} />
```
Crash--: Can you add it to proptypes ? ;) https://github.com/cozy/cozy-drive/blob/fab5d468047ab2ebb5c50892980e58d7d002193e/src/viewer/PdfJsViewer.jsx#L186

PR #1651, opened 2019-05-06
Public viewer fix

y-lohse:
The first commit is self-explanatory i think.

For the second one: when importing from `react-pdf/dist/entry.webpack`, react-pdf loads a webworker that it creates through a webpack configuration. This worker is created at the root of the webpack output, and because this file is already transpiled, we can't change that (or at least I didn't manage to).

So the solution is to load the worker through *our* webpack configuration, so we can change the output directory.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -24,6 +24,19 @@ if (isDrive && target === 'mobile')
 if (target !== 'mobile')
   configurationFiles.push(require('./webpack/appicon.config.js'))
 const extraConfig = {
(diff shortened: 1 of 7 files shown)


y-lohse:
[Viewer PDF vue publique - page blanche Firefox+Safari](https://trello.com/c/QEgqLaHr/1930-viewer-pdf-vue-publique-page-blanche-firefoxsafari)

PR #1653, opened 2019-05-06
Add appMetadata and contacts doctype version to CozyClient

cedricmessiant:
So that `CozyClient` can use those infos to add `cozyMetadata` on contacts it creates.

There are several `CozyClient` instantiations in cozy-drive (mobile, photos, public, ...), we probably need to add those infos everywhere?

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -21,7 +21,8 @@
   },
   "snapshotSerializers": ["enzyme-to-json/serializer"],
   "transform": {
-    "^.+\\.(js|jsx)?$": "<rootDir>/test/jestLib/babel-transformer.js"
+    "^.+\\.(js|jsx)?$": "<rootDir>/test/jestLib/babel-transformer.js",
+    "^.+\\.webapp$": "<rootDir>/test/jestLib/json-transformer.js"
   },
   "transformIgnorePatterns": ["node_modules/(?!cozy-ui)/"],
   "testMatch": ["**/(*.)(spec|test).js?(x)"],
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -7,6 +7,8 @@ import { schema, DOCTYPE_FILES } from 'drive/lib/doctypes'
 export const getLang = () =>
   navigator && navigator.language ? navigator.language.slice(0, 2) : 'en'
 import { isMobileApp, getDeviceName } from 'cozy-device-helper'
+import appMetadata from 'drive/appMetadata'
+
 export const getOauthOptions = () => {
   return {
     redirectURI: isMobileApp() ? 'cozydrive://auth' : 'http://localhost',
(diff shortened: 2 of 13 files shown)


cedricmessiant:
[(𝐌) Partage par groupe 🎓](https://trello.com/c/a3u5bsyP/1803-%F0%9D%90%8C-partage-par-groupe-%F0%9F%8E%93)

Crash--:
In mobile & photo, pretty sure. 

Don't know for public

cedricmessiant:
I added `appMetadata` for all cozy clients we create. This is ready for review

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -7,6 +7,8 @@ import { schema, DOCTYPE_FILES } from 'drive/lib/doctypes'
 export const getLang = () =>
   navigator && navigator.language ? navigator.language.slice(0, 2) : 'en'
 import { isMobileApp, getDeviceName } from 'cozy-device-helper'
+import appMetadata from 'drive/appMetadata'
```
Crash--: Hmm, we already have https://github.com/cozy/cozy-drive/blob/master/app.config.js#L51 

We should use only one, not both. https://github.com/cozy/cozy-drive/blob/6969df9c512053b0374db6c656a9bfbb5af85511/src/drive/mobile/lib/cozy-helper.js#L17

jest.config.json
```diff
@@ -21,7 +21,8 @@
   },
   "snapshotSerializers": ["enzyme-to-json/serializer"],
   "transform": {
-    "^.+\\.(js|jsx)?$": "<rootDir>/test/jestLib/babel-transformer.js"
+    "^.+\\.(js|jsx)?$": "<rootDir>/test/jestLib/babel-transformer.js",
+    "^.+\\.webapp$": "<rootDir>/json-transformer.js"
```
Crash--: Why no put this json-transformer in /test/jestLib/ ?

cedricmessiant: :+1:

src/drive/mobile/lib/cozy-helper.js
cedricmessiant: ok, so the idea is to generate `appMetadata` in `app.config.js`?

cedricmessiant: Ok, I use `appMetadata.version` everywhere.

I'm just not sure what to do about this:

```
  plugins: [
    new DefinePlugin({
      __PIWIK_SITEID__: 8,
      __PIWIK_TRACKER_URL__: JSON.stringify('https://matomo.cozycloud.cc'),
      __APP_VERSION__: JSON.stringify(pkg.version)
    })
  ]
```

Is `__APP_VERSION__` needed by piwik/matomo? I guess I can replace `pkg.version` by `appMetadata.version` here too.

cedricmessiant: well, given that the file is `app.config.js` I wont do this change

PR #1656, opened 2019-05-07
Share with cozy url

cedricmessiant:
We can use the cozy url to share documents to users with a cozy but without email, The stack will find the email to send the notification. It also handles contacts that have a cozy in groups.


cedricmessiant:
[(𝐒) Partage via url Cozy 🎓](https://trello.com/c/woo5qwaR/1884-%F0%9D%90%92-partage-via-url-cozy-%F0%9F%8E%93)

PR #1663, opened 2019-05-09
Testcafe : add timestamp in log

CozyKourai:
Using log-prefix to add timestamp to testcafé log. 

I choose to use `toTimeString` to display the date-time in format ` [14:23:45 GMT+0200 (CEST)]`
This can be [change to any date format ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Conversion_getter)
Maybe `toISOString` is better because more precise (`[2019-05-09T08:58:50.487Z]`), but I think it's less readable

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,7 @@
     "url-polyfill": "1.1.0",
     "webpack-node-externals": "1.7.2",
     "whatwg-fetch": "3.0.0",
-    "minilog": "3.1.0"
+    "minilog": "3.1.0",
+    "log-prefix": "0.1.1"
   }
 }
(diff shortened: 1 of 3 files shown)


CozyKourai:
[Testcafe: betterLogs](https://trello.com/c/AiHyF7KZ/1963-testcafe-betterlogs)

y-lohse:
Why do we want this?

CozyKourai:
@y-lohse To have more info on what's taking a long time in test, Or, when it's failing does the previous step took longer than before, etc..
This was actually suggested by our friends from 'infra'

cedricmessiant:
I prefer `toISOString` to be able to see the date and not only the time

y-lohse:
Luckily our tests are not running over several days so I'm not sure the full date is useful :p

PR #1665, opened 2019-05-09
refactor: Removed unused code

y-lohse:
We can do without this stylefile now.


Crash--:
It's not what I call "unused code". It's a refactoring to use utilities :D

y-lohse:
Well the `require` statement was the unused part. The css rule I just moved :)

PR #1666, opened 2019-05-09
fix: Use email if public_name and name are not available

cedricmessiant:
When Bob shares a document to Alice and Alice look at this document's sharings in her cozy's, Alice's cozy may only have the recipient's emails to display the avatar of the recipients. This PR fixes this use case.


cedricmessiant:
[(𝐒) Partage via url Cozy 🎓](https://trello.com/c/woo5qwaR/1884-%F0%9D%90%92-partage-via-url-cozy-%F0%9F%8E%93)

PR #1667, opened 2019-05-10
fix:change exif to 6 april

CozyKourai:
Those 2 images are sometimes in a different order.
image

This is not because of the upload as they are uploaded in differents test already
The problem was that those 2 photos have exactly the same exif value on date and time.
So I change the date for IMG0 to 6 april 2019 (while the other JPG image is 5 april 2019)


CozyKourai:
[Fix: Photos order](https://trello.com/c/h1ZpCEwk/1968-fix-photos-order)

PR #1669, opened 2019-05-10
feat: Use the cozy-ui viewer

y-lohse:
Todo:

- Integrate the drive-specific CallToAction for the NoViewer
- Remove old viewer files
- PDF files have trouble loading at the moment ?
- Handling the pdfjs worker on public pages (basically https://github.com/cozy/cozy-drive/pull/1651 but as an optional feature for the viewer in ui)

Edit: all done.


y-lohse:
[Intégration du viewer cozy-ui dans drive/photos](https://trello.com/c/hDMhH2WW/1967-int%C3%A9gration-du-viewer-cozy-ui-dans-drive-photos)

app.config.js
```diff
@@ -28,6 +28,7 @@ const extraConfig = {
     rules: [
       {
         test: /\.worker(\.entry)\.js$/,
+        issuer: { not: [/node_modules\//] },
```
y-lohse: We want to use our custom worker loader, unless the worker is loaded by a dependency, in which case we leave it alone.

Crash--: Can you add this comment in the file? Since it a JS file, let's enjoy being able to write comments :D

src/drive/web/modules/public/LightFileViewer.jsx
```diff
@@ -1,14 +1,19 @@
...
+import { Viewer } from 'cozy-ui/transpiled/react'
 import PropTypes from 'prop-types'
 
 import PublicToolbar from './PublicToolbar'
 import styles from 'drive/web/modules/viewer/barviewer.styl'
 
+import createWorker from 'react-pdf/dist/pdf.worker.entry.js'
+import { pdfjs } from 'react-pdf'
+
+pdfjs.GlobalWorkerOptions.workerPort = createWorker()
```
Crash--: Do you think we need to write a not about that in the UI doc? 

Do you think is there any value to mutualize that? Since we used it also in LightFolder? should we create a `<PublicViewer>` wraping `<UiViewer>` ?

src/drive/web/modules/viewer/FileOpenerExternal.jsx
```diff
@@ -75,8 +76,11 @@ export class FileOpener extends Component {
             <Viewer
               files={[file]}
               currentIndex={0}
-              onChange={doNothing}
-              onClose={withCloseButtton ? this.navigateToDrive : null}
+              onChangeRequest={doNothing}
+              onCloseRequest={withCloseButtton ? this.navigateToDrive : null}
+              renderFallbackExtraContent={file => (
+                <Fallback file={file} t={t} />
```
Crash--: Can we add a comment about why we need to pass `t`? It's pretty fresh now, in a few months I doubt

src/drive/web/modules/viewer/styles.styl
```diff
@@ -0,0 +1,57 @@
+@require 'settings/breakpoints.styl'
+@require 'settings/palette.styl'
+
+.pho-viewer-noviewer-cta
+    position         relative
+    border-radius    .5rem
+    background-color rgba(255, 255, 255, 0.05)
```
Crash--: Do we need a css var for that?

src/drive/web/modules/viewer/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+
+    h3:before
+        content      ''
+        position     relative
+        top          .2rem
+        width        1.5rem
+        height       1.5rem
+        margin-right .75rem
+        display      inline-block
+        background   embedurl('./icons/icon-magic-trick.svg') center center / cover no-repeat
```
Crash--: Oh no. We still have those :'(

package.json
```diff
@@ -134,7 +136,6 @@
     "react-dropzone": "5.1.1",
     "react-markdown": "3.4.1",
```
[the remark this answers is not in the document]
y-lohse: It's still used in the mobile auth part I believe, so i figured we can remove it when we've switched to the npm module of cozy-auth ?

src/drive/web/modules/public/LightFileViewer.jsx
y-lohse: > Do you think we need to write a not about that in the UI doc?

Probably uncommon but it can't hurt.

>Do you think is there any value to mutualize that? Since we used it also in LightFolder? should we create a <PublicViewer> wraping <UiViewer> ?

Yup.

src/drive/web/modules/viewer/styles.styl
y-lohse: I haven't actually looked at that in this PR, but mayyybe? If the color scheme is inverted then this would'nt work too well.

cc @GoOz what do you think?

GoOz: Well we don't have variables for colors with alpha transparency except the one for the layer of modals.
There's no easy way to do this with CSS custom properties sadly and I'm not sure adding in UI new color variables for every transparency need is a good solution. But I don't really have a good solution either… :|

src/drive/web/modules/viewer/styles.styl
GoOz: Hey it IS in an `embedurl`. Can you do it with an `Icon` component instead? because in my PR about cleaning icons I removed the color from the SVG.

y-lohse: Yes it can definitely be done with an Icon component. But that's for another PR.

Crash--: Can you create a card to not forget it?

src/drive/web/modules/viewer/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+        &:before
+            content       ''
+            position      relative
+            top           .3rem
+            border-radius 1rem
+            width         1.5rem
+            height        1.5rem
+            margin-right  .75rem
+            display       inline-block
+            background    dodgerBlue embedurl('./icons/icon-check.svg') center center no-repeat
```
Crash--: this one too https://github.com/cozy/cozy-drive/pull/1669/files#r285450699

src/drive/web/modules/viewer/FilesViewer.jsx
```diff
@@ -96,39 +97,34 @@ class FilesViewer extends Component {
...
-        )
-      }
+    if (currentIndex === -1 && !this.state.currentFile) {
+      return (
+        <Overlay>
+          <Spinner size="xxlarge" middle noMargin color="white" />
+        </Overlay>
+      )
+    } else {
+      const hasCurrentIndex = currentIndex != -1
```
cedricmessiant: `!==`

src/drive/web/modules/viewer/NoViewerButton.jsx
```diff
@@ -0,0 +1,75 @@
...
+DownloadButton.contextTypes = {
+  client: PropTypes.object.isRequired
+}
+
+const NoViewerButton = ({ file, t }) => {
+  if (isMobileApp()) return <OpenWithCordovaButton t={t} file={file} />
+  else return <DownloadButton t={t} file={file} />
+}
+
+export default NoViewerButton
```
cedricmessiant: one component = one file?

y-lohse:
Merging despite the problematic test, apparently not related to these changes..?

PR #1670, opened 2019-05-10
Fix getDisplayName and getInitials

cedricmessiant:
- if sharing has been done by url, we may only have the member's `public_name`
- also add a default values for initials and display name if name, public_name or email are not available (for sharing members for whom sharing has been done by url and that have not yet accepted the sharing)

### DIFF
--- a/src/sharing/components/Recipient.jsx
+++ b/src/sharing/components/Recipient.jsx
@@ -49,14 +49,16 @@ export const RecipientsAvatars = ({
           size={size}
         />
       )}
-      {reversedRecipients.slice(0, MAX_DISPLAYED_RECIPIENTS).map(recipient => (
-        <Avatar
-          key={`key_avatar_${recipient.email}`}
-          text={Contact.getInitials(recipient)}
-          size={size}
-          textId={Contact.getDisplayName(recipient)}
-        />
-      ))}
+      {reversedRecipients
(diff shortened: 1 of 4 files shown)


cedricmessiant:
We should probably have some sort of `getRecipientDisplayName`  and use it in `getInitials` and `getDisplayName` but I was not able to find a satisfying solution. Either we have to pass the translation function to the model or to do tricky stuff that make the code harder to understand

src/sharing/components/Recipient.jsx
```diff
@@ -166,7 +166,7 @@ const Recipient = (props, { client, t }) => {
   const { instance, isOwner, status, ...rest } = props
   const isMe =
     (isOwner && status === 'owner') || instance === client.options.uri
-  const name = Contact.getDisplayName(rest)
+  const name = t(Contact.getDisplayName(rest))
```
Crash--: really not a big fan of that.

cedricmessiant: We could either : 
1- pass `t()` to `getDisplayName`
2- pass the default value (already translated) to `getDisplayName` when we need it (and use `''` as default)

2 seems ok but I would like to avoid to change every call to `Contact.getDisplayName` because this case where we have no name/email/public_name seems to append only in certains cases for recipients in "Shared with me" folder

PR #1675, opened 2019-05-13
refacto: viewer private & public scenario

CozyKourai:
Change private viewer scenario in order to check one file for each viewer type (instead of checking all of the 25 files uploaded) in order to reduce CI time.
Now time to build drive and run tests is around ~21min while before this PR and #1679 it need around ~30min

### DIFF
--- a/testcafe/tests/drive/viewer-feature.js
+++ b/testcafe/tests/drive/viewer-feature.js
@@ -3,207 +3,149 @@ import logger from '../helpers/logger'
 import {
   TESTCAFE_DRIVE_URL,
   setDownloadPath,
-  getFilesWithExt,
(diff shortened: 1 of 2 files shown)


CozyKourai:
[SCN - viewer Drive - révision](https://trello.com/c/oN4Q8mqD/1764-scn-viewer-drive-r%C3%A9vision)

testcafe/tests/drive/viewer-feature.js
```diff
@@ -1,209 +1,152 @@
 import { driveUser } from '../helpers/roles'
 import logger from '../helpers/logger'
+//import { Role } from 'testcafe'
```
y-lohse: If it's not used…

PR #1662, opened 2019-05-09
chore: Use minilog

Crash--:
I tried to put minilog as a redux-logger but some functionalities where missing

PR #1676, opened 2019-05-13
Upload conflict

cedricmessiant:
Handle conflicts when we upload a file via :
- Import files button in toolbar
- Import files action in action menu
- dropzone

Uses `handle-conflicts` flag to activate the feature.

Note that we need `sharingState` to be able to know if a file is shared (if a file is shared, user can't update it). To do this, we use a `withSharingState` HOC and pass the `sharingState` directly to the `uploadFiles` action creator. We could use the same `mapDispatchToProps` in `UploadButton` and in `Dropzone` or even the whole `compose(...)` function to avoid duplication, WDYT?

### DIFF
--- a/src/drive/locales/en.json
+++ b/src/drive/locales/en.json
@@ -444,6 +444,8 @@
     "alert": {
       "success": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
       "success_conflicts": "%{smart_count} file uploaded with %{conflictNumber} conflict(s). |||| %{smart_count} files uploaded with %{conflictNumber} conflict(s).",
+      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} file(s) updated.",
+      "updated": "%{smart_count} file updated. |||| %{smart_count} files updated.",
       "errors": "Errors occurred during the file upload.",
       "network": "You are currenly offline. Please try again once you're connected."
     }
--- a/src/drive/locales/en.json
+++ b/src/drive/locales/en.json
@@ -444,7 +444,7 @@
     "alert": {
       "success": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
       "success_conflicts": "%{smart_count} file uploaded with %{conflictNumber} conflict(s). |||| %{smart_count} files uploaded with %{conflictNumber} conflict(s).",
-      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} file(s) updated.",
+      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} updated.",
       "updated": "%{smart_count} file updated. |||| %{smart_count} files updated.",
       "errors": "Errors occurred during the file upload.",
       "network": "You are currenly offline. Please try again once you're connected."
--- a/src/drive/locales/en.json
+++ b/src/drive/locales/en.json
@@ -444,8 +444,10 @@
     "alert": {
       "success": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
       "success_conflicts": "%{smart_count} file uploaded with %{conflictNumber} conflict(s). |||| %{smart_count} files uploaded with %{conflictNumber} conflict(s).",
-      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} updated.",
+      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated. |||| %{smart_count} files uploaded and %{updatedCount} updated.",
+      "success_updated_conflicts": "%{smart_count} file uploaded, %{updatedCount} updated and %{conflictCount} conflict(s). |||| %{smart_count} files uploaded, %{updatedCount} updated and %{conflictCount} conflict(s).",
       "updated": "%{smart_count} file updated. |||| %{smart_count} files updated.",
+      "updated_conflicts": "%{smart_count} file updated with %{conflictCount} conflict(s). |||| %{smart_count} files updated with %{conflictCount} conflict(s).",
       "errors": "Errors occurred during the file upload.",
       "network": "You are currenly offline. Please try again once you're connected."
     }
--- a/src/drive/web/modules/navigation/duck/actions.jsx
+++ b/src/drive/web/modules/navigation/duck/actions.jsx
@@ -255,37 +255,51 @@ export const uploadFiles = (files, dirId, sharingState) => dispatch => {
 }
 
 const uploadQueueProcessed = (
-  loaded,
+  created,
   quotas,
   conflicts,
   networkErrors,
   errors,
   updated
 ) => dispatch => {
+  const conflictCount = conflicts.length
+  const createdCount = created.length
+  const updatedCount = updated.length
   if (quotas.length > 0) {
     // quota errors have their own modal instead of a notification
     dispatch(showModal(<QuotaAlert />))
-  } else if (conflicts.length > 0) {
-    Alerter.info('upload.alert.success_conflicts', {
-      smart_count: loaded.length,
-      conflictNumber: conflicts.length
-    })
   } else if (networkErrors.length > 0) {
     Alerter.info('upload.alert.network')
   } else if (errors.length > 0) {
     Alerter.info('upload.alert.errors')
-  } else if (updated.length > 0 && loaded.length === 0) {
-    Alerter.success('upload.alert.updated', {
-      smart_count: updated.length
+  } else if (updatedCount > 0 && createdCount > 0 && conflictCount > 0) {
+    Alerter.success('upload.alert.success_updated_conflicts', {
+      smart_count: createdCount,
+      updatedCount,
+      conflictCount
     })
-  } else if (updated.length > 0 && loaded.length > 0) {
+  } else if (updatedCount > 0 && createdCount > 0) {
     Alerter.success('upload.alert.success_updated', {
-      smart_count: loaded.length,
-      updatedCount: updated.length
+      smart_count: createdCount,
+      updatedCount
+    })
+  } else if (updatedCount > 0 && conflictCount > 0) {
+    Alerter.success('upload.alert.updated_conflicts', {
+      smart_count: updatedCount,
+      conflictCount
+    })
+  } else if (conflictCount > 0) {
+    Alerter.info('upload.alert.success_conflicts', {
+      smart_count: createdCount,
+      conflictNumber: conflictCount
+    })
+  } else if (updatedCount > 0 && createdCount === 0) {
+    Alerter.success('upload.alert.updated', {
+      smart_count: updatedCount
     })
   } else {
     Alerter.success('upload.alert.success', {
-      smart_count: loaded.length
+      smart_count: createdCount
     })
   }
 }
--- a/src/drive/web/modules/navigation/duck/actions.jsx
+++ b/src/drive/web/modules/navigation/duck/actions.jsx
@@ -225,11 +225,12 @@ export const getFileDownloadUrl = async id => {
   return `${cozy.client._url}${link}`
 }
 
-export const uploadFiles = (files, dirId) => dispatch => {
+export const uploadFiles = (files, dirId, sharingState) => dispatch => {
   dispatch(
     addToUploadQueue(
       files,
       dirId,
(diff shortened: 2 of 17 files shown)


cedricmessiant:
[(𝐋) Upload conflit - étape 1 - écraser fichier non partagé](https://trello.com/c/0cTBor9A/1761-%F0%9D%90%8B-upload-conflit-%C3%A9tape-1-%C3%A9craser-fichier-non-partag%C3%A9)

src/drive/web/modules/upload/index.js
```diff
@@ -36,8 +38,11 @@ const status = action => {
   }
 }
 
-const item = (state, action) =>
-  Object.assign({}, state, { status: status(action) })
+const item = (state, action = { isUpdate: false }) =>
+  Object.assign({}, state, {
```
y-lohse: Nit, but we don't use `Object.assign` anymore, we can use object destructuring instead.

src/drive/web/modules/upload/index.js
```diff
@@ -85,20 +91,33 @@ export const processNextFile = (
...
+      error = uploadError
     }
 
-    const status =
-      statusError[error.status] ||
-      (/Failed to fetch$/.exec(error.toString()) && NETWORK) ||
-      FAILED
+    if (error) {
+      // eslint-disable-next-line no-console
+      console.warn(error)
```
y-lohse: Nit: I think we started using `minilog`.

cedricmessiant: yep, but it is not merged : https://github.com/cozy/cozy-drive/pull/1662

src/drive/locales/en.json
```diff
@@ -444,6 +444,8 @@
     "alert": {
       "success": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
       "success_conflicts": "%{smart_count} file uploaded with %{conflictNumber} conflict(s). |||| %{smart_count} files uploaded with %{conflictNumber} conflict(s).",
+      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} file(s) updated.",
```
y-lohse:
```suggestion
      "success_updated": "%{smart_count} file uploaded and %{updatedCount} updated(s). |||| %{smart_count} files uploaded and %{updatedCount} updated.",
```

"5 files uploaded and 2 updated"… just a suggestion to avoid the file(s).

src/drive/web/modules/upload/index.js
```diff
@@ -138,30 +173,55 @@ const uploadFile = async (client, file, dirID) => {
...
+  const parentDirectory = resp.data
+  return `${parentDirectory.path}/${file.name}`
+}
+
+export const overwriteFile = async (client, file, path, dirID) => {
+  const statResp = await client.collection('io.cozy.files').statByPath(path)
+  const fileId = statResp.data.id
+  const resp = await client
+    .collection('io.cozy.files')
+    .updateFile(file, { dirId: dirID, fileId })
```
y-lohse: I think we could use the `dirId` from `statResp `, no?

src/drive/web/modules/upload/index.js
```diff
@@ -171,12 +231,26 @@ const getErrors = queue => filterByStatus(queue, FAILED)
...
 export const getProcessed = state =>
   getUploadQueue(state).filter(f => f.status !== PENDING)
 export const getSuccessful = state => getLoaded(getUploadQueue(state))
 
+export const selectors = {
+  getConflicts,
+  getErrors,
+  getQuotaErrors,
+  getNetworkErrors,
+  getLoaded,
```
y-lohse: Do we still want to export `getLoaded`?

src/drive/web/modules/upload/index.js
cedricmessiant: :+1:

src/drive/web/modules/upload/index.js
cedricmessiant: Good point, it probably better to keep it in here!

src/drive/web/modules/upload/UploadButton.jsx
```diff
@@ -43,4 +48,16 @@ const UploadButton = ({ label, disabled, onUpload, className }) => (
...
+  }
+})
+
+export default compose(
+  withSharingState,
+  connect(
+    null,
+    mapDispatchToProps
+  )
+)(UploadButton)
```
Crash--: Can we add proptypes for this `UploadButton` component ? Since we're using its `ownProps` I think this can be very useful ;)

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -225,15 +225,23 @@ export const getFileDownloadUrl = async id => {
   return `${cozy.client._url}${link}`
 }
 
-export const uploadFiles = (files, dirId) => dispatch => {
+export const uploadFiles = (files, dirId, sharingState) => dispatch => {
```
Crash--: Can we add a jsdoc here to describe the arguments?

src/drive/web/modules/upload/index.js
```diff
@@ -36,8 +41,11 @@ const status = action => {
   }
 }
 
-const item = (state, action) =>
-  Object.assign({}, state, { status: status(action) })
+const item = (state, action = { isUpdate: false }) => ({
+  ...state,
+  isUpdate: action.isUpdate,
```
Crash--: isUpdate = the action dispatched when we need to update the doc? Or is it a boolean to know if the item has been uploaded after a conflict ? 

Can we have a comment ?

src/drive/web/modules/upload/index.js
```diff
@@ -85,22 +95,50 @@ const processNextFile = (
...
+        if (
+          flag('handle-conflicts') &&
+          !isShared(sharingState, { path }) &&
+          !hasSharedParent(sharingState, { path })
+        ) {
+          const uploadedFile = await overwriteFile(client, file, path)
+          fileUploadedCallback(uploadedFile)
+          dispatch({ type: RECEIVE_UPLOAD_SUCCESS, file, isUpdate: true })
+          error = null
+        }
```
Crash--: Not blocking but I would prefer to have a kind of `retry` system instead of having to deal with the "same thing" in an `catch` block.

src/drive/web/modules/upload/index.js
```diff
@@ -138,30 +176,55 @@ const uploadFile = async (client, file, dirID) => {
   return resp.data
 }
 
+export const getFilePath = async (client, file, dirID) => {
```
Crash--: Should we rename to something more accurate? Does `parentDirectory.path` return the fullpath / absolute path of the folder?

src/drive/web/modules/upload/index.js
```diff
@@ -138,30 +176,55 @@ const uploadFile = async (client, file, dirID) => {
...
+  const resp = await client.collection('io.cozy.files').get(dirID)
+  const parentDirectory = resp.data
+  return `${parentDirectory.path}/${file.name}`
+}
+
+export const overwriteFile = async (client, file, path) => {
+  const statResp = await client.collection('io.cozy.files').statByPath(path)
```
Crash--: Is there any reason to use statByPAth() instead of a simple `get`? Since `get` returns fullpath it can be easier to deal with?

src/drive/web/modules/upload/index.js
```diff
@@ -138,30 +176,55 @@ const uploadFile = async (client, file, dirID) => {
...
+    )
+  )
 }
 
 export const purgeUploadQueue = () => ({ type: PURGE_UPLOAD_QUEUE })
 
 export const onQueueEmpty = callback => (dispatch, getState) => {
   const queue = getUploadQueue(getState())
-  const loaded = getLoaded(queue)
+  const uploaded = getUploaded(queue)
```
Crash--: why are we removing `loaded`?

src/drive/web/modules/upload/index.js
```diff
@@ -171,12 +234,25 @@ const getErrors = queue => filterByStatus(queue, FAILED)
 const getQuotaErrors = queue => filterByStatus(queue, QUOTA)
 const getNetworkErrors = queue => filterByStatus(queue, NETWORK)
 const getLoaded = queue => filterByStatus(queue, LOADED)
```
Crash--: If we remove `loaded` we should remove the getter too?

cedricmessiant: the `getLoaded` is used in `getUploaded` and `getUpdated`

src/drive/web/modules/upload/index.js
cedricmessiant: because we have to make a distinction between uploaded files (new) and updated files

src/drive/web/modules/upload/index.js
cedricmessiant: I don't think I can use `get` to retrieve file id and dir id from the file's path

src/drive/web/modules/upload/index.js
cedricmessiant: :+1:

cedricmessiant: It's a boolean to know if it is an update after a conflict. At first, I called it `isConflict` but it's not good. Maybe `isOverwrite` but I still feel that `isUpdate` is better since we use "update" in the messages we display to the user. We could use `action.status` for this and modify the `status()` function above to return a different status if the file has been updated. Something like : 

```
    case RECEIVE_UPLOAD_SUCCESS:
      return action.status === 'update' ? UPDATED : LOADED
```

Thus, we could return `getLoaded` and use `filterByStatus()` in `getUploaded` / `getUpdated` selectors. I will need to change `getSuccessful` too if we do this.

src/drive/web/modules/upload/index.js
Crash--: I think jsdoc will be useful here. I expected that `file` was a file object.

src/drive/web/modules/upload/index.js
Crash--: If we can do with the current attributes, I would prefer this (suggested) approach

src/drive/web/modules/upload/index.js
Crash--: So maybe `created` and `updated` are better? Since even if we `update` we `upload`. No?

src/drive/web/modules/upload/index.js
cedricmessiant: I'm not sure I understand what it would look like

src/drive/web/modules/upload/index.js
cedricmessiant: Yes, `parentDirectory.path == '/Administratif/Whatever`
Something more accurate like `getFileFullpath` ?

src/drive/web/modules/upload/index.js
cedricmessiant: :+1:

src/drive/web/modules/upload/index.js
Crash--: Something calling itself a few times with new parameters? (In this case juste two times). But forget my comment ;)

src/drive/web/modules/upload/index.js
Crash--: yep

Crash--: yep

cedricmessiant:
Done on my side. Still need an approval. 
I could use `minilog` but I would need to copy `lib/logger.js` file from the minilog PR. In any case, we will have a conflict on `upload/index.js` when we rebase the minilog branch so we'll fix it at this moment.

cedricmessiant:
I had to push 2 new commits to fix toast and upload queue in some cases

src/drive/web/modules/upload/UploadQueue.jsx
```diff
@@ -29,45 +30,40 @@ const Pending = translate()(props => (
...
+    statusIcon = <Spinner class="u-ml-half" color={palette['dodgerBlue']} />
+  } else if (status === CANCEL) {
+    statusIcon = (
+      <Icon class="u-ml-half" icon="cross" color={palette['monza']} />
+    )
+  } else if (ERROR_STATUSES.includes(status)) {
+    error = true
+    statusIcon = (
+      <Icon class="u-ml-half" icon="warning" color={palette['monza']} />
+    )
```
Crash--: We'll have a warning icon for a few ms before we retry? Or do we skip it and put a new loading one? (just curious)

cedricmessiant: Good point. I only see a loading icon even if I slow down the network. And in fact it makes sense since the status in redux never changes during the conflict management process.

Crash--: Thank for the clarification :)

PR #1677, opened 2019-05-14
feat: Handle user selected media buckets

lmeunier:
This is a WIP and a strong review is needed.
 To test this pull-request, you need an other pull-request on the cordova-plugin-list-library-items
 repo: https://github.com/cozy/cordova-plugin-list-library-items/pull/23

Questions:
- what should we do if a backup is currently running and the user
changes the media bucket lists?
- should a backup be started when the user select a new media bucket?

TODO:
- better UI/navigation
- i18n
- add relevant loggers
- tests
- iOS


src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
```diff
@@ -0,0 +1,108 @@
...
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      mediaBuckets: []
+    }
+  }
+
+  createHandle(item) {
+    const self = this
```
Crash--: Is there a good reason to do that?  Can't we use this.props.add/delMediaBucket() directly?

tpetitberghien:
Hi @lmeunier 

Thanks for this! I will let our tech guys check your pull request.

Regarding your questions product oriented:

`- what should we do if a backup is currently running and the user changes the media bucket lists?
`
Backup is not synchronisation, so we dont remove any file for the moment. Fine for me to keep files already backed up into the Cozy. However, if user change media bucket, we should stop at the end of the current file backed up and restart from the new media bucket.

`- should a backup be started when the user select a new media bucket?`
I think so, as we do have in "Settings" view of the app, user can start manually a backup. As soon as he select a media bucket (and already have permission to access to files of the device and settings of the app enable backup), backup of the new media bucket should start.

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
```diff
@@ -0,0 +1,108 @@
...
+import { getMediaBuckets as getAllMediaBuckets } from 'drive/mobile/lib/media'
+
+const { BarCenter } = cozy.bar
+
+class SelectMediaBuckets extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      mediaBuckets: []
```
Crash--: since we use add() and delete(), we should declare our initial state as a Set()

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
```diff
@@ -0,0 +1,108 @@
...
+      })
+    }
+  }
+
+  componentDidMount() {
+    var self = this
+    getAllMediaBuckets().then(data => {
+      self.setState({
+        mediaBuckets: data
+      })
```
Crash--: no need to self = this ;) 

You can also write : 
```jsx 
async componentDidMount() {
 const data = await getAllMediaBuckets()
 this.setState({mediaBuckets: data}}
}```

File: src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx [hunk 4]
```diff
@@ -0,0 +1,108 @@
...
+    })
+  }
+
+  render() {
+    return (
+      <Main>
+        <ScrollToTop />
+        <BarCenter>
+          <h2 className={styles['settings__title']}>Select Media Buckets</h2>
```
Crash--: You started to use string here : https://github.com/cozy/cozy-drive/pull/1677/files#diff-496c2803b6373b077bd43538e90ef51cR362 

To be able to use it, you have to : 
```jsx 
Component {
 const { t } = this.props 
}
export translate()(Component)
```

The translate HOC will inject a `t` props. And after that you can use : 
t('MyString')

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
```diff
@@ -0,0 +1,108 @@
...
+        </BarCenter>
+        {this.state.mediaBuckets.map(item => (
```
Crash--: Can we map on a `Set` ? It seems not https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
```diff
@@ -0,0 +1,108 @@
...
+              {item}
+            </label>
+            <label
+              htmlFor={item}
+              className={styles['settings__subcategory__item']}
+            >
+              <Toggle
+                id={item}
+                checked={this.props.selectedBuckets.has(item)}
+                onToggle={this.createHandle(item)}
```
Crash--: OK now I understand. You can `bind` this.createHandle to `this` in order to have the right context when you call it. 

```jsx 
constructor(){
this.createHandle = this.createHandle.bind(this, item) 
}
```

Or : 
```jsx 
onToggle={ () => this.createHandle(item) }
```
since arrow function auto bind to this

lmeunier:
> I think we can only target Android ATM since we don't have any issue in iOS. ;)

Shouldn't we hide the "Media buckets" panel on iOS?

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: We can't use this.props.add/delMediaBucket() directly because the Toggle component use only one event with the state the component (boolean `checked`).

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: Fixed

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: Indeed, self=this is not needed.
Thanks for the rewrite, I'm not yet comfortable with async/await. :)

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: Fixed

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: You are right.
The `Set` is now converted to an array via the ES6 spread operator.

src/drive/mobile/modules/settings/components/SelectMediaBuckets.jsx
lmeunier: I used an arrow function with a `checked` argument :
```
onToggle={checked => this.createHandle(item, checked)}
```

Crash--:
> Shouldn't we hide the "Media buckets" panel on iOS?

Yep. Let's hide on iOS. You have a `isIOSApp()` somewhere ;)

src/drive/mobile/lib/media.js
```diff
@@ -141,3 +142,24 @@ export const getPhotos = async () => {
 
   return Promise.resolve(defaultReturn)
 }
+
+export const getMediaBuckets = async () => {
```
Crash--: We should only call this method on Android since `listMediaBuckets` is not available on iOS

src/drive/mobile/modules/settings/duck/reducers.js
```diff
@@ -4,18 +4,22 @@ import {
...
   offline: false,
   analytics: false,
   backupImages: false,
-  wifiOnly: true
+  wifiOnly: true,
+  mediaBuckets: new Set(['Camera'])
 }
 
 const settings = (state = initialState, action) => {
+  state = { ...initialState, ...state }
```
Crash--: @y-lohse Do you have an idea of why @lmeunier has to do that to have access to its `mediaBuckets`? Do we persist `settings` store?

y-lohse: Yes we do that [here](https://github.com/cozy/cozy-drive/blob/master/src/drive/store/configureStore.js#L39).

We also [have this](https://github.com/cozy/cozy-drive/blob/master/src/drive/store/persistedState.js#L30) still hanging around, but it shouldn't be triggered anymore.

I'm not sure it's useful to keep this line though, the reducer will simply add the `mediaBuckets` entry when it's needed?

Crash--: So to persist this new Settings, we have to add it here : https://github.com/cozy/cozy-drive/blob/master/src/drive/store/configureStore.js#L39 

With this modification, even with a kill / reboot of the app we will still have the buckets ;)

y-lohse: I don't think that's needed? [This line](https://github.com/cozy/cozy-drive/blob/master/src/drive/store/configureStore.js#L42) already saves the complete settings. So once the buckets get in the state, they should be automatically saved. But they won't be there on the first boot.

I also wonder if using a `Set` will be a problem, not sure it will be properly deserialized.

lmeunier: This line is needed when `state.mobile.settings` exists but does not have the `mediaBuckets` entry. In this case, [getSetting(state, "mediaBucket")](https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/modules/settings/duck/reducers.js#L39) always returns `undefined` (instead of the default `Set(['Camera'])`) and an exception is thrown on the "Select Media Buckets" page.
This is typically the case when a user is already using the mobile app (`state.mobile.settings` exists) and the app is updated with this feature. Without this line, `state.mobile.settings.mediaBuckets` would still be undefined.

src/drive/mobile/lib/media.js
lmeunier: Fixed

https://github.com/cozy/cozy-drive/compare/5da40f02a715ac92fd0b811a52b5f9e8536dc879..8bbeeace9e475c88b9c93ef8ab34263c9c572e9f

src/drive/mobile/modules/settings/duck/reducers.js
Crash--: So to do the things right, we should migrate the persisted state to add this default value, right? 

This default value should only target `Android`. We can create a `addDefaultMediaBucket` migration (https://github.com/cozy/cozy-drive/blob/master/src/drive/store/persistedState.js#L59) to add this default value. 

WDYT?

y-lohse: Would work, or we could do something like:

```
export const getMediaBuckets = state => getSetting(state, 'mediaBucket') || new Set(['Camera']))
```

… over here : https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/modules/settings/duck/reducers.js#L54

I don't really mind which one :)

lmeunier: If I'm not wrong, we have 3 choices:
1. keep the line `state = { ...initialState, ...state }`
2. create a `addDefaultMediaBucket` migration to add the default value
3. add something like `export const getMediaBuckets = state => getSetting(state, 'mediaBucket') || new Set(['Camera']))`

So, which one should be implemented?

y-lohse: I'm not comfortable with option 1 because we're not supposed to mutate the current `state`. I think 2 is a bit more complicated than 3 and it seems ike we'll forget removing the migration afterwards… so I would go for option 3. @Crash-- you've looked at this PR more than me, WDYT?

Crash--: 3 sounds good ;)

Crash--:
Will try to test & build this PR tomorrow ;)

src/drive/mobile/modules/settings/duck/reducers.js
lmeunier: Fixed in the last force-push:
https://github.com/cozy/cozy-drive/compare/1e292de87acc3b74bee2e22a6334dceea6ab257a..ecb6127e4b63984bc9b569e11c8fc0834167a7dd

lmeunier:
> Will try to test & build this PR tomorrow ;)

I've just rebased my branch from your master. This PR should be up to date.

Crash--:
1 month later... 

Sorry for the delay! I just tried your PR yesterday and it works great on an Xperia device!

lmeunier:
> Sorry for the delay! I just tried your PR yesterday and it works great on an Xperia device!

No worries :)

> Is it possible to switch on several buckets? It seems during my test that if I activate one, others are deactivated.

It should be possible to switch on several buckets. If not, it's a bug.
And indeed, it was a bug. Thank you for noticing it :)
Fixed in the last [force-push](https://github.com/cozy/cozy-drive/compare/61c685cd186b0b3f479d566285be7ef04c1cf727..8c6f86abdb1af58a778dec9758393875acbd309f#diff-0cd5a682fa43aad066eea3d19b54370f).

> Can you add a little padding to the global view ? Something like `<div className="u-p-1"></div>` should already be better ;)

Fixed in the last [force-push](https://github.com/cozy/cozy-drive/compare/61c685cd186b0b3f479d566285be7ef04c1cf727..8c6f86abdb1af58a778dec9758393875acbd309f#diff-5893c9bb7cb9d90ce01565f8a9a8c627).

> It appears that I had a lot of buckets. Specially if we have a few music folders on the device. Can we do something about that?

I think you have album covers in your music folders, and these album covers are listed as images.
Based on this [stackexchange thread](https://android.stackexchange.com/questions/69168/how-do-i-exclude-album-art-from-the-gallery), if you rename your album cover images to `Folder.jpg` or `AlbumArt.jpg` they should not be listed. Can you try this?

Also, you can try to install a bunch of Gallery apps (you should find some on the f-droid repo), just to check if these apps consider your music folders as an image bucket.

> It also appears that we have a "Cozy Drive" bucket :o

What's inside this "Cozy Drive" bucket? I don't have one on my android emulator where I do all my tests.

Crash--:
Few more remarks: 

- When we upload a file, we use the filename to check if the file is already uploaded. It works great if we have only one folder to sync, but now that we have multiple folders... we can have the same name several times. Do you know if the bucket name is unique? 

This one is very problematic and we should find a way to fix it. 

- It seems that a DCIM folders is not see as a bucket folder. Investing to see if others app are working

> What's inside this "Cozy Drive" bucket? I don't have one on my android emulator where I do all my tests.

In fact in Cozy bucket... we have the files downloaded to be available offline ^^

> I think you have album covers in your music folders, and these album covers are listed as images.
Based on this stackexchange thread, if you rename your album cover images to Folder.jpg or AlbumArt.jpg they should not be listed. Can you try this?

Works great!

lmeunier:
Hi,

> When we upload a file, we use the filename to check if the file is already uploaded. It works great if we have only one folder to sync, but now that we have multiple folders... we can have the same name several times. Do you know if the bucket name is unique?

Bucket names are not unique. If I'm not mistaken, a bucket is just a folder containing images. So you can have multiple buckets with the same name (and multiple photos with the same name).
That being said, buckets have a unique ID (column BUCKET_ID). I don't know how we can use it.

With the current deployed cozy-app, there are cases where images are synced from multiple folders, and thus it's possible that filenames are duplicated and some photos not synced:
* it's possible to have a "Camera" folder in the Internal *and* External storage (it's the default when Android ask you to move contents from Internal storage to SD card)
* when a user changes it's phone, the new phone have an empty "Camera" folder and the file counter is reset to 1, so photos named DSC_0001.JPG, DSC_0002.JPG, etc will appear in this folder

In both cases, some photos won't be synced to cozy-stack because a photo with the same name is already present on cozy-stack.

I was thinking about it in the last few days. The filename is not sufficient to determine if a photo is already synced to cozy-stack. We need to use other informations.
* file date / photo taken date (from EXIF): not reliable enough, these informations can be altered but the photo is still the same
* bucket ID: I've never used them, I don't know if it could be reliable, useless when photos are moved to an other folder, not present on the cozy-stack side
* file content: from the file content, we can compute a CRC or a hash (md5/sha1/...), but it may be cpu and I/O intensive on smartphone, and we would have to patch cozy-stack to expose this information
* file size: should be easy to extract this information on both cozy-app and cozy-stack, not the most reliable information to uniquely identify a file, but coupled with the filename it should be enough
* ...

So we now know how to uniquely identify a photo even if a photo with the same name is already on cozy-stack. We need to upload this photo to cozy-stack, so we need to find a new unique name for this photo on cozy-stack.

Again, we have a lot of choices here:
* timestamp prefix
* UUID prefix
* counter: for example DSC_0001_<counter>.JPG
* information used to uniquely identify the photo (from the list just above): for example, <file_size>_<filename.jpg> or <md5sum>_<filename.jpg>
* subfolder (but what would be the name of the subfolder?)
* ...

With all this in mind, we also have to think about photos already synced to cozy-stack. I don't want to have duplicated photos on my cozy-stack. And to be honest, I don't know how to handle thise case. :)


> It seems that a DCIM folders is not see as a bucket folder. Investing to see if others app are working

I don't understand. You mean that you have a DCIM folder with photos inside, and this bucket is not listed in the cozy-app?
For my tests, I use the [Camera Roll](https://f-droid.org/en/packages/us.koller.cameraroll/) app.

> > What's inside this "Cozy Drive" bucket? I don't have one on my android emulator where I do all my tests.
> 
> In fact in Cozy bucket... we have the files downloaded to be available offline ^^

Ok, one solution would be to add an exception for the "Cozy Drive" bucket and not show it in the buckets list.

nono:
Ping @Crash--

nono:
ping @Crash--

PR #1680, opened 2019-05-15
Breadcrumb fix

y-lohse:
I deleted a variable that seemed unused, but the file was `required` elsewhere, and stylus doesn't complain about undefined variables.

### DIFF
--- a/src/drive/web/modules/navigation/breadcrumb.styl
+++ b/src/drive/web/modules/navigation/breadcrumb.styl
@@ -3,7 +3,7 @@
 @require 'settings/breakpoints.styl'
 @require 'settings/z-index.styl'
 
-@require '../../../styles/topbar.styl'
+$coz-bar-size = 3rem
 
 .fil-path-backdrop
   flex 1 1 auto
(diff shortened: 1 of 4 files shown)


y-lohse:
[iOS+Android 1.18.3 : Regression visuelle fil d'ariane](https://trello.com/c/RSWPnw2E/1974-iosandroid-1183-regression-visuelle-fil-dariane)

src/drive/web/modules/navigation/breadcrumb.styl
```diff
@@ -32,15 +31,15 @@
     height     $coz-bar-size
     border     0
     flex-shrink 0
-    background embedurl('../assets/icons/icon-arrow-left.svg') center center no-repeat
+    background embedurl('../../../assets/icons/icon-arrow-left.svg') center center no-repeat
```
Crash--: Why this change.

y-lohse: This is so weird. Once I removed `@require '../../../styles/topbar.styl'`, the build started failing… because indeed theses pathes were wrong. Why do they work when the `@require '../../../styles/topbar.styl'` line is present is beyond me.

PR #1685, opened 2019-05-20
chore: Add HMR to Drive web

Crash--:
It brings HMR only to Drive web ATM.

We'll have an error `window is not defined` but will be fixed in the next version of cozy-scripts. (first PR here : https://github.com/CPatchane/create-cozy-app/pull/1222) 

Enjoy


src/drive/targets/browser/index.jsx
```diff
@@ -61,11 +61,23 @@ document.addEventListener('DOMContentLoaded', () => {
...
+
+const AppComponent = props => (
+  <I18n lang={props.lang} polyglot={props.polyglot}>
+    <CozyProvider store={props.store} client={props.client}>
+      <Router history={props.history} routes={AppRoute} />
+    </CozyProvider>
+  </I18n>
+)
+
+const HotedApp = hot(module)(AppComponent)
```
y-lohse: What's `module` here?

Crash--: This is the "old" API. But compatible with parcel. And it seems to work well. I didn't try the new one (without `module`). We could try for a new entry. It seems to be more resilient on JS errors... 

`module` is the global `webpack` object. https://webpack.js.org/api/module-variables/

PR #1686, opened 2019-05-20
fix: Request email permissions for the webapp

y-lohse:
The mobile app uses the same permissions as the web app, so we need email permissions for both.

Arguably, we don't need the mobile specific code anymore..? Should we remove [this](https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/lib/cozy-helper.js#L26) and everything that goes with it, or leave it as a backup?

### DIFF
--- a/src/drive/targets/mobile/InitAppMobile.jsx
+++ b/src/drive/targets/mobile/InitAppMobile.jsx
@@ -33,8 +33,7 @@ import {
   initBar,
   restoreCozyClientJs,
   resetClient,
-  getOauthOptions,
-  permissions
+  getOauthOptions
 } from 'drive/mobile/lib/cozy-helper'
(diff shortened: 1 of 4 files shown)


y-lohse:
[App - 403 envoi messages](https://trello.com/c/22X8mMDq/1964-app-403-envoi-messages)

Crash--:
Let's remove the code ! 💪

src/drive/targets/mobile/InitAppMobile.jsx
```diff
@@ -238,8 +237,7 @@ class InitAppMobile {
     const root = document.querySelector('[role=application]')
     const onboarding = {
       oauth: {
-        ...getOauthOptions(),
-        scope: permissions
+        ...getOauthOptions()
```
Crash--: Should be enough since in the new cozy-auth version, we don't use the scope :)

PR #1687, opened 2019-05-20
Fix warnings in js console

cedricmessiant:
Remove 2 props related warnings in js console


cedricmessiant:

PR #1691, opened 2019-05-20
chore: Replace p by a div since we now have figure child

Crash--:
`<p>` HTML tag cannot contain block elements like `<figure>`, it triggers a DOM warning in some browsers console, hence the `<div>` replacement.

See #991 

This PR is not about using UI component plz ;)


Crash--:

PR #1692, opened 2019-05-20
fix: Don't use setState if unmounting

Crash--:
We don't want to use `setState` on an unmounted component.

### DIFF
--- a/src/components/pushClient/Banner.jsx
+++ b/src/components/pushClient/Banner.jsx
@@ -22,20 +22,23 @@ class BannerClient extends Component {
   state = {
     mustShow: false
   }
+  constructor(props) {
+    super(props)
+    this.willUnmount = false
+  }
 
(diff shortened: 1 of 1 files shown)


Crash--:

src/components/pushClient/Banner.jsx
```diff
@@ -23,19 +23,21 @@ class BannerClient extends Component {
     mustShow: false
   }
 
-  async componentWillMount() {
+  componentWillMount() {
+    this.willUnmount = false
     if (Config.promoteDesktop.isActivated !== true) return
```
y-lohse: I think this line should be in the `componentDidMount`?

Crash--: Hmmm, not really since the render() has already been called in the DidMount. i'll move this line to the render method itself and remove the need of willmount ;)

PR #1695, opened 2019-05-21
Handle conflicts when we move a file

cedricmessiant:
Remaining:
- add unit tests on `cancelMove` and `moveEntries`
- we could have a `CozyFile.restore` method in `cozy-doctypes` that just calls the collection's method but do we want this?

Also, I opened a discussion on trello about error handling


src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -75,7 +77,9 @@ class MoveModal extends React.Component {
 
     try {
       await Promise.all(
-        entries.map(entry => this.moveEntry(entry._id, entry.dir_id))
+        entries.map(
+          async entry => await this.moveEntry(entry._id, entry.dir_id)
```
Crash--: don't need to add async / await here, right? Since they are declared in the method?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -90,10 +94,37 @@ class MoveModal extends React.Component {
     }
   }
 
-  moveEntry = (entryId, destinationId) => {
-    return this.context.client
-      .collection('io.cozy.files')
-      .updateFileMetadata(entryId, { dir_id: destinationId })
+  moveEntry = async (entryId, destinationId, entry) => {
```
Crash--: jsdoc ;)

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -90,10 +94,37 @@ class MoveModal extends React.Component {
...
+        } else {
+          const sbpResp = await client
+            .collection('io.cozy.files')
+            .statByPath(targetPath)
+          // FIXME: strange but we don't have _rev in statByPath response so we have to do a get:
+          const fileToDeleteResp = await client
+            .collection('io.cozy.files')
+            .get(sbpResp.data.id)
+          const fileToDelete = fileToDeleteResp.data
+          await client.collection('io.cozy.files').destroy(fileToDelete)
```
Crash--: destroy => ça envoie à la corbeille ou ça supprime réellement ?

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: I will definitely when the code is OK :-)

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: visiblement ça envoie à la corbeille, je ne crois pas que c'est ce qu'on veut

src/drive/web/modules/move/MoveModal.jsx
Crash--: No need to `await` 

We have an `async` map, we return `moveEntry` which is a `promise`. So `awaiting` `promise.all` is enough. Nop?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -89,10 +97,39 @@ class MoveModal extends React.Component {
...
+            .statByPath(targetPath)
+          await client.collection('io.cozy.files').destroy(resp.data)
+          return await client
+            .collection('io.cozy.files')
+            .updateFileMetadata(entryId, { dir_id: destinationId })
+        }
+      } else {
+        throw e
+      }
+    }
```
Crash--: Am I going to say it? Hmmm. Should we move this function to doctype?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -89,10 +97,39 @@ class MoveModal extends React.Component {
...
+  moveEntry = async (entryId, destinationId, filename) => {
+    const { client } = this.context
+    try {
+      return await client
+        .collection('io.cozy.files')
+        .updateFileMetadata(entryId, { dir_id: destinationId })
+    } catch (e) {
+      if (flag('handle-move-conflicts') && e.status === 409) {
+        const targetPath = await CozyFile.getFullpath(destinationId, filename)
+        if (this.props.sharingState.sharedPaths.includes(targetPath)) {
```
Crash--: Is there a way to not having to rely on props.sharingState here? And having a new arg or something?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -89,10 +97,39 @@ class MoveModal extends React.Component {
...
+   * @param   {string} filename      - The file's name
+   * @returns {Promise}              - A promise that returns the move action response if resolved or an Error if rejected
+   *
+   */
+  moveEntry = async (entryId, destinationId, filename) => {
+    const { client } = this.context
+    try {
+      return await client
+        .collection('io.cozy.files')
+        .updateFileMetadata(entryId, { dir_id: destinationId })
```
Crash--: I don't think we need to await in this method (except when we have to wait a result before going on).

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -89,10 +97,39 @@ class MoveModal extends React.Component {
...
+    } catch (e) {
+      if (flag('handle-move-conflicts') && e.status === 409) {
+        const targetPath = await CozyFile.getFullpath(destinationId, filename)
+        if (this.props.sharingState.sharedPaths.includes(targetPath)) {
+          throw e
```
Crash--: I don't think we should `throw e` since it'll throws `conflict error`. This is not the error we want to receive ? Personnally I would prefer to receive : "throw Can't move since there is a conflict and ${target} is shared`

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: If we want to move this method out of here, we could pass `sharedPaths` as an argument (not really pretty). Or maybe we could pass a `overwriteIfConflict` argument but in this case, we have to move the path calculation and flag check out of the function. I'm OK with this.

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: I thought about it this morning but I don't have a clear opinion on this. I'm not sure that we need a way to move a file except in `MoveModal`.
The main difficulty I see if we do this is that we need to be able to know which files have been deleted so that we can restore them in case of move cancel. I'm about to use the component's state to keep track of the files we delete while cancellation is possible.

src/drive/web/modules/move/MoveModal.jsx
Crash--: I read the code again. I think it is coherent to rely on props since the component is coupled to the sharingProvider, BUT I think `moveEntry` should not be part of the component but more part of the "model". 

How will you know if the file is shared with a `overwriteIfConflict` argument? You will check if the file is shared before calling `moveEntry` ?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -74,8 +86,12 @@ class MoveModal extends React.Component {
     const { t } = this.context
 
     try {
+      // TODO: restore the deleted files
       await Promise.all(
-        entries.map(entry => this.moveEntry(entry._id, entry.dir_id))
+        entries.map(
+          async entry =>
+            await CozyFile.move(entry._id, { folderId: entry.dir_id })
```
Crash--: we can remove this await

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -121,9 +131,17 @@ class MoveModal extends React.Component {
...
 
     return (
-      <Modal size={'xlarge'} closable={false} overflowHidden mobileFullscreen>
+      <Modal
+        size={'xlarge'}
+        closable={false}
+        overflowHidden
+        mobileFullscreen
+        into="body"
+        aria-label={t('Move.modalTitle')}
```
Crash--: 👍

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -36,15 +41,22 @@ class MoveModal extends React.Component {
...
 
     try {
       this.setState({ isMoveInProgress: true })
       await Promise.all(
-        entries.map(entry => this.moveEntry(entry._id, folderId))
+        entries.map(async entry => {
+          const targetPath = await CozyFile.getFullpath(folderId, entry.name)
+          const force =
+            flag('handle-move-conflicts') && !sharedPaths.includes(targetPath)
+          await CozyFile.move(entry._id, { folderId }, force)
```
Crash--: we can remove this await

cedricmessiant:
[(𝐋) Upload conflit - étape 1 - écraser fichier non partagé](https://trello.com/c/0cTBor9A/1761-%F0%9D%90%8B-upload-conflit-%C3%A9tape-1-%C3%A9craser-fichier-non-partag%C3%A9)

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -71,30 +91,44 @@ class MoveModal extends React.Component {
...
+
       Alerter.info(
         t('Move.cancelled', {
           subject: entries.length === 1 ? entries[0].name : '',
           smart_count: entries.length
         })
       )
+
+      if (restoreErrorsCount) {
+        Alerter.error(
```
y-lohse: Doesn't that lead to "flashing" alert messages, that cover each other up, if there are both sucesses and errors?

cedricmessiant: :+1: Changed!

cedricmessiant:
@Crash-- @y-lohse I added a few commits since your reviews to remove warnings, improve/fix stuff. Now waiting for https://github.com/cozy/cozy-client/pull/452 to be merged before we can merge this.

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -1,67 +1,87 @@
...
+
   navigateTo = folder => {
     this.setState({ folderId: folder.id })
   }
 
+  registerCancelable = promise => {
+    const cancelableP = cancelable(promise)
+    this.promises.push(cancelableP)
+    return cancelableP
+  }
```
Crash--: to follow @ptbrowne advice we should have something like : 

```jsx 
registerCancelable = promise => {
if(!this.promises) this.promises = []
const cancelableP = cancelable(promise)
this.promises.push(cancelableP)
return cancelableP
}
```

```jsx 
unregisterCancelable(){
this.promises.forEach(p => p.cancel())
this.promises = [] 
}
```

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -70,31 +90,46 @@ class MoveModal extends React.Component {
...
+          )
+        )
+      )
+      const fileCollection = client.collection(CozyFile.doctype)
+      let restoreErrorsCount = 0
+      await Promise.all(
+        trashedFiles.map(id => {
+          try {
+            this.registerCancelable(fileCollection.restore(id))
+          } catch (e) {
```
Crash--: we will have lint error.

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -165,11 +206,20 @@ class MoveModal extends React.Component {
 }
 
 MoveModal.propTypes = {
-  entries: PropTypes.array
+  client: PropTypes.object.isRequired,
+  displayedFolder: PropTypes.object.isRequired,
+  entries: PropTypes.array,
+  t: PropTypes.func.isRequired,
+  sharingState: PropTypes.object // to know how to handle conflicts
```
Crash--: Maybe we can be more explicit for this comment. It's pretty clear now, but in a few months " why on earth we need sharingState to handle conflicts? It doesn't make any sens" :D

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: not sure it applies here since we are cancelling the promises and not really unregistering them. But I agree that a separate method `cancelAllPromises` would be nice

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: Why? (No lint error locally or on Travis)

src/drive/web/modules/move/MoveModal.jsx
cedricmessiant: :+1:

src/drive/web/modules/move/MoveModal.jsx
Crash--: :o ? We don't have "no-used-var" rule ? 

I'm sure we're adding a new more line in our drive console :D

src/drive/web/modules/move/MoveModal.jsx
Crash--: In fact, we have to unregister them. Since the contructor is not called everytime, we should be sure to reset our promise array on unmount

PR #1696, opened 2019-05-21
fix: Remove lot of messages in the console

Crash--:
- filter on message sent by react-dev-tools 
- unique key on intent search


Crash--:
Yep. Check a look at the update ;). Or check the trello card :D

Crash--:
48 passed (14m 45s)
✨  Done in 896.47s.

PR #1697, opened 2019-05-21
fix: Remove check for corrupted files

Crash--:
We can now remove this part, since we have finished the migration path


Crash--:
[Désactiver code correction fichiers corrompus 64ko](https://trello.com/c/ZzgRtawr/1753-d%C3%A9sactiver-code-correction-fichiers-corrompus-64ko)

PR #1707, opened 2019-05-22
fix: Remove double / in paths (root directory)

cedricmessiant:
When `mydoc.odt` was in the root folder, the calculated path was `//mydoc.odt` and thus, we did not detect that the file was shared

### DIFF
--- a/src/drive/web/modules/upload/index.js
+++ b/src/drive/web/modules/upload/index.js
@@ -204,7 +204,7 @@ const uploadFile = async (client, file, dirID) => {
 export const getFileFullpath = async (client, file, dirID) => {
   const resp = await client.collection('io.cozy.files').get(dirID)
   const parentDirectory = resp.data
-  return `${parentDirectory.path}/${file.name}`
+  return `${parentDirectory.path}/${file.name}`.replace('//', '/')
 }
 
 /*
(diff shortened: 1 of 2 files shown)


cedricmessiant:
Maybe `lodash.trimEnd` is better (i.e more specific) for this: https://lodash.com/docs/4.17.11#trimEnd

src/drive/web/modules/upload/index.js
```diff
@@ -204,7 +204,7 @@ const uploadFile = async (client, file, dirID) => {
 export const getFileFullpath = async (client, file, dirID) => {
   const resp = await client.collection('io.cozy.files').get(dirID)
   const parentDirectory = resp.data
-  return `${parentDirectory.path}/${file.name}`
+  return `${parentDirectory.path}/${file.name}`.replace('//', '/')
```
Crash--: We can't have a filename with `//` in it?

cedricmessiant: I get an error from the stack when I try to use `/` in a filename (rename). Also, I can't use `/` in a filename on linux so I can't test an upload with `/` in filename

Crash--:
Can we close this one? @cedricmessiant

Crash--:
ping @cedricmessiant

nono:
Ping @cedricmessiant

Crash--:
We have done that in https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/File.spec.js#L56

PR #1710, opened 2019-05-23
Fix/class name

Crash--:
class is not a valid react attribute. className is


yarn.lock
```diff
@@ -10409,6 +10409,7 @@ minilog@3.1.0:
 "minilog@https://github.com/cozy/minilog.git#master":
   version "3.1.0"
   resolved "https://github.com/cozy/minilog.git#f01f7d9dfe20981177dd34b9662c2f077d818f82"
+  integrity sha1-0tDxiHyjY9Gs8OqG1cTfKTs/tnU=
```
y-lohse: Why is this change here?

Crash--: Because we merged a PR without editing this file in fact. This is needed ;) 

I closed the PR fixing this issue this morning this I've also the commit in this branch.

PR #1714, opened 2019-05-24
fix: Path in root directory (use CozyFile model)

cedricmessiant:
Depends on https://github.com/cozy/cozy-libs/pull/454

### DIFF
--- a/src/drive/web/modules/upload/index.js
+++ b/src/drive/web/modules/upload/index.js
@@ -4,6 +4,7 @@ import logger from 'lib/logger'
 import flag from 'cozy-flags'
 
 import { hasSharedParent, isShared } from 'sharing/state'
+import { CozyFile } from 'models'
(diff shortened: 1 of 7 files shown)


src/drive/web/modules/upload/index.js
```diff
@@ -99,6 +100,11 @@ export const processNextFile = (
       'Upload module needs a cozy-client instance to work. This instance should be made available by using the extraArgument function of redux-thunk'
     )
   }
+
+  if (!CozyFile.cozyClient) {
+    CozyFile.registerClient(client)
+  }
+
```
Crash--: Seems a bit weird to register the client here in a middle of this function. No?

cedricmessiant: moved after cozy client initializations

Crash--:
tests are failing :(

PR #1718, opened 2019-05-27
fix: Remove wait(DOWNLOAD_DELAY) from checkLocalFile

CozyKourai:
- Remove delay before checking for downloads.
- Add a loop to check for file every 10ms. Error will be throw if the file still not exist after `DOWNLOAD_DELAY`
- increase `DOWNLOAD_DELAY` to avoid false error as we don't have to wait the whole delay anymore

### DIFF
--- a/testcafe/tests/helpers/utils.js
+++ b/testcafe/tests/helpers/utils.js
@@ -92,11 +92,26 @@ export const getLastExecutedCommand = ClientFunction(
 )
 
 //@param{string} filepath : Expected full path to file
-export async function checkLocalFile(filepath) {
-  await t.wait(data.DOWNLOAD_DELAY)
-  await t.expect(fs.existsSync(filepath)).ok(`${filepath} doesn't exist`)
-  logger.debug(`${filepath} exists on local drive`)
+export async function checkLocalFile(t, filepath) {
+  for (var i = 0; i < data.DOWNLOAD_CHECK_RETRIES; i++) {
+    await t.wait(10)
+    if (fs.existsSync(filepath)) {
+      logger.info(
+        `${filepath} exists on local drive. Waited for a total of ${i.toString()} milliseconds`
+      )
+      break
+    }
+  }
+  //Will throw error if file was not download
+  await t
+    .expect(fs.existsSync(filepath))
+    .ok(
+      `${filepath} doesn't exist after ${
+        data.DOWNLOAD_CHECK_RETRIES
+      } milliseconds`
+    )
 }
+
 //@param{string} filepath : Expected full path to file
 export async function deleteLocalFile(filepath) {
   fs.unlink(filepath, function(err) {
--- a/testcafe/tests/drive/file_sharing_scenario.js
+++ b/testcafe/tests/drive/file_sharing_scenario.js
@@ -77,8 +77,8 @@ fixture`Drive : Access a file public link, download the file, and check the 'cre
     await setDownloadPath(data.DOWNLOAD_PATH)
     console.groupEnd()
   })
-  .afterEach(async () => {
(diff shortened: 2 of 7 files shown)


CozyKourai:
[Retirer le timeout du download](https://trello.com/c/XrGhnWHx/1991-retirer-le-timeout-du-download)

testcafe/tests/helpers/utils.js
```diff
@@ -92,11 +92,22 @@ export const getLastExecutedCommand = ClientFunction(
...
+      logger.info(
+        `${filepath} exists on local drive. Waited for a total of ${i.toString()} microseconds`
+      )
+      return
+    }
+  }
+  //Will throw error if file was not download
+  await t
+    .expect(fs.existsSync(filepath))
+    .ok(`${filepath} doesn't exist after ${data.DOWNLOAD_DELAY} microseconds`)
```
Crash--: hmm ok c'est quand l'assertion est bonne, non ? Vu le message, ça devrait être nok ?

CozyKourai: `.ok()` vérifie que l'assertion est bonne en effet, mais testcafé n'affiche le message que lorsque le test a failed.

testcafe/tests/helpers/utils.js
```diff
@@ -92,11 +92,22 @@ export const getLastExecutedCommand = ClientFunction(
...
-  await t.expect(fs.existsSync(filepath)).ok(`${filepath} doesn't exist`)
-  logger.debug(`${filepath} exists on local drive`)
+export async function checkLocalFile(t, filepath) {
+  for (var i = 0; i < data.DOWNLOAD_DELAY; i++) {
+    await t.wait(10)
+    if (fs.existsSync(filepath)) {
+      logger.info(
+        `${filepath} exists on local drive. Waited for a total of ${i.toString()} microseconds`
+      )
+      return
```
y-lohse: Why `return`? Did you mean `break`?

testcafe/tests/helpers/utils.js
```diff
@@ -92,11 +92,22 @@ export const getLastExecutedCommand = ClientFunction(
...
-export async function checkLocalFile(filepath) {
-  await t.wait(data.DOWNLOAD_DELAY)
-  await t.expect(fs.existsSync(filepath)).ok(`${filepath} doesn't exist`)
-  logger.debug(`${filepath} exists on local drive`)
+export async function checkLocalFile(t, filepath) {
+  for (var i = 0; i < data.DOWNLOAD_DELAY; i++) {
+    await t.wait(10)
+    if (fs.existsSync(filepath)) {
+      logger.info(
+        `${filepath} exists on local drive. Waited for a total of ${i.toString()} microseconds`
```
y-lohse: I think it's milliseconds and not microseconds, but whatever. However be aware that the `fs.existsSync` also takes time, so with the current code we will be waiting for longer than `DOWNLOAD_DELAY`. I guess `DOWNLOAD__CHECK_RETRIES` or something like that might be more accurate.

testcafe/tests/helpers/utils.js
CozyKourai: Changed to `break`

testcafe/tests/helpers/utils.js
CozyKourai: `DOWNLOAD_DELAY` is now `DOWNLOAD_CHECK_RETRIES`

testcafe/tests/helpers/utils.js
y-lohse: Alright, let's see if the tests still pass after this change, since before they weren't running at all :/

Crash--: ~~return it's the same as break in a for loop, no?~~

Oh no. You right! OMG

PR #1725, opened 2019-05-31
Use FileInput from cozy-ui

cedricmessiant:
See https://github.com/cozy/cozy-ui/pull/1021

Except the aria-label change, I think it's OK. I tried to fix the fact that we display the icon via MenuItem in photos but styles are not easy to fix. Also, I removed translate() in UploadButton in photos because it seems we don't need it.

### DIFF
--- a/src/drive/web/modules/upload/UploadButton.jsx
+++ b/src/drive/web/modules/upload/UploadButton.jsx
@@ -2,51 +2,25 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { compose } from 'redux'
-import { Icon } from 'cozy-ui/react'
+import { Icon, FileInput } from 'cozy-ui/react'
 
 import withSharingState from 'sharing/hoc/withSharingState'
 import { uploadFiles } from 'drive/web/modules/navigation/duck'
 
-const styles = {
-  parent: {
-    position: 'relative',
-    overflow: 'hidden'
-  },
-  input: {
-    position: 'absolute',
-    top: 0,
-    left: 0,
-    opacity: 0,
-    width: '100%',
-    height: '100%',
-    zIndex: 1
-  }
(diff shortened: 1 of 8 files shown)


src/drive/web/modules/upload/UploadButton.jsx
```diff
@@ -2,51 +2,25 @@ import React from 'react'
...
-
 const UploadButton = ({ label, disabled, onUpload, className }) => (
-  <label
-    role="button"
-    disabled={disabled}
+  <FileInput
     className={className}
-    style={styles.parent}
+    data-test-id="upload-btn"
+    label={label}
```
y-lohse: Seeing this a bit late, but I guess a `label` prop makes it sound like a label will be visually rendered when it's about a11y. Maybe `aria-label` would have been better?

cedricmessiant: I agree. I will make a PR in cozy-ui

cedricmessiant:
Except the `aria-label` change, I think it's OK. I tried to fix the fact that we display the icon via `MenuItem` in photos but styles are not easy to fix. Also, I removed `translate()` in `UploadButton` in photos because it seems we don't need it.

cedricmessiant:
[(𝐒) Composant select file](https://trello.com/c/02UskpST/2008-%F0%9D%90%92-composant-select-file)

src/drive/web/modules/upload/UploadButton.jsx
cedricmessiant: well in fact, there is no `label` prop in `FileInput`, thanks :-)

cedricmessiant: There is no `label` prop in `FileInput` component in fact. So the label we pass here is just passed to the hidden input so I don't think we need this in the end.

PR #1732, opened 2019-06-04
Feat/topbar enhanced

GoOz:
Using `AppTitle` on Drive & Photos and cleaning a bit.

### DIFF
--- a/src/photos/components/Topbar.jsx
+++ b/src/photos/components/Topbar.jsx
@@ -3,7 +3,7 @@ import styles from '../styles/topbar.styl'
 
 import React, { Component } from 'react'
 import { translate } from 'cozy-ui/react/I18n'
-import { withBreakpoints } from 'cozy-ui/react'
+import { withBreakpoints, AppTitle } from 'cozy-ui/react'
 import { withRouter } from 'react-router'
 import PropTypes from 'prop-types'
 import flow from 'lodash/flow'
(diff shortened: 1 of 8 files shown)


src/photos/styles/topbar.styl
```diff
@@ -31,7 +31,7 @@ $coz-bar-size = 3rem
     background   embedurl('../assets/icons/icon-back.svg')
     cursor       pointer
 
-    +tiny-screen()
+    +small-screen()
```
GoOz: This one is just a random fix of a bug I stumbled upon. The return arrow was messed up between mobile & desktop viewport. This fixes it.

src/photos/targets/public/App.jsx
```diff
@@ -80,7 +80,7 @@ export class App extends Component {
                     styles['--hide-bar']
                   )}
                 >
-                  <h2 className={styles['pho-content-title']}>{album.name}</h2>
```
y-lohse: Shouldn't the css rule be removed from the stylus file?

src/photos/components/Topbar.jsx
```diff
@@ -66,9 +66,9 @@ EditableAlbumName.propTypes = {
 }
 
 const TopbarTitle = ({ children }) => (
-  <h2 data-test-id="pho-content-title" className={styles['pho-content-title']}>
+  <AppTitle data-test-id="pho-content-title" className="pho-content-title">
```
y-lohse: `className="pho-content-title"` is probably wrong though, we should use the css module version?

GoOz: Yeah but… I did… What's going on here? 🤔

src/photos/targets/public/App.jsx
GoOz: No because it's still used in https://github.com/cozy/cozy-drive/blob/master/src/photos/components/Topbar.jsx for when there's a input displayed (for renaming) and there's some style with it.

Crash--:
@GoOz Some snapshots are failing

GoOz:
@Crash-- I can't see the issue…

nono:
Ping @y-lohse

PR #1735, opened 2019-06-05
fix: Empty input[type=file] so that the user can upload the same file

cedricmessiant:
We have to empty the value of the input[type=file] so that the user can upload the same file twice. If the input is uncontrolled, it leads to a strange behavior because nothing happens

### DIFF
--- a/src/drive/web/modules/upload/UploadButton.jsx
+++ b/src/drive/web/modules/upload/UploadButton.jsx
@@ -15,6 +15,7 @@ const UploadButton = ({ label, disabled, onUpload, className }) => (
     multiple
     onChange={onUpload}
     data-test-id="upload-btn"
+    value={[]} // always erase the value to be able to re-upload the same file
   >
     <span>
       <Icon icon="upload" />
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/upload/UploadButton.jsx
```diff
@@ -15,6 +15,7 @@ const UploadButton = ({ label, disabled, onUpload, className }) => (
     multiple
     onChange={onUpload}
     data-test-id="upload-btn"
+    value={[]} // always erase the value to be able to re-upload the same file
```
Crash--: should we put this fix in UI?

cedricmessiant: That's where I put it at first but I'm not sure. If we use `hidden={false}`, we don't want it but I think we may want it in all other cases. It depends of the desired behavior in fact. If the UI displays the value in some way, it seems OK that nothing happens if the same file is put in the input (that's the behavior of the browser's widget)

PR #1739, opened 2019-06-10
chore: Remove unused dep

Crash--:
It seems that we don't use this dep ?


y-lohse:
Yes I think we do, it's a peer dep of cozy-ui.

Crash--:
Thanks!

PR #1751, opened 2019-06-27
refactor: Use splitName from Models

Crash--:
Need to wait https://github.com/cozy/cozy-libs/pull/531/files#diff-2fcc2368cf903f66a7df40a3d0c06848R81 

But the idea is to use our new split function from the models ;)


nono:
#531 has been merged. Can we resume the work on this PR?

PR #1764, opened 2019-07-19
refactor: Use createContext API from react, not from third packages

Crash--:
I think create-react-context was there for historical reason. Since we use React now, we can use it from React directly


nono:
Ping @Crash--

PR #1767, opened 2019-07-23
fix: Always display upload item

Crash--:
We had a condition to display or not the upload item if we are in a shared menu (ie we already have a primary action button) or not. But with this condition, if we are in a shared menu, we didn't have the upload item anymore. 

The current fix is to display the upload button anytime in the more menu.

### DIFF
--- a/src/drive/styles/toolbar.styl
+++ b/src/drive/styles/toolbar.styl
@@ -67,7 +67,6 @@
                 right    1rem
                 top      .1rem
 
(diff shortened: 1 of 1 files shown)


Crash--:
[Dossier partagé - bouton Import non affiché](https://trello.com/c/EzXOh9p5/2034-dossier-partag%C3%A9-bouton-import-non-affich%C3%A9)

PR #1770, opened 2019-07-29
fix(deps): Update cozy-bar to v7.6.0

drazik:
The cozy-bar now bundles its own cozy-ui, which should fix the stylesheets conflicts we had between the cozy-bar's cozy-ui and app's cozy-ui.

See https://github.com/cozy/cozy-bar/pull/608 for more infos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "babel-preset-cozy-app": "1.2.2",
     "classnames": "2.2.6",
     "copy-text-to-clipboard": "1.0.4",
-    "cozy-bar": "6.28.5",
+    "cozy-bar": "7.6.0",
     "cozy-ci": "0.2.0",
     "cozy-client": "6.53.0",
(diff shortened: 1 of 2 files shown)


drazik:
There are some visual differences, but the main one is a dropdown content that is totally different. This does not seem to be related to this PR. See https://visualreview.cozycloud.cc/#/2/4/803.

There's also a subtle difference on the "Settings" button in the bar, but nothing breaking, I don't think this is blocking.

Crash--:
> There are some visual differences, but the main one is a dropdown content that is totally different. This does not seem to be related to this PR. See https://visualreview.cozycloud.cc/#/2/4/803.

Yep, it's from a previous PR thanks. 

>There's also a subtle difference on the "Settings" button in the bar, but nothing breaking, I don't think this is blocking. 

Was already there. Thanks

PR #1777, opened 2019-08-02
Improve clustering performances

Gara64:
This improves the overall performances of the clustering, by: 
1. Requesting new files with an index instead of the changes
2. Use cursor-based pagination for albums
3. Use a bigger files batch limit, as allowed by the 2 previous improvements (1000 vs 200)

On tests performed on my local machine, the clustering performed 6x faster on a large dataset (+5k photos).

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "chrome-remote-interface": "0.27.1",
     "cordova": "7.1.0",
     "cordova-android": "7.1.4",
+    "cozy-jobs-cli": "1.8.14",
     "enzyme": "3.9.0",
     "enzyme-adapter-react-16": "1.14.0",
     "enzyme-to-json": "3.3.5",
--- a/src/photos/ducks/clustering/files.js
+++ b/src/photos/ducks/clustering/files.js
@@ -1,44 +1,19 @@
 import { cozyClient, log } from 'cozy-konnector-libs'
 import { DOCTYPE_FILES, DOCTYPE_ALBUMS } from 'drive/lib/doctypes'
 
-export const getChanges = async (lastSeq, limit) => {
-  log('info', `Get changes on files since ${lastSeq}`)
-  const result = await cozyClient.fetchJSON(
-    'GET',
-    `/data/${DOCTYPE_FILES}/_changes?include_docs=true&since=${lastSeq}`
-  )
-  // Filter the changes to only get non-trashed images.
-  const photosChanges = result.results
-    .map(res => {
-      return { doc: res.doc, seq: res.seq }
-    })
-    .filter(res => {
-      return (
-        res.doc.class === 'image' &&
-        !res.doc._id.includes('_design') &&
-        !res.doc.trashed
-      )
-    })
-    .slice(0, limit)
-
-  const newLastSeq =
-    photosChanges.length > 0
-      ? photosChanges[photosChanges.length - 1].seq
-      : null
-  const photos = photosChanges.map(photo => photo.doc)
-  return { photos, newLastSeq }
-}
-
-export const getFilesFromDate = async date => {
-  // Note a file without a metadata.datetime would not be indexed: this is not
-  // a big deal as this is only used to compute parameters
+export const getFilesFromDate = async (
+  date,
+  { indexDateField, limit = 0 } = {}
+) => {
+  log('info', `Get files from ${date}`)
+  const dateField = indexDateField || 'metadata.datetime'
   const filesIndex = await cozyClient.data.defineIndex(DOCTYPE_FILES, [
-    'metadata.datetime',
+    dateField,
     'class',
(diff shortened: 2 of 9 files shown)


src/photos/ducks/clustering/files.js
```diff
@@ -71,24 +50,29 @@ export const getFilesByAutoAlbum = async album => {
...
-    const result = await cozyClient.data.fetchReferencedFiles(album, {
-      skip: skip,
-      wholeResponse: true
-    })
+    const key = [DOCTYPE_ALBUMS, album._id]
+    const cursor = [key, startDocid]
+    const result = await cozyClient.data.fetchReferencedFiles(
+      album,
+      { cursor },
+      'id'
```
y-lohse: I think this should be `_id`, `id` shouldn't even be in the couchdb.

src/photos/ducks/clustering/settings.js
```diff
@@ -106,11 +106,12 @@ export const updateParamsPeriod = async (setting, params, photos) => {
...
-export const updateSettingStatus = async (setting, count, changes) => {
-  log('info', 'Update setting for last seq', changes.newLastSeq)
+export const updateSettingStatus = async (setting, count, lastDate) => {
+  log('info', `Update setting for last date ${lastDate}`)
   const evaluationCount =
     count > 0 ? setting.evaluationCount + count : setting.evaluationCount
-  const lastSeq = changes.newLastSeq
-  const newSetting = { ...setting, evaluationCount, lastSeq }
+  const runs = setting.runs ? setting.runs + 1 : 1
+  const newSetting = { ...setting, evaluationCount, lastDate, runs }
```
y-lohse: This is a breaking change for existing cluster? Do we mind?

src/photos/ducks/clustering/utils.js
```diff
@@ -12,22 +12,30 @@ export const prepareDataset = (photos, albums = []) => {
...
       photo.timestamp = hours
       // For each photo, we need to check the clusterid, i.e. the auto-album
       // referenced by the file. If there is none, the photo wasn't clustered before
-      if (!photo.clusterId && file.referenced_by) {
-        const ref = file.referenced_by.find(
+      if (
+        !photo.clusterId &&
+        file.relationships &&
+        file.relationships.referenced_by &&
+        file.relationships.referenced_by.data
```
y-lohse: If we have lodash around, this sort of things can be simplified with `has(file, 'relationships.referenced_by.data')`

package.json
```diff
@@ -76,6 +76,7 @@
     "chrome-remote-interface": "0.27.1",
     "cordova": "7.1.0",
     "cordova-android": "7.1.4",
+    "cozy-jobs-cli": "1.8.14",
```
y-lohse: We should avoid adding a dependency that isn't used anywhere. If it's a dev tool, I would either keep it as a global dependency on my machine, or add the corresponding scripts to the package.json.

src/photos/ducks/clustering/files.js
Gara64: Actually, this is not the id of the document, but the sort type, see https://github.com/cozy/cozy-client-js/blob/3326271ca0c5796b1c2d660731d7f20e5bb1c845/src/relations.js#L33

src/photos/ducks/clustering/settings.js
Gara64: Yes it is, but there is no existing clusters in prod (there are just some instances on int that we can reset)

y-lohse:
(Approved the PR because it's mergeable as it is, but feel free to make the lodash and cozy-jobs-cli changes for extra points!)

package.json
Gara64: Hm, I tend to disagree for this, this is a dev tools especially needed for the service, so for me it's relevant to be included in the devDep. It is actually done like this in banks: https://github.com/cozy/cozy-banks/blob/master/package.json#L103

y-lohse: In that case it should be used for something in the repo, either in a script, or a documentation. Otherwise the tool is only usable by a single person who happens to know how to use it.

(The other reason why adding the usage somewhere in the repo is important is because if someone in 3 months decides to remove outdated dependencies, this one would get removed because it has no apparent usage.)

Gara64: Fair enough for the documentation :)
Would it be ok in a README in the service [directory](https://github.com/cozy/cozy-drive/tree/master/src/photos/targets/services)?

src/photos/ducks/clustering/utils.js
Gara64: It's actually more `get(file, 'relationships.referenced_by.data')` because `has` will pass if data is null, which we don't want

package.json
y-lohse: Yes!

PR #1778, opened 2019-08-09
Use cozy-client for clustering

Gara64:
This removes all references of cozy-client-js to use only cozy-client in the clustering service.

A positive side-effect  is the removal of the heavy cozy-konnector-libs dependency.

### DIFF
--- a/src/photos/ducks/clustering/albums.js
+++ b/src/photos/ducks/clustering/albums.js
@@ -1,4 +1,4 @@
-import { cozyClient, log } from 'cozy-konnector-libs'
+import log from 'cozy-logger'
 import { DOCTYPE_ALBUMS } from 'drive/lib/doctypes'
 import uniq from 'lodash/uniq'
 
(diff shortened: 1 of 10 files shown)


src/photos/ducks/clustering/albums.js
```diff
@@ -58,64 +73,64 @@ const addAutoAlbumReferences = async (photos, album) => {
...
+    .sortBy([
+      {
+        name: 'desc'
       }
-    }
+    ])
+
+  let result = await client.query(query)
+  let hydrated = client.hydrateDocuments(DOCTYPE_ALBUMS, result.data)
+  let autoAlbums = hydrated
```
Crash--: Why do we need to `hydrateDocuments` ? `query` should do it by itself?

src/photos/ducks/clustering/albums.js
```diff
@@ -58,64 +73,64 @@ const addAutoAlbumReferences = async (photos, album) => {
...
+  let skip = autoAlbums.length
+  while (result.next) {
+    result = await client.query(query.offset(skip))
+    hydrated = client.hydrateDocuments(DOCTYPE_ALBUMS, result.data)
+    autoAlbums = autoAlbums.concat(hydrated)
+    skip = autoAlbums.length
   }
```
Crash--: We have a `queryAll` method now: https://github.com/cozy/cozy-client/commit/932cdc18ae18762eba90f229810f531563110f27#diff-405c53e28840707e47a25b26d2ed602bR445

src/photos/ducks/clustering/files.js
```diff
@@ -41,41 +35,33 @@ export const getFilesFromDate = async (
...
-export const getAllPhotos = async () => {
-  const files = await cozyClient.data.findAll(DOCTYPE_FILES)
+export const getAllPhotos = async client => {
+  // This does not use pagination but is significantly faster
+  // Note this is only used once, for init
+  const results = await client
+    .getStackClient()
+    .collection('io.cozy.files')
+    .all({ limit: null })
+  const files = results.data
```
Crash--: Why are we using stackClient? By doing this, we shortcut the PouchLink (so no offline for instance) :/. Is it only for performance reason?

src/photos/ducks/clustering/albums.js
Gara64: From what I recall discussion with @y-lohse, hydrateDocuments is performed when we use a `<Query>`, otherwise, it must be done manually

src/photos/ducks/clustering/files.js
Gara64: Is it possible to call the [all](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L40) method without stackClient? 
I wanted to use this to request the stack on the `_all_docs` endpoint, which seemed dramatically faster for large datasets.

I didn't know about the bypassing of PouchLink, however this code  it is only used for clustering init, where we query all the files to compute the parameters, so it's probably not crucial for offline?

src/photos/ducks/clustering/albums.js
Crash--: Created an issue on CozyClient (https://github.com/cozy/cozy-client/issues/493) . I tend to think we should fix this behavior.

src/photos/ducks/clustering/files.js
Crash--: I think there is no other way to call `_all_docs` :(

ptbrowne: Have you tried with .limit(null) on a query passed to cozyClient.query() ?

Gara64: Right, it works with a `limitBy(null)`

PR #1779, opened 2019-08-12
fix: Handle section dates

Gara64:
This fixes situations where non-clustered photos wouldn't be displayed in the timeline because the date section (which are days for non-clustered-yet photos) was not properly managed.

### DIFF
--- a/src/photos/ducks/timeline-clusters/dates.js
+++ b/src/photos/ducks/timeline-clusters/dates.js
@@ -48,7 +48,7 @@ const addYear = (f, date) => {
 export const isSameMonth = (f, newerDate, olderDate) => {
   const newer = formatDate(f, newerDate)
   const older = formatDate(f, olderDate)
-  return differenceInCalendarMonths(newer, older) < 1
+  return differenceInCalendarMonths(newer, older) === 0
 }
 
 export const isSameDay = (f, newerDate, olderDate) => {
@@ -60,7 +60,7 @@ export const isSameDay = (f, newerDate, olderDate) => {
 export const isSameHour = (f, newerDate, olderDate) => {
   const newer = formatDate(f, newerDate)
   const older = formatDate(f, olderDate)
-  return differenceInHours(newer, older) < 1
+  return differenceInHours(newer, older) === 0
 }
 
 export const isEqualOrNewer = (newerDate, olderDate) => {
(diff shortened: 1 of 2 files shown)


src/photos/ducks/timeline-clusters/dates.js
```diff
@@ -20,7 +20,8 @@ import {
  */
 const formatDate = (f, date, formatter) => {
   const [year, month, day] = date.substr(0, 10).split('-')
-  const [hours, minutes, seconds] = date.substr(11, 8).split(':')
+  const [hours, minutes, seconds] =
+    date.length > 10 ? date.substr(11, 8).split(':') : [0, 0, 0]
```
Crash--: Are we sure to also have at least 10 char in the date (to have year/month/day)? Or should we do something similar for that? 

When can we have date without `hours / minutes / seconds` ? Is it because we don't have these informations in the exif datas? 

Since we are setting an arbitrary value to `h:m:s`, do we really want to display 

My Section Title ⠂00h ? 

```
 let titleWithHours = section.title + ' ⠂' + formatH(f, startPeriod) + 'h'
      if (!isSameHour(f, endPeriod, startPeriod)) {
        titleWithHours += '-' + formatH(f, endPeriod) + 'h'
      }
```

src/photos/ducks/timeline-clusters/dates.js
```diff
@@ -53,7 +54,7 @@ export const isSameMonth = (f, newerDate, olderDate) => {
 export const isSameDay = (f, newerDate, olderDate) => {
   const newer = formatDate(f, newerDate)
   const older = formatDate(f, olderDate)
-  return differenceInCalendarDays(newer, older) < 1
+  return differenceInCalendarDays(newer, older) === 0
```
Crash--: Should we do the same modification with isSameHour / isSameMonth ? Why doing this only on sameDay ?

src/photos/ducks/timeline-clusters/dates.js
Gara64: > Are we sure to also have at least 10 char in the date 
Yes, the format used is always Y-M-D, like `2019-01-01`

> When can we have date without hours / minutes / seconds ? 
When there is a "daily" section, i.e. photos not clustered yet. For these photos, we extract the date from `metadata.datetime` and group them per day, like `2019-01-01` 

>Since we are setting an arbitrary value to h:m:s, do we really want to display
My Section Title ⠂00h ?

We only show the hours when there are more than one cluster in the same day. The `h:m:s` forced to 0 will only apply to photos without clusters, that will be grouped by day

src/photos/ducks/timeline-clusters/dates.js
Gara64: Right!

PR #1780, opened 2019-08-12
fix(Photos): Give an abject to splitFilename

Crash--:


src/photos/ducks/upload/UploadQueue.jsx
```diff
@@ -20,7 +20,8 @@ const Pending = translate()(props => (
 ))
 
 const Item = translate()(({ file, status }) => {
-  const { filename, extension } = CozyFile.splitFilename(file)
+  //In splitFileName we expect to get an object with a name property
+  const { filename, extension } = CozyFile.splitFilename({ name: file })
```
y-lohse: Maybe we should rename [the prop](https://github.com/cozy/cozy-drive/blob/8d9021427e9a360d974dd5f52fe4f72393bc85d8/src/photos/ducks/upload/UploadQueue.jsx#L149) to fileName to make things more explicit ?

Crash--: In fact, I just passed the full file object instead of just the filename. Same API everywhere :)

y-lohse:
I still think renaming the prop would be worthwhile, but LGTM :)

PR #1809, opened 2019-09-02
feat: Use a better index

Crash--:
indexing by dir_is first since we're looking for files in a specific dir_id. The order of the indexes are very important 

See: https://docs.cozy.io/en/cozy-stack/pouchdb-quirks/#indexing-more-than-one-field
> Furthermore, the order in which fields are indexed on a multi-index is significant, most notably when it comes to sorting.

### DIFF
--- a/src/drive/web/modules/move/MoveModal.jsx
+++ b/src/drive/web/modules/move/MoveModal.jsx
@@ -152,7 +152,8 @@ export class MoveModal extends React.Component {
           $ne: TRASH_DIR_ID
         }
       })
-      .sortBy([{ type: 'asc' }, { name: 'asc' }])
+      .indexFields(['dir_id', 'type', 'name'])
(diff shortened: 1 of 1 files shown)


src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -152,7 +152,8 @@ export class MoveModal extends React.Component {
           $ne: TRASH_DIR_ID
         }
       })
-      .sortBy([{ type: 'asc' }, { name: 'asc' }])
+      .indexFields(['dir_id', 'type', 'name'])
+      .sortBy([{ dir_id: 'asc' }, { type: 'asc' }, { name: 'asc' }])
```
y-lohse: Do we need the new sort option here?

Crash--: Yes because we need to sort by the same fields that the ones indexed.

y-lohse: Uh, ok.

PR #1810, opened 2019-09-02
feat: Improve parameters recomputing

Gara64:
We used to store the period of the photos concerned by a set of parameters, to be able to know which photos' period were clustered with which set of parameters, to retrieve the correct parameters for new photos (e.g. the photo with the datetime 2019-02-01 will be clustered with the params having a `period` between 2019-01-01 and 2019-03-01).

Additionally, we kept track of how many new photos were clustered since the last re-evaluation of params, to start a new params recomputing when  `evaluationCount > EVALUATION_THRESHOLD`.

However, this was not efficient, because `evaluationCount` was incremented for each new photo, even though the datetime was not necessary newer than the last params period, while a new params recomputing will only be on photos newer than the last params period, causing [unecessary attemps](https://github.com/cozy/cozy-drive/blob/a8f705ef3b45084e88c08b91e78fcebc43425cf0/src/photos/targets/services/onPhotoUpload.js#L137-L139). 

So, we introduce an `evaluation` period,  which is checked to know if it is worth to increment the `evaluationCount`.

### DIFF
--- a/src/photos/ducks/clustering/consts.js
+++ b/src/photos/ducks/clustering/consts.js
@@ -1,4 +1,3 @@
-export const THRESHOLD = 314
 export const DEFAULT_PERIOD = 'all'
 export const DEFAULT_MODE = 'default'
 export const SETTING_TYPE = 'clustering'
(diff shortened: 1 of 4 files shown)


Gara64:
@ptbrowne Yep, I added tests

PR #1816, opened 2019-09-06
feat: Activate service for % of instances

Gara64:
This activates the clustering service, by using the real trigger condition, i.e. when a new photo is uploaded. However, we only run it for a fixed 5% of instances, in order to make a progressive deployment ; we plan to increase this % progressively.

### DIFF
--- a/src/photos/ducks/clustering/utils.js
+++ b/src/photos/ducks/clustering/utils.js
@@ -92,11 +92,11 @@ const hashCode = toHash => {
 }
 
 /**
- *  Pick the instance if it matches the condition, met a percentage of cases
+ *  Returns true if `instance` is chosen to be part of a progressive rollout, according to `percentage`
  *  @param {string} instance - The string to hash
  *  @param {number} percent - The percent of instances that should match
  *  @returns {boolean} If the instance is picked or not
  */
-export const pickInstance = (instance, percent) => {
+export const isPartOfProgressiveRollout = (instance, percent) => {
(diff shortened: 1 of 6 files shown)


src/photos/ducks/clustering/utils.js
```diff
@@ -74,3 +73,30 @@ export const convertDurationInMilliseconds = duration => {
...
+  for (i = 0; i < toHash.length; i++) {
+    chr = toHash.charCodeAt(i)
+    hash = (hash << 5) - hash + chr
+    hash = hash & hash // Convert to 32-bit integer
+  }
+  return hash
+}
+
+/**
+ *  Pick the instance if it matches the condition, met a percentage of cases
```
ptbrowne:
```suggestion
 *  Returns true if `instance` is chosen to be part of a progressive rollout, according to `percentage`
```

src/photos/ducks/clustering/utils.js
```diff
@@ -74,3 +73,30 @@ export const convertDurationInMilliseconds = duration => {
...
+ *  @param {string} instance - The string to hash
+ *  @param {number} percent - The percent of instances that should match
+ *  @returns {boolean} If the instance is picked or not
+ */
+export const pickInstance = (instance, percent) => {
```
ptbrowne: I think we could have a more descriptive name for a function returning a boolean.

```suggestion
export const isPartOfProgressiveRollout = (instance, percent) => {
```

PR #1832, opened 2019-09-11
Fix/photos

Crash--:
Since the merge of the "clustered timeline", we have few bugs : 
- can't create an album anymore. `f` was not defined. So I use `f` from `translate()`
- can't add photos during the creation of an album (also, selection bar was visible when selecting). To fix that, we needed to re-export the right component


Crash--:
Closed in favor of #1833

PR #1837, opened 2019-09-17
Feat/scanner

Crash--:
Ajout de la feature "Scanner" un document 1ere version

### DIFF
--- a/src/drive/targets/mobile/config.xml
+++ b/src/drive/targets/mobile/config.xml
@@ -75,15 +75,6 @@
     </platform>
     <preference name="ShowSplashScreenSpinner" value="false" />
     <preference name="AutoHideSplashScreen" value="false" />
-    <edit-config file="*-Info.plist" mode="merge" target="NSCameraUsageDescription">
-        <string>need camera access to take pictures</string>
-    </edit-config>
-    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryUsageDescription">
-        <string>need photo library access to get pictures from there</string>
-    </edit-config>
-    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryAddUsageDescription">
-        <string>need photo library access to save pictures there</string>
-    </edit-config>
     <engine name="android" spec="^7.1.4" />
     <engine name="ios" spec="^4.5.5" />
     <plugin name="cordova-plugin-splashscreen" spec="git+https://github.com/cozy/cordova-plugin-splashscreen.git#e10cadcd07b613f745370ba023ac0c52cf415648" />
--- a/src/drive/targets/mobile/config.xml
+++ b/src/drive/targets/mobile/config.xml
@@ -75,6 +75,15 @@
     </platform>
     <preference name="ShowSplashScreenSpinner" value="false" />
     <preference name="AutoHideSplashScreen" value="false" />
+    <edit-config file="*-Info.plist" mode="merge" target="NSCameraUsageDescription">
+        <string>need camera access to take pictures</string>
+    </edit-config>
+    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryUsageDescription">
+        <string>need photo library access to get pictures from there</string>
+    </edit-config>
+    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryAddUsageDescription">
+        <string>need photo library access to save pictures there</string>
+    </edit-config>
     <engine name="android" spec="^7.1.4" />
     <engine name="ios" spec="^4.5.5" />
     <plugin name="cordova-plugin-splashscreen" spec="git+https://github.com/cozy/cordova-plugin-splashscreen.git#e10cadcd07b613f745370ba023ac0c52cf415648" />
@@ -104,9 +113,6 @@
     <plugin name="cordova-ios-plugin-no-export-compliance" spec="git+https://github.com/mikaoelitiana/cordova-ios-plugin-no-export-compliance.git#915556bf7713e9733005977af4fa4bfba40f203f" />
     <plugin name="io.cozy.jsbackgroundservice" spec="https://github.com/cozy/cordova-jsbackgroundservice.git#1.3.0" />
     <plugin name="com-darryncampbell-cordova-plugin-intent" spec="^1.1.1" />
-    <plugin name="io.cozy.plugins.listlibraryitems" spec="https://github.com/cozy/cordova-plugin-list-library-items.git#1.5.2">
-        <variable name="PHOTO_LIBRARY_USAGE_DESCRIPTION" value=" " />
-    </plugin>
     <plugin name="com.lampa.startapp" spec="^0.1.4" />
     <plugin name="cordova-plugin-queries-schemes" spec="https://github.com/cozy/cordova-plugin-queries-schemes.git#f7cc50ec542d7eeab6857e366dceb2f039f5dcd9" />
     <plugin name="cordova-plugin-keyboard" spec="https://github.com/cozy/cordova-plugin-keyboard.git#4fa4850e6f0ea78e1e2ce573ccde5531a2353edb" />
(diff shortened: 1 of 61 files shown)


src/drive/targets/mobile/config.xml
```diff
@@ -75,6 +75,15 @@
...
     <preference name="AutoHideSplashScreen" value="false" />
+    <edit-config file="*-Info.plist" mode="merge" target="NSCameraUsageDescription">
+        <string>need camera access to take pictures</string>
+    </edit-config>
+    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryUsageDescription">
+        <string>need photo library access to get pictures from there</string>
+    </edit-config>
+    <edit-config file="*-Info.plist" mode="merge" target="NSPhotoLibraryAddUsageDescription">
+        <string>need photo library access to save pictures there</string>
+    </edit-config>
```
Crash--: need to see how we handle that ATM

y-lohse: Have a look at this https://github.com/cozy/cozy-drive/blob/master/src/drive/targets/mobile/translations/app/en.json#L8 ;)

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+  const offlineMessage = () => {
+    return alert(t('Scan.error.offline'))
+  }
+  const uploadingMessage = () => {
+    return alert(t('Scan.error.uploading'))
+  }
+  const actionOnClick = (() => {
+    if (status === 'uploading') return uploadingMessage
+    if (!online) return offlineMessage
+    return onClick
```
y-lohse: Probably worth transforming it into a class.

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+  })()
+
+  return (
+    <span className="u-pl-1 u-flex u-pt-half u-pb-half" onClick={actionOnClick}>
```
y-lohse: Shouldn't this be `actionOnClick()`?

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+      <Icon icon="camera" />
+      <span className="u-pl-half">{t('Scan.scan_a_doc')}</span>
+    </span>
+  )
+})
+class ScanItem extends Component {
```
y-lohse: `Item` keeps popping up but I don't think the name is very clear.

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+          }}
+          generateName={() => {
+            const date = new Date()
+            return `Scan_${date.toISOString()}.jpg`
+          }}
+          onConflict={'rename'}
+          onBeforeUpload={() => stopMediaBackup()}
+          onFinish={() => startMediaBackup()}
+        >
+          {({ status, error, onClick, filename, onClear, online }) => {
```
y-lohse: Why does `online` come from the Scanner ?

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+                <ScanItemMenu
+                  status={status}
+                  onClick={onClick}
+                  online={online}
+                />
+              )
+            }
+            if (!filename)
+              return (
+                <ScanItemMenu
```
y-lohse: This is the same as `error`, isn't it?

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
y-lohse: And maybe `onClick` could be renamed to something like `startScanner`?

src/drive/web/modules/drive/Toolbar/components/Scanner.jsx
```diff
@@ -0,0 +1,210 @@
+import React from 'react'
+
+import { withClient } from 'cozy-client'
+
+import { CozyFile } from 'cozy-doctypes'
+
+class Scanner extends React.Component {
+  state = {
+    status: 'iddle',
```
y-lohse:
```suggestion
    status: 'idle',
```

src/drive/web/modules/drive/Toolbar/components/Scanner.jsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+  onClick = () => {
+    this.defaultPluginConfig = {
+      quality: 70,
+      destinationType: window.navigator.camera.DestinationType.FILE_URI,
+      sourceTypes: window.navigator.camera.PictureSourceType.CAMERA,
+      correctOrientation: true
+    }
+
+    window.navigator.camera.getPicture(this.onSuccess, this.onFail, {
```
y-lohse: We should have a try/catch here or something, in case the plugin is not available.

src/drive/web/modules/drive/Toolbar/components/Scanner.jsx
```diff
@@ -0,0 +1,210 @@
...
+            this.setState({ error: err })
+            this.setState({ status: 'done' })
+            console.error('error getting fileentry file!' + err)
+          }
+        )
+      },
+      error => {
+        this.setState({ error })
+        this.setState({ status: 'done' })
+      }
```
y-lohse: This whole section would be clearer with separate functions instead of nested callbacks imo.

src/drive/web/modules/drive/Toolbar/components/Scanner.jsx
y-lohse: And maybe we can expose constants for the different statuses and conflict strategies? Avoids hard to detect typos.

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
Crash--: Nop. We invoke immediately the method instancing `actionOnClick`. So in `actionOnClick` it's bind to the right method. No need to invoke it again. `onClick` will trigger the invokation

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
Crash--: I though it's easier that the Scanner handles by itself the online / offline status. In a near future, Scanner will be able to write in localStorage the content of the scan if offline and then sync it later. 

Agreed with `startScanner` thanks for the reminder :)

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
Crash--: I don't think so. 

We can have a filename since we create it before a remote request and then having an error in the request. For instance `NO_QUOTA` or some other errors

Crash--: Oh now I see what you mean! Yeah ATM this is the same display, let's remove this condition and add it again later if needed :)

src/drive/web/modules/drive/Toolbar/components/PortaledQueue.jsx
```diff
@@ -0,0 +1,22 @@
...
+        queue={file ? [file] : []}
+        doneCount={doneCount}
+        successCount={successCount}
+        t={t}
+        purgeQueue={onClear}
+      />,
+      document.querySelector('[role=application]')
+    )
+  }
+}
```
ptbrowne: This component's responsibility seems unclear to me. By its name, it seems like that it only does the portalling but it also

- changes the `purgeQueue` to `onClear`.
- changes the `queue` to `file`

I think we should limit/simplify the components responsibility.

- Why do we need to change the props ?
- I think UploadQueue should receive the `t` function by itself ?

Can you add a docstring explaining the responsibilities ?

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
```diff
@@ -0,0 +1,109 @@
...
+  startMediaBackup,
+  cancelMediaBackup
+} from 'drive/mobile/modules/mediaBackup/duck'
+
+import { connect } from 'react-redux'
+/**
+ *
+ */
+
+const ScanItemMenu = translate()(({ status, onClick, t, online }) => {
```
ptbrowne: ScanMenuItem ?

src/drive/web/modules/drive/Toolbar/components/ScanItem.jsx
ptbrowne: I think this component should have a docstring.

src/drive/web/modules/trash/Toolbar.jsx
```diff
@@ -54,7 +53,7 @@ const Toolbar = ({
     >
       <Button
         theme={'danger-outline'}
-        className={classNames(styles['u-hide--mob'])}
+        className={'u-hide--mob'}
```
ptbrowne:
```suggestion
        className='u-hide--mob'
```

PR #1853, opened 2019-09-30
chore: Use critical logs

Gara64:
Use `critical` level of logs, which is the only one interpreted by the stack to say if the job is actually failing. This is useful to monitor the logs.


Crash--:
We can close it since the commit was marger with the release branch

PR #1866, opened 2019-10-16
Fix/crash when no meta

Crash--:
- Since we're trying to display the label even if no label => crash 
- Also updated UI to have a better Alerter support on iPhone X+

### DIFF
--- a/src/drive/web/modules/actionmenu/MobileActionMenu.jsx
+++ b/src/drive/web/modules/actionmenu/MobileActionMenu.jsx
@@ -65,14 +65,19 @@ const MenuHeaderFile = ({ file, t }) => {
               {extension}
             </span>
           </div>
-          <div className="u-coolGrey u-fz-tiny u-fs-normal u-flex u-flex-items-center">
-            <Icon icon="qualify" size="10" />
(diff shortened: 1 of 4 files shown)


src/drive/web/modules/actionmenu/MobileActionMenu.jsx
```diff
@@ -65,14 +65,19 @@ const MenuHeaderFile = ({ file, t }) => {
...
-          <div className="u-coolGrey u-fz-tiny u-fs-normal u-flex u-flex-items-center">
-            <Icon icon="qualify" size="10" />
-            <Caption
-              className={classNames(styles['fil-mobileactionmenu-category'])}
-            >
-              {t(`Scan.items.${file.metadata.label}`)}
-            </Caption>
-          </div>
+          {file.metadata &&
+            file.metadata.label && (
```
y-lohse: I find `has(file, 'metadata.label')` more elegant and less error-prone, but not blocking for me.

Crash--: I'm waiting for file?.metadata.?label :D

PR #1869, opened 2019-10-21
fix: Don't open file/folder when selection isActive

Crash--:
We should not be able to open / view a file when we're in `selection mode`. 

- Before the prop was read during the didmount() and not during the callback 
- Also, we had a proptypes for `selectionModeActive` but this props was never passed... So added the HOC


Crash--:
I updated the comment to reflect the 2 fixed bugs

PR #1870, opened 2019-10-22
Feat/thumbnail real time

Crash--:
- cozy-realtime upgrade 
- use withClient instead of context to get cozy-client instance 
- No need to handle manually the process of refresh token 

- Add realtime for thumbnail in order to display it when the thumbnail is created


src/components/Image/ImageLoader.jsx
```diff
@@ -12,20 +15,32 @@ const LOADED = 'LOADED'
...
+    this.realtime = new Realtime({ client })
+    this.type = 'io.cozy.files.thumbnails'
+    this.realtime.subscribe('created', this.type, this.handlerCreate)
+  }
+
+  /**
+   * Reload the link when realtime tell us that the
+   * thumbnail is created. By default size === small
+   */
+  handlerCreate = doc => {
```
y-lohse: Nit: handleCreate or createHandler would be more correct english :)

src/components/Image/ImageLoader.jsx
```diff
@@ -35,8 +50,13 @@ class ImageLoader extends React.Component {
       this.img.onload = this.img.onerror = null
       this.img.src = ''
     }
+    this.unsubscribeRealtime()
   }
 
+  unsubscribeRealtime = () => {
+    this.realtime &&
```
y-lohse: Can `this.realtime` be `undefined` at this stage ?

src/drive/web/modules/navigation/RealtimeFiles.jsx
```diff
@@ -18,23 +19,14 @@ import logger from 'lib/logger'
 export class RealtimeFiles extends React.Component {
   realtimeListener = null
   pouchListener = null
-  static contextTypes = {
-    client: PropTypes.object.isRequired
-  }
+
   componentWillMount() {
```
y-lohse: Could we change this to `didMount` while we're here, or what that involve other changes?

src/drive/web/modules/navigation/RealtimeFiles.spec.jsx
```diff
@@ -31,86 +30,58 @@ describe('RealTimeFiles', () => {
...
+    const addFile = jest.fn()
+    const deleteFile = jest.fn()
+    const component = mount(
+      <RealtimeFiles
+        client={client}
+        files={files}
+        updateFile={updateFile}
+        addFile={addFile}
+        deleteFile={deleteFile}
+        // location={{ pathname: 'recent' }}
```
y-lohse: Can be deleted I guess :)

src/drive/web/modules/navigation/RealtimeFiles.spec.jsx
```diff
@@ -31,86 +30,58 @@ describe('RealTimeFiles', () => {
...
+      .fn()
+      .mockReturnValueOnce(false)
+      .mockReturnValue(true)
+    component.instance().onDocumentChange({ _id: '2' })
+    expect(deleteFile).toBeCalled()
+
+    //Doc was not present, but the current view => add
+    component.instance().isInCurrentView = jest.fn().mockReturnValue(true)
+    component.instance().onDocumentChange({ _id: '4' })
+    expect(addFile).toBeCalled()
```
y-lohse: There's a potential issue that these test don't cover: if the update test also triggers a call to `deleteFile`, it won't be detected. I think the safest way to test these would be to reset the mocks after each test, and to also test stuff like `expect(deleteFile).not.toBeCalled()`.

src/components/Image/ImageLoader.jsx
Crash--: fixed (and also in our doc : https://github.com/cozy/cozy-libs/pull/855)

src/components/Image/ImageLoader.jsx
Crash--: I think we can have a race condition between the didMount() and unmount() so I added a check to be sure

PR #1872, opened 2019-10-24
feat: Increase your quota

Crash--:
Add a new button when we're out of quota to upgrade our current plan

### DIFF
--- a/src/drive/web/modules/upload/QuotaAlert.jsx
+++ b/src/drive/web/modules/upload/QuotaAlert.jsx
@@ -1,15 +1,32 @@
 import React from 'react'
+import get from 'lodash/get'
 import Modal from 'cozy-ui/react/Modal'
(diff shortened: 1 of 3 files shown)


src/drive/web/modules/upload/QuotaAlert.jsx
```diff
@@ -1,15 +1,32 @@
...
-    secondaryAction={onClose}
-  />
-)
-
-export default translate()(QuotaAlert)
+const buildPremiumLink = (uuid, managerUrl) =>
+  `${managerUrl}/cozy/instances/${uuid}/premium`
+
+const QuotaAlert = ({ t, onClose, client }) => {
+  const instanceInfo = useWithInstance(client)
```
y-lohse: Lol I think `useInstance` or `withInstance` might be enough, no need to hook the hocs :D

PR #1875, opened 2019-10-28
fix: Better icons

Crash--:
We now use Icon from cozy-ui for ou public page


src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -177,14 +181,16 @@ class PublicToolbar extends React.Component {
...
       const sourceId = response.data.attributes.source_id
       const sharingId = sourceId.split('/')[1]
-
+      console.log('sharingId', sharingId)
       const { sharecode } = getQueryParameter()
 
       const link = client
         .collection('io.cozy.sharings')
         .getDiscoveryLink(sharingId, sharecode)
+      console.log('link', link)
```
ptbrowne: There are 3 console.logs :) The linter should catch those?

PR #1876, opened 2019-10-28
feat: Display QuotaAlert when needed

Crash--:
- Dispatch a quotaAlert when needed 
- Add <ModalManager at the root of the layout since it can be called from every where 
- Should use `QuotaAlert` from UI (see https://github.com/cozy/cozy-ui/pull/1218 but let's do that in an other PR when merged) 

- Disk usage should be in cozy-client (see https://github.com/cozy/cozy-client/issues/553)

### DIFF
--- a/src/photos/ducks/upload/index.jsx
+++ b/src/photos/ducks/upload/index.jsx
@@ -1,9 +1,12 @@
+import React from 'react'
 import { combineReducers } from 'redux'
 import logger from 'lib/logger'
 
 import Alerter from 'cozy-ui/react/Alerter'
 
 import UploadQueue from './UploadQueue'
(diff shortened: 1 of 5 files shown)


src/photos/ducks/upload/index.jsx
```diff
@@ -70,11 +75,24 @@ const processNextFile = callback => async (dispatch, getState) => {
     dispatch({ type: RECEIVE_UPLOAD_SUCCESS, file })
   } catch (error) {
     logger.log(error)
+    const statusError = {
+      409: CONFLICT,
```
y-lohse: It doesn't look like CONFLICT is being used here at all?

Crash--: thanks !

PR #1888, opened 2019-11-04
fix: Overflow hidden only on mobile

Crash--:
We recently applied an overflow hidden to fix the scroll issue on Android. But since, we broke the scroll on a web browser. Let's add this fix only on a mobileApp ;)


Crash--:
done in https://github.com/cozy/cozy-drive/pull/1890

PR #1889, opened 2019-11-05
Feat/cozy auth

Crash--:
Currently being tested

Several points:
- Use withClient() instead of this.context.client
- Use cozy-auth to handle auth 
- Add support for universal link 
- New way to handle the push to /onboarding. We store an `onboarded` boolean in the localStorage 
- Add a migration function to handle the old and new way to connect


src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
```diff
@@ -1,108 +1,93 @@
...
-  onboardingInformationsPropTypes,
-  onboardingPropTypes
-} from '../../../../authentication/OnboardingPropTypes'
+import appBooted from '../../../targets/mobile/'
 class DriveMobileRouter extends Component {
-  static contextTypes = {
-    client: Proptypes.object.isRequired
+  async componentDidMount() {
+    //Wait for the app to be booted to avoid race condition between cordova & JS
+    await appBooted
```
y-lohse: I'm not sure how that works, I thought awaiting in componentDidMount didn't prevent the initial rendering of the component..?

src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
```diff
@@ -1,108 +1,93 @@
...
-          appRoutes={appRoutes}
+          protocol={PROTOCOL}
+          appTitle={SOFTWARE_NAME}
+          universalLinkDomain={getUniversalLinkDomain()}
+          appSlug={appMetadata.slug}
           history={history}
-          onAuthenticated={this.afterAuthentication}
+          onAuthenticated={async () => {
+            return await this.afterAuthentication()
+          }}
```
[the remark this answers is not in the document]
Crash--: fixed, thanks

src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
Crash--: Oh you have totally right! changed to use a state instead

y-lohse:
Last commit LGTM 👍

PR #1892, opened 2019-11-08
fix: Use get

Crash--:
🤕

### DIFF
--- a/src/components/pushClient/Supportus.jsx
+++ b/src/components/pushClient/Supportus.jsx
@@ -24,17 +24,13 @@ const Supportus = ({ t, client }) => {
     'context.data.attributes.enable_premium_links'
   )
   const diskUsage = withDiskUsage(client)
+  const quota = get(diskUsage, 'data.attributes.quota', false)
   /**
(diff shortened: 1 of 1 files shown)


ptbrowne:
It seems this component could use a test if it can break. BTW I thought it was supposed to be renamed to SupportUs ?

src/components/pushClient/Supportus.jsx
```diff
@@ -24,17 +24,13 @@ const Supportus = ({ t, client }) => {
     'context.data.attributes.enable_premium_links'
   )
   const diskUsage = withDiskUsage(client)
+  const quota = get(diskUsage, 'diskUsage.data.attributes.quota', false)
```
y-lohse:
```suggestion
  const quota = get(diskUsage, 'data.attributes.quota', false)
```

I think..?

Crash--:
> It seems this component could use a test if it can break. BTW I thought it was supposed to be renamed to SupportUs ?

I don't consider a NIT as a change to be done.

Crash--:
thanks @y-lohse ! 😓

PR #1898, opened 2019-11-20
fix: Support US

Crash--:
- Bad quota fixed 
- Changed string
- Rename files


src/components/pushClient/SupportUs.jsx
```diff
@@ -30,7 +30,7 @@ const Supportus = ({ t, client }) => {
    * if quota < 50Gb then, the user is freemium
    * if managerUrl then user is not self hosted
    */
-  if (enable_premium_links && managerUrl && uuid && quota && quota < 50000000)
+  if (enable_premium_links && managerUrl && uuid && quota && quota < 5000000000)
```
ptbrowne: It seems like the comment could be removed if helper functions were extracted.

```
const isFreemiumUser = instanceInfo => ...
const isSelfHosted = instanceInfo => ...
const arePremiumLinksEnabled = instanceInfo => ...
```

This would remove the risk that the comment get old and outdated and it encourages reuse of functions.

Crash--: https://github.com/cozy/cozy-client/pull/567

Crash--:
Have added tests on https://github.com/cozy/cozy-client/pull/567 

Have added a todo to use the model as soon as the other PR is merged :)

PR #1901, opened 2019-11-20
fix: iOS shared action

Crash--:
To fix the native shared menu, I used the `url` attribute of
cordova social plugin. It works great but only with one file, so
instead of having an error, let's hide this functionnality if we
select more than one file

### DIFF
--- a/src/drive/web/modules/drive/Container.jsx
+++ b/src/drive/web/modules/drive/Container.jsx
@@ -28,6 +28,7 @@ import {
 } from 'drive/mobile/modules/offline/duck'
 import { extractFileAttributes } from 'drive/web/modules/navigation/duck/async'
 import styles from 'drive/styles/actionmenu.styl'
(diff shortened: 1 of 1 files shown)


src/drive/web/modules/drive/Container.jsx
```diff
@@ -102,11 +103,13 @@ const mapDispatchToProps = (dispatch, ownProps) => {
           __TARGET__ === 'mobile'
             ? {
                 action: files => dispatch(exportFilesNative(files)),
-                displayCondition: files =>
-                  files.reduce(
+                displayCondition: files => {
+                  if (isIOSApp()) return files.length === 1
```
y-lohse: This will display it for a folder too right?

Crash--: Nice catch! Thanks. Fixed

Crash--:
Merging since this in a build issue with our android version on travis

PR #1902, opened 2019-11-20
refactor: Use cozy-ui transpiled

Crash--:
We still make reference to cozy-ui/react. This way we uniformise our call and only make call to cozy-ui/transpiled 

We had an alias in our babel config, but since we need to mock a few ui's functions, if a cozy-module was using the transpiled version of UI, our mock didn't work. 

Now that everyone is using transpiled version, it fixes this issue.


ptbrowne:
> We had an alias in our babel config, but since we need to mock a few ui's functions, if a cozy-module was using the transpiled version of UI, our mock didn't work.

I couldn't understand this ? What was the probleme ?

Crash--:
> I couldn't understand this ? What was the problem ?

The problem was that in some of our test we have a mock for the MuiTheme since the method doesn't work well in Jest. 

```
jest.mock('cozy-ui/react/utils/color', () => ({
  getCssVariableValue: () => '#fff'
}))
```

But some of internal libraries use `cozy-ui/transpiled/react/utils/color` so the mock didn't work for this path... 

Instead of adding a new alias for our jest config, I preferred to unified our call to UI and only import it from the transpiled version. 

Before this PR we had in a few components something like : 
```
import Text from 'cozy-ui/react/Text';
import { Modal } from 'cozy-ui/transpiled/react';
```

It was time to uniform

PR #1905, opened 2019-11-22
fix: Add margin-bottom to SupportUS button

Crash--:
add bottom margin, specially needed if we don't hav e the "Desktop pushClient button" since this one give us margin...


ptbrowne:
We should not have to rely on other elements giving margin bottoms, I think it is a smell.

Crash--:
> We should not have to rely on other elements giving margin bottoms, I think it is a smell. 

No kidding? The issue has be opened since 1year ^^ (https://github.com/cozy/cozy-ui/issues/797#issuecomment-540572530)

PR #1908, opened 2019-11-25
feat: Track usage of versioning and CTA

Crash--:


ptbrowne:
Maybe the trackEvent function could be xported by cozy-ui/react/helpers. The conditional on `tracker = getTracker; if (tracker) { ... }` could be extracted there, and at the call point, I think it would be clearer  to keep the event.

`trackEvent(['Drive', 'Versioning', 'OpenVersioningModal'])` 

Reading this, I could be mistaken that the trackEvent function is the same between the 3 files.

Crash--:
Started the work here: https://github.com/cozy/cozy-ui/pull/1267

PR #1918, opened 2019-12-05
Feat/open with

Crash--:
TODO : 
- [x] correct lint on InitAppMobile 
- [ ] squash / fixup commits 
- [x] Add test for the upload/index component


src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
```diff
@@ -30,27 +30,27 @@ class DriveMobileRouter extends Component {
...
     const oauthClient = client.getStackClient()
-    oauthClient.onTokenRefresh = async token => {
-      restoreCozyClientJs(
-        client.getStackClient().uri,
-        client.getStackClient().oauthOptions,
-        token
-      )
-      this.props.dispatch(setToken(token))
-      await initBar(client)
+    oauthClient.onTokenRefresh = async (/*token*/) => {
```
y-lohse:
```suggestion
    oauthClient.onTokenRefresh = async () => {
```

src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
```diff
@@ -30,27 +30,27 @@ class DriveMobileRouter extends Component {
...
       client.getStackClient().oauthOptions,
       client.getStackClient().token
     )
+    setToken(accesstoken)
     await initBar(client)
-    this.props.saveServerUrl(client.getStackClient().uri)
+    saveServerUrl(client.getStackClient().uri)
     setCozyUrl(client.getStackClient().uri)
-    this.props.saveCredentials(client, accesstoken)
+    saveCredentials(client, accesstoken)
```
y-lohse: `saveCredentials ` also calls `setToken `, right? Maybe this could use a comment or a little more cleaning up?

src/drive/targets/mobile/InitAppMobile.jsx
```diff
@@ -187,6 +167,43 @@ class InitAppMobile {
...
+    function initError(err) {
+      //eslint-disable-next-line
+      console.log("init failed: " + err);
+    }
+
+    cordova.openwith.init(initSuccess, initError)
+    cordova.openwith.addHandler(this.openWithHandler)
+  }
+  //We write the items in localStorage and then push
+  //a to a specific route
```
y-lohse:
```suggestion
  //to a specific route
```

src/drive/web/modules/upload/UploadQueue.jsx
```diff
@@ -151,14 +151,16 @@ export class UploadQueue extends Component {
...
+            {queue.map((item, index) => {
+              return (
+                <Item
+                  key={`key_queue_${index}`}
+                  file={item.file}
+                  isDirectory={item.isDirectory}
+                  status={item.status}
+                />
+              )
+            })}
```
y-lohse: I don't think we need this change :)

src/drive/web/modules/upload/index.js
```diff
@@ -236,6 +240,32 @@ export const overwriteFile = async (client, file, path) => {
...
+export const uploadFilesFromNative = (
+  files,
+  folderId,
+  uploadFilesSuccessCallback
+) => async dispatch => {
+  dispatch({
+    type: ADD_TO_UPLOAD_QUEUE,
+    files: files
+  })
+  //!TODO Promise.All
```
y-lohse: Any reason not to do it now?

src/drive/targets/mobile/config.xml
```diff
@@ -17,6 +17,7 @@
         <preference name="android-minSdkVersion" value="21" />
         <preference name="StatusBarBackgroundColor" value="#D6D8DA" />
         <preference name="AndroidPersistentFileLocation" value="Compatibility" />
+        <preference name="AndroidLaunchMode" value="singleTask" />
```
y-lohse: Is there a chance this breaks something? Like opening the app from banks?

src/drive/mobile/modules/upload/index.jsx
```diff
@@ -0,0 +1,166 @@
...
+  state = {
+    items: null,
+    folderId: ROOT_DIR_ID
+  }
+  async componentDidMount() {
+    const items = await localforage.getItem('importedFiles')
+    this.setState({ items })
+  }
+
+  async uploadAndState() {
```
y-lohse: I don't understand the `AndState` in the name, what does it mean?

src/drive/mobile/modules/upload/index.jsx
```diff
@@ -0,0 +1,166 @@
...
+    const { items, folderId } = this.state
+    const { router, uploadFilesFromNative } = this.props
+    const filesForQueue = generateForQueue(items)
+    uploadFilesFromNative(filesForQueue, folderId, this.callbackSuccess)
+    setTimeout(() => router.push(`/folder/${folderId}`), 500)
```
y-lohse: Why the arbitrary 500ms timeout?

src/drive/mobile/modules/upload/index.jsx
```diff
@@ -0,0 +1,166 @@
...
+      })
+      .indexFields(['dir_id', 'type', 'name', '_id'])
+      .sortBy([{ dir_id: 'asc' }, { type: 'asc' }, { name: 'asc' }])
+  }
+  breadcrumbQuery = client => {
+    return client.get('io.cozy.files', this.state.folderId)
+  }
+  onClose = () => {
+    const { router } = this.props
+    router.push('/')
```
y-lohse: Is there a point of doing `localforage.removeItem('importedFiles')` here as well?

src/drive/mobile/modules/upload/index.jsx
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  render() {
+    const { items, folderId, uploadInProgress } = this.state
+    const { t } = this.props
+    if (items === null) return null
```
y-lohse: Wouldn't it make more sense to start rendering the rest of the modal, even while the `items` aren't ready? That way we can start the queries earlier. And we can delay showing the file list until the items are ready for example.

src/drive/web/modules/upload/index.js
Crash--: I didn't produce any test with 10 concurrent uploads. I don't how it'll react on a slow or shitty network. 

I'll add a note about it in the todo. I think this should be done when working on our upload system one day ;)

src/drive/targets/mobile/config.xml
Crash--: Nop. I have added a description for this specific commit.

https://github.com/cozy/cozy-drive/pull/1918/commits/80f0aa9cbcb5d163bb9f12bb54a7406fa0459588

src/drive/mobile/modules/upload/index.jsx
Crash--: Me neither. At first I had to setter a state maybe...

src/drive/mobile/modules/upload/index.jsx
Crash--: We can discuss about the 500, and I think 50 is ok. Will change it.

The setTimeout() is here to be sure that the first dispatch was done. I think that during my tests I had a concurrent action resulting in errors... Will add a comment

src/drive/mobile/modules/upload/index.jsx
Crash--: Yes! If app exited before the end of the upload, next time the user share something to cozy-drive we'll want to upload previous no imported files. And this is not what we want since path or else can have changed. Also this makes the functionally more resilient if something goes wrong.

src/drive/mobile/modules/upload/index.jsx
Crash--: Good point!

src/drive/mobile/modules/upload/index.jsx
Crash--: You absolutely right!

src/drive/mobile/modules/upload/index.jsx
ptbrowne: (Comment or `const NO_CONCURRENT_ACTION_DELAY = 500`, or both ?)

src/drive/web/modules/upload/index.js
```diff
@@ -236,6 +240,32 @@ export const overwriteFile = async (client, file, path) => {
...
+  for (var i = 0; i < files.length; ++i) {
```
y-lohse: Nit:

```suggestion
  for (const i = 0; i < files.length; ++i) {
```

`var` is so old school 😄

PR #1919, opened 2019-12-06
fix: Replace for back button on Android

Crash--:
We need to use history.replace() instead of history.push() since we can go back on Android with the back button. And in that case, we go back to the ImportModal. This is not what we want.

Thanks @Gara64 for the bug report ;)


Gara64:
:tada:

PR #1926, opened 2019-12-12
fix: setStore order before bar.init

Crash--:
Since cozy-bar now use cozy-client, we need to setStore before launching the bar init.


Crash--:
Closed since it's in 1.18.23 ;)

PR #1928, opened 2019-12-17
feat: Don't open Note on Notes if not mine

Crash--:
Since Notes app doesn't handle sharing at the moment, we don't want top open the note in Notes if the file is note created on my cozy (aka the file is shared with me). If the file is shared, then we display the markdown in the viewer.

### DIFF
--- a/src/drive/web/modules/drive/files.js
+++ b/src/drive/web/modules/drive/files.js
@@ -1,4 +1,6 @@
 //!TODO Put this to cozy-client/models/files
+import get from 'lodash/get'
+
 const FILE_TYPE = 'file'
 const DIR_TYPE = 'directory'
 export const ALBUMS_DOCTYPE = 'io.cozy.photos.albums'
@@ -20,4 +22,15 @@ export const isReferencedByAlbum = file => {
     }
   }
   return false
+}
+/**
+ *
+ * @param {object} file io.cozy.files
+ * @param {object} client CozyClient instance
+ * @return boolean If the file was created on my cozy
+ */
+export const isNoteMine = (file, client) => {
+  const myCozyUrl = client.getStackClient().uri
+  const noteCreatedOnCozy = get(file, 'cozyMetadata.createdOn')
+  return noteCreatedOnCozy.startsWith(myCozyUrl)
 }
--- a/src/drive/web/modules/navigation/FileExplorer.jsx
+++ b/src/drive/web/modules/navigation/FileExplorer.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux'
 import { withRouter } from 'react-router'
 import { compose } from 'redux'
 
-import { queryConnect, models } from 'cozy-client'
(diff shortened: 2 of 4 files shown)


src/drive/web/modules/drive/files.js
```diff
@@ -21,3 +23,14 @@ export const isReferencedByAlbum = file => {
...
+/**
+ *
+ * @param {object} file io.cozy.files
+ * @param {object} client CozyClient instance
+ * @return boolean If the file was created on my cozy
+ */
+export const isNoteMine = (file, client) => {
+  const myCozyUrl = client.getStackClient().uri
+  const noteCreatedOnCozy = get(file, 'cozyMetadata.createdOn')
+  return noteCreatedOnCozy.startsWith(myCozyUrl)
```
y-lohse: A bit hacky but ok :) I guess it should be possible to see in the sharing data who the creator is.

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -176,11 +178,19 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
...
+  }
   if (availableOffline) {
     return dispatch(openLocalFile(file))
   }
 
   if (models.file.isNote(file)) {
+    //This is a tmp hack since Notes doesn't handle sharing properly.
+    if (!isNoteMine(file, props.client)) {
+      return redirectToViewer()
+    }
```
y-lohse: There are 3 calls to `redirectToViewer ()` and the `return` is a bit misleading, since it's only there to stop the function execution.

It could be rewritten as

```
const isNote = models.file.isNote(file)
const isMyNote = isNote && isNoteMine(file, props.client)

if (isMyNote) { … } else { redirectToViewer() }
```

src/drive/web/modules/drive/files.js
y-lohse: That said, `createdOn` isn't a standard metadata field? https://github.com/cozy/cozy-doctypes/tree/master/docs#document-metadata Are you sure it exists?

Crash--: https://github.com/cozy/cozy-doctypes/pull/125 ;)

Crash--: >A bit hacky but ok :) I guess it should be possible to see in the sharing data who the creator is.

Definitely hacky, will be changed in the next few days I think ;)

Crash--: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md#cozymetadata

It's specific for file

src/drive/web/modules/navigation/FileExplorer.jsx
Crash--: Thanks for the suggestion. Need to call `redirectToViewer` twice but much more cleaner :)

PR #1940, opened 2020-01-16
Feat/cozy sharings

Crash--:
Use new version of cozy-sharing and remove old code

We had a bug when opening "Sharing" view before. We displayed "no sharing" because we received an empty sharring array from the consumer. We use the new `hasLoadedAtLeastOnePage` props to know if at least one page (pagination) has been loaded. We can now make the difference between empty sharing and not yet loaded sharing


src/drive/targets/public/index.jsx
```diff
@@ -16,7 +16,7 @@ import LightFolderView from 'drive/web/modules/public/LightFolderView'
 import LightFileViewer from 'drive/web/modules/public/LightFileViewer'
 import ErrorShare from 'components/Error/ErrorShare'
 import { configureReporter, setCozyUrl } from 'drive/lib/reporter'
-import getSharedDocument from 'sharing/getSharedDocument'
+import getSharedDocument from 'cozy-sharing/dist/getSharedDocument'
```
y-lohse: Ideally we should export `getSharedDocument ` from the index, it avoids assuming things about the file structure of the module.

src/drive/web/modules/navigation/AppRoute.jsx
```diff
@@ -18,6 +18,9 @@ import {
 import { Container as Trash } from 'drive/web/modules/trash'
 import FileHistory from '../../../../components/FileHistory'
 import UploadFromMobile from 'drive/mobile/modules/upload'
+
+import 'cozy-sharing/dist/stylesheet.css'
```
y-lohse: Should we import this with the other css files?

Crash--: I'm wondering if we shouldn't move the imports to this file. Currently we're duplicating import in targets/browser/index, targets/mobile/index, public/index. Let me see if I create a new commit to do that ;)

Crash--: Done in https://github.com/cozy/cozy-drive/pull/1940/commits/1b5e09da32733948bc92bf538269d6110c498db8

PR #1942, opened 2020-01-21
fix: Create StyledApp component to inject CSS and style

Crash--:
In a previous commit, I was importing CSS in `AppRoute` component. But this component is not loaded when we are in a public context. I created a `StyledApp` which only import styles / css

### DIFF
--- a/src/drive/web/modules/drive/StyledApp.jsx
+++ b/src/drive/web/modules/drive/StyledApp.jsx
@@ -1,0 +1,7 @@
+import 'cozy-ui/transpiled/react/stylesheet.css'
+import 'cozy-sharing/dist/stylesheet.css'
+//eslint-disable-next-line
+import mainStyles from 'drive/styles/main.styl'
+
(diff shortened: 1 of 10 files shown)


src/drive/web/modules/drive/StyledApp.jsx
```diff
@@ -0,0 +1,7 @@
+import 'cozy-ui/transpiled/react/stylesheet.css'
+import 'cozy-sharing/dist/stylesheet.css'
+//eslint-disable-next-line
+import mainStyles from 'drive/styles/main.styl'
+
+const StyleApp = ({ children }) => children
+export default StyleApp
```
y-lohse: Do we really need the component here? I would think the imports are enough no?

Crash--: Yep, but I plan to have a centralized way to handle the bootstrap of the public / private app. It'll be in this file (after a rename) ;)

PR #1944, opened 2020-01-22
refactor: Use Q instead of client.all

Crash--:


src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -219,12 +219,11 @@ export const handleFileOpen = (file, availableOffline, props, dispatch) => {
   }
 }
 
-const appsQuery = client => client.all('io.cozy.apps')
 export default compose(
   translate(),
   queryConnect({
     apps: {
-      query: appsQuery,
+      query: Q('io.cozy.apps'),
```
ptbrowne: I think you have to do `() => Q('io.cozy.apps')`

Crash--: Hmm it seems to work without  🤔

ptbrowne: Ah oui non c'est OK https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/Query.jsx#L28-L29

ptbrowne:
LGTM

PR #1945, opened 2020-01-28
Release/1.19.0

Crash--:
Needs an update of cozy-sharing

### DIFF
--- a/src/drive/web/modules/move/MoveModal.jsx
+++ b/src/drive/web/modules/move/MoveModal.jsx
@@ -1,17 +1,22 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { compose } from 'redux'
+import { connect } from 'react-redux'
+
 import { Modal } from 'cozy-ui/transpiled/react'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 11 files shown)


src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -201,13 +207,17 @@ export class MoveModal extends React.Component {
...
-          onConfirm={this.moveEntries}
-          onClose={onClose}
-          targets={entries}
-          currentDirId={folderId}
-          isMoving={isMoveInProgress}
-        />
+        <RefreshableSharings>
+          {({ refresh }) => (
+            <Footer
+              onConfirm={() => this.moveEntries(refresh)}
```
y-lohse: Have you considered something like `onConfirm={async () => { await this.moveEntries(); refresh()}}`? That way `moveEntries` doesn't have to know about a callback.

Crash--: I think your proposal is so much better. Thanks!

Crash--: In fact we can't since we need also to call the callback after a cancel...

PR #1955, opened 2020-02-19
Feat/open notes

Crash--:
- When we click on a Notes file, we call the stack to build the URL for the Notes app 
- If we click on a sharing link and this sharing only shared a Notes, let's do the same


src/drive/web/modules/drive/files.js
```diff
@@ -24,13 +24,29 @@ export const isReferencedByAlbum = file => {
...
-  const myCozyUrl = client.getStackClient().uri
-  const noteCreatedOnCozy = get(file, 'cozyMetadata.createdOn')
-  return noteCreatedOnCozy.startsWith(myCozyUrl)
+export const fetchUrlToOpenANote = async (client, file) => {
+  const {
+    data: { note_id, subdomain, protocol, instance, sharecode, public_name }
+  } = await client
+    .getStackClient()
+    .collection('io.cozy.notes')
+    .fetchURL({ _id: file.id })
```
y-lohse: No way to use `client.query` here?

src/drive/web/modules/navigation/FileExplorer.jsx
```diff
@@ -187,32 +190,12 @@ export const handleFileOpen = (file, availableOffline, props, dispatch) => {
...
-    } else {
-      //If something went wrong previoulsy like the request to /apps has failled
-      //we display the note within the viewer
-      redirectToViewer()
+  const { client } = props
+  if (isNote) {
+    try {
+      window.location.href = await fetchUrlToOpenANote(client, file)
+    } catch (e) {
+      Alerter.error('alert.offline')
```
y-lohse: We don't want to redirect to the viewer in this case?

Crash--: Not yet, next PR. I've to update the Viewer to display a few lines on it. And for that I need to upgrade UI -> XP modal and so on :p

Crash--:
> Cool! Testing the output of fetchUrlToOpenANote might be a good addition to this PR :) 

I'm currently adding this method to Cozy-Client and tests will be done there.

src/drive/web/modules/drive/files.js
Crash--: Will fix it in cozy-client, thanks :)

PR #1956, opened 2020-02-20
feat: Add a button to create a note from Drive

Crash--:
Add a button in the MoreMenuAction on Drive to create a Note. 

- [x] Need an upgrade of Cozy-Client (https://github.com/cozy/cozy-client/pull/647/files) 
- [x] Upgrade FR transifex 
- [x] Waiting for the SVG


src/drive/web/modules/drive/Toolbar/components/useAppLinkWithStoreFallBack.jsx
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+  return {
+    fetchStatus,
+    isInstalled,
+    url
+  }
+}
+
+export default useAppLinkWithStoreFallback
```
y-lohse: Reminds me of something 🤓

package.json
```diff
@@ -112,6 +112,7 @@
     "cozy-scanner": "0.2.0",
     "cozy-scripts": "1.13.2",
     "cozy-sharing": "^1.8.0",
+    "cozy-stack-client": "11.2.0",
```
y-lohse: This is not a dev dep, right? We didn't add it only for tests?

The main problem I see with this is that we now need to upgrade both packages manually :(

src/drive/web/modules/drive/Toolbar/components/useAppLinkWithStoreFallBack.jsx
Crash--: Let's me add it in Cozy-Client ;).

package.json
Crash--: Nope. This is not for the test :/

src/drive/web/modules/drive/Toolbar/components/useAppLinkWithStoreFallBack.jsx
Crash--: https://github.com/cozy/cozy-client/pull/651

PR #1957, opened 2020-02-26
fix: Generate the return url from universallink helper

Crash--:
Repeat after me: don't use window.location.href to get access to the current url. It doesn't work on mobile 🤦‍♂ (file:///....../index.html). 

Instead, use the UI helper to generate the correct Universal Link 

Toolbar Menu: we've done some rework of the Toolbar item to create some "features groups". First one : creation (new note, new folder, upload files, scan file). Second : Folder's action : download / share / select items. Third: Deletion


src/drive/web/modules/drive/Toolbar.jsx
```diff
@@ -87,6 +82,11 @@ class Toolbar extends Component {
             </Item>
           )}
         <hr />
+        <NotRootFolder>
+          <Item>
+            <ShareItem />
+          </Item>
+        </NotRootFolder>
```
ptbrowne: Why this reordering ? Is it intended ?

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
```diff
@@ -1,16 +1,31 @@
...
 import styles from 'drive/styles/toolbar.styl'
-import { withClient, models, useAppLinkWithStoreFallback } from 'cozy-client'
+import useCapabilities from 'lib/hooks/useCapabilities'
 
 const CreateNoteItem = ({ client, t, displayedFolder }) => {
   const { fetchStatus, url, isInstalled } = useAppLinkWithStoreFallback(
     'notes',
     client
   )
+  const capabilities = useCapabilities(client)
```
ptbrowne: Friendly reminder that useCapabilities does not use a fetchPolicy meaning that we fetch the settings every time a CreateNoteItem is displayed.

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
```diff
@@ -1,16 +1,31 @@
...
+    capabilities,
+    'capabilities.attributes.flat_subdomains'
+  )
 
+  const returnUrl = generateUniversalLink({
+    slug: 'drive',
+    cozyUrl: client.getStackClient().uri,
+    subDomainType: isFlatDomain ? 'flat' : 'nested',
+    nativePath: `/files/${displayedFolder ? displayedFolder.id : null}`
+  })
```
ptbrowne: Could you add a test for this 😗 ?

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
Crash--: It's displayed all the time, so only one request or one more if the client change ;). 

I'm wondering if I should move this hook to cozy-client in order to not repeat that everywhere and also if one day we have a useQuery() with fetch policy we could be able to update all our app with that :)

src/drive/web/modules/drive/Toolbar.jsx
Crash--: Yep. PR's description updated

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
ptbrowne: I think it has its place in cozy-client, yes :)

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
```diff
@@ -25,7 +40,7 @@ const CreateNoteItem = ({ client, t, displayedFolder }) => {
           window.location.href = await models.note.generatePrivateUrl(
```
y-lohse: Is it ok to use `window.location.href` for the redirection? Shouldn't we use `open(url)` or something similar?

Crash--: Yep, it's ok. Our apps are not configured to handle this kind of opening, so it opens the browser.

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
Crash--: I'm not convinced by this test, I just write a defensive code, and I don't know if it worth it

Crash--: If you really want this test, I'll try to take time to write few others for the toolbar.

PR #1963, opened 2020-03-06
Feat/shortcut

Crash--:
Lot of small stuff in there! 

- First, I tried to remove the import from the root of UI. We should avoid them since tree shaking can't be done efficiently 
- Cozy-UI upgrades so I upgrade cozy-scanner and history modal to work with the new ExperimentalDialog 

For the shortcuts: 
- Created a new page to redirect to make the redirection. This was the only way (to use a href) to open a new tab on browser. For mobile user, we don't need to open a new tab, we just need to redirect the location.href
- Added a new FileOpener component that handles the opening of a file. Currently other than shortcut is still using the old way. But the idea is to have the opening actions in that file. 

Still needs: https://github.com/cozy/cozy-ui/pull/1401


src/drive/web/modules/filelist/generateFileUrl.js
```diff
@@ -0,0 +1,27 @@
+import { generateWebLink } from 'cozy-client'
+
+export const generateFileUrl = ({ file, client, isFlatDomain }) => {
```
y-lohse: Maybe it should be `generateShortcutUrl ` since it only handle that kind of files?

src/drive/web/modules/navigation/ExternalRedirect.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { withClient } from 'cozy-client'
+import useFetchShortcut from 'drive/web/modules/filelist/useFetchShortcut'
+import { Empty } from 'cozy-ui/transpiled/react'
+import { IconSprite } from 'cozy-ui/transpiled/react'
```
y-lohse: Nit: those 2 are also imported from the index :)

src/drive/web/modules/filelist/FileOpener.jsx
```diff
@@ -0,0 +1,79 @@
...
+      <a
+        href={fileUrl}
+        rel="noopener noreferrer"
+        target="_blank"
+        style={{
+          width: '100%',
+          display: 'contents',
+          textDecoration: 'none',
+          color: 'var(--coolGrey)',
+          alignItems: 'center'
```
y-lohse: That's… quite a bit of inline styles 😁

PR #1966, opened 2020-03-10
fix: Since the use of cozyclient hook, data is in data...

Crash--:
Title is self explaining I think... ^^

### DIFF
--- a/src/components/FileHistory/HistoryModal.jsx
+++ b/src/components/FileHistory/HistoryModal.jsx
@@ -39,7 +39,7 @@ const HistoryModal = ({
   const capabilities = useCapabilities(client)
   const isFileVersioningEnabled = get(
     capabilities,
-    'capabilities.attributes.file_versioning'
+    'capabilities.data.attributes.file_versioning'
   )
   return (
     <ExperimentalDialog onClose={router.goBack}>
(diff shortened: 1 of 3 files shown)


ptbrowne:
Mmh, I think it would be better if it was done directly through the hook, the hook should hide this difference, and it would be easier for the user not to deal with the data field.

Crash--:
> Mmh, I think it would be better if it was done directly through the hook, the hook should hide this difference, and it would be easier for the user not to deal with the data field.

 I agree with you in the sense that we need to have something somewhere to hide this complexicity. 
I'm not sure the hooks is the right place to do it. The previous idea behind our hooks was to be compatible with the query / useQuery API and their returns. 

So two solutions: 
- We normalize in the hooks the response. App has nothing to do just const { isFlat } = useCapabilities()
- We add a model to do that. App has to do : const capabilities = useCapabilities(); const {isFlat} = model.capabilities.normalize(capabilities)

Crash--:
> Mmh, I think it would be better if it was done directly through the hook, the hook should hide this difference, and it would be easier for the user not to deal with the data field.

 I agree with you in the sense that we need to have something somewhere to hide this complexicity. 
I'm not sure the hooks is the right place to do it. The previous idea behind our hooks was to be compatible with the query / useQuery API and their returns. 

So two solutions: 
- We normalize in the hooks the response. App has nothing to do just const { isFlat } = useCapabilities()
- We add a model to do that. App has to do : const capabilities = useCapabilities(); const {isFlat} = model.capabilities.normalize(capabilities)

PR #1970, opened 2020-03-16
Feat/rename input

Crash--:
3 things in this PR : 
- First we select the filename when we rename a file 
- Second, we disable the FileOpener when renaiming 
- Third, we display a confirmation Modal if the extension has changed

### DIFF
--- a/src/drive/web/modules/filelist/FilenameInput.jsx
+++ b/src/drive/web/modules/filelist/FilenameInput.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
+import { CozyFile } from 'cozy-doctypes'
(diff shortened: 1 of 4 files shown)


src/drive/web/modules/filelist/FilenameInput.jsx
```diff
@@ -65,22 +63,30 @@ export default class FilenameInput extends Component {
...
   }
-
+  handleFocus() {
+    const { name } = this.props
+    const { filename } = CozyFile.splitFilename({ name, type: 'file' })
+    //Since we're mounting the component and focusing it at the same time
+    // let's add a small timeout to be sure the ref is populated
+    setTimeout(() => {
+      this.textInput.current.setSelectionRange(0, filename.length)
+    }, 5)
```
y-lohse: This is a bit meh, 5ms is really arbitrary. `useEffect` is cool for things like that but would require "unclassing" the component.

PR #1979, opened 2020-03-19
Cozy client migration

y-lohse:
We would like to start testing the new view using cozy-client, so I'd like to merge this. Most of the code is still a work in progress so I'm not sure how much sense a review makes at this point. Everything new is behind a flag, the old routes should be untouched.


src/drive/lib/ModalContext.jsx
```diff
@@ -0,0 +1,28 @@
...
+    </ModalContext.Provider>
+  )
+}
+
+export const ModalStack = () => {
+  const { modalStack } = useContext(ModalContext)
+
+  if (modalStack.length === 0) return null
+  else return modalStack[modalStack.length - 1]
+}
```
ptbrowne, line 28: Tiens, ca ressemble à la ModalStack dans cozy-ui non ?

y-lohse: Oui, je me suis dit la meme chose en l'écrivant. Il y a surement quelque chose à mutualiser a ce niveau la.

y-lohse:
This PR is ready for review. Just to be clear, not everything in there is finished, but we would like to merge what is already there so we can work on the same codebase and so that other people can start testing the new view. So the review is more about identifying any red flags than a defitinitive code review.

- Changes to existing files should be fairly small: it's either splitting up components into multiple files, or exporting tweaked versions of existing components. There should be no breaking changes in existing files.
- There are new fIles for the view using cozy-client. Most of them are in `src/drive/web/view/Drive`. I think it's a good time for feedback about the general architecture. On the other hand, this code still needs tests, and may be moved around (for example all the pure logic code in `useActions.jsx`) in future PRs.

ptbrowne:
Thanks for all the explanations in video. LGTM and we continue to work on this on other PRs.

PR #1982, opened 2020-03-23
fix: Dropzone used a no longer existent icon

ptbrowne:
Icon was renamed to cloud a long time ago


ptbrowne:
Fixed, I think we could add an explicit "b2b" mention in those files.

Crash--:
latest commit for this file was "fix: Better Personalization" 🤷‍♂

PR #1984, opened 2020-03-24
Fix/toolbar alignement

Crash--:
- Fix issue with the copy / past emails address 
- Fix alignement see screenshots
Capture d’écran 2020-03-24 à 08 14 53
Capture d’écran 2020-03-24 à 08 15 00
Capture d’écran 2020-03-24 à 08 15 12

### DIFF
--- a/src/drive/styles/toolbar.styl
+++ b/src/drive/styles/toolbar.styl
@@ -55,20 +55,18 @@
             color var(--pomegranate)
             background embedurl('../assets/icons/icon-trash-red.svg') 1rem center no-repeat
         &.fil-action-share
+            position relative
             background embedurl('../assets/icons/icon-share-dark.svg') 1rem center no-repeat
         &.fil-action-create-note
             background embedurl('../assets/icons/icons-16-icon-note.svg') 1rem center no-repeat
         &.fil-action-create-shortcut
             background embedurl('../assets/icons/icon-link.svg') 1rem center no-repeat
-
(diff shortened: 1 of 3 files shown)


src/drive/styles/toolbar.styl
```diff
@@ -92,7 +91,9 @@
 
     .fil-public-download
         display none
-
+    .fil-toolbar-menu
+        .fil-toolbar-menu-recipients
+            display inline-flex !important
```
ptbrowne: Why is it needed ?

Crash--: because I've a conflict with the previous media query (@media (max-width: 63.938rem)). Its style are loaded before the small-screen() and I've a display none at the end :/ 

I can play with the declaration order, I'm pretty just it will work but... 

Is there a way in UI to have a media query between max-width: 63 & min-width > small-screen?

Crash--: Capture d’écran 2020-03-24 à 09 26 17

y-lohse: I think [tablet()](https://github.com/cozy/cozy-ui/blob/master/stylus/settings/breakpoints.styl#L193) is what you are looking for?

Crash--: medium-device() doest the trick

Crash--: tablet is exactly what I needed thanks!

PR #1985, opened 2020-03-25
Fix/real time plugin

Crash--:
before: 
Capture d’écran 2020-03-25 à 07 50 47
after:
Capture d’écran 2020-03-25 à 07 54 40

Not changed in RealTimeFile since it needs to rewrite a test

### DIFF
--- a/src/components/Image/ImageLoader.jsx
+++ b/src/components/Image/ImageLoader.jsx
@@ -25,7 +25,7 @@ class ImageLoader extends React.Component {
     const { client } = this.props
     this._mounted = true
     this.loadNextSrc()
-    this.realtime = client.plugins.realtime.realtime
+    this.realtime = client.plugins.realtime
     this.type = 'io.cozy.files.thumbnails'
     this.realtime.subscribe('created', this.type, this.handleCreate)
   }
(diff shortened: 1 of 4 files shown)


src/components/Image/ImageLoader.jsx
```diff
@@ -26,7 +25,7 @@ class ImageLoader extends React.Component {
     const { client } = this.props
     this._mounted = true
     this.loadNextSrc()
-    this.realtime = new Realtime({ client })
+    this.realtime = client.plugins.realtime.realtime
```
ptbrowne: We should use directly the instance of the plugin.

`client.plugins.realtime`

The instance of the plugin proxies to the CozyRealtime instance.
See https://github.com/cozy/cozy-notes/commit/6f0f92d833e014d06f125ba52ff3a7fc0bacac33#r38037516 for more context.

This is especially important in situations where we login/logout and keep the same client since we have to re-instantiate the CozyRealtime instance. THis is not done yet in the plugin but should be.

When CozyRealtime is reinstantiated, we'll have a new instance in `realtime.realtime`, but we'll keep a reference to the old instance here.

PR #2006, opened 2020-04-27
Fix/scanner2

Crash--:
- Before Scanner was using the old `ExperimentalModal` that was calling MuiCozyTheme by itself. Now that this behavior is removed on the UI's side, let's call `MuiCozyTheme` from Drive since this content is displayed in the Cozy Bar. (see https://github.com/cozy/cozy-bar/issues/683#issuecomment-620020307)

- Since cozy-scanner use `lang` to use the right translation file, we need to pass it to the barcontext provider

- UploadQueue was reworked recently let's adapt the code to it

### DIFF
--- a/src/drive/web/modules/drive/Toolbar/components/PortaledQueue.jsx
+++ b/src/drive/web/modules/drive/Toolbar/components/PortaledQueue.jsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import ReactDOM from 'react-dom'
-import { UploadQueue } from 'drive/web/modules/upload/UploadQueue'
+import { DumbUploadQueue as UploadQueue } from 'drive/web/modules/upload/UploadQueue'
 import { translate } from 'cozy-ui/transpiled/react'
 class PortaledQueue extends Component {
   render() {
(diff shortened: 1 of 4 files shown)


src/drive/web/modules/drive/Toolbar/components/PortaledQueue.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import ReactDOM from 'react-dom'
-import { UploadQueue } from 'drive/web/modules/upload/UploadQueue'
```
ptbrowne, line 3: I am not sure what's happening here, did it break before ? Does this needs a test ?

Crash--: It works before, but just launching webpack tells us 

```
/src/drive/web/modules/drive/Toolbar/components/PortaledQueue.jsx 39:55-66
"export 'UploadQueue' was not found in 'drive/web/modules/upload/UploadQueue' 
```

Crash--:
I will spend a few hours on the bar and fix the context issue and add tests over there. After that let's see if I can add test for the scanner's integration

PR #2017, opened 2020-05-12
Read-write actions on public view

y-lohse:
With this PR, the public sharing view supports actions that require write permissions (such as renaming files, uploading, etc).

In order to re-use the menu items that we have in the protected pages, I migrated them all to use ActionMenu from cozy-ui. This is also why the PR is fairly big, apologies for that.


src/drive/web/modules/public/Actions.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
```
y-lohse, line 1: Most of this file is lifted from https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/drive/Container.jsx — I didn't spend too much time on it because it's been rewritten in #1979

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
```diff
@@ -8,21 +8,37 @@ import {
...
   )
+
+  let notesAppUrl = ''
+  const { fetchStatus, url, isInstalled } = useAppLinkWithStoreFallback(
+    'notes',
+    client
+  )
+  if (fetchStatus === 'loaded') {
+    notesAppUrl = url
+  } else if (fetchStatus === 'errored') {
```
y-lohse, line 34: I don't know if the commit message is explicit enough, but on the public page we don't have permissions to query the list of apps, and this is one of the only places where we really need it. I think the fallback to a generated link is ok, the only issue is if the notes app is not installed, then we won't have a link to the store.

src/drive/web/modules/drive/Toolbar/components/UploadButtonItem.jsx
```diff
@@ -0,0 +1,22 @@
...
+    disabled={isDisabled}
+    displayedFolder={displayedFolder}
+    label={t('toolbar.item_upload')}
+    className={classNames(styles['c-btn'], 'u-hide--mob')}
+  />
+)
+
+export default compose(
+  toolbarContainer,
+  translate()
```
ptbrowne, line 21: Nit: we could use useI18n at some point. This can be done via a codemod.

src/drive/web/modules/drive/Toolbar/components/ScanWrapper.jsx
```diff
@@ -60,66 +60,69 @@ const ScanMenuItem = translate()(({ status, onClick, t, online }) => {
...
-            const tracker = getTracker()
-            if (tracker) {
-              tracker.push(['trackEvent', 'Drive', 'Scanner', 'Finished'])
-            }
-            startMediaBackup()
-          }}
-        >
-          {({ status, error, startScanner, filename, onClear, online }) => {
-            if (error || !filename) {
+        <div onClick={e => e.stopPropagation()}>
```
ptbrowne, line 66: Nit: The ScanWrapper is renderered as part of the ActionMenu hierarchy ? If so, I understand the solution here, but I think it would be cleaner to have it outside from the ActionMenu.

ptbrowne:
Ah I forgot, when on a PR I now always check to see if some spec files have been changed 😈 😇 , and for such a big change, I am a bit surprised to see that none have been changed. Is there a particular reason for this ?

I find it also difficult to write test files, particularly for UI stuff. I am more and more convinced that Argos would be good to have on our apps.

src/drive/web/modules/public/LightFolderView.jsx
```diff
@@ -5,6 +5,7 @@ import { withRouter } from 'react-router'
 import { Content, Overlay } from 'cozy-ui/transpiled/react'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
 import { models, withClient } from 'cozy-client'
+import { ModalManager } from 'react-cozy-helpers'
```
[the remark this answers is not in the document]
Crash--: https://github.com/cozy/cozy-drive/blob/master/app.config.js#L46 🤷‍♂️

.travis.yml
```diff
@@ -89,9 +89,11 @@ jobs:
         - yes | sdkmanager "platforms;android-27" # see https://github.com/travis-ci/travis-ci/issues/8874#issuecomment-350350607
         - scripts/decrypt.sh
         - scripts/signing.sh
-        - curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
+        - curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
```
y-lohse, line 92: Without this change, travis was installing node 10.14.0, but one of our deps (pretty-format) requires >= 1.14.2. I did not find a way to use a different 10;x version, so switched to the nest LTE. In theory this has no consequence but I don't like that we have to use 2 different versions of node. Should we also upgrade the web version?

src/drive/web/modules/drive/Toolbar/components/UploadButtonItem.jsx
y-lohse: Yep 👍 I will do this in a separate PR to avoid bloating this one, it's already big 😅

src/drive/web/modules/drive/Toolbar/components/ScanWrapper.jsx
y-lohse: >  The ScanWrapper is renderered as part of the ActionMenu hierarchy ? 

Yes, and I agree, it shouldn't be rendered elsewhere. I'd rather do it in a separate PR though, because changing this is more complicated than `ShortcutItem` for example, due to the API of `<Scanner />`.

src/drive/targets/public/index.jsx
```diff
@@ -107,13 +109,20 @@ const init = async () => {
...
+                    <Route
+                      path="file/:fileId/revision"
+                      component={FileHistory}
+                    />
+                  </Route>
+                </Route>
+                <Route path="external/:fileId" component={ExternalRedirect} />
+                <Redirect from="/*" to={`files/${sharedDocumentId}`} />
+              </Router>
+            </SharingProvider>
```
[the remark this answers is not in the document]
y-lohse: Yep, good point, changed it.

y-lohse:
> Ah I forgot, when on a PR I now always check to see if some spec files have been changed 😈 😇 , 
and for such a big change, I am a bit surprised to see that none have been changed. Is there a particular reason for this ?
>
> I find it also difficult to write test files, particularly for UI stuff. I am more and more convinced that Argos would be good to have on our apps.

Yeah the changes were mostly UI related, to screenshots would make more sense here. But I'll add some :)

.travis.yml
y-lohse: > I don't know why we don't have to upgrade the web version too?

The web version uses a node image from travis, which installs node 10.20.0 and satisfies the requirement from pretty-format.
[Here's the PR](https://github.com/facebook/jest/pull/9423) — the change is from 2 weeks ago.

> Also which package update is responsible for the pretty-format update?

I can check this, but it's a jest package. I don't like our chances of being able to downgrade it.

We use [the recommended way to install node on debian](https://github.com/nodesource/distributions/blob/master/README.md) but I don't see an option to change the version.
We could also switch to node 12 for the web project?

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
y-lohse: > What happens if click on the button with no network?

It will try to open notes :/. 

> Can't we do a check if we're in a public context or not?

Yep, we can! The offline behavior would be the same though. But it's probably a better solution.

> (and no the commit was not self explaining, maybe you can add your comment to it ;))

👍

src/drive/web/modules/drive/Toolbar/components/MoreMenu.jsx
```diff
@@ -0,0 +1,76 @@
...
+  breakpoints: { isMobile }
+}) => {
+  const [menuIsVisible, setMenuVisible] = useState(false)
+  const anchorRef = React.createRef()
+
+  const openMenu = useCallback(() => setMenuVisible(true))
+  const closeMenu = useCallback(() => setMenuVisible(false))
+
+  return (
+    <div>
```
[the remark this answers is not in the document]
y-lohse: Nope, the ActionMenu uses it's container for positioning.

src/drive/web/modules/drive/Toolbar/components/UploadItem.jsx
```diff
@@ -1,27 +1,42 @@
 import React from 'react'
-import classNames from 'classnames'
-import UploadButton from 'drive/web/modules/upload/UploadButton'
```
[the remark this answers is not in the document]
y-lohse: No :( It is still used outside of the menu, for example on the root folder.

.travis.yml
Crash--: > The web version uses a node image from travis, which installs node 10.20.0 and satisfies the requirement from pretty-format.

Why not using the same thing so? 

I don't think we want to upgrade to node 12 yet.

src/drive/web/modules/drive/Toolbar/delete/DeleteItem.jsx
```diff
@@ -1,28 +1,29 @@
...
 import { translate } from 'cozy-ui/transpiled/react/I18n'
+import { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu'
+import Icon from 'cozy-ui/transpiled/react/Icon'
 import deleteContainer from './delete'
+
 const DeleteItem = translate()(
   ({ t, isSharedWithMe, displayedFolder, trashFolder, onLeave }) =>
     isSharedWithMe ? (
-      <a
+      <ActionMenuItem
```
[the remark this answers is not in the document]
y-lohse: Not at the moment no. The ActionMenu doesn't inject props on it's children, but we could add support for ActionMenuItems themes in the future.

.travis.yml
y-lohse: > Why not using the same thing so?

We're using the `android` image of travis at the moment… which makes sense for building an android image 😁. Otherwise we'd have to install a lot of dependencies and build tools manually.

> I don't think we want to upgrade to node 12 yet.

Why not though? The only issue I see is that the other apps still use node 10 and not all of them can be upgraded straight away.

src/drive/web/modules/drive/Toolbar/components/MoreMenu.jsx
```diff
@@ -73,4 +74,4 @@ const MoreMenu = ({
   )
 }
 
-export default withBreakpoints()(MoreMenu)
+export default translate()(withBreakpoints()(MoreMenu))
```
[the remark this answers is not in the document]
y-lohse: We don't, good catch!

package.json
```diff
@@ -24,7 +24,7 @@
     "lint:js": "eslint '{src,test}/**/*.{js,jsx}'",
     "testcafe:drive": "node testcafe/runner-drive.js",
     "testcafe:photos": "node testcafe/runner-photos.js",
-    "test": "env USE_REACT=true cozy-scripts test --config=jest.config.json",
+    "test": "env NODE_ENV='test' cozy-scripts test --config=jest.config.json",
```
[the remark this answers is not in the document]
y-lohse: Did it in the commit message! But in case you're genuinely curious about this and don't want to check the commit message, it's because of https://github.com/cozy/cozy-ui/pull/1365

src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx
Crash--: cc @poupotte pour info ;)

package.json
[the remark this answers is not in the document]
Crash--: YOU FORCE PUSHED, right 🤣 Sorry for the noise

.travis.yml
Crash--: I'm not very confident about this node 12 upgrade for now, but let's use that since it's just here to build dev APK. We can test it like that

src/drive/web/modules/public/LightFolderView.jsx
```diff
@@ -82,6 +85,15 @@ class DumbFolderView extends React.Component {
       })
   }
 
+  loadPermissions = async () => {
+    const { client } = this.props
+    const response = await client
+      .collection('io.cozy.permissions')
+      .getOwnPermissions()
+    const permission = get(response, 'data.attributes.permissions.files', {})
```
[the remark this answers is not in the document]
y-lohse: I'll use permission.fetchOwn once https://github.com/cozy/cozy-client/pull/701  is merged.

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -236,13 +236,18 @@ export const getFileDownloadUrl = async id => {
  * @param {Object} sharingState - The sharing context (provided by SharingContext.Provider)
  * @returns {function} - A function that dispatches addToUploadQueue action
  */
-export const uploadFiles = (files, dirId, sharingState) => dispatch => {
+export const uploadFiles = (
+  files,
+  dirId,
+  sharingState,
+  fileUploadedCallback = () => null
```
Crash--, line 243: comments to add

PR #2019, opened 2020-05-20
Release/1.24.0

Crash--:
- Stop / Resume backup when importing a file 
- RegisterPluginClient everywhere


Crash--:
updated commit message with : 
"Since the user import the file manually we consider
that this upload should have priority over the sync
process. So we stop it and restart it as soon as the
upload is done."

PR #2021, opened 2020-05-22
Public view fixes

y-lohse:
Rebasing #2017 was not straightforward and it appears I did loose a few changes in the process.

See individual commit messages for details.

### DIFF
--- a/src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.spec.jsx
+++ b/src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.spec.jsx
@@ -9,8 +9,6 @@ import MuiCozyTheme from 'cozy-ui/transpiled/react/MuiCozyTheme'
 import ShortcutCreationModal from './ShortcutCreationModal'
 import AppLike from '../../../../../../../test/components/AppLike'
 const tMock = jest.fn()
-jest.mock('cozy-ui/transpiled/react/utils/color', () => ({
-  getCssVariableValue: () => '#fff'
 }))
 
 function createMatchMedia(width) {
@@ -36,7 +34,7 @@ describe('ShortcutCreationModal', () => {
     tMock.mockImplementation(key => key)
   })
 
-  it('should show the new account form', async () => {
+  it('should show the new shortcut form', async () => {
     const { getByLabelText, getByText } = render(
       <AppLike client={client}>
         <MuiCozyTheme>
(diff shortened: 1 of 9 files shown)


src/drive/web/modules/public/LightFileList.jsx
```diff
@@ -23,7 +24,9 @@ const LightFileList = ({
...
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(LightFileList)
+export default withRouter(
+  connect(
+    mapStateToProps,
+    mapDispatchToProps
+  )(LightFileList)
+)
```
Crash--, line 41: I don't understand this change. It seems we don't use `router` in this component and we don't pass it neither to its children. Can you explain why this is needed?

src/drive/web/modules/public/LightFileList.jsx
```diff
@@ -19,14 +22,20 @@ const LightFileList = ({
...
 )
 
 export default withRouter(
   connect(
-    mapStateToProps,
+    state => ({
+      ...mapStateToProps,
+      isRenaming: isRenaming(state),
+      renamingFile: getRenamingFile(state)
+    }),
```
Crash--, line 38: We already those props in the mapStateToProps https://github.com/cozy/cozy-drive/blob/01748efdf7a932959ac6b4636414cbabd924c154/src/drive/web/modules/public/Actions.jsx#L20

src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.spec.jsx
```diff
@@ -69,6 +77,8 @@ describe('ShortcutCreationModal', () => {
         }
       }
     )
+    expect(Alerter.error).toHaveBeenCalledTimes(1)
+    await waitFor(() => expect(Alerter.success).toHaveBeenCalledTimes(1))
```
Crash--, line 81: We should also add a test with a onCreated callback?

src/drive/web/modules/public/LightFileList.jsx
y-lohse: Ah yes thanks, good catch. I know realize  they shouldn't be in `Actions` since they aren't used by any of the actions anyway.

src/drive/web/modules/public/LightFileList.jsx
y-lohse: It is used [here](https://github.com/cozy/cozy-drive/blame/fix-public/src/drive/web/modules/public/Actions.jsx#L54-L55), in the `mapDispatchToProps` of he component. But I agree, it's not very obvious. Do you think it would be better to include the `Actions` file in this one to make it more explicit? Otherwise we can add a comment imo.

Crash--: The best solution is to not rely on `ownprops.something` (since it can be dangerous for the perf)  but I don't know how to handle that.

PR #2025, opened 2020-05-28
Public navigation fixes

y-lohse:
The navigation on the public view uses `/files/:folderId` format, whereas the private app uses `/folder/:folderId`. This trips up some internal functions like [getFolderUrl](https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/public/LightFolderView.jsx#L102).

This PR makes both public and private app use the same format. This solves 2 problems:

1. Navigating on the public view changes the URL, meaning you can use the browsers back button etc
2. Reloading the content of the view actually reloads the currently displayed folder, and not the first opened folder

### DIFF
--- a/src/drive/targets/public/index.jsx
+++ b/src/drive/targets/public/index.jsx
@@ -112,7 +112,8 @@ const init = async () => {
             ) : (
               <Router history={hashHistory}>
                 <Route component={PublicLayout}>
-                  <Route path="files(/:folderId)" component={LightFolderView}>
+                  <Redirect from="/files/:folderId" to="/folder/:folderId" />
+                  <Route path="folder(/:folderId)" component={LightFolderView}>
                     <Route
                       path="file/:fileId/revision"
                       component={FileHistory}
--- a/src/drive/web/modules/public/LightFolderView.jsx
+++ b/src/drive/web/modules/public/LightFolderView.jsx
@@ -27,7 +27,7 @@ import { FILES_FETCH_LIMIT } from 'drive/constants/config'
 import Viewer from 'drive/web/modules/viewer/PublicViewer'
 import { isMobileApp } from 'cozy-device-helper'
 
-class DumbFolderView extends React.Component {
+export class DumbFolderView extends React.Component {
   state = {
     revoked: false,
     viewerOpened: false,
@@ -112,7 +112,7 @@ class DumbFolderView extends React.Component {
     return (
       <Main isPublic>
         <Topbar>
-          <Breadcrumb isPublic onFolderOpen={this.props.fetchFolder} />
+          <Breadcrumb isPublic onFolderOpen={this.navigateToFolder} />
           <PublicToolbar
             files={[this.props.displayedFolder]}
             isFile={false}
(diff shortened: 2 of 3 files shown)


ptbrowne:
No tests need to be updated or added after this change ?

y-lohse:
> No tests need to be updated or added after this change ?

Nothing to change no, and I don't know what would make sense to add as it's primarily configuration change…

src/drive/web/modules/public/LightFolderView.jsx
```diff
@@ -112,7 +112,7 @@ class DumbFolderView extends React.Component {
     return (
       <Main isPublic>
         <Topbar>
-          <Breadcrumb isPublic onFolderOpen={this.props.fetchFolder} />
+          <Breadcrumb isPublic onFolderOpen={this.navigateToFolder} />
```
Crash--, line 115: Where do we still use fetchFolder?

y-lohse: It's used in `loadCurrentFolder` and `navigateToFolder`, ie. on the initial mount, view reload and navigation changes. But it is not used in the jsx anymore, only the class functions.

src/drive/targets/public/index.jsx
```diff
@@ -112,7 +112,8 @@ const init = async () => {
             ) : (
               <Router history={hashHistory}>
                 <Route component={PublicLayout}>
-                  <Route path="files(/:folderId)" component={LightFolderView}>
+                  <Redirect from="/files/:folderId" to="/folder/:folderId" />
+                  <Route path="folder(/:folderId)" component={LightFolderView}>
```
Crash--, line 116: Did you check that we have nothing that is based on the previous `files/` url? In some files, I know we do few things with the url.

src/drive/web/modules/public/LightFolderView.jsx
Crash--: We can add a test to click on the breadcrumb and check if the URL has changed in the way we want, no?

src/drive/targets/public/index.jsx
y-lohse: You might be thinking of https://github.com/cozy/cozy-drive/blame/public-nav/src/drive/web/modules/public/Actions.jsx#L46 ? In that case it's the /folder/:id/file/:id sub-route, and it's ok with that change.

I don't think there's more, but in any case the `Redirect` should take care of everything.

Crash--: I was more thinking about the FileExplorer file https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/navigation/FileExplorer.jsx but may be we are not using it here.

src/drive/web/modules/public/LightFolderView.jsx
y-lohse: I tried to add a test like this, but it requires rendering pretty much the whole app, including a complete store and router. I don't see a way to do it in a test env in a reasonable amount of time.

I added a test that uses shallow rendering instead, that tests that calling `navigateToFolder` calls `router.push` with the correct path. It wouldn't have caught the breadcrumb problem, but we would have seen that the URL was called with `/folder/123` instead of `/files/123` prior to 3585b9b42d5486d70c15542559efde198c069a07

src/drive/targets/public/index.jsx
y-lohse: Indeed, this component isn't used on the public view.

PR #2026, opened 2020-05-29
fix: More defensive code.

Crash--:
We can have FSCK's issue mostly `index_orphan_tree` or `filesystem_missing`.
In that case, we can receive an empty doc without path. So let's be
more defensive to not crash the `recent's view`.

We display the file in this view but without its folder below. The user will have an error when trying to download it. But it's better than crashing the view

### DIFF
--- a/src/drive/web/modules/navigation/duck/actions.jsx
+++ b/src/drive/web/modules/navigation/duck/actions.jsx
@@ -164,6 +164,27 @@ export const fetchMoreFiles = (folderId, skip, limit) => {
     }
   }
 }
+/**
+ *
+ * @param {Array} files
+ * @param {Array} parentFolders
+ */
+export const mergeFilesWithParents = (files, parentFolders) => {
+  const filesWithPath = files.map(file => {
+    const parentFolder = parentFolders.rows.find(row => row.id === file.dir_id)
+    /* We can have FSCK's issue mostly index_orphan_tree or filesystem_missing.
+      In that case, we can receive an empty doc without path. So let's be
+      more defensive to not crash the recent view */
+    //!TODO Here we are able to detect an FSCK error. We should add a flag to the file
+    //to inform the user
+    const path =
+      parentFolder && parentFolder.doc && parentFolder.doc.path
+        ? parentFolder.doc.path
+        : ''
+    return { ...file, path, id: file._id }
+  })
+  return filesWithPath
+}
 
 export const fetchRecentFiles = () => {
   return async (dispatch, getState) => {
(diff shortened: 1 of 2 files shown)


Crash--:
It's just a WIP at the moment. sorry for the noise.

Crash--:
Bon pour review. On a statué sur ce qu'on voulait faire ;)

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -164,6 +164,27 @@ export const fetchMoreFiles = (folderId, skip, limit) => {
...
+    const parentFolder = parentFolders.rows.find(row => row.id === file.dir_id)
+    /* We can have FSCK's issue mostly index_orphan_tree or filesystem_missing.
+      In that case, we can receive an empty doc without path. So let's be
+      more defensive to not crash the recent view */
+    //!TODO Here we are able to detect an FSCK error. We should add a flag to the file
+    //to inform the user
+    const path =
+      parentFolder && parentFolder.doc && parentFolder.doc.path
+        ? parentFolder.doc.path
+        : ''
```
ptbrowne, line 183: Is the file shown to the user when its path is empty ? What is the meaning of having an empty path ? Should we have a sentinel value like FSCK_PATH_PROBLEM instead of empty string ?

src/drive/web/modules/navigation/duck/actions.jsx
```diff
@@ -164,6 +164,27 @@ export const fetchMoreFiles = (folderId, skip, limit) => {
     }
   }
 }
+/**
+ *
```
ptbrowne, line 168: Can you add a description of the goal of the function ? I don't understand why it is related to the recent view.

ptbrowne: Should this method be part of the FileCollection ?

src/drive/web/modules/navigation/duck/actions.jsx
Crash--: See the TODO... I don't want to do a one shot fix for this kind of error and we need to brainstorm on how to handle those cases.

Crash--: > Is the file shown to the user when its path is empty ? What is the meaning of having an empty path ? 

Described in the PR...

src/drive/web/modules/navigation/duck/actions.jsx
Crash--: Yes, and this is why this is not ported to CC yet. My goal was not to rewrite this part yet, since it uses this kind of code https://github.com/cozy/cozy-drive/blob/5f13f23069199c355385b5d68bd0ded42a4ee316/src/drive/web/modules/navigation/duck/actions.jsx#L203 based on  https://github.com/cozy/cozy-drive/blob/5f13f23069199c355385b5d68bd0ded42a4ee316/src/drive/web/modules/navigation/duck/async.js#L288 and it will be removed for the migration.

nono:
What is the status of this PR?

src/drive/web/modules/navigation/duck/actions.jsx
ptbrowne: OK, thanks for the explanation.

PR #2030, opened 2020-06-10
fix: Await the unlink to be sure to have destroyed the Pouch.

Crash--:
Thanks to this PR (https://github.com/cozy/cozy-bar/pull/689), I checked
the code in Drive to handle the logout since we have this kind of issue
(not having the pouch really deleted). I think it can help.

Needs https://github.com/cozy/cozy-bar/pull/689

### DIFF
--- a/src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
+++ b/src/drive/mobile/modules/authorization/DriveMobileRouter.jsx
@@ -61,8 +61,8 @@ class DriveMobileRouter extends Component {
     }
   }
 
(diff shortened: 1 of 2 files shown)


ptbrowne:
It seems the logout is not awaited. It should be otherwise awaiting the unlink will not wok.

https://github.com/cozy/cozy-drive/blob/master/src/drive/mobile/modules/authorization/duck/actions.js#L15

Can you add a test testing that client.logout is correctly awaited ?

nono:
What is the status of this PR?

PR #2033, opened 2020-06-15
Test actions on files

y-lohse:
This PR adds tests for the actions on files (follow up on #2031 among others).


src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+
+      mockSharingContextValue.hasWriteAccess.mockReturnValue(true)
+      const renameAction = getAction('rename')
+      expect(renameAction.displayCondition(['abc'])).toBe(true)
+      expect(renameAction.displayCondition(['abc', 'def'])).toBe(false)
+    })
+
+    it('dispatches a rename action when activated', async () => {
+      const renameAction = getAction('rename')
+      const spy = jest.spyOn(rename, 'startRenamingAsync')
```
y-lohse, line 290: Ideally I would have liked to test the call to `dispatch`, but somehow I can't get a reference to the `dispatch` returned by `useDispatch`. Even things like `const spy = jest.spyOn(mockStore, 'dispatch')` don't get called..?

src/drive/web/modules/actions/utils.js
```diff
@@ -0,0 +1,159 @@
...
+    : 'error.download_file.offline'
+}
+
+const openFileDownloadError = error => {
+  return isMissingFileError(error)
+    ? 'mobile.error.open_with.missing'
+    : 'mobile.error.open_with.offline'
+}
+
+export const downloadFiles = async (files, client) => {
```
ptbrowne, line 35: I would prefer that as a convention, for functions that need to be passed a client, we put `client` as the first argument. I already do this for all the code I write, and I think it helps since argument order here is not obvious.

src/drive/web/modules/actions/utils.js
```diff
@@ -0,0 +1,159 @@
...
+
+    Alerter.success('alert.trash_file_success')
+  } catch (err) {
+    if (!isAlreadyInTrash(err)) {
+      Alerter.error('alert.try_again')
+    }
+  }
+}
+
+export const exportFilesNative = async (files, client, filename) => {
```
ptbrowne, line 90: For example at this point, why is client in the middle ? I think with the `client` first rule, it would be a bit easier to remember.

src/drive/web/modules/actions/utils.js
```diff
@@ -0,0 +1,159 @@
...
+}
+
+export const openFileWith = async (file, client, filename) => {
+  if (isMobileApp() && window.cordova.plugins.fileOpener2) {
+    let fileData
+    try {
+      fileData = await client.collection('io.cozy.files').fetchFileContent(file)
+    } catch (error) {
+      Alerter.error(openFileDownloadError(error))
+      throw error
```
ptbrowne, line 147: Why do we rethrow the error here ?

src/drive/web/modules/actions/utils.js
```diff
@@ -0,0 +1,159 @@
...
+    const blob = await fileData.blob()
+    try {
+      await saveAndOpenWithCordova(blob, filename)
+    } catch (error) {
+      Alerter.error('mobile.error.open_with.noapp')
+    }
+  } else {
+    Alerter.error('mobile.error.open_with.noapp')
+  }
+}
```
ptbrowne, line 159: I think it would be a good time to add docstrings to those functions.

src/drive/web/modules/actions/utils.js
```diff
@@ -0,0 +1,159 @@
...
+    const filenameToUse = filename ? filename : file.name
+    const localFile = await saveFileWithCordova(blob, filenameToUse)
+    return localFile.nativeURL
+  })
+
+  try {
+    Alerter.info('alert.preparing', {
+      duration: Math.min(downloadAllFiles.length * 2000, 6000)
+    })
+    const urls = await Promise.all(downloadAllFiles)
```
ptbrowne, line 106: Nit: we hould use a Promise pool here otherwise we may have problems (too many connections for example) on a case where there are a lot of files.

src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+    router: {
+      push: jest.fn()
+    },
+    location: { pathname: '/folder' }
+  }
+  const mockClient = createMockClient({})
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    global.__TARGET__ = 'desktop'
```
ptbrowne, line 50: Is it a real target ? I thought __TARGET__ could only be "browser" or "mobile". Recently I had to add it to the globals exposed by jest. I put "browser" so I think the above line is not needed. I think we should explicitly set it when setting it to "mobile".

src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+    {
+      canMove: true
+    }
+  ]
+  const getAction = (actionKey, hookArgs = defaultHookArgs) => {
+    const { result } = renderActionsHook(...hookArgs)
+    return result.current[actionKey]
+  }
+
+  it('returns actions keyed by icon', () => {
```
ptbrowne, line 81: If it is a good moment to have the actions keyed by id, I think we should do it now and repeat potentially the icon in the action spec. I do not think having the icon as key is a good idea.

src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+      })
+
+      it('is visible for a single file on iOS', () => {
+        global.window.cordova = { platformId: 'ios' }
+        const downloadAction = getAction('download')
+        expect(
+          downloadAction.displayCondition([{ id: 'abc', type: 'file' }])
+        ).toBe(true)
+        expect(
+          downloadAction.displayCondition([{ id: 'abc', type: 'folder' }])
```
ptbrowne, line 155: Nit: At this point, it was not completely clear to that displayCondition was not `displaying` a condition of type "[{ id: 'abc', type: 'folder' }]" then I understood that displayCondition was the *name* for the function and not the action of this function. I think "shouldBeDisplayed" would be easier to read.

src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+            { id: 'def', type: 'file' }
+          ])
+        ).toBe(false)
+        expect(
+          downloadAction.displayCondition([
+            { id: 'abc', type: 'file' },
+            { id: 'def', type: 'folder' }
+          ])
+        ).toBe(false)
+        global.window.cordova = null
```
ptbrowne, line 169: global.window.cordova will not be cleaned up if the test fail, prefer to setup/cleanup in beforeEach, afterEach.

src/drive/web/modules/actions/useActions.spec.jsx
```diff
@@ -0,0 +1,422 @@
...
+          { type: 'file', id: 'abc' },
+          { type: 'file', id: 'def' }
+        ])
+      ).toBe(false)
+      expect(
+        offlineAction.displayCondition([{ type: 'folder', id: 'abc' }])
+      ).toBe(false)
+    })
+  })
+})
```
ptbrowne, line 422: I think that's good that all the action are now tested on several situations. However, I think it's difficult to get an overall picture of what's going on, situation per situation.

I think it could be good that at some point we make a test plan with:

- sets of known files / folders with checks
- personas (read/write desktop, read/write mobile, read desktop, read mobile)

And matrix test all different situations and check the displayed actions.

Then we could have a test like this.

```
const { displayedActions } = setup(fileSet1, {
   selected: ['file1', 'file2'],
   platform: 'android'
})
expect(displayedActions.toMatch(['share', 'openwith']))
```

- I find it easier to think in terms of situation than action per action.
- I think it helps when the test case can be read by a non technical person

ptbrowne, line 422: I think that's good that all the action are now tested on several situations. However, I think it's difficult to get an overall picture of what's going on, situation per situation.

I think it could be good that at some point we make a test plan with:

- sets of known files / folders with checks
- personas (read/write desktop, read/write mobile, read desktop, read mobile)

And matrix test all different situations and check the displayed actions.

Then we could have a test like this.

```
const { displayedActions } = setup(fileSet1, {
   selected: ['file1', 'file2'],
   platform: 'android'
})
expect(displayedActions.toMatch(['share', 'openwith']))
```

- I find it easier to think in terms of situation than action per action.
- I think it helps when the test case can be read by a non technical person

src/drive/web/modules/actions/useActions.spec.jsx
ptbrowne: startRenamingAsync is a bit more down the line that dispatch right ? I think I prefer to have it this way, since we could think of `dispatch` as an implementation detail ?

y-lohse: It is further down the line, but it also is an action creator so in a way it's couple to `dispatch`. Though we could change the implementation of `startRenamingAsync` without breaking the test too.

src/drive/web/modules/actions/useActions.spec.jsx
y-lohse: Ah you're correct, it's `browser` and not `desktop`. The new global isn't merged yet, it's in #2032, but I can use the same approach 👍

src/drive/web/modules/actions/useActions.spec.jsx
y-lohse: I think actions should be keyed by id and not icon, in fact in the definition they are keyed by id. But the component in cozy-ui expects icons names as keys and a `displayCondition` function. I'll create an issue in cozy-ui about that.

In the meantime, I think it makes sense to keep the name `displayCondition` for "grep-ability", and for the keys… we could change the key further down the line, but we'd have to do it in multiple places, so I suspect that would end up being more confusing.

src/drive/web/modules/actions/useActions.spec.jsx
y-lohse: I think that's a good idea. It was easier to write them as a list like this, but we can definitely keep the structure you suggest in mind for a future improvement.

src/drive/web/modules/actions/utils.js
y-lohse: Good catch, I don't think it has any business being here anymore.

PR #2037, opened 2020-06-19
createFolderV2

ptbrowne:
Use client.create to create a folder.

- Add selector to get the current folder list of files from cozy-client
  query
- Use this selector to know if we are creating a duplicate directory
- Create a folder through client.create for the newly create directory
  to be stored in the redux store


src/drive/web/modules/selectors.js
```diff
@@ -1,5 +1,10 @@
 export const getCurrentFolderId = rootState => {
-  return rootState.router.params.folderId
+  if (rootState.router.params.folderId) {
+    return rootState.router.params.folderId
+  } else if (rootState.router.location.pathname == '/folder') {
+    return 'io.cozy.files.root-dir'
```
Crash--, line 9: you can use `ROOT_DIR_ID`  from 'constants/config` or something similar

src/drive/web/modules/views/Drive/index.jsx
```diff
@@ -89,14 +89,8 @@ const DriveView = ({ folderId, router, children }) => {
...
-        {false && (
-          <Toolbar
-            folderId={null}
-            canUpload={true}
-            canCreateFolder={true}
-            disabled={false}
-          />
-        )}
+        {/* TODO do not have props hardcoded */}
+        <Toolbar canUpload={true} canCreateFolder={true} disabled={false} />
```
Crash--, line 93: we don't need to pass `folderId` ? You got it from the `getCurrentFolderId` selector ?

src/drive/web/modules/selectors.spec.jsx
```diff
@@ -0,0 +1,100 @@
...
+  })
+})
+
+describe('getFolderContent', () => {
+  it('should return content from cozy client queries', async () => {
+    const { store } = await setup()
+    const state = store.getState()
+    const files = getFolderContent(state, 'folderid123456')
+    expect(files.length).toBe(10)
+  })
```
Crash--, line 33: Can we add a test to describe what we receive if the query was not executed yet?

src/drive/web/modules/navigation/duck/reducer.js
```diff
@@ -35,11 +38,6 @@ import {
   OPEN_FOLDER_FROM_TRASH_FAILURE
 } from 'drive/web/modules/trash/actions'
 
-import {
-  OPEN_FOLDER_FROM_SHARINGS_FAILURE,
-  OPEN_FOLDER_FROM_SHARINGS,
-  OPEN_FOLDER_FROM_SHARINGS_SUCCESS
-} from 'drive/web/modules/drive/SharingsContainer'
```
Crash--, line 42: So we can remove them from `SharingsContainer` ?

src/drive/web/modules/selectors.js
```diff
@@ -10,3 +13,34 @@ export const getCurrentFolderId = rootState => {
...
+  const folderContentQueries = getFolderContentQueries(rootState, folderId)
+  if (folderContentQueries.length > 0) {
+    const mostRecentQueryResults = maxBy(
+      folderContentQueries,
+      x => x.lastUpdate
+    )
+    return mostRecentQueryResults.data
+  } else {
+    return null
+  }
```
Crash--, line 53: I though, I don't know exactly how, that this is the job of cozyclient to handle that kind of stuff. Shouldn't we use getDocumentFromState or something? 🤔

src/drive/web/modules/queries.js
```diff
@@ -42,4 +47,25 @@ const buildQuery = ({ currentFolderId, type, sortAttribute, sortOrder }) => ({
...
+  )
+  const otherType = type === 'directory' ? 'file' : 'directory'
+  const otherQueryId = formatFolderQueryId(
+    otherType,
+    folderId,
+    sortAttribute,
+    sortOrder
+  )
+  return otherQueryId
+}
```
Crash--, line 69: Are you considering that the query `directory 123 name desc` returns the same values than `files 123 name desc`? How is that even possible to have the both queries?

ptbrowne: No, those are two different queries, one resolving with the directories of the folder, the other with the files.

src/drive/web/modules/navigation/duck/reducer.js
ptbrowne: Yes, they are removed.

src/drive/web/modules/views/Drive/index.jsx
ptbrowne: Yes.

src/drive/web/modules/selectors.js
ptbrowne: Mmh I am not sure that something like this should be in cozy-client. Here we use applicative knowledge to know that different queries can be merged as a single query. Maybe we could achieve something similar with a query group (this would be a new notion in cozy-client), but I am not sure that we have clear ideas on this.

src/drive/web/modules/selectors.spec.jsx
ptbrowne: I added one test, thanks for the suggestion.

PR #2040, opened 2020-06-25
fix: Download a single file

y-lohse:
Downloading a single file was broken because `client.getDownloadLinkById` returns a full URL, whereas the cozy-client-js version returned the path only, without the domain name.


ptbrowne:
:shipit: ;)

PR #2041, opened 2020-06-25
Recent view

Crash--:
- Migrate the Recent View to Cozy-Client 
- Use cozy-client instead of ccjs in the PushBanner 
- Refactor the FileViewerV2 in order to be usable for folder / non folder view

/!\ Lot of code duplication in this PR, but we wanted to test our views before trying to figure out an API. In the next PR, we've refactored the code ;)


src/drive/web/modules/queries.js
```diff
@@ -28,7 +28,12 @@ export const formatFolderQueryId = (
   return `${type} ${folderId} ${sortAttribute} ${sortOrder}`
 }
 
-const buildQuery = ({ currentFolderId, type, sortAttribute, sortOrder }) => ({
+const buildDriveQuery = ({
```
ptbrowne, line 31: I do not find it that much clearer, but it sure seems that buildQuery is not specific enough. I suggest `buildFolderContentQuery` ?

Crash--: I want to rename `DriveView` to `FolderView` and rename this query to `FolderQuery`. 

I'm currently fixing a few bugs in FilesViewer, I'll do it in a later commit

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+  buildRecentQuery,
+  buildParentsByIdsQuery
+} from 'drive/web/modules/queries'
+
+const useFilesQueryWithPath = query => {
+  const result = useQuery(query.definition, query.options)
+  const resultData = result.data || []
+
+  const dirIds = uniq(resultData.map(({ dir_id }) => dir_id))
+  const parentsQuery = buildParentsByIdsQuery(dirIds)
```
ptbrowne: Ici à mon avis on pourrait peut être utiliser une relationship 1-1 entre le fichier et son parent. Ca permettrait d'avoir directement le parent au niveau de l'objet `file`, on aurait pas besoin de faire le `map` un peu plus bas.

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+    router.push(`/folder/${folderId}`)
+  })
+
+  const navigateToFile = useCallback(file => {
+    router.push(`/recent/file/${file.id}`)
+  })
+
+  return (
+    <Main>
+      <ModalStack />
```
ptbrowne, line 55: Je me demande pourquoi la ModalStack est pas dans Main ?

Crash--: C'est en train d'être refactoriser. 

On va avoir au final un truc du genre :
``` 
return (
    <FolderView>
      <FolderViewHeader>
        <TrashBreadcrumb />
        <TrashToolbar />
      </FolderViewHeader>
      <FolderViewBody actions={actions} queries={{}} canSort />
      {children}
    </FolderView>
  )
```

src/drive/web/modules/views/Recent/index.jsx
Crash--: On va surtout harmoniser le fait que la stack nous retourne le path pour un type: file aussi. C'est en cours de discussion.

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+                  withSharedBadge={true}
+                  isFlatDomain={true}
+                  thumbnailSizeBig={isBigThumbnail}
+                />
+              ))}
+              {result.hasMore && <LoadMore fetchMore={result.fetchMore} />}
+            </div>
+          )}
+        </FileListBodyV2>
+      </FileListv2>
```
ptbrowne, line 98: Nit: v2 and V2

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+      {children}
+    </Main>
```
ptbrowne, line 100: I am a little bit concerned here that we repeat all the logic from the Drive view, I wonder if we could reuse the same exact views between the Recent and Drive views, only with query and options of the toolbar changed ?

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+const RecentViewWithProvider = props => (
+  <SharingProvider doctype="io.cozy.files" documentType="Files">
+    <ThumbnailSizeContextProvider>
+      <RouterContextProvider>
+        <ModalContextProvider>
+          <RecentView {...props} />
+        </ModalContextProvider>
+      </RouterContextProvider>
+    </ThumbnailSizeContextProvider>
+  </SharingProvider>
```
ptbrowne, line 113: It seems that this would be repeated between the Drive view and the Recent view, could we find a way to share both contexts ?

src/components/pushClient/index.js
```diff
@@ -21,8 +21,8 @@ export const isIOS = () =>
 export const DESKTOP_BANNER = 'desktop_banner'
 export const NOVIEWER_DESKTOP_CTA = 'noviewer_desktop_cta'
 
-export const isClientAlreadyInstalled = async () => {
-  const resp = await cozy.client.fetchJSON('GET', '/settings/clients')
+export const isClientAlreadyInstalled = async client => {
+  const resp = await client.query(client.get('io.cozy.settings', 'clients'))
```
ptbrowne, line 24: No tests seem to have been changed by this change, I think it would be a good time to add one, what do you think ?

src/drive/web/modules/selectors.js
```diff
@@ -1,19 +1,20 @@
...
-  } else if (rootState.router.location.pathname == '/folder') {
+  } else if (get(rootState, 'router.location.pathname') == '/folder') {
     return ROOT_DIR_ID
   }
   return null
 }
 
 export const getCurrentFileId = rootState => {
-  return rootState.router.params.fileId
+  return get(rootState, 'router.params.fileId', null)
```
ptbrowne, line 17: :+1:

src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  return {
+    ...result,
+    data: resultData.map(file => ({
+      ...file,
+      displayedPath: get(parentsDocsById[file.dir_id], 'path')
+    }))
+  }
+}
```
ptbrowne, line 28: I repeat since the file has moved, I think using a 1-1 relationship from file to parent would be better since:

1. It would express the relationship in code
2. It would not rely on React hooks/queries for additional data fetching

src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+  if (parentsResult.fetchStatus === 'loaded' && parentsResult.data.length > 0) {
+    parentsDocsById = keyBy(parentsResult.data, '_id')
+  }
+
+  return {
+    ...result,
+    data: resultData.map(file => ({
+      ...file,
+      displayedPath: get(parentsDocsById[file.dir_id], 'path')
```
ptbrowne, line 25: Is this a real attribute from `io.cozy.files` ? If not, I think we should not add custom attributes to objects from the client, for use in downstream views. With relationships, this could be expressed as a function

```
const getParentPath = file => {
  return file.parent.path
}
```

src/drive/web/modules/views/Recent/index.jsx
Crash--: https://github.com/cozy/cozy-drive/pull/2041/files#r446060353

ptbrowne: Is this going to land in this PR ?

src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -0,0 +1,126 @@
...
+    totalCount,
+    useQueryResultAttributes,
+    path = '/test',
+    dir_id = 'io.cozy.files.root-dir',
+    updated_at = '2020-05-14T10:33:31.365224+02:00'
+  } = {}) => {
+    const { store, client } = setupStoreAndClient({
+      selection: {
+        selected: {}
+      }
```
ptbrowne, line 31: I think you should wrap the selection slice into the `initialStoreState` attribute.

src/drive/web/modules/views/Recent/index.jsx
ptbrowne: Ok ca n'empêche qu'une relation `parent` serait sûrement utile. Vous en discutez ou de la stack qui retourne le `path` (je n'ai pas vu passé) ?

src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -0,0 +1,126 @@
...
+        const childrenDontHaveText = Array.from(node.children).every(
+          child => !hasText(child)
+        )
+        return hasText(node) && childrenDontHaveText
+      })
+    }
+    //Get the HTMLElement containing the filename
+    const el0 = getByText(`foobar0`)
+    //check if the filename is displayed with the extension
+    expect(getByTextWithMarkup(`foobar0.pdf`))
```
ptbrowne, line 99: It seems an expectation is missing here.

```
expect(..).toBeDefined() # ?
```

src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(el0)
```
ptbrowne, line 100: Same here.

src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -0,0 +1,126 @@
...
+    expect(fileRow0.getElementsByTagName('time')[0].dateTime).toEqual(
+      updated_at
+    )
+    //check the path to the parent's folder
+    const linkElement0 = fileRow0.getElementsByClassName('fil-file-path')[0]
+    expect(linkElement0.textContent).toEqual(path)
+
+    expect(linkElement0.href.endsWith(`#/folder/${dir_id}`)).toBe(true)
+
+    //check if the ActionMenu is displayed
```
ptbrowne, line 113: Nit: I think we should format comments with `// Check if the ActionMenu is displayed`.

src/drive/web/modules/views/Recent/index.jsx
Crash--: Nop, already too much work done on this one. Coming with the Trash view

ptbrowne: OK thanks. Next time, I think it'd be helpful to have this kind of information in the description.

src/drive/web/modules/views/Recent/index.jsx
ptbrowne: I guess this will be changed in next PRs https://github.com/cozy/cozy-drive/pull/2041#discussion_r446099915

src/drive/web/modules/views/Recent/index.spec.jsx
Crash--: `getBy*` throw when the predicate was not found.  So if it's doesn't throw, it works. Let's remove the expect() here it doesn't do anything, you right.

ptbrowne: You can write `expect(() => getByTextWithMarkup('foobar0.pdf')).not.toThrow()`.

src/components/pushClient/index.js
ptbrowne: Done in #2043

PR #2042, opened 2020-06-26
feat: Use cozy-client for download folder

ptbrowne:
The download files action had already been coded for the selection. So this
PR is only a matter of using this action for the download folder action
on the current folder.

I also added a beginning of tests for the MoreMenu. I think we should test
all items there instead of one by one, this will be easier to understand
I think, and initialisation code will be mutualised.


src/drive/web/modules/drive/Toolbar/toolbar.jsx
```diff
@@ -9,9 +9,10 @@ const mapStateToProps = state => {
   const displayedFolder = flag('drive.client-migration.enabled')
     ? getDisplayedFolder(state)
     : state.view.displayedFolder
+
   return {
     displayedFolder,
-    notRootfolder: displayedFolder && displayedFolder.id !== ROOT_DIR_ID,
+    notRootfolder: !displayedFolder || displayedFolder.id !== ROOT_DIR_ID,
```
Crash--, line 15: NIT: Je trouve l'ancienne version bien plus facile à lire

ptbrowne: Euh, c'est pas la même condition

PR #2043, opened 2020-06-26
test: Add test for PushClient

Crash--:
Add a test for the isClientAlreadyInstalled

### DIFF
--- a/src/components/pushClient/index.spec.js
+++ b/src/components/pushClient/index.spec.js
@@ -1,0 +1,32 @@
+import { isClientAlreadyInstalled, DESKTOP_SOFTWARE_ID } from './index'
+import CozyClient from 'cozy-client'
+
+describe('isClientAlreadyInstalled', () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+  test('isClientAlreadyInstalled is true', async () => {
+    const client = new CozyClient({})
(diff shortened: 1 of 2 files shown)


src/components/pushClient/index.spec.js
```diff
@@ -0,0 +1,32 @@
...
+      {
+        attributes: {
+          software_id: DESKTOP_SOFTWARE_ID
+        }
+      }
+    ])
+    const isInstalled = await isClientAlreadyInstalled(client)
+    expect(isInstalled).toBe(true)
+  })
+  it('test isClientAlreadyInstalled is not isntalled', async () => {
```
ptbrowne, line 20:
```suggestion
  it('test isClientAlreadyInstalled is not installed', async () => {
```

src/components/pushClient/index.spec.js
```diff
@@ -0,0 +1,32 @@
+import { isClientAlreadyInstalled, DESKTOP_SOFTWARE_ID } from './index'
+import CozyClient from 'cozy-client'
+
+describe('isClientAlreadyInstalled', () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+  it('test isClientAlreadyInstalled is true', async () => {
```
ptbrowne, line 8:
```suggestion
  it('isClientAlreadyInstalled is true', async () => {
```

PR #2044, opened 2020-06-26
Trash view

y-lohse:
This PR adds the trash view, and refactors the existing one to reduce duplicated code.


src/drive/web/modules/views/Trash/Breadcrumb.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+import { MobileAwareBreadcrumbV2 as Breadcrumb } from 'drive/web/modules/navigation/Breadcrumb/MobileAwareBreadcrumb'
+
+const RecentBreadcrumb = () => {
```
ptbrowne: TrashBreadcrumb

src/drive/web/modules/views/Trash/index.jsx
```diff
@@ -0,0 +1,117 @@
...
+      <RouterContextProvider>
+        <ModalContextProvider>
+          <TrashView {...props} />
+        </ModalContextProvider>
+      </RouterContextProvider>
+    </ThumbnailSizeContextProvider>
+  </SharingProvider>
+)
+
+export default TrashViewWithProvider
```
ptbrowne: Nit: As a reviewer, I expect this file to be refactored in the future, but for the sake of the reviewer, I think it'd be a bit better if the refactor was done before the new view was created, so that I do not have to guess that it's going to be refactored afterwards.

src/drive/web/modules/views/Trash/useFilesQueryWithPath.jsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  return {
+    ...result,
+    data: resultData.map(file => ({
+      ...file,
+      displayedPath: get(parentsDocsById[file.dir_id], 'path')
+    }))
+  }
+}
```
ptbrowne: This seems to be a copy paste of another file, I do not see why the trash view needs to have a specific `useFilesQueryWithPath`.

src/drive/web/modules/trash/components/DestroyConfirm.jsx
```diff
@@ -24,11 +24,14 @@ const DestroyConfirm = ({ t, fileCount, confirm, onClose }) => {
...
-          .then(onClose)
-          .catch(onClose)
-      }
+      primaryAction={async () => {
+        try {
+          await confirm()
+          onClose()
+        } catch {
+          onClose()
+        }
```
ptbrowne, line 33: Nit: `finally` can be used here so that only onClose can be used.

src/drive/web/modules/views/Trash/useActions.jsx
```diff
@@ -0,0 +1,48 @@
...
+            fileCount={files.length}
+            confirm={async () => {
+              for (const file of files) {
+                await client
+                  .collection('io.cozy.files')
+                  .deleteFilePermanently(file.id)
+              }
+            }}
+            onClose={popModal}
+          />
```
ptbrowne, line 40: I wonder why the confirm function needs to be passed here. I think DestroyConfirm could have a `files` prop, and could provide the `confirm` implementation to downstream components (instead of having the confirm implementation here. This way it would be testable on its own, and would not need to be wired up through `useActions`. And the fileCount would be taken directly from the `files` property.

src/drive/web/modules/views/Trash/useActions.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const actions = {
+    restore: {
+      icon: 'restore',
+      action: async files => {
+        for (const file of files) {
+          await client.collection('io.cozy.files').restore(file.id)
+        }
+        refresh()
+      }
```
ptbrowne, line 24: I think both the restore action function an the destroy action implemenation can be extracted from here.

src/drive/web/modules/views/Trash/TrashFolderView.jsx
```diff
@@ -0,0 +1,102 @@
...
+        canSort
+      />
+      {children}
+    </FolderView>
+  )
+}
+
+export default connect(state => ({
+  currentFolderId: getCurrentFolderId(state)
+}))(TrashFolderView)
```
ptbrowne, line 101: I have trouble understanding this commit, it seems that a lot of code is added. Is this duplication of existing code ?

src/drive/web/modules/views/Trash/TrashFolderView.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const navigateToFolder = useCallback(folderId => {
+    router.push(`/trash/${folderId}`)
+  })
+
+  const navigateToFile = useCallback(file => {
+    router.push(`/trash/file/${file.id}`)
+  })
+
+  const actions = useActions()
```
ptbrowne, line 69: I think it'd be a bit more understandable if useActions was the same function between all the views, and the view would provide the name of the actions it can display.

```
const TRASH_FOLDER_ACTIONS = [
  'deletePermanently',
  'restore'
]
const actions = useActions(TRASH_FOLDER_ACTIONS)
```

Here I am worried about both useActions drifting in how they work.

src/drive/web/modules/views/Trash/index.jsx
```diff
@@ -52,12 +53,14 @@ export const TrashView = ({ router, children }) => {
...
   const actions = useActions()
+  const { t } = useI18n()
+
   return (
     <Main>
       <RealTimeQueries doctype="io.cozy.files" />
       <ModalStack />
       <Topbar>
-        <Breadcrumb />
+        <Breadcrumb path={[{ name: t('breadcrumb.title_trash') }]} />
```
ptbrowne, line 36: 👍 I find it easier to understand this way as it is more explicit.

src/drive/web/modules/views/Trash/index.jsx
```diff
@@ -56,69 +29,21 @@ export const TrashView = ({ router, children }) => {
...
-        <ModalContextProvider>
-          <TrashView {...props} />
-        </ModalContextProvider>
-      </RouterContextProvider>
-    </ThumbnailSizeContextProvider>
-  </SharingProvider>
-)
-
-export default TrashViewWithProvider
+export default TrashView
```
ptbrowne: Mmmh it seems that a lot of previously added code was removed, great :)

src/drive/web/modules/views/Drive/index.jsx
```diff
@@ -1,50 +1,51 @@
...
+        id: get(displayedFolder, 'dir_id')
+      },
+      {
+        id: displayedFolder.id,
+        name: displayedFolder.name
+      }
+    ],
+    'id'
+  )
+    .filter(({ id }) => Boolean(id))
```
ptbrowne, line 38: Why is this filter important ? I think a little comment could be helpful here.

src/drive/web/modules/views/Drive/index.jsx
```diff
@@ -1,50 +1,51 @@
...
 
-import RealTimeQueries from './RealTimeQueries'
+const getBreadcrumbPath = (t, displayedFolder) =>
+  uniqBy(
+    [
+      {
+        id: ROOT_DIR_ID
+      },
+      {
+        id: get(displayedFolder, 'dir_id')
```
ptbrowne, line 29: Could we add the `name` here ? It'd be easier to know where we are.

ptbrowne: Ah I see, we do not have the parent path here. Having the parent relationship on `files` would help here I think.

src/drive/web/modules/navigation/AppRoute.jsx
```diff
@@ -101,15 +101,15 @@ const AppRoute = (
...
-        }
-      >
+      <Route path="trash">
+        <IndexRoute
+          component={
+            flag('drive.client-migration.enabled') ? TrashView : LegacyTrash
+          }
+        />
         <Route path="file/:fileId" component={FilesViewer} />
+        <Route path=":folderId" component={TrashFolderView} />
```
ptbrowne, line 111: Nit: This commit could have been squashed with the commit introducing the bug.

test/components/AppLike.jsx
```diff
@@ -18,10 +22,51 @@ export const TestI18n = ({ children }) => {
...
+            <ThumbnailSizeContextProvider>
+              <ModalContext.Provider
+                value={modalContextValue || mockModalContextValue}
+              >
+                {children}
+              </ModalContext.Provider>
+            </ThumbnailSizeContextProvider>
+          </RouterContext.Provider>
+        </SharingContext.Provider>
+      </TestI18n>
```
ptbrowne, line 69: 👍 Cool to have AppLike be as close as possible a the App !

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -71,25 +72,25 @@ const FolderViewBody = ({
...
                       displayedFolder={null}
-                      onFolderOpen={navigateToFolder}
-                      onFileOpen={navigateToFile}
                       withSelectionCheckbox={true}
-                      withFilePath={false}
                       withSharedBadge={true}
                       isFlatDomain={true}
+                      onFolderOpen={navigateToFolder}
+                      onFileOpen={navigateToFile}
+                      withFilePath={withFilePath}
```
ptbrowne, line 104: Nit: Previously, props were almost sorted an this is not the case anymore. I think it's good when similar props are grouped, especially when there are a lot of them like here.

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -53,68 +29,21 @@ export const RecentView = ({ router, children }) => {
...
+        navigateToFolder={navigateToFolder}
+        navigateToFile={navigateToFile}
+        actions={actions}
+        queryResults={[result]}
+        canSort={false}
+        withFilePath={true}
+      />
       {children}
-    </Main>
+    </FolderView>
```
ptbrowne, line 45: Really cool :)

src/drive/web/modules/trash/Toolbar.jsx
```diff
@@ -50,7 +50,11 @@ const Toolbar = ({
           {isMobile && (
             <>
               <ActionMenuItem
-                onClick={() => emptyTrash()}
+                onClick={() =>
+                  flag('drive.client-migration.enabled')
+                    ? client.collection('io.cozy.files').emptyTrash()
```
ptbrowne, line 55: Naive question : Is the query representing the trash correctly updated here ? How ?

src/drive/web/modules/views/Drive/RealTimeQueries.jsx
```diff
@@ -29,7 +29,10 @@ const RealTimeQueries = ({ doctype }) => {
         dispatchChange(document, Mutations.updateDocument)
       }
       const dispatchDelete = document => {
-        dispatchChange(document, Mutations.deleteDocument)
+        dispatchChange(
+          { ...document, _deleted: true },
+          Mutations.deleteDocument
+        )
```
ptbrowne, line 35: What was the problem otherwise ?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -49,27 +51,42 @@ const FolderViewBody = ({
...
+        )}
         <FileListBodyV2 selectionModeActive={false}>
           <AddFolder />
           {isInError && <Oops />}
           {isLoading && <FileListRowsPlaceholder />}
-          {isEmpty && <EmptyDrive canUpload={true} />}
+          {/* TODO FolderViewBody should not have the responsability to chose 
+          which empty component to display. It should be done by the "view" itself. 
+          But adding a new prop like <FolderViewBody emptyComponent={} 
+          is not good enought too */}
```
ptbrowne, line 81: Why is the emptyComponent prop not good ? 

s/enought/enough
s/too/? (not grammatically correct)

src/drive/web/modules/trash/Toolbar.jsx
Crash--: See your comment https://github.com/cozy/cozy-drive/pull/2044#discussion_r446298402 :D 

We receive the `DELETE_DOCUMENT` from the realtime. And then, we add the `_deleted:true` attribute to the file, and CozyClient has a build in mechanism to hide those documents with `deleted: true`

ptbrowne: I find it brittle to rely on the realtime for this, I find it a bit of a hack. Suppose we turn it off after a little bit of idle time, since it can be costly to keep a realtime connection to our servers, and then the trash would not work. In one of my PRs I had to do an adhoc receiveQueryResult to correctly keep the client store up-to-date. It's not perfect but I think I'd prefer it over the reliance on the realtime.

Here the main thing that bothers me is is that it's not very explicit that we rely on the realtime to keep cozy-client's collection updated.

src/drive/web/modules/views/Trash/Breadcrumb.jsx
ptbrowne: This file is removed in a later commit, this comment does not apply.

src/drive/web/modules/views/Trash/useFilesQueryWithPath.jsx
ptbrowne: Removed in a later commit.

PR #2045, opened 2020-06-26
test: Dropzone test

ptbrowne:
~~Dropzone get displayedFolder from the state~~

Dropzone gets displayed folder as prop.

~~I don't know where is sharing state supposed to come from, I don't see it specified in FolderView's Dropzone :thinking:~~

EDIT: It seems to be coming from withSharingState HOC.


ptbrowne:
@y-lohse Do you know where is the sharing state is supposed to come from ?

ptbrowne:
I've edited my PR since FolderView passes the displayedFolder to Dropzone.

PR #2047, opened 2020-06-26
Share view

y-lohse:
This PR migrates the Sharings View. Only the last 2 commits are relevant, the other ones are from #2044 .

### DIFF
--- a/src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
+++ b/src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
@@ -39,7 +39,7 @@ const mapStateToProps = state => ({
 
 const FilesViewerWithFolderId = connect(mapStateToProps)(FilesViewerWithQuery)
 
-const SharingsContainer = props => (
+const FilesViewerSharing = props => (
   <SharedDocuments>
     {({ sharedDocuments }) => (
       <FilesViewerWithFolderId {...props} sharedDocumentIds={sharedDocuments} />
@@ -47,4 +47,4 @@ const SharingsContainer = props => (
   </SharedDocuments>
 )
 
-export default SharingsContainer
+export default FilesViewerSharing
--- a/src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
+++ b/src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
@@ -7,7 +7,7 @@ import { connect } from 'react-redux'
 import FilesViewer, {
   FilesViewerLoading
 } from 'drive/web/modules/viewer/FilesViewerV2'
-import SharedDocuments from 'cozy-sharing/dist/components/SharedDocuments'
+import withSharedDocumentIds from './withSharedDocumentIds'
 
 const FilesViewerWithQuery = ({ sharedDocumentIds, ...props }) => {
   const filesQuery = buildSharingsQuery(sharedDocumentIds)
@@ -39,12 +39,4 @@ const mapStateToProps = state => ({
 
 const FilesViewerWithFolderId = connect(mapStateToProps)(FilesViewerWithQuery)
 
-const FilesViewerSharing = props => (
-  <SharedDocuments>
-    {({ sharedDocuments }) => (
-      <FilesViewerWithFolderId {...props} sharedDocumentIds={sharedDocuments} />
-    )}
-  </SharedDocuments>
-)
-
-export default FilesViewerSharing
+export default withSharedDocumentIds(FilesViewerWithFolderId)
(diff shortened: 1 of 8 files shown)


ptbrowne:
> Only the last 2 commits are relevant, the other ones are from #2044 .

You can change the base branch of the PR ;)

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const SharingsContainer = props => (
+  <SharedDocuments>
+    {({ sharedDocuments }) => (
+      <FilesViewerWithFolderId {...props} sharedDocumentIds={sharedDocuments} />
+    )}
+  </SharedDocuments>
+)
+
+export default SharingsContainer
```
ptbrowne, line 50: Nit: I find it always a bit surprising when a file's default export is not named as the file's name.

Nit: It seems a withSharedDocuments HOC could be shared between this file and Sharings/index.jsx file

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
```diff
@@ -0,0 +1,50 @@
...
+  FilesViewerLoading
+} from 'drive/web/modules/viewer/FilesViewerV2'
+import SharedDocuments from 'cozy-sharing/dist/components/SharedDocuments'
+
+const FilesViewerWithQuery = ({ sharedDocumentIds, ...props }) => {
+  const filesQuery = buildSharingsQuery(sharedDocumentIds)
+  const results = useQuery(filesQuery.definition, filesQuery.options)
+  const { router } = useRouter()
+  if (results.data) {
+    const viewableFiles = results.data.filter(f => f.type !== 'directory')
```
ptbrowne, line 17: I think we do not need the filter since the query is already on the files.

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { useRouter } from 'drive/lib/RouterContext'
+import { buildSharingsQuery } from 'drive/web/modules/queries'
+import { getCurrentFileId } from 'drive/web/modules/selectors'
+import { connect } from 'react-redux'
+import FilesViewer, {
+  FilesViewerLoading
+} from 'drive/web/modules/viewer/FilesViewerV2'
+import SharedDocuments from 'cozy-sharing/dist/components/SharedDocuments'
+
+const FilesViewerWithQuery = ({ sharedDocumentIds, ...props }) => {
```
ptbrowne, line 12: Nit: SharingsFileviewer or FilesViewerSharings to keep the same name as the file ?

Nit: Can you add a docstring ?

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
```diff
@@ -0,0 +1,50 @@
...
+  const filesQuery = buildSharingsQuery(sharedDocumentIds)
```
ptbrowne, line 13: I am wondering if the query or its results should not come directly from the parent since they have a clear parent/child relationship and since they use the same query. Wouldn't it be simpler to use directly the raw FilesViewer in the SharingsView ?

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
y-lohse: Forgot to renameit!

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
y-lohse: This particular query is on both files and folders, so we do need it :)

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
y-lohse: I wonder about this too, but we also want the viewer to have it's own URL so we have to rely on react-router to a degree for this. It's a good point but I'd rather tackle this in another PR when we decide to do the change.

ptbrowne: OK, no problem.

src/drive/web/modules/views/Sharings/FilesViewerSharings.jsx
ptbrowne: nit: I think what confused me was the variable named `filesQuery`, it could maybe be renamed to sharingQuery ?

y-lohse: Well it's also used in the recent view, and it's in the sense of the `files` doctype, so I do think the name is good as it is :)

ptbrowne: OK

ptbrowne:
So... no tests ? Are they coming later ?

y-lohse:
Yes! Most of the view is using the same components as Recent/Drive which have tests, so most of the rendering does in fact have test coverage. We'll be adding tests for the specifics of the various views in another PR.

PR #2048, opened 2020-06-29
Update cozy-scripts to 5

ptbrowne:
I could not upgrade completely to 4 due to errors (I do not remember
them though but I feel like they should not be so hard to overcome).

This is a first step towards using the latest version. After that,
I think it will be a good thing to add the SpeedMeasurePlugin and
learn more about what is taking so much time in webpack (I personally
have 7s recompile time on save).


ptbrowne:
There is a problem for the CreateNoteItem where a hook is called conditionally. https://travis-ci.org/github/cozy/cozy-drive/jobs/703071426#L309

However I haven't found a place where CreateNoteItem is rendered with hasAppsPermissions set to false, is the `false` part of the if still relevant ?

y-lohse:
> However I haven't found a place where CreateNoteItem is rendered with hasAppsPermissions set to false, is the false part of the if still relevant ?

This is for the public view, but not used at the moment. We plan on adding it again but I think we can use git's history when the time comes, so I think it's ok to delete that part.

ptbrowne:
Fixed in 6b30e4b0. Tell me if the PR is OK for you. I want to merge it ASAP because it can cause conflicts easily.

Crash--:
On attend de stabiliser la migration avant de merger cette PR. Je prendrai le taff de rebase :)

Crash--:
Currently updating to CS 4. Need to wait for https://github.com/cozy/create-cozy-app/pull/1334

To fix: 
Also one error that seems to appear from our intents stuff
Capture d’écran 2020-07-21 à 14 13 35

Also few warnings 
Capture d’écran 2020-07-21 à 14 45 13

Crash--:
Our migration to CC is not finished yet, so still in stand by

Crash--:
todo perso: module.hot.accept

Crash--:
@ptbrowne I think it's good to review now 👍

package.json
```diff
@@ -15,7 +15,7 @@
     "build:photos": "npm-run-all --parallel build:photos:browser",
     "build:photos:browser": "COZY_APP_SLUG=photos cozy-scripts build --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
     "watch:photos": "npm run watch:photos:browser",
-    "watch:photos:browser": "COZY_APP_SLUG=photos cozy-scripts watch --debug --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
+    "watch:photos:browser": "COZY_APP_SLUG=photos cozy-scripts start --no-stack --hot --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
```
ptbrowne, line 18: I think --no-stack and --hot are by defualt right now ?

app.config.js
```diff
@@ -69,6 +69,7 @@ if (
   configurationFiles[0].multiple.services.__mergeStrategy.strategy[
     'resolve.modules'
   ] = 'replace'
+
```
ptbrowne, line 72: We can remove this commit I guess.

yarn.lock
```diff
@@ -14448,7 +14448,7 @@ react-fast-compare@^3.0.1:
   resolved "https://registry.yarnpkg.com/react-fast-compare/-/react-fast-compare-3.2.0.tgz#641a9da81b6a6320f270e89724fb45a0b39e43bb"
   integrity sha512-rtGImPZ0YyLrscKI9xTpV8psd6I8VAtjKCzQDlzyDvqJA8XOW78TXYQwNRNd8g8JZnDu8q9Fu/1v4HPAVwVdHA==
 
-react-hot-loader@^4.12.19, react-hot-loader@^4.3.11:
+react-hot-loader@^4.3.11:
```
ptbrowne, line 14435: Why is there still a react-hot-loader here ? Is it wrongly declared in one of our dependency ?

ptbrowne (commented):
LGTM !

yarn.lock
Crash--: => Found "react-hot-loader@4.12.21"
info Reasons this module exists
   - "cozy-ui" depends on it
   - Hoisted from "cozy-ui#react-hot-loader"
   - Hoisted from "cozy-bar#cozy-ui#react-hot-loader"

I guess the bar is not up to date with your latest change in UI (https://github.com/cozy/cozy-ui/commit/9e6f388df56209e4f6320225ab8c7082c90d75cf) 


(and thanks for to remind me I needed to check the lock file 👍 )

Crash--: after a clean of my lock file:

=> Found "react-hot-loader@4.13.0"
info Reasons this module exists
   - "cozy-bar#cozy-ui" depends on it
   - Hoisted from "cozy-bar#cozy-ui#react-hot-loader"

way better ;)

PR #2049, opened 2020-06-29
fix: Only params from matching route  are extracted to redux

y-lohse:
So far, we extracted the first router params we found into the redux store. But several routes could be matching, with different sets of params, and only the first in the list would be extracted, not necessarily matching the correct route.

With this PR, we check for the most probable route and extract those params.

### DIFF
--- a/src/drive/store/connectedRouter.js
+++ b/src/drive/store/connectedRouter.js
@@ -1,7 +1,8 @@
 /**
  * Tools to store part of the react-router state inside redux
  */
-import { getParams } from 'react-router/lib/PatternUtils'
+import { matchPattern } from 'react-router/lib/PatternUtils'
+import fromPairs from 'lodash/fromPairs'
 
 const LOCATION_CHANGED = 'HISTORY.LOCATION_CHANGED'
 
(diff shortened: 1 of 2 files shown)


src/drive/store/connectedRouter.spec.js
```diff
@@ -3,10 +3,12 @@ import { connectStoreToHistory, createReducer } from './connectedRouter'
...
 
 const routes = [
-  '/folder/:folderId/file/:fileId',
-  '/files/:folderId/file/:fileId',
   '/files/:folderId',
-  '/folder/:folderId'
+  '/folder/:folderId',
+  '/files/:folderId/file/:fileId',
+  '/folder/:folderId/file/:fileId',
+  'sharings/files/:fileId',
```
ptbrowne, line 10: Is it expected to have a leading / here ?

src/drive/store/connectedRouter.js
```diff
@@ -13,10 +14,14 @@ const createParamsExtractor = routes => {
...
-        return params
-      }
+      const match = matchPattern(route, pathname)
+      if (match && match.remainingPathname === '')
+        return fromPairs(
+          match.paramNames.map((paramName, index) => [
+            paramName,
+            match.paramValues[index]
+          ])
+        )
```
ptbrowne, line 24: I though react-router would use the first route that matches ? To have the same kind of behavior, shouldn't we use the same thing, and sort of routes so that the matching is the same as react-router ?

src/drive/store/connectedRouter.spec.js
y-lohse: Nope, good catch!

src/drive/store/connectedRouter.js
y-lohse: From what I understand, it uses the first route that matches without any `remainingPathname`. `matchPattern ` seemed like the highest possible level I could borrow in terms of having a matching behavior, anything above that expects a full router configuration to work with.

ptbrowne: OK thanks

PR #2050, opened 2020-06-29
Trash view/viewer

Crash--:
Even the root of the TrashView is a folder. Let's use only one "TrashFolderView" component and only one route to handle the Trash view (we've done the same for the "Drive View"). 

Add the FileViewerTrash to the TrashView. We can now view the file in the viewer even in we're in the trash


src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -36,7 +36,10 @@ const FolderViewBody = ({
     setSortOrder({ sortAttribute: attribute, sortOrder: order })
   )
 
-  const isInError = queryResults.some(query => query.fetchStatus === 'error')
+  const isInError = queryResults.some(query => {
+    console.log('query', query)
```
ptbrowne, line 40: Careful, a log is here.

Crash--: look at the associated commit message ^^"

Crash--: I don't know why the test is green on my machine and not on travis.

PR #2051, opened 2020-06-29
refactor: Modular view actions

y-lohse:
This PR refactors the actions for our various views, so that each view can easily pick and choose which actions to use without duplicating anything.

We switch from `useActions(options) // -> a hard-coded array of actions` to `useActions([delete, upload, etc], options)`.


src/drive/web/modules/actions/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+  isAnyFileReferencedByAlbum,
+  exportFilesNative,
+  downloadFiles,
+  trashFiles,
+  openFileWith,
+  restoreFiles
+} from './utils'
+
+export const share = ({ hasWriteAccess }) => {
+  const { pushModal, popModal } = useContext(ModalContext)
```
ptbrowne, line 33: I think we are going in the right direction with the modular actions.

But, I am unconfortable with the fact that we use hooks outside rendering functions. Hooks are really weird and I do not think that it's a good idea to have them far away from the rendering function. I think an object containing { modalContext, client } could be passed to actions.

y-lohse: I went with that approach, the options object is created in the component that calls `useAction`, so all the hook calls are close to the actions. And views that only use a few options can import just what they need in the options.

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -26,7 +42,26 @@ export const RecentView = ({ router, children }) => {
...
+    client,
+    pushModal,
+    popModal,
+    refresh,
+    dispatch,
+    router,
+    location,
+    hasWriteAccess: false,
+    canMove: true
+  }
```
ptbrowne, line 60: Nit: Since those are directly as props, I think this can be renamd into actionsProps. Similarly, at the implementation it, s/actionsOptions/actionsProps.

ptbrowne:
🎉 Great, I think the code is in a much better state now !

PR #2052, opened 2020-06-30
feat: Migrate RenameFile to CozyClient

Crash--:
I duplicated & simplified the reducer, it'll be easier to remove after
the end of the migration.

We now only keep 2 actions in the reducer : start and abort renaming.
The rest is done by the RenameInput component itself

Used the client.save() to update the redux store directly.


src/drive/web/modules/drive/RenameInputV2.jsx
```diff
@@ -0,0 +1,41 @@
...
+import { withClient } from 'cozy-client'
+import Alerter from 'cozy-ui/transpiled/react/Alerter'
+
+import FilenameInput from 'drive/web/modules/filelist/FilenameInput'
+import { abortRenaming } from './renameV2'
+
+const updateFileNameQuery = async (client, file, newName) => {
+  return client
+    .collection('io.cozy.files')
+    .updateAttributes(file.id, { name: newName })
```
ptbrowne, line 13: I think it'd be better to use `client.save()` here as the document would be updated in the store without waiting for the realtime.

```
client.save({ ...file, name: newName })
```

src/drive/web/modules/drive/RenameInputV2.jsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const mapDispatchToProps = dispatch => ({
+  abortRenaming: () => dispatch(abortRenaming())
+})
+
+export default connect(
+  null,
+  mapDispatchToProps
+)(withClient(RenameInput))
```
ptbrowne, line 41: Nit: useClient could be used to have one less level of nesting. At the end of the migration, the withClient to useClient codemod could be used.

src/drive/web/modules/drive/RenameInputV2.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const RenameInput = ({ abortRenaming, file, client }) => {
+  return (
+    <FilenameInput
+      name={file.name}
+      onSubmit={async newName => {
+        try {
+          await updateFileNameQuery(client, file, newName)
+        } catch {
+          Alerter.error('alert.folder_name', { folderName: name })
```
ptbrowne, line 24: Is it expected to have "folderName" here (instead of fileName) ?

src/drive/web/modules/drive/RenameInputV2.jsx
```diff
@@ -0,0 +1,41 @@
...
+import FilenameInput from 'drive/web/modules/filelist/FilenameInput'
+import { abortRenaming } from './renameV2'
+
+const updateFileNameQuery = async (client, file, newName) => {
+  return client
+    .collection('io.cozy.files')
+    .updateAttributes(file.id, { name: newName })
+}
+
+const RenameInput = ({ abortRenaming, file, client }) => {
```
ptbrowne, line 16: Nit: s/abortRenaming/onAbort.

src/drive/web/modules/drive/RenameInputV2.jsx
Crash--: Yep, we're doing it ;).

src/drive/web/modules/drive/RenameInputV2.jsx
Crash--: I used the existing string. I created a new one. And thanks for the comment, it appears that I used the old `name` instead of `newName`...

ptbrowne:
💯 for the test with react test utils.

PR #2055, opened 2020-07-01
Public view

y-lohse:
This PR still needs tests but other than that it's ready to review. You'll especially notice the `usePublicFilesQuery.jsx` file which is… less than ideal.

To recap:, this is the public view, where we may only have permissions to read/write on one specific io.cozy.files instead of the doctype as a whole. 

This means that:

- We can't use `Q().where()`as it translates into a POST request. The only method we can use is FileCollection.statById and use the `included` results.
- `Q('io.cozy.files').getById(id)` calls `statById`, but we also need to paginate the `included` results. This is impossible with the DSL as far as I can tell, so we can't use a query.
- Not using a query means changes to the store are not automatically reflected in the UI
- We also don't have realtime notifications to update the store

So we end up with a custom hook that emulates the query API but is not really one.  
The `included` results in statById are added to the cozy-client store, so there may be a way to retrieve them through a query, but we couldn't find one. `Q().getByIds()` is forbidden on the stack, and setting a fetch policy of `noFetch` leaves the query perpetually pending. Maybe we could write a selector that directly looks into the cozy-client store and retrieves the documents by id… but frankly having the `POST /files/find` route open on the cozy-stack would solve all these problems.

In order to reflect changes to the UI, we drill a `refreshFolderContent` prop to bits of code that write to `io.cozy.files`. A possible improvement would be to add this function to a dedicated context and rely on `useContext` instead of prop drilling to pass it around, but it's a late idea and I'm not even sure it's much better.


src/drive/web/modules/selectors.js
```diff
@@ -28,6 +28,14 @@ export const getDisplayedFolder = rootState => {
   return null
 }
 
+export const getParentFolder = (rootState, parentFolderId) => {
+  if (parentFolderId) {
+    const doc = getDocumentFromState(rootState, 'io.cozy.files', parentFolderId)
```
Crash--, line 33: This is useful only if the parent has been loaded first. Is it always the case? What happen if we're accessing a subfolder directly?

src/drive/web/modules/views/Public/index.jsx
```diff
@@ -0,0 +1,185 @@
...
+  const client = useClient()
+
+  const [viewerOpened, setViewerOpened] = useState(false)
+  const [currentViewerIndex, setCurrentViewerIndex] = useState(null)
+
+  const filesResult = usePublicFilesQuery(currentFolderId)
+  const files = filesResult.data
+  const viewableFiles = files.filter(f => f.type !== 'directory')
+
+  const refreshFolderContent = () => filesResult.forceRefetch() // We don't have enough permissions to rely on the realtime notifications or on a cozy-client query to update the view when something changes, so we realod the view instead
```
Crash--, line 70:
```suggestion
  const refreshFolderContent = () => filesResult.forceRefetch() // We don't have enough permissions to rely on the realtime notifications or on a cozy-client query to update the view when something changes, so we relaod the view instead
```

src/drive/web/modules/views/Public/index.jsx
```diff
@@ -0,0 +1,185 @@
...
+  const navigateToFile = useCallback(async file => {
+    const isNote = models.file.isNote(file)
+    if (isNote) {
+      try {
+        const noteUrl = await models.note.fetchURL(client, file)
+        const url = new URL(noteUrl)
+        if (!isMobileApp()) {
+          url.searchParams.set('returnUrl', window.location.href)
+        }
+        window.location.href = url.toString()
```
Crash--, line 85: I'm wondering... Did we do that for the private view? Why can't we put that in the FileOpener?

src/drive/web/modules/views/Public/usePublicFilesQuery.jsx
```diff
@@ -0,0 +1,76 @@
...
+  // Most stack routes are off-limit when we have a read-only token, so we use a simple GET to load the folder content.
+  // no query because we need to paginate the included files
+  const { included, links } = await client
+    .collection('io.cozy.files')
+    .statById(folderId, {
+      'page[cursor]': cursorToUse
+    })
+
+  const nextLink = get(links, 'next', '')
+  const nextLinkURL = new URL(`http://example.com${nextLink}`)
```
Crash--, line 15: What is this exemple.com?

src/drive/web/modules/selectors.js
y-lohse: > What happen if we're accessing a subfolder directly?

You're right, in that case we don't show a parent breadcrumb. This is where we would need a dedicated network request for the breadcrumb.

src/drive/web/modules/views/Public/usePublicFilesQuery.jsx
y-lohse: I've clarified this part of the code and added a comment.

PR #2059, opened 2020-07-03
fix: Show confirm modal before emptying the trash

y-lohse:
On the migrated version, we were deleting the trash straight away instead of showing a confirmation modal.

~Still needs unit tests~

### DIFF
--- a/src/drive/web/modules/trash/Toolbar.spec.jsx
+++ b/src/drive/web/modules/trash/Toolbar.spec.jsx
@@ -1,0 +1,53 @@
+import React from 'react'
+import { render, fireEvent, act } from '@testing-library/react'
+import { waitForElementToBeRemoved } from '@testing-library/dom'
+import flag from 'cozy-flags'
+import { createMockClient } from 'cozy-client'
+import { ModalContextProvider, ModalStack } from 'drive/lib/ModalContext'
+
+import AppLike from 'test/components/AppLike'
+import { Toolbar } from './Toolbar'
+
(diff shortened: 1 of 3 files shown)


src/drive/web/modules/trash/Toolbar.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+    const actuallyCalledModal = mockModalContextValue.pushModal.mock.calls[0][0]
+    expect(mockModalContextValue.pushModal).toHaveBeenCalledWith(
+      <EmptyTrashConfirm
+        onConfirm={actuallyCalledModal.props.onConfirm}
+        onClose={actuallyCalledModal.props.onClose}
+      />
+    )
+
+    actuallyCalledModal.props.onConfirm()
```
Crash--, line 52: Why not `fireEvent.click(getByText(confirm))` (i don't know the exact string)and then check if emptyTrash has been called?

y-lohse: Because we have mocked the modal context and aren't rendering the ModalStack, so the confirmation modal is never actually rendered during the test.

Crash--: So... Why are we mocking the Modal context? :D Why not been as close as possible than the use of this component ? And instead of checking if the pushModal was called with the right component, we check if we have the text "are you sure to delete the files" ?

PR #2060, opened 2020-07-06
fix: Display SyncStatus on mobile

Crash--:
Add MediaBackupProgression / FirstUpload & RatingModal on the new Drive View 

Added few tests for the startMediaBackup 
Use CozyClient to update the last Sync date

### DIFF
--- a/src/drive/mobile/modules/mediaBackup/duck/index.spec.js
+++ b/src/drive/mobile/modules/mediaBackup/duck/index.spec.js
@@ -1,0 +1,142 @@
+import { startMediaBackup } from './index'
+import { createMockClient } from 'cozy-client/dist/mock'
+import { isAuthorized, getPhotos } from 'drive/mobile/lib/media'
+jest.mock('drive/mobile/lib/media', () => ({
+  ...jest.requireActual('drive/mobile/lib/media'),
+  isAuthorized: jest.fn(),
+  getPhotos: jest.fn()
+}))
+
+import { setBackupImages } from 'drive/mobile/modules/settings/duck'
+jest.mock('drive/mobile/modules/settings/duck', () => ({
+  ...jest.requireActual('drive/mobile/modules/settings/duck'),
+  setBackupImages: jest.fn()
+}))
+
+import { getUploadDir, uploadPhoto } from './files'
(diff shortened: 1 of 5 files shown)


src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -51,6 +56,13 @@ const FolderViewBody = ({
 
   return (
     <>
+      {__TARGET__ === 'mobile' && (
+        <div>
+          {currentFolderId === ROOT_DIR_ID && <MediaBackupProgression />}
```
y-lohse, line 61: I think it would be better to add these components in DriveView rather than in the generic file with a condition :)

Crash--: I'm wondering if we should not display this message on each views since the backup can have a big impact on the app performance... 🤔 

But OK, let's move this one on the driveview for now

src/drive/mobile/modules/settings/RatingModal.jsx
```diff
@@ -80,9 +80,9 @@ RatingModal.propTypes = {
 }
 
 // sub-components
-const EnjoyCozy = (props, context) => {
-  const { onReply } = props
-  const { t } = context
+const EnjoyCozy = props => {
+  const { onReply, t } = props
```
y-lohse, line 84: Nit: we could use `useI18n` here instead of relying on props :)

src/drive/mobile/modules/mediaBackup/duck/index.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      attributes: {
+        path: uploadDirPath
+      }
+    })
+  })
+  it('starts the mediabackup with one photo to upload', async () => {
+    getPhotos.mockResolvedValue(mockedPhotoOnDevice)
+    mockedGetState.mockReturnValue(mockedState)
+    isAuthorized.mockResolvedValue(true)
+    const req = startMediaBackup(true)
```
y-lohse, line 65: It might be good to have variables for these 2, just to know what they do.

PR #2063, opened 2020-07-07
fix: Trash query for pouch

y-lohse:
Pouch needs all fields in the selector, like the other queries.


ptbrowne:
Is it something that could be done at cozy-client level ? If so, could you open an issue ?

Crash--:
Thanks @ptbrowne done in https://github.com/cozy/cozy-client/issues/731

PR #2067, opened 2020-07-09
fix: Don't subscribe to Store change, but save it before closing the app

Crash--:
Before, we were subscribing to all the store changes. resulting in a lot of localStorage write...

Now we only saveState when the app goes in background.


src/drive/targets/mobile/InitAppMobile.jsx
```diff
@@ -173,6 +174,18 @@ class InitAppMobile {
         text: t('mobile.notifications.backup_paused')
       })
     } */
+    const currentState = store.getState()
```
y-lohse, line 177: Can we add a comment explaining that onPause is in fact also called before the app is closed?

PR #2069, opened 2020-07-10
The big cleanup

y-lohse:
This PR removes code that is no longer needed after completing the migration, and there's a lot of it.

The commits that jump out are 2ebfb3f 3ee3e88 620fedf because they migrate some stuff that fell through the cracks. The rest is just removing files and I'm not sure how to really help with the review. Some parts of the folder structure don't make too much sense anymore but I figured we can move things around in a later PR.


src/drive/web/modules/move/Explorer.jsx
```diff
@@ -1,18 +1,19 @@
...
 const Explorer = ({ children }) => (
-  <FileList canSort={false} fileActions={null} withSelectionCheckbox={false}>
-    <div
-      data-test-id="fil-content-modal"
-      className={fileListStyles['fil-content-body']}
-    >
-      {children}
-    </div>
-  </FileList>
+  <FileListv2>
```
ptbrowne: No need for the V2 anymore right ?

src/drive/web/modules/drive/Toolbar/components/DownloadButtonItem.jsx
```diff
@@ -26,10 +24,7 @@ const DownloadButtonItem = ({ t, displayedFolder, downloadAll }) => {
 const mapDispatchToProps = (dispatch, ownProps) => ({
   downloadAll: folder => {
     const client = ownProps.client
-    const action = flag('drive.client-migration.enabled')
-      ? () => downloadFilesV2(client, folder)
-      : () => dispatch(downloadFiles(folder))
-    return action()
+    return downloadFilesV2(client, folder)
```
ptbrowne: No need for V2

src/drive/web/modules/trash/Toolbar.jsx
```diff
@@ -38,18 +34,16 @@ export const Toolbar = ({
...
-          />
-        )
-      })
-    : useCallback(emptyTrash)
+  const onEmptyTrash = useCallback(() => {
+    pushModal(
+      <EmptyTrashConfirm
+        onClose={popModal}
+        onConfirm={() => {
+          client.collection('io.cozy.files').emptyTrash()
```
ptbrowne, line 42: This should be awaited ?

ptbrowne: What happens if there is an error during emptying of the trash ?

y-lohse: Good catch, the migration on this is incomplete.

PR #2072, opened 2020-07-16
Variety of fixes

y-lohse:
Several fixes for issues that have popped up after completing the migration. Individual commits should give a good idea of what's going on.

Still needs https://github.com/cozy/cozy-client/pull/740 but ready to review.

### DIFF
--- a/src/drive/lib/RealTimeQueries.jsx
+++ b/src/drive/lib/RealTimeQueries.jsx
@@ -1,12 +1,15 @@
 import { memo, useEffect } from 'react'
-import { useClient, Mutations } from 'cozy-client'
+import { useClient, Mutations, models } from 'cozy-client'
 import { receiveMutationResult } from 'cozy-client/dist/store'
 
 const RealTimeQueries = ({ doctype }) => {
   const client = useClient()
 
   const dispatchChange = (document, mutationDefinitionCreator) => {
-    const response = { data: { ...document, _type: doctype } }
+    const response = {
+      data: models.file.normalize(document)
+    }
+
     const options = {}
     client.dispatch(
       receiveMutationResult(
--- a/src/drive/lib/registerClientPlugins.js
+++ b/src/drive/lib/registerClientPlugins.js
@@ -1,0 +1,7 @@
+import { RealtimePlugin } from 'cozy-realtime'
+
+const registerClientPlugins = client => {
+  client.registerPlugin(RealtimePlugin)
+}
+
+export default registerClientPlugins
--- a/src/drive/mobile/lib/cozy-helper.js
+++ b/src/drive/mobile/lib/cozy-helper.js
@@ -1,18 +1,19 @@
-/* global cozy, document */
+/* global cozy, */
 import { LocalStorage as Storage } from 'cozy-client-js'
 import CozyClient, { StackLink } from 'cozy-client'
 import PouchLink from 'cozy-pouch-link'
 import { RealtimePlugin } from 'cozy-realtime'
-import { isMobileApp, isIOSApp } from 'cozy-device-helper'
+import { isMobileApp, isIOSApp, getDeviceName } from 'cozy-device-helper'
 
 import { SOFTWARE_ID, SOFTWARE_NAME } from './constants'
 import { disableBackgroundService } from './background'
 import { schema, DOCTYPE_FILES } from 'drive/lib/doctypes'
-export const getLang = () =>
-  navigator && navigator.language ? navigator.language.slice(0, 2) : 'en'
-import { getDeviceName } from 'cozy-device-helper'
 import appMetadata from 'drive/appMetadata'
 import { getRedirectUri } from 'drive/mobile/lib/redirect'
+
+export const getLang = () =>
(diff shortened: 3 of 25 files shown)


src/drive/mobile/lib/cozy-helper.js
```diff
@@ -1,18 +1,19 @@
-/* global cozy, document */
+/* global cozy, */
```
ptbrowne, line 1:
```suggestion
/* global cozy */
```

src/drive/targets/mobile/InitAppMobile.jsx
```diff
@@ -23,11 +23,8 @@ import {
   startHeartBeat,
   stopHeartBeat
 } from 'drive/mobile/lib/tracker'
-import {
-  getLang,
-  initClient,
-  registerClientPlugins
-} from 'drive/mobile/lib/cozy-helper'
+import { getLang, initClient } from 'drive/mobile/lib/cozy-helper'
+import registerClientPlugins from 'drive/lib/registerClientPlugins'
```
ptbrowne, line 27: Could  registerClientPlugins be in initClient ?

src/drive/lib/registerClientPlugins.js
```diff
@@ -0,0 +1,7 @@
+import { RealtimePlugin } from 'cozy-realtime'
+
+const registerClientPlugins = client => {
+  client.registerPlugin(RealtimePlugin)
+}
+
+export default registerClientPlugins
```
ptbrowne, line 7: I wonder why this is not done at the place where the client is initialized ?

src/drive/lib/RealTimeQueries.jsx
```diff
@@ -1,12 +1,15 @@
 import { memo, useEffect } from 'react'
-import { useClient, Mutations } from 'cozy-client'
+import { useClient, Mutations, models } from 'cozy-client'
```
ptbrowne, line 2: File global remark: At this point if we think that it's useful we can move this file to cozy-client ?

src/drive/web/modules/drive/DeleteConfirm.jsx
```diff
@@ -61,4 +79,4 @@ const DeleteConfirmWithSharingContext = ({ files, ...rest }) =>
     </SharedDocument>
   )
 
-export default translate()(DeleteConfirmWithSharingContext)
+export default DeleteConfirmWithSharingContext
```
ptbrowne, line 82: Remark: theere is a codemod to do the change from translate() to useI18n() automatically. It might be good to run on the whole drive codebase.

```
jscodeshift --parser babel -t ~/code/cozy/cozy-ui/codemods/use-i18n-hook/index.js src --extensions js,jsx
```

src/drive/web/modules/actions/index.jsx
```diff
@@ -67,7 +67,8 @@ export const trash = ({
   pushModal,
   popModal,
   hasWriteAccess,
-  refresh
+  refresh,
+  refreshFolderContent
```
ptbrowne, line 71: Why is the refreshFolderContent not done in the `refresh` callback passed from above ?

src/drive/targets/mobile/InitAppMobile.jsx
y-lohse: No because initClient is also used in the background process where we don't want the plugin to be registered. We also want to register the plugin on the web where we don't use initClient at all. All in all I think merging the 2 together would make the code harder to understand and maintain.

src/drive/lib/RealTimeQueries.jsx
y-lohse: I'll create a PR in cozy-client.

src/drive/targets/mobile/InitAppMobile.jsx
ptbrowne: >  All in all I think merging the 2 together would make the code harder to understand and maintain.

Why ? I think that having the same initialization code makes it easier to understand. Typically I think what is important to understand are the differences between mobile and web. Here since this is done in two different files, it's hard to compare, don't you think ? And why is the web version not using initClient ? Doesn't it have to initialize a client ?

y-lohse: The client initialization for web and mobile are wildly different (different links, token handling, etc). We could have a single function that inits cozy client for all contexts, but it would have a bunch of `if mobile / if background / etc` in it, which I don't think is preferable.

The other reason why I don't think we should make this change is that we need to import a lot of files to init cozy-client on mobile, most of which are not relevant at all on the web and should ideally not be in the bundle. The file structure in drive is not ideal, and it's easy to end up including way too many things because of chained imports. This is what caused the bug I'm addressing here in the first place — accidentally including cozy-helper in the intents bundle broke it.

ptbrowne: OK, fine by me.

src/drive/web/modules/drive/DeleteConfirm.jsx
y-lohse: Right, i'll do a separate PR for that!

PR #2075, opened 2020-07-20
Fix versions view

y-lohse:
The file versions views on Recent and Sharings were still broken, because we were accidentally passing the global `location` object to our actions instead of the `location` object provided by react-router, which includes the pathname of the current view etc.

I also had to tweak the sharing route configuration a little so that all components are in the right hierarchy — they all need to be children of `/sharing` in order for the nav  to stay active, but once inside a shared folder we want a different main route.

### DIFF
--- a/src/drive/web/modules/views/Public/index.jsx
+++ b/src/drive/web/modules/views/Public/index.jsx
@@ -56,6 +56,7 @@ const PublicFolderView = ({
   currentFolderId,
   parentFolder,
   router,
+  location,
   children
 }) => {
   const client = useClient()
--- a/src/drive/web/modules/views/Public/index.spec.jsx
+++ b/src/drive/web/modules/views/Public/index.spec.jsx
@@ -1,9 +1,11 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
+import { Router, hashHistory, Route, Redirect } from 'react-router'
 
 import { setupStoreAndClient } from 'test/setup'
 import AppLike from 'test/components/AppLike'
 import usePublicFilesQuery from './usePublicFilesQuery'
+import FileHistory from 'components/FileHistory'
 
 import { generateFileFixtures, getByTextWithMarkup } from '../testUtils'
 
@@ -12,9 +14,10 @@ import PublicFolderView from './index'
 jest.mock('./usePublicFilesQuery', () => {
   return jest.fn()
 })
+jest.mock('components/FileHistory', () => () => <div>FileHistory stub</div>)
 jest.mock('components/pushClient')
 
-describe('Recent View', () => {
+describe('Public View', () => {
   const setup = () => {
     const { store, client } = setupStoreAndClient()
 
@@ -26,7 +29,12 @@ describe('Recent View', () => {
 
     const rendered = render(
       <AppLike client={client} store={store}>
-        <PublicFolderView />
+        <Router history={hashHistory}>
+          <Redirect from="/" to="/folder/123" />
+          <Route path="folder(/:folderId)" component={PublicFolderView}>
+            <Route path="file/:fileId/revision" component={FileHistory} />
+          </Route>
+        </Router>
       </AppLike>
     )
     return { ...rendered, client }
@@ -51,7 +59,7 @@ describe('Recent View', () => {
       refreshFolderContent: jest.fn()
     })
 
-    const { getByText } = setup()
+    const { getByText, findByText } = setup()
     // Get the HTMLElement containing the filename if exist. If not throw
     const el0 = getByText(`foobar0`)
     // Check if the filename is displayed with the extension. If not throw
@@ -68,5 +76,11 @@ describe('Recent View', () => {
     expect(
       document.querySelectorAll("[data-test-id='fil-actionmenu-inner']").length
     ).toEqual(1)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    await expect(findByText('FileHistory stub')).resolves.toBeTruthy()
   })
 })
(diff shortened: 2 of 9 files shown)


src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -96,5 +100,11 @@ describe('Recent View', () => {
     expect(
       parentDiv1.getElementsByClassName('fil-file-path')[0].textContent
     ).toEqual(path)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    waitFor(() => expect(getByText('Current version')).toBeTruthy())
```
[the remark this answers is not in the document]
Crash--: I don't think you need to `waitFor` no? 

`getByText` waits by default. Something like 

```
expect(() => {
getByText()
}).not.toThrow();
```

should be enough

src/drive/web/modules/views/Sharings/index.jsx
```diff
@@ -31,7 +31,12 @@ import {
 import { buildSharingsQuery } from 'drive/web/modules/queries'
 import { useFilesQueryWithPath } from '../Recent/useFilesQueryWithPath'
 
-export const SharingsView = ({ router, sharedDocumentIds, children }) => {
+export const SharingsView = ({
+  router,
+  location,
+  sharedDocumentIds,
+  children
+}) => {
```
Crash--, line 39: you should also add the test to test the versions

src/drive/web/modules/views/Recent/index.spec.jsx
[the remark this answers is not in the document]
y-lohse: That was my first attempt, but it doens't work. I also don't think `getByText` waits, that's `findByText`. I'm not sure of the difference between `findByText`  and `waitFor(getByText)` though.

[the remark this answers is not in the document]
y-lohse: Cool I'll use findBy then since it's the same.

Event hough the `expect` is not needed, I find it good to add them as it makes it explicit that the line is really testing something, and not a forgotten line.

src/drive/web/modules/views/Sharings/index.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+          </Route>
+        </Route>
+      </Router>
+    </AppLike>
+  )
+  return { ...rendered, client }
+}
+
+describe('Sharings View', () => {
+  it('tests the recent view', async () => {
```
Crash--, line 49:
```suggestion
  it('tests the sharing view', async () => {
```

src/drive/web/modules/views/Public/index.jsx
```diff
@@ -56,6 +56,7 @@ const PublicFolderView = ({
   currentFolderId,
   parentFolder,
   router,
+  location,
   children
 }) => {
   const client = useClient()
```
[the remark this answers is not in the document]
Crash--: Hmmm, is it done ?

src/drive/web/modules/views/Public/index.spec.jsx
```diff
@@ -68,5 +75,11 @@ describe('Recent View', () => {
     expect(
       document.querySelectorAll("[data-test-id='fil-actionmenu-inner']").length
     ).toEqual(1)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    expect(findByText('Current version')).toBeTruthy()
```
ptbrowne, line 83: findBy methods need to be awaited. Here you are checking that a Promise is truthy.

You can 

```
await findByText('Current version')

// or the more explicit
await expect(findByText('Current version')).toResolve()
```

src/drive/web/modules/views/Public/index.spec.jsx
```diff
@@ -68,5 +75,11 @@ describe('Recent View', () => {
     expect(
       document.querySelectorAll("[data-test-id='fil-actionmenu-inner']").length
     ).toEqual(1)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    expect(findByText('Current version')).resolves.toBeTruthy()
```
ptbrowne, line 83: You need to await this expect.

src/drive/web/modules/views/Recent/index.spec.jsx
```diff
@@ -96,5 +100,11 @@ describe('Recent View', () => {
     expect(
       parentDiv1.getElementsByClassName('fil-file-path')[0].textContent
     ).toEqual(path)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    expect(findByText('Current version')).resolves.toBeTruthy()
```
ptbrowne, line 108: This expect must be awaited

await expect(findByText('Current version')).resolves.toBeTruthy()

src/drive/web/modules/views/Sharings/index.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+    const parentDiv1 = el1.closest('.fil-file')
+    expect(
+      parentDiv1.getElementsByClassName('fil-file-path')[0].textContent
+    ).toEqual(path)
+
+    // navigates  to the history view
+    const historyItem = getByText('History')
+    fireEvent.click(historyItem)
+
+    expect(findByText('Current version')).resolves.toBeTruthy()
```
ptbrowne, line 111: This expect must be awaited.

y-lohse:
Good catch on that test, it turned out to be a false positive. The history modal was rendered, but because it does it's own data fetching it was returning `null`, and the string we were looking up wasn't there.

Because the test renders a large tree of component, it becomes hard to mock the return values for cozy-client for all components. I didn't manage to mock the expected return values for the FileHistory component, so instead I stub it in the test to make it easier to lookup. The navigation itself is still tested though, so the initial bug would have been caught by the test.

PR #2076, opened 2020-07-21
Fix selection

y-lohse:
The selection on the public page was slightly broken, because we used to store only file ids and get the full file object by searching in the cozy-client redux store. However due to limitations on the public page, we don't use a query and so the documents are not in the store.

Instead the selection now stores the full object instead  of just the id.

I also enabled the trash and rename actions on the Recent and Sharings view… because we can.

### DIFF
--- a/src/drive/web/modules/selection/duck.js
+++ b/src/drive/web/modules/selection/duck.js
@@ -1,8 +1,5 @@
 import { combineReducers } from 'redux'
-import get from 'lodash/get'
-
-const getFileByIdFromCozyClient = (state, id) =>
-  get(state, ['cozy', 'documents', 'io.cozy.files', id])
+import uniqBy from 'lodash/uniqBy'
 
 // constants
 const SELECT_ITEM = 'SELECT_ITEM'
@@ -12,21 +9,21 @@ const SHOW_SELECTION_BAR = 'SHOW_SELECTION_BAR'
 const HIDE_SELECTION_BAR = 'HIDE_SELECTION_BAR'
 
 // selectors
-export const isSelected = (state, id) =>
-  state.selection.selected.find(_id => _id === id) !== undefined
-export const getSelectedIds = state => state.selection.selected
+export const isSelected = (state, file) =>
+  state.selection.selected.find(({ id }) => id === file.id) !== undefined
 export const isSelectionBarVisible = state =>
   state.selection.selected.length !== 0 || state.selection.isSelectionBarOpened
-export const getSelectedFiles = state =>
-  getSelectedIds(state).map(id => getFileByIdFromCozyClient(state, id))
+export const getSelectedFiles = state => {
+  return state.selection.selected
+}
 
 // actions
 export const showSelectionBar = () => ({ type: SHOW_SELECTION_BAR })
 export const hideSelectionBar = () => ({ type: HIDE_SELECTION_BAR })
 export const toggleSelectionBar = () => ({ type: TOGGLE_SELECTION_BAR })
-export const toggleItemSelection = (id, selected) => ({
+export const toggleItemSelection = (file, selected) => ({
   type: selected ? UNSELECT_ITEM : SELECT_ITEM,
-  id
+  file
 })
 
 // reducers
@@ -36,10 +33,9 @@ const selected = (state = [], action) => {
   }
   switch (action.type) {
     case SELECT_ITEM:
-      return [...state, action.id]
+      return uniqBy([...state, action.file], 'id')
     case UNSELECT_ITEM: {
-      const idx = state.indexOf(action.id)
-      return [...state.slice(0, idx), ...state.slice(idx + 1)]
+      return state.filter(({ id }) => action.file.id !== id)
     }
     case HIDE_SELECTION_BAR:
       return []
--- a/src/drive/web/modules/selection/duck.js
+++ b/src/drive/web/modules/selection/duck.js
@@ -1,5 +1,5 @@
 import { combineReducers } from 'redux'
-import uniqBy from 'lodash/uniqBy'
+import omit from 'lodash/omit'
 
 // constants
 const SELECT_ITEM = 'SELECT_ITEM'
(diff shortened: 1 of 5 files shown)


src/drive/web/modules/selection/duck.js
```diff
@@ -36,10 +33,9 @@ const selected = (state = [], action) => {
   }
   switch (action.type) {
     case SELECT_ITEM:
-      return [...state, action.id]
+      return uniqBy([...state, action.file], 'id')
```
Crash--, line 36: Can we add a todo to say that we should always index / normalize by id? In order to not having to deal with uniq / map filter?

src/drive/web/modules/selection/duck.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+  isSelectionBarVisible,
+  toggleItemSelection,
+  isSelected,
+  getSelectedFiles
+} from './duck.js'
+
+describe('selection store', () => {
+  const setup = () => createStore(combineReducers({ selection: reducer }))
+
+  const item1 = { id: 1, name: 'item 1' }
```
Crash--, line 15: Don't we have a few FILE_MOCK? It will be more relevant, no?

src/drive/web/modules/selection/duck.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+    // selecting one item
+    store.dispatch(toggleItemSelection(item1, false))
+    expect(isSelected(store.getState(), item1)).toBe(true)
+    expect(isSelected(store.getState(), item2)).toBe(false)
+    expect(isSelectionBarVisible(store.getState())).toBe(true)
+
+    // reselecting the same item does nothing
+    store.dispatch(toggleItemSelection(item1, false))
+    expect(isSelected(store.getState(), item1)).toBe(true)
```
Crash--, line 48: Hmm, I'm not agreed with that. If we select an already selected item it should be removed, no? 

OK the second argument seems to be "selected or not". Can we have a variable for that? In order to understand at the first read the code? 

But I still disagree with the comment / text : "reselecting the same item does nothing" is not accurate.

src/drive/web/modules/selection/duck.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+    expect(isSelectionBarVisible(store.getState())).toBe(true)
+
+    // deselecting the first item
+    store.dispatch(toggleItemSelection(item1, true))
+    expect(isSelected(store.getState(), item1)).toBe(false)
+    expect(isSelected(store.getState(), item2)).toBe(true)
+    expect(isSelectionBarVisible(store.getState())).toBe(true)
+
+    // deselecting the same item again does nothing
+    store.dispatch(toggleItemSelection(item1, true))
```
Crash--, line 63: same here.

src/drive/web/modules/views/Recent/index.jsx
```diff
@@ -55,11 +54,11 @@ export const RecentView = ({ router, children }) => {
     dispatch,
     router,
     location,
-    hasWriteAccess: false,
+    hasWriteAccess: true,
     canMove: true
   }
   const actions = useActions(
-    [share, download, trash, open, rename, move, qualify, versions, offline],
+    [download, trash, open, rename, move, qualify, versions, offline],
```
Crash--, line 61: we can't share anymore on the recent view?

src/drive/web/modules/views/Sharings/index.jsx
```diff
@@ -56,11 +55,11 @@ export const SharingsView = ({ router, sharedDocumentIds, children }) => {
     dispatch,
     router,
     location,
-    hasWriteAccess: false,
+    hasWriteAccess: true,
     canMove: true
   }
   const actions = useActions(
-    [share, download, trash, open, rename, move, qualify, versions, offline],
+    [download, trash, open, rename, move, qualify, versions, offline],
```
Crash--, line 62: same in the sharings view ? 🤔

src/drive/web/modules/selection/duck.js
```diff
@@ -12,21 +9,21 @@ const SHOW_SELECTION_BAR = 'SHOW_SELECTION_BAR'
 const HIDE_SELECTION_BAR = 'HIDE_SELECTION_BAR'
 
 // selectors
-export const isSelected = (state, id) =>
-  state.selection.selected.find(_id => _id === id) !== undefined
-export const getSelectedIds = state => state.selection.selected
+export const isSelected = (state, file) =>
+  state.selection.selected.find(({ id }) => id === file.id) !== undefined
```
ptbrowne, line 13: It seems like the selected is a list. It seems like changing it to be a hash would improve the perf of the existence checks into O(1).

src/drive/web/modules/selection/duck.js
ptbrowne: Again, it seems like a Map(id => object) would be better suited as the data structure for the selected slice.

ptbrowne:
I think it would be better not to keep the full objects in the redux store, as it is a good practice to keep the minimal amount of information in each slice. Having the full object in the selected slice can lead to usage of stale data. Where does the public data live ? Shoulnd't we fix that instead ?

src/drive/web/modules/views/Recent/index.jsx
y-lohse: It was never an option.

src/drive/web/modules/views/Sharings/index.jsx
y-lohse: Same answer, it was never displayed there.

y-lohse:
> Where does the public data live ? Shoulnd't we fix that instead ?

Sorry, I forgot to address this. The public data lives only in memory, as a result of [a call to `statById`](https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/views/Public/usePublicFilesQuery.jsx#L8). This is absolutely something we should fix, but we're not quit sure about the approach yet, as it likely involves some heavy works on the cozy-stack side. One thing is sure though, it won't happen in the next couple of days, hence this fix.

PR #2077, opened 2020-07-22
Fix/improv

Crash--:
I open the PR to not forget to work on later (adding tests mainly) 

edit by @y-lohse : it's ready now!


y-lohse:
I added some tests, mostly for 301dc8eaa2e7ad7ccd151a1c796285e5e10a85da. I gave up on testing 221f680a946b99e2e6918c08667ba7e92095c5d8 because it basically requires creating a mocked low level filesystem API, but I think we can live without it since the calling code is now tested.

PR #2078, opened 2020-07-22
fix: Close the menu after the scanner is called

y-lohse:
We currently have a problem where the more menu in the top corner stays open after we start the cozy-scanner. This was introduced as a workaround [here](https://github.com/cozy/cozy-drive/pull/2017/files#r423532174) — the `Scanner` component needs to be rendered in order to run the scanner, but the ActionMenu closes itself after an action is selected, which unmounts `Scanner`.

So i moved `Scanner` outside of the `ActionMenu` so that it can stay in the render tree independently from the menu. However the scanner also exposes a `startScan` function that needs to be called from inside the menu. To make this function accessible, the `ScanWrapper` now renders a context provider, and the corresponding menu item is a consumer of that context. This way the scanner and the menu still work together, but without needing to be rendered together.

### DIFF
--- a/src/drive/web/modules/drive/Toolbar/components/MoreMenu.jsx
+++ b/src/drive/web/modules/drive/Toolbar/components/MoreMenu.jsx
@@ -15,6 +15,7 @@ import CreateNoteItem from './CreateNoteItem'
 import CreateShortcut from './CreateShortcut'
 import DownloadButtonItem from './DownloadButtonItem'
 import ShareItem from '../share/ShareItem'
+import StartScanner from './StartScanner'
 import ScanWrapper from './ScanWrapper'
 
 const MoreMenu = ({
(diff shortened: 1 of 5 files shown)


src/drive/web/modules/drive/Toolbar/components/StartScanner.jsx
```diff
@@ -0,0 +1,45 @@
...
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import { getTracker } from 'cozy-ui/transpiled/react/helpers/tracker'
+import { SCANNER_UPLOADING } from 'cozy-scanner'
+import { ScannerContext } from './ScanWrapper'
+
+const StartScanner = ({ t }) => {
+  const { status, online, startScanner } = useContext(ScannerContext)
+
+  const offlineMessage = () => {
+    return alert(t('Scan.error.offlinee'))
```
ptbrowne, line 13: Is it expected to use window.alert instead of the Alerter ?

y-lohse: Apparently yes! This code only runs on mobile and the native alerts look good and give the app a more native feel.

src/drive/web/modules/drive/Toolbar/components/MoreMenu.jsx
```diff
@@ -30,45 +31,50 @@ const MoreMenu = ({
   const openMenu = useCallback(() => setMenuVisible(true))
   const closeMenu = useCallback(() => setMenuVisible(false))
 
+  const ConditionalScanner = isMobileApp() ? ScanWrapper : React.Fragment
```
ptbrowne, line 34: Shouldn't it be the responsibility of the ScanWrapper to render only its children on desktop ?

src/drive/web/modules/drive/Toolbar/components/MoreMenu.spec.jsx
```diff
@@ -1,64 +1,95 @@
...
+    })
+
+    afterAll(() => {
+      window.navigator.camera = cameraObject
+    })
+
+    it('opens and closes the scanner', async () => {
+      const { getByText, queryByText } = await setup()
+
+      isMobileApp
```
Crash--, line 83: What is that?

src/drive/web/modules/drive/Toolbar/components/MoreMenu.spec.jsx
```diff
@@ -1,64 +1,93 @@
...
 
-const sleep = duration => new Promise(resolve => setTimeout(resolve, duration))
+jest.mock('cozy-device-helper', () => ({
+  ...jest.requireActual('cozy-device-helper'),
+  isMobileApp: jest.fn().mockReturnValue(false)
+}))
 
 mockCozyClientRequestQuery()
 
+configure({ testIdAttribute: 'data-test-id' })
```
ptbrowne, line 23: This should be done in jest.setup no ?

PR #2079, opened 2020-07-24
Opening files

y-lohse:
Something we missed during the migration is that some files are not opened in the viewer. Shortcuts for example are opened in new tabs, and notes are opened in the notes app.

@Crash-- I reunited both file opening logic in one place, I'm curious to see what you think about it. 

I still need to add some tests before this can be merged.


src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -53,12 +79,16 @@ const FolderViewBody = ({
...
-        window.location.href = resp.data.attributes.url
+        if (isMobileApp()) {
+          const resp = await client.query(
+            Q('io.cozy.files.shortcuts').getById(file.id)
+          )
+          window.location.href = resp.data.attributes.url
+        } else {
+          const url = generateFileUrl({ file, client, isFlatDomain: true })
+          window.open(url, '_blank')
+        }
```
Crash--, line 104: We introduced a regression in this commit, no? If we're renaming a file and we click on it, it'll follow the link, no?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -53,12 +79,16 @@ const FolderViewBody = ({
...
-        )
-        window.location.href = resp.data.attributes.url
+        if (isMobileApp()) {
+          const resp = await client.query(
+            Q('io.cozy.files.shortcuts').getById(file.id)
+          )
+          window.location.href = resp.data.attributes.url
+        } else {
+          const url = generateFileUrl({ file, client, isFlatDomain: true })
+          window.open(url, '_blank')
```
Crash--, line 103: sometimes we use `system`, sometimes `self` and now `blank`. Hard to follow

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -21,6 +21,32 @@ import SelectionBar from 'drive/web/modules/selection/SelectionBar'
...
+        cozyUrl: client.getStackClient().uri,
+        pathname: '/',
+        slug: 'drive',
+        hash: `external/${file.id}`,
+        subDomainType: isFlatDomain ? 'flat' : 'nested'
+      })
+    }
+    return webLink
+  }
+}
```
Crash--, line 55: Why did we move this method in this file?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -53,12 +79,16 @@ const FolderViewBody = ({
...
-      //Should only be called if mobile
       if (isShortcut) {
-        const resp = await client.query(
-          Q('io.cozy.files.shortcuts').getById(file.id)
-        )
-        window.location.href = resp.data.attributes.url
+        if (isMobileApp()) {
+          const resp = await client.query(
+            Q('io.cozy.files.shortcuts').getById(file.id)
+          )
```
Crash--, line 99: no try  /catch ?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -53,12 +79,16 @@ const FolderViewBody = ({
...
+          window.location.href = resp.data.attributes.url
+        } else {
+          const url = generateFileUrl({ file, client, isFlatDomain: true })
```
Crash--: true on flatdomain ? 🤔 We have method now to get the right value, right?

src/drive/web/modules/filelist/File.jsx
```diff
@@ -321,7 +320,6 @@ const File = props => {
       >
         <FileThumbnail
           file={attributes}
-          withSharedBadge={withSharedBadge}
```
Crash--, line 327: This is not an `unused props`  as described in the commit, but more an always true prop now, right?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
y-lohse: I think it's better to have all the file opening logic in the same place, rather than split between FileExplorer and File as it was before. For example [this comment](https://github.com/cozy/cozy-drive/blob/v1.25.0/src/drive/web/modules/navigation/FileExplorer.jsx#L193) was related to the [`if !isMobile` in FIle.jsx](https://github.com/cozy/cozy-drive/blob/v1.25.0/src/drive/web/modules/filelist/FileOpener.jsx#L25), but with the 2 being very far away it was hard to understand.

src/drive/web/modules/views/Folder/FolderViewBody.jsx
y-lohse: Yep, it's corrected in a later commit.

src/drive/web/modules/filelist/File.jsx
y-lohse: You're right, I'll edit the commit title

src/drive/web/modules/views/Folder/FolderViewBody.jsx
y-lohse: Do we use `self` at all? But yeah we do use `system` in one place, not sure if we could use `blank` instead there.

src/drive/web/modules/views/Folder/FolderViewBody.jsx
Crash--: You totally right about the opening stuff. My concern was more about the `generateFileUrl` method. I'm more and more convinced that JS only stuff should not be in a JSX file. I'm sure that `generateFileUrl` can be used outside of this component

y-lohse: Oh ok! Yep no problem I'll move it in a standalone file.

src/drive/web/modules/views/Folder/FolderViewBody.jsx
Crash--: Yep in QuotaAlert https://github.com/cozy/cozy-ui/blob/master/react/QuotaAlert/index.jsx#L34 / 
https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/upload/QuotaAlert.jsx#L34 but I can't remember why :/

y-lohse:
I ended up testing the file opening logic in pure js instead of using react-testing-lib because we listen for taps on files through hammer.js and I didn't manage to trigger hammer's listener in the test env.

src/drive/web/modules/filelist/FileOpener.jsx
```diff
@@ -40,9 +32,7 @@ const FileOpener = ({
               toggle(ev.srcEvent)
             } else {
               ev.srcEvent.stopImmediatePropagation()
-              if (!shouldOpenInANewTab) {
-                open(ev.srcEvent, file)
-              }
+              if (isActive) open(ev.srcEvent, file)
```
Crash--, line 35: Seems weird. isActive is when we're renaiming?

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -178,8 +178,8 @@ const FolderViewBody = ({
                     <File
                       key={file._id}
                       attributes={file}
-                      withSelectionCheckbox={true}
-                      withSharedBadge={true}
+                      withSelectionCheckbox
+                      withSharedBadge
```
Crash--, line 127: you can remove this `withSharedBadge` props

PR #2081, opened 2020-07-28
docs: JDK / Gradle / Java home

Crash--:


Crash--:
JDK sure it's not (at least not in this version and Cordova needs this version...). Gradle I'm not sure, but @AVuillaume has to install it manually

AVuillaume:
- The output of the command `java -version` (JDK) was 14.0.2, the most recent one.
After setting the env variable JAVA_HOME, I could not compile the project, because the version of the JDK had to be 1.8.
(Error was `Error: Requirements check failed for JDK 1.8`.)
I installed the JDK 1.8, re-set the env variable using **export JAVA_HOME=`/usr/libexec/java_home -v 1.8.0`**.
(The ` characters are missing in the command above due to GitHub's code format ^).

- The second error I encountered was Gradle not installed. I fixed that by using `brew install gradle`.

- Finally, I haven't accepted the license agreements, this was fixed by using:
`yes | ~/Library/Android/sdk/tools/bin/sdkmanager --licenses`
(Location of sdkmanager may change for some installations).

But then I was able to compile everything without any problems.

nono:
@Crash-- can we merge this PR?

PR #2082, opened 2020-07-29
Toolbar fixes

y-lohse:
We had a crash on the sharings view on mobile, because we tried to pass `displayedFolder.id` to the scanner while `displayedFolder` was null.

- ScanWrapper now detects if it isn't inside a folder and exits if it shouldn't be rendered
- StartScanner exits if there is no ScannerContext available
- There is a toolbar on the recent view, to make it consistent with other views
- I refactored the `NotRootFolder` thing (again). Hopefully now it's less of a headache to work with.


src/drive/web/modules/drive/Toolbar/components/MoreMenu.spec.jsx
```diff
@@ -87,5 +88,19 @@ describe('MoreMenu', () => {
       expect(queryByText('Save the doc')).toBeNull()
       expect(queryByText('Scan a doc')).toBeNull()
     })
+
+    it('is not displayed outsie of a folder', async () => {
```
Crash--, line 92:
```suggestion
    it('is not displayed outsdie of a folder', async () => {
```

PR #2083, opened 2020-07-31
fix: Rename files after upload

y-lohse:
Files that have just been uploaded don't have a `meta.rev` field, but they do have a plain `_rev` field. Trying to read `meta.rev` was causing an error when renaming a file just after it had been uploaded.

### DIFF
--- a/src/drive/web/modules/filelist/FilenameInput.jsx
+++ b/src/drive/web/modules/filelist/FilenameInput.jsx
@@ -98,7 +98,8 @@ class FilenameInput extends Component {
     //Since we're mounting the component and focusing it at the same time
     // let's add a small timeout to be sure the ref is populated
     setTimeout(() => {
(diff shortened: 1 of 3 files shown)


src/drive/web/modules/filelist/FilenameInput.jsx
```diff
@@ -98,7 +98,8 @@ class FilenameInput extends Component {
     //Since we're mounting the component and focusing it at the same time
     // let's add a small timeout to be sure the ref is populated
     setTimeout(() => {
-      this.textInput.current.setSelectionRange(0, filename.length)
+      if (this.textInput.current)
```
y-lohse, line 101: Mildly related, but in the tests the component is unmounted before the end of the timeout, so `this.textInput.current` becomes undefined.

PR #2090, opened 2020-08-05
fix: Added exif creation date of asset sent back to java

AVuillaume:
**Description**
This PR is related to [this PR](https://github.com/cozy/cordova-plugin-list-library-items/pull/31).
This change sends back to the java code the asset's creation date.

**How does it work ?**
This line of code `createdDate: file['creationDate'],`, adds the asset's creation date. With this, the java code use this date to format it and send it to the stack through the headers.
For more information about this, please check out [this PR](https://github.com/cozy/cordova-plugin-list-library-items/pull/31).

### DIFF
--- a/src/drive/mobile/lib/media.js
+++ b/src/drive/mobile/lib/media.js
@@ -51,6 +51,7 @@ const generatePayloadForNative = async ({
     libraryId: file['id'],
     mimeType: file['mimeType'],
     filePath: file['filePath'],
+    createdDate: file['creationDate'],
     httpMethod: method,
     headers: {
       Authorization: 'Bearer ' + token,
--- a/src/drive/mobile/lib/media.js
+++ b/src/drive/mobile/lib/media.js
@@ -40,7 +40,7 @@ export const requestAuthorization = async () => {
 }
 /**
  * Generates an http payload for the native part.
- * The uri and file variables are from the native part, 
+ * The uri and file variables are from the native part,
  * used to return a payload with headers.
  * @param {any} file - The file to upload, used to get the id, mimeType, and filePath.
  * @param {any} uri - The uri, used to get the serverUrl.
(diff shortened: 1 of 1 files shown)


Crash--:
It seems that on iOS the Objective-C also pass this `creationDate`(https://github.com/cozy/cordova-plugin-list-library-items/blob/master/src/ios/ListLibraryItems.m#L256) to the JS. So the JS will also send back this created to Objective-C, but we don't use it since we get the date from the PHAsset directly. https://github.com/cozy/cordova-plugin-list-library-items/blob/master/src/ios/ListLibraryItems.m#L178 

I'm wondering if we should not do the `Date` manipulation on the JS side to have only one place where we create an RFC 1123 Date (instead of having one implementation on Android and one more on iOS). 

What do you think?

y-lohse:
> but we don't use it since we get the date from the PHAsset directly.

Why can't we do the same on android and are doing this back and forth with the js instead?

Crash--:
> Why can't we do the same on android and are doing this back and forth with the js instead?

For "performance" reasons : 
- On Android we had to query the MediaStore, aka make an SQL request to fetch the informations. I don't want to make a query on each item we have to upload. => A solution can be to store the list results and use it in the uploadItem. Like that we just have to read the in memory data without making a request. (but I still think the solution mentioned above it's better) 
- On iOS we have to do this query since we have to map on the assets contained in the PHAsset (to get the right asset (aka the photo for a Live Photo for instance). 

Adding a date in the payload has no perf impact for now and can add the ability to centralise the way we manipulate it.

src/drive/mobile/lib/media.js
```diff
@@ -51,6 +51,7 @@ const generatePayloadForNative = async ({
     libraryId: file['id'],
     mimeType: file['mimeType'],
     filePath: file['filePath'],
+    createdDate: file['creationDate'],
```
ptbrowne, line 54: Nit: It'd be nice to have a small docstring for this function.

nono:
Ping @Crash--

nono:
Ping @Crash--

Crash--:
Thanks for the reminder. Let's see how it behaves in the next beta

PR #2095, opened 2020-08-24
fix: Added device name to JS payload

AVuillaume:
**Description**
This PR is iOS and Android related. It sends the asset's creation date and the device name to the JS side.
The related changes on the native side are available in this [PR](https://github.com/cozy/cordova-plugin-list-library-items/pull/34).
**How does it work ?**
To sync an asset, both iOS and Android native side sends additional information to the JS side, through a data structure (Dictionary for iOS, JSON object for Android).
Asset's creation date and device name are set in a data structure, and sent here, to the JS side.

### DIFF
--- a/src/drive/mobile/lib/media.js
+++ b/src/drive/mobile/lib/media.js
@@ -116,8 +116,6 @@ export const uploadLibraryItem = async (
       file: libraryItem,
       uri
     })
-    console.log("JS PART: The id of asset is: " + libraryItem['id']);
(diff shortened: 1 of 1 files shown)


nono:
Ping @Crash--

Crash--:
This work is in standby ATM. We need to go deeper on this subject https://github.com/cozy/cozy-doctypes/pull/135 before doing anything in drive.

PR #2096, opened 2020-08-24
fix: Display Empty Trash button even on "tablet"

Crash--:
u-hide--mob hides all the elements less than "desktop" size (< 1024px). isMobile refers to max-width: 768px... 

see https://github.com/cozy/cozy-ui/issues/1528

### DIFF
--- a/src/drive/web/modules/trash/Toolbar.jsx
+++ b/src/drive/web/modules/trash/Toolbar.jsx
@@ -83,14 +83,15 @@ export const Toolbar = ({
       className={styles['fil-toolbar-trash']}
       role="toolbar"
     >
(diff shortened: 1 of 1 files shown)


JF-Cozy (approved):
LGTM 👍

y-lohse (approved):
Cool, thanks! Could that have been caught by a test?

Crash--:
> Could that have been caught by a test?

If we start to test isTablet, maybe. And by test, I mean a react-testing-lib one with a get to check if the element is displayed or not. But we don't test isTablet, and I'm not sure we want to start to invest in it in the next weeks.

PR #2097, opened 2020-08-24
fix: Reduce query's cache

Crash--:
Thumbnails links have a validity of 10 minutes. So
we need to have a cache lesser than the one from
the stack. Like that, we make a new request to get
new links before the end of the previous ones.

(https://github.com/cozy/cozy-doctypes/pull/134)


JF-Cozy (approved):
LGTM 👍

PR #2120, opened 2020-10-15
fix: Don't display trashed file or folder in Sharing View

Crash--:
We don't want to display shared file that are in the trash in the "Sharing View"

### DIFF
--- a/src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
+++ b/src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
@@ -3,10 +3,14 @@ import get from 'lodash/get'
 import uniq from 'lodash/uniq'
 import keyBy from 'lodash/keyBy'
 import { buildParentsByIdsQuery } from 'drive/web/modules/queries'
+import { TRASH_DIR_ID } from 'drive/constants/config'
 
+export const excludeTrashedFiles = files => {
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
```diff
@@ -3,10 +3,14 @@ import get from 'lodash/get'
 import uniq from 'lodash/uniq'
 import keyBy from 'lodash/keyBy'
 import { buildParentsByIdsQuery } from 'drive/web/modules/queries'
+import { TRASH_DIR_ID } from 'drive/constants/config'
 
+export const excludeTrashedFiles = files => {
+  return files.filter(f => f.dir_id !== TRASH_DIR_ID && f.trashed !== true)
+}
```
ptbrowne, line 10: Nit: I think having a `isFromTrash` function instead of excludeTrashedFiles would be a bit better, as it would be more composable.

From the caller I think that

```
(result.data || []).filter(isFromTrash)
```

is as expressive as

```
excludeTrashedFiles(result.data || []) 
```

The difference is that the isFromTrash function can be more easily reused.

ptbrowne (approved):

src/drive/web/modules/views/Recent/useFilesQueryWithPath.jsx
Crash--: Love the idea thanks!!! 77b887cf9

PR #2140, opened 2020-11-03
refactor: Use ActionMenu from Cozy-UI and remove old Menu related stuff

Crash--:
We had this old menu in Drive. Before having UI's ActionMenu. Since UI now handles Mobile / Desktop let's use that ;)


ptbrowne (approved):

Crash--:
Merging since the red CI is because of the android build failure to download yarn

PR #2142, opened 2020-11-03
fix: Add sharing permission

Crash--:
This permission should had been added in a PR but this was lost during the departure of Y.

### DIFF
--- a/src/drive/targets/manifest.webapp
+++ b/src/drive/targets/manifest.webapp
@@ -145,6 +145,11 @@
       "verbs": ["POST"],
       "selector": "worker",
       "values": ["sendmail"]
-    }
+    },
+    "sharings": {
+      "description": "Required to have access to the sharings in realtime",
+      "type": "io.cozy.sharings",
+      "verbs": ["GET"]
+    }	
   }
 }


src/drive/targets/manifest.webapp
```diff
@@ -145,6 +145,11 @@
       "verbs": ["POST"],
       "selector": "worker",
       "values": ["sendmail"]
-    }
+    },
+    "sharings": {
+      "description": "Required to have access to the sharings in realtime",
+      "type": "io.cozy.sharings",
+      "verbs": ["GET"]
```
ptbrowne, line 152: Shouldn't we have a POST here also ? To create a sharing ?

Crash--: Nop. https://docs.cozy.io/en/cozy-stack/sharing/#post-sharings

> To create a sharing, no permissions on io.cozy.sharings are needed: an application can create a sharing on the documents for whose it has a permission.

We just need the permission to have a READ access from the realtime. Also, no need to have special permissions to get sharings for io.cozy.files for instance, since drive has access to io.cozy.files (sharing check the permission on the related doctype in fact)

ptbrowne: Thanks !

PR #2146, opened 2020-11-05
fix: Use snarkdown instead of react-markdown

Crash--:
Remove the use of react-markdown. Let's use snarkdown instead. It's much lighter. Also React-markdown adds a wrapper (p) we don't want in some case. 

https://bundlephobia.com/result?p=snarkdown@2.0.0 vs https://bundlephobia.com/result?p=react-markdown@5.0.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,6 @@
     "react-autosuggest": "9.3.4",
     "react-dom": "16.13.1",
     "react-dropzone": "5.1.1",
-    "react-markdown": "3.6.0",
     "react-measure": "2.5.0",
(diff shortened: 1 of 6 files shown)


JF-Cozy (approved):

JF-Cozy:
nit : on pourrait peut-être mettre une petite note de pourquoi `snarkdown` à la place de `ReactMarkdown` ?

Crash--:
https://bundlephobia.com/result?p=snarkdown@2.0.0 vs https://bundlephobia.com/result?p=react-markdown@5.0.2

PR #2149, opened 2020-11-09
feat: Show "shared" in action menu when a document is shared

JF-Cozy:
instead of "share"


Crash--:
I'm closing it since we fixed other way the issue thanks!

PR #2151, opened 2020-11-10
fix: Add favicon for Public pages

JF-Cozy:
Next step is to use cozy-script to do it automatically, and for every apps

### DIFF
--- a/webpack/appicon.config.js
+++ b/webpack/appicon.config.js
@@ -9,6 +9,19 @@ module.exports = {
         from: `src/${app}/targets/vendor/assets/app-icon.svg`,
         to: 'public/app-icon.svg'
       },
+      {
+        from: `src/${app}/targets/vendor/assets/favicon*`,
+        to: 'public/',
+        flatten: true
+      },
+      {
+        from: `src/${app}/targets/vendor/assets/apple-touch-icon.png`,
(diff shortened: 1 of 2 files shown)


webpack/appicon.config.js
```diff
@@ -9,6 +9,19 @@ module.exports = {
...
+        flatten: true
+      },
+      {
+        from: `src/${app}/targets/vendor/assets/apple-touch-icon.png`,
+        to: 'public/apple-touch-icon.png'
+      },
+      {
+        from: `src/${app}/targets/vendor/assets/safari-pinned-tab.svg`,
+        to: 'public/safari-pinned-tab.svg'
+      },
```
Crash--, line 24: can you try an `yarn build:photos` and see if it doesn't crash? I don't know exactly the behavior of CopyPlugin and specially "what the plugin does if the file doesn't exist"

Crash--: it seems to work even without any warning :o https://travis-ci.com/github/cozy/cozy-drive/jobs/432119850

Crash-- (approved):

webpack/appicon.config.js
JF-Cozy: Oui le build fonctionne bien :)

PR #2152, opened 2020-11-10
Qualification migration service

paultranvan:
We used to qualify files by manually setting specific attributes in the metadata, either in some konnectors, either through cozy-scanner.
We recently introduced a new API [here](https://github.com/cozy/cozy-client/pull/820) by using a qualification model that enforce new qualification rules.
This service migrates files qualified in the old way with the new qualification model.

Keep in my mind that:
* This PR needs this [fix](https://github.com/cozy/cozy-stack/pull/2743) for the permissions
* Eventually we want cozy-scanner and konnectors to use the new qualification API
* Some developer documentation should be added

### DIFF
--- a/src/drive/lib/doctypes.js
+++ b/src/drive/lib/doctypes.js
@@ -2,6 +2,7 @@ import { Contact, Group } from 'models'
 import extraDoctypes from 'drive/lib/extraDoctypes'
 
 export const DOCTYPE_FILES = 'io.cozy.files'
+export const DOCTYPE_FILES_SETTINGS = 'io.cozy.files.settings'
 export const DOCTYPE_ALBUMS = 'io.cozy.photos.albums'
 export const DOCTYPE_PHOTOS_SETTINGS = 'io.cozy.photos.settings'
 export const DOCTYPE_APPS = 'io.cozy.apps'
--- a/src/drive/lib/migration/qualification.js
+++ b/src/drive/lib/migration/qualification.js
@@ -1,4 +1,4 @@
-import { models } from 'cozy-client'
+import { models, Q } from 'cozy-client'
 import log from 'cozy-logger'
 const { Qualification } = models.document
 const { saveFileQualification } = models.file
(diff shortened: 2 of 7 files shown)


src/drive/lib/doctypes.js
```diff
@@ -2,6 +2,7 @@ import { Contact, Group } from 'models'
 import extraDoctypes from 'drive/lib/extraDoctypes'
 
 export const DOCTYPE_FILES = 'io.cozy.files'
+export const DOCTYPE_FILES_SETTINGS = 'io.cozy.files.settings'
```
Crash--, line 5: can you add the info in cozy-doctypes please ?

src/drive/lib/migration/qualification.js
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * Query the files indexed on their updatedAt date.
+ *
+ * @param {object} client - The CozyClient instance
+ * @param {string} date - The starting date to query
+ * @param {number} limit - The maximum number of files to return
+ */
+export const queryFilesFromDate = async (client, date, limit) => {
+  const query = client
+    .find('io.cozy.files')
```
Crash--, line 16: client.find() is deprecated, you should use Q('io.cozy.files').where() (see https://github.com/cozy/cozy-client/pull/830)

Crash-- (approved):
Let's test it!

src/drive/lib/doctypes.js
paultranvan: Oh right, forgot it

src/drive/lib/migration/qualification.js
paultranvan: Just when I got used to this API, it is deprecated, my story with js :laughing:

paultranvan: Done with 0290faa65425b519a88fb9d7dc290624c5410a4e

src/drive/lib/doctypes.js
paultranvan: https://github.com/cozy/cozy-doctypes/pull/155

PR #2159, opened 2020-11-16
Fix/public view banner

Crash--:
- Use translation for the Close Button 
- rewrite to functional component
- add a loading state to display or not the banner to avoid flickering between PublicToolbarByLink or PublicToolbarCozytoCozy


src/drive/web/modules/public/PublicToolbar.jsx
```diff
@@ -30,68 +30,79 @@ const isShortcutCreatedOnTheRecipientCozy = included => {
...
-        <PublicToolbarByLink
-          files={files}
-          hasWriteAccess={hasWriteAccess}
-          refreshFolderContent={refreshFolderContent}
-        />
-      )
-    }
+          const { data: sharing } = await client
+            .getStackClient()
+            .fetchJSON('GET', `/sharings/${sharingId}`)
```
ptbrowne, line 65: Nit: Seems like a lot of business logic here ? Is there a way to extrct the discovery link setup from the component ? Would it make sense ?

ptbrowne (commented):
No tests have been created or modified, do tests exist ? I think it's a good time to add tests if they do not exist.

src/drive/web/modules/public/PublicToolbar.jsx
Crash--: It's a good idea! a38ded2 and added test

src/drive/web/modules/public/useSharingInfos.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+  })
+  afterAll(() => {
+    window.location = location
+  })
+
+  it('returns the right infos when using useSharingInfo', async () => {
+    const discoveryLink = '/link'
+    getDiscoveryLinkMock.mockReturnValue(discoveryLink)
+    getOwnPermissionsMock.mockResolvedValue(mockedPermissionsWithInstance)
```
ptbrowne: Nit: I personnally prefer to mock at a lower level, on StackClient::fetchJSON for example. Here we are dependent on the implementation.

src/drive/web/modules/public/useSharingInfos.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(result.current.loading).toEqual(true)
+
+    await act(() => waitForNextUpdate())
+    expect(result.current.loading).toEqual(false)
+    expect(result.current.isSharingShortcutCreated).toEqual(true)
+    expect(result.current.sharing).toEqual(mockSharing.data)
+    expect(result.current.discoveryLink).toEqual(discoveryLink)
+  })
+
+  it('returns the right infos when using useSharingInfo', async () => {
```
ptbrowne, line 99: The two tests have the same description, what is different between them ?

ptbrowne: Can you add "without discoveryLink" ?

src/drive/web/modules/public/useSharingInfos.jsx
```diff
@@ -0,0 +1,70 @@
...
+          const sourceId = response.data.attributes.source_id
+          const sharingId = sourceId.split('/')[1]
+          const { sharecode } = getQueryParameter()
+
+          const link = client
+            .collection('io.cozy.sharings')
+            .getDiscoveryLink(sharingId, sharecode)
+          const { data: sharing } = await client.query(
+            Q('io.cozy.sharings').getById(sharingId)
+          )
```
ptbrowne, line 38: THe business logic after the try could be extracted into a fetchDiscoveryLink function ?

ptbrowne (approved):

src/drive/web/modules/public/useSharingInfos.spec.js
Crash--: mocking on fetchJSON makes the test dependent to the implementation too because we'll mock the first call to return X, the second to return Y and the third to return Z. Also it makes. it harder to follow / understand since outside method (I mean in cozy-client for instance) can also makes call to fetchJSON without any notice. 

How do you handle that?

src/drive/web/modules/public/useSharingInfos.jsx
Crash--: I don't think so because we do severals things here : 
- fetchPermissions
- check if the shortcutIsCreated
- getDiscoveryLink
- fetchSharing 

However, I can extract the method the get the sharingId from the fetch permissions call response 👍

src/drive/web/modules/public/useSharingInfos.spec.js
ptbrowne: When I mock fetchJSON I usually do:

```
fetchJSON.mockImplementation((method, route) => {
  if (method === 'GET' && route == '/permissions') { return mockPermissions; }
  else if (method === 'GET && route == 'otherRoute') { return otherStuff }
  else { throw new Error(`${method}:${route} is not implemented in tests`) }
})
```

Crash--: Interesting, thanks!

PR #2160, opened 2020-11-16
feat: Only display Share for the ShareMenuItem

Crash--:
We only want to display actions in this Menu. So we use only infinitive verbs.


Crash--:
Merging since this is small & scoped modification

PR #2162, opened 2020-11-17
Fix/modal

Crash--:
first round to remove Modal in favor of Dialog. Some modals still remaining ATM


Crash--:
Since there is nothing in particular in this PR, I'm merging it. Don't hesitate to comment even after merged if needed :)

PR #2169, opened 2020-11-19
fix: Regression with isomorphic fetch in cozy-client

paultranvan:
This is related to https://github.com/konnectors/libs/pull/724


paultranvan:
Actually, the only time we use isomorphic-fetch is on node context, i.e. service/konnectors. So we prefer to directly replace it with node-fetch

PR #2170, opened 2020-11-19
chore: Upgrade cozy-scanner

paultranvan:
Use new cozy-scanner version, including the new qualification


Crash--:
thanks! (merging it since I'm in a middle of a new release process ;))

PR #2177, opened 2020-11-26
chore: Upgrade cozy-client

paultranvan:
This upgrades cozy-client in order to use [this fix](https://github.com/cozy/cozy-client/pull/832). The qualification migration service was broken because of it.

### DIFF
--- a/src/drive/targets/services/qualificationMigration.js
+++ b/src/drive/targets/services/qualificationMigration.js
@@ -7,6 +7,9 @@ import {
   queryFilesFromDate,
   getMostRecentUpdatedDate
 } from 'drive/lib/migration/qualification'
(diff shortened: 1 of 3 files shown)


src/drive/targets/services/qualificationMigration.js
```diff
@@ -7,6 +7,9 @@ import {
   queryFilesFromDate,
   getMostRecentUpdatedDate
 } from 'drive/lib/migration/qualification'
+import fetch from 'node-fetch'
```
Crash--, line 10: Is it really needed? If yes, you should add a dep to node-fetch and not relying on the fact that the dep is here because another dep needs it

Crash-- (approved):

PR #2195, opened 2020-12-10
fix: Query returns docs in data object

paultranvan:
The returned object is not the same between `.query` and `.queryAll`: there is a `data` object for the former but not for the later. Maybe we should change that to avoid future mistakes like this one?


paultranvan:
Duplicate with #2196

PR #2209, opened 2021-01-04
fix: Rename insertDocuments to addReferences

Crash--:
Since https://github.com/cozy/cozy-client/pull/807/files#diff-1f726d94368a619efa8c1839270760a5eeb041e23208fab4c23781eba88d1092L66
we renamed insertDocuments to addReferences.

Todo add a test


Crash--:
fixed in #2213

PR #2210, opened 2021-01-05
Fix certification icon width on firefox

JF-Cozy:
Certification icon has width 0 on firefox. The span is needed to make the tooltip working.


ptbrowne (approved):
I'm wondering if a div would not be simpler (if it's supported by the tooltip).

LGTM otherwise.

JF-Cozy:
> I'm wondering if a div would not be simpler (if it's supported by the tooltip).
> 
> LGTM otherwise.

Le comportement est le même avec un div, il faut `u-w-100` aussi

PR #2213, opened 2021-01-05
Fix clustering

paultranvan:
The clustering service is currently broken since a cozy-client upgrade with breaking change in the HasManyFiles relationships: 
* `insertDocuments` becomes `addReferences`
* `removeDocuments` becomes `removeReferences`

This PR:
* Fix the API change
* Add somes tests to avoid such situation in the future
* Improve the error handling to avoid hiding errors and be alerted when the service actually fails
* Fix existing queries

### DIFF
--- a/src/photos/ducks/clustering/albums.spec.js
+++ b/src/photos/ducks/clustering/albums.spec.js
@@ -12,19 +12,19 @@ client.create = jest.fn().mockImplementation((doctype, album) => {
 client.mutate = jest.fn()
 
 describe('album', () => {
-  const photos = [
-    {
-      datetime: '2020-01-01',
-      clusterId: 'fakeAlbumId1',
-      id: '1'
-    },
-    {
-      datetime: '2020-01-01',
-      clusterId: 'fakeAlbumId1',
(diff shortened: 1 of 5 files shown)


Crash-- (approved):
Thanks for the first tests ;)

src/photos/ducks/clustering/albums.spec.js
```diff
@@ -0,0 +1,74 @@
...
+      datetime: '2020-01-01',
+      clusterId: 'fakeAlbumId1',
+      id: '1'
+    },
+    {
+      datetime: '2020-01-01',
+      clusterId: 'fakeAlbumId1',
+      id: '2'
+    }
+  ]
```
ptbrowne, line 26: Nit: This "photos" variable can be scoped to the first test.

ptbrowne (approved):
Nice !

src/photos/ducks/clustering/albums.spec.js
paultranvan: Indeed, fixed with 05c982e

PR #2217, opened 2021-01-11
Fixes for clustering service

paultranvan:
This fixes 2 behaviors with the photos clustering service:

* The query to get the new photos to clusterize was based on `file.created_at`, which is the date from exif data (or the upload date if it does not exist): the photos with a `file.created_at` older than the last execution date were not clusterized. We now use `file.cozyMetadata.createdAt`, to always rely on the upload date.
* When the service is running, a lock is set in order to avoid parallel executions of the service. In some cases, the lock was never released, causing the service to be never run again. We add a safeguard to bypass the lock if it has been set for more than 24h.


Crash-- (approved):
Can it be interesting to log something when the lock is here for too long? In order to have data to know when the case happens? 

Or at least, to know the % of instance concerned by a "too long lock" to know the severity of the issue?

JF-Cozy (approved):

paultranvan:
@Crash-- Good idea, I added a warning with 9898169 (and turned previous warnings into info), so it'll be possible to monitor it on metabase

PR #2222, opened 2021-01-13
fix: Permission issue on Android 10

Crash--:
For some devices on Android 10 (not all, I don't know why yet)
we have a permission error during the sync process resulting in

```
Sentry by " Backup error: okhttp3.internal.http2.StreamResetException:
 stream was reset: REFUSED_STREAM".
```

When debuging the error was clearer:

```
Exception while getting FileInputStream: java.io.FileNotFoundException:
/storage/emulated/0/DCIM/Camera/IMG_20210113_064202.jpg: open failed:
EACCES (Permission denied)
```

Since Android 10 (API 29) Android has a new mechanism to get
access to the device storage. It's called "scoped storage".

For now we can have the old behavior by requesting the legacy
system but it'll break when we'll need to target API 30.

Also this legacy flag is only supported by buildTools >= 29
and we need to compile againt sdk 29 to have it. So I've added
a few args to the cli in order to use the right versions
based on https://cordova.apache.org/docs/en/latest/guide/platforms/android/#setting-gradle-properties
recommandation

```
Apps that run on Android 11 but target Android 10 (API level 29)
can still request the requestLegacyExternalStorage attribute.
This flag allows apps to temporarily opt out of the changes
associated with scoped storage, such as granting access to different
directories and different types of media files.
After you update your app to target Android 11, the system ignores
the requestLegacyExternalStorage flag.
```
see https://developer.android.com/about/versions/11/privacy/storage


Crash--:
Il semblerait qu'avec cette modif, je ne sois plus compatible avec une bon % des téléphones qu'on supportait... 🤒 

Je vais creuser un peu

PR #2230, opened 2021-01-22
fix: Missing last execution timestamp

paultranvan:
We used to store a timestamp of the last execution of the service,
which is useful to know if the service should actually be run, depending on the 'running' and 'postponed' status.
However, this timestamp was not always set, leading to missing
information.
Therefore, we prefer to remove this timestamp and now use the
`cozyMetadata.udpatedAt` field, that we are sure is always set, as the
photo settings are updated at the beggining of the execution.

Note the doctype format should be updated accordingly: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.photos.md#clustering-settings

### DIFF
--- a/src/photos/ducks/clustering/files.js
+++ b/src/photos/ducks/clustering/files.js
@@ -12,6 +12,7 @@ export const getFilesFromDate = async (client, date, { limit = 0 } = {}) => {
     .partialIndex({
       trashed: false
     })
+    .indexFields(['class', 'cozyMetadata.createdAt'])
     .sortBy([{ class: 'asc' }, { 'cozyMetadata.createdAt': 'asc' }])
     .limitBy(limit)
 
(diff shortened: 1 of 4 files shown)


Crash-- (commented):
OK, but if `lastExecution` was not set, it means that there is a bug somewhere, no? Here, we use the updated_at, ok, but this field is only updated if the document is updated. Are we sure it is always the case? 

Is it related to the `postponed` or to the `lock` not being removed?

paultranvan:
@Crash-- Yes the bug the bug was relying on `lastExecution` while it wasn't set.

> Here, we use the updated_at, ok, but this field is only updated if the document is updated. Are we sure it is always the case?

The setting document is updated at the [beginning of the service](-https://github.com/cozy/cozy-drive/blob/master/src/photos/targets/services/onPhotoUpload.js#L217).
 
 > Is it related to the postponed or to the lock not being removed?
 Much probably, yes

Crash--:
> Much probably, yes

What do we need to have a glue in order to fix it? `lastExecution` was set at the end of the flow? It means that we don't arrive at the end of the execution right?

paultranvan:
> lastExecution was set at the end of the flow? It means that we don't arrive at the end of the execution right?

Yes, and that is why we don't want to rely on `lastExecution` anymore. At least with `cozyMetadata.updatedAt` we are sure that this field is always set.
Furthermore, we now better catch the execution failures: https://github.com/cozy/cozy-drive/blob/master/src/photos/targets/services/onPhotoUpload.js#L283

Crash-- (approved):

PR #2237, opened 2021-02-01
fix: Call the hooks in the right order

Crash--:
If not, we got undefined for currentIndex when opening
the viewer. Let's set the index before opening it.

Tried to add a test on the component, but I can't
handle it since we've a react-test-lib test but
file-opener is done with HammerJS. Click will not
fire the right event... We should test our
hooks call in order to test the component

### DIFF
--- a/src/drive/web/modules/views/Public/index.jsx
+++ b/src/drive/web/modules/views/Public/index.jsx
@@ -109,8 +109,8 @@ const PublicFolderView = ({
         Alerter.error('alert.offline')
       }
     } else {
-      setViewerOpened(true)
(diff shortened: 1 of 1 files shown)


ptbrowne:
For the test, it could be possible to mock FolderViewBody

const FolderViewBody = ({ navigateToFile }) => {
  returns files.map(file => <div onClick={() => navigateToFile(file)}>{ file }</div>)
 }

ptbrowne (approved):

Crash--:
Interesting approach @ptbrowne thanks for the idea. I'll test that later 👍

PR #2239, opened 2021-02-03
fix: Remove deep dependecies in CertificationsIcons useCallback

JF-Cozy:
Deep dependencies in useCallback seems to be a tricky usage, because it's checked on mount (unreachable prop object will crash JS)

### DIFF
--- a/src/drive/web/modules/filelist/cells/FileName.jsx
+++ b/src/drive/web/modules/filelist/cells/FileName.jsx
@@ -17,9 +17,10 @@ import { CozyFile } from 'models'
 
 import styles from 'drive/styles/filelist.styl'
 
-const CertificationsIcons = ({ attributes }) => {
+export const CertificationsIcons = ({ attributes }) => {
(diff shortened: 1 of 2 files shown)


ptbrowne:
LGTM. Is there already a test for this ?

JF-Cozy:
> LGTM. Is there already a test for this ?

Le parent File (qui est la ligne) est testé mais aucune des cellules elles-mêmes

ptbrowne:
Can you add a test please ?

JF-Cozy:
> Can you add a test please ?

@ptbrowne c'est fait :)

ptbrowne (approved):
Thanks for the test, I wonder just if AppIcon could directly take the data-testid. I checked and for now the AppIcon does not support this. It could be a good improvement for the future.

JF-Cozy:
> Thanks for the test, I wonder just if AppIcon could directly take the data-testid. I checked and for now the AppIcon does not support this. It could be a good improvement for the future.

@ptbrowne https://github.com/cozy/cozy-ui/issues/1730

PR #2241, opened 2021-02-08
feat: Use partial index for negative queries

paultranvan:
As recommended
[here](https://docs.cozy.io/en/tutorials/data/advanced/#partial-indexes),
we use partial indexes for negative queries. This is notably better to
have lighter indexes (the _id is not indexed anymore) and thus, more
efficient queries.

### DIFF
--- a/src/drive/web/modules/queries.js
+++ b/src/drive/web/modules/queries.js
@@ -41,9 +41,13 @@ const buildDriveQuery = ({
     Q('io.cozy.files')
       .where({
         dir_id: currentFolderId,
-        _id: { $ne: TRASH_DIR_ID },
         type,
         [sortAttribute]: { $gt: null }
+      })
+      .partialIndex({
+        _id: {
+          $ne: TRASH_DIR_ID
+        }
       })
       .indexFields(['dir_id', 'type', sortAttribute])
       .sortBy([
@@ -190,11 +194,13 @@ export const buildMoveOrImportQuery = (client, dirId) => ({
     Q('io.cozy.files')
       .where({
         dir_id: dirId,
-        _id: {
-          $ne: TRASH_DIR_ID
-        },
         type: { $gt: null },
         name: { $gt: null }
+      })
+      .partialIndex({
+        _id: {
+          $ne: TRASH_DIR_ID
+        }
       })
       .indexFields(['dir_id', 'type', 'name'])
       .sortBy([{ dir_id: 'asc' }, { type: 'asc' }, { name: 'asc' }])
--- a/src/drive/web/modules/queries.js
+++ b/src/drive/web/modules/queries.js
@@ -187,8 +187,7 @@ const buildTrashQuery = ({
 
 export const buildMoveOrImportQuery = (client, dirId) => ({
   definition: () =>
-    client
-      .find('io.cozy.files')
+    Q('io.cozy.files')
       .where({
         dir_id: dirId,
         _id: {


Crash-- (approved):
Thanks! Hope that remove the warnings in the console ;).

I'm wondering... Since we create a new index for this request, should we delete the old one? What are your thoughts on this question?

ptbrowne:
I do not think we should put the responsibility of taking care of deleting indexes on app developers. If this is a problem, AMHA the stack could periodically remove indexes that have not been used for a long time.

paultranvan:
@Crash-- Ideally the old index should be removed as it will take disk space for nothing, but I agree with @ptbrowne that the developer shouldn't bother with deleting the old index. 
Currently, I don't think we have a way to find unused indexes as CouchDB does not provide any info about a mango index use/state. But I guess we could add the info by the stack (e.g. save a timestamp when an index is used): this would help to detect unused indexes and eventually remove them

ptbrowne:
>  But I guess we could add the info by the stack

Yep, that's what I was thinking about.

Crash--:
> I do not think we should put the responsibility of taking care of deleting indexes on app developers. If this is a problem, AMHA the stack could periodically remove indexes that have not been used for a long time.


Personally, I disagree with that statement. Cozy-Stack can have the info of when the index has been used for the last time (by adding information somewhere). But she doesn't know why. Is it because the user has not used the app recently? Is it because he didn't go to the specific page of the app that needs this index? Is it because the app has been uninstall? Is it because the index is not used anymore by the app? 

If we start to add this kind of magic removal, we'll end up having "old applications" that are even slower than today. Because we'll need to build the index from scratch if the user starts to reuse the app. 

And in that case we're not talking about PouchDB... At some point, the app dev will have to take care of it. 

Currently, we start to think to have some kind of warmup queries that will be run in a cron way to keep few indexes up to date. Those queries will be made by the app developers... (I don't know if we'll end up implementing this feature or not)

To me, if you are responsible to create you should be responsible to delete when it's needed. App developers are responsible to create the indexes via cozy-client. They know how to index, they should be responsible to remove them when it's possible aka when they know that the index is changing. 

To add information about indexes : 
177Go  indexes 
500Go  datas

ptbrowne:
The space data you showed is interesting, thanks to have put it there. I think it is optimistic to think that without any incentive, or data like this, an app developer will take care of that. People already have a lot to think about. If indexes are easy to create, they will be created. If this is hard to know which indexes are in use by all apps on Cozy and if it's OK to remove it, I expect developers will not remove them.

> If we start to add this kind of magic removal, we'll end up having "old applications" that are even slower than today.

I do not see a problem with an app having a slow startup if it has been a long time you did not use it. For example Trello shows a "Doing a bit of bookkeeping here, please wait.." if this has been a long time that the user has used the app. This is a strategy that is often used : put hot data in a cache, keep stale data on disk.

Having a space for app developers to see statistics on their app performance, index use etc... could be a good way I think to make developers [aware](https://www.youtube.com/watch?v=8ECgLA2MjFo) of the impact of their indexes. What makes it a bit difficult is that indexes are shared across apps.

I'd love to have the opinion of @nono or @sblaisot on the subject.

nono:
My opinion: currently, we don't care about unused indexes. In the long run, we will have to clean them, but for now, it is not very important. What is costly is building the index, not keeping it on disk (at least at our scale). One day, we will want to clean them to optimize our costs, but we have other priorities for the moment.

paultranvan:
I referenced this discussion in https://github.com/cozy/cozy-client/issues/873 so we can close this PR and keep track of the discussion

PR #2245, opened 2021-02-11
fix: Use fetchPolicy correctly with specific metadata query

JF-Cozy:
and with buildOnlyFolderQuery too


ptbrowne (approved):
LGTM

I know this is not part of this PR but I could not help notice the `client.get` above the first fixed fetchPolicy. It's good to keep an eye for those and fix them while we fix nearby code.

JF-Cozy:
Effectivement. Je fais une carte tech transverse pour les `client.get` et les `getById(doctype/id)`

PR #2247, opened 2021-02-12
Replace client.get() by Q().getById()

JF-Cozy:
see commit message


package.json
```diff
@@ -8,14 +8,15 @@
...
     "watch:drive:mobile:hot": "COZY_APP_SLUG=drive NODE_ENV=mobile:development  cozy-scripts start  --mobile --manifest src/drive/targets/manifest.webapp --src-dir src/drive ",
     "watch:drive:standalone": "COZY_APP_SLUG=drive NODE_ENV=mobile:development cozy-scripts start  --manifest src/drive/targets/manifest.webapp --src-dir src/drive",
     "build:photos": "npm-run-all --parallel build:photos:browser",
     "build:photos:browser": "COZY_APP_SLUG=photos cozy-scripts build --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
     "watch:photos": "npm run watch:photos:browser",
-    "watch:photos:browser": "COZY_APP_SLUG=photos cozy-scripts start --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
+    "watch:photos:browser": "COZY_APP_SLUG=photos cozy-scripts watch --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
+    "watch:photos:browser:hot": "COZY_APP_SLUG=photos cozy-scripts start --manifest src/photos/targets/manifest.webapp --build-dir build/photos --src-dir src/photos",
```
ptbrowne, line 19: Nit: Hot reloading should be the default and not the exception now. In banks we have a "cold" script. But I think the best would be to use the verb start as in cozy-create-app and create-react-app. start for hot, watch for cold.

src/components/FileHistory/index.jsx
```diff
@@ -15,7 +16,7 @@ const trackEvent = () => {
 const FileHistory = ({ params: { fileId } }) => {
   trackEvent()
   return (
-    <Query query={client => client.get('io.cozy.files', fileId)}>
+    <Query query={() => Q('io.cozy.files').getById(fileId)}>
```
ptbrowne, line 19: Not part of this PR : both queries could be named and have a fetch policy.

ptbrowne (approved):
Thanks 👍 The majority of requests I've seen here are not named, this is something easy that could be done in the future.

package.json
JF-Cozy: Ok pour partir sur des `yarn start:drive` et `yarn watch:drive` pour le hot/cold 👍

src/components/FileHistory/index.jsx
JF-Cozy: https://github.com/cozy/cozy-drive/issues/2248

PR #2288, opened 2021-03-23
Fix/photos4

Crash--:
Started by fixing issues on Photo (adding / removing...) finished by removing the use of Modal, the use of confirm, and the store access


src/photos/ducks/timeline/components/Timeline.jsx
```diff
@@ -74,18 +73,32 @@ class Timeline extends Component {
...
+          count={selected.length}
+          related={belongsToAlbums(selected)}
+          onClose={this.closeModal}
+          confirm={async () => {
+            await Promise.all(
+              selected.map(p => this.props.deletePhoto(p))
+            ).then(clearSelection)
+            this.closeModal()
+          }}
+        />
```
ptbrowne, line 98: For the record, it is a bit tricky to put a component in the state since it will not get the react tree updates. I do not think there is a bug here but it can lead to problems (for ex: https://github.com/cozy/cozy-banks/pull/2042/commits/8ef1ee106c0756d9aeaa4598374cb93b4aa300c0 where we cache an object in the state and then it no longer is updated from the store).

src/photos/ducks/albums/components/AddToAlbumModal.jsx
```diff
@@ -26,33 +27,40 @@ class AddToAlbumModal extends Component {
...
+            <div>
+              <CreateAlbumForm
+                onSubmitNewAlbum={name =>
+                  createAlbum(name, photos)
                     .then(onDismiss)
                     .then(onSuccess)
                 }
               />
+              <Divider className="u-ml-0 u-maw-100 u-mt-1" />
+              {isFetchingAlbums && <Loading loadingType="albums_fetching" />}
```
ptbrowne, line 45: Nit: we should use directly `t` here and not let Loading do it for us.

```
<Loading loadingType={t("albums_fetching")} />
```

ptbrowne (approved):
🎉  Thanks !

src/photos/ducks/timeline/components/Timeline.jsx
Crash--: Thanks! 5b8218d2c1e791da63e50e8cce6a277a88c5a94c

PR #2289, opened 2021-03-24
fix: Add correct toolbar on public file viewer

JF-Cozy:
Must have the Viewer toolbar on desktop
and the cozy-bar "toolbar" (it's the '...' menu) on mobile


Crash-- (commented):
Since we have a lot of "regressions" each time we edit this file, we should add a few tests

Crash-- (approved):

PR #2293, opened 2021-03-25
Fix/delete album

Crash--:
Make a mistake in the previous PR resulting on calling a use case we can't have and crashing the app. So I fixed the issue, and remove the dead code.


JF-Cozy (approved):

src/photos/ducks/albums/components/AlbumPhotos.jsx
```diff
@@ -80,62 +78,35 @@ class AlbumPhotos extends Component {
       .collection('io.cozy.files')
       .downloadArchive(photos.map(({ _id }) => _id), 'selected')
   }
-  closeModal = () => {
+  closeDestroyConfirmModal = () => {
     this.setState({
-      displayModal: false
+      displayModalDestroyConfirm: false
```
ptbrowne, line 83: displayDestroyConfirmModal to have the convention that closeDestroyConfirmModal

ptbrowne (approved):
Is there a need to add a test to be sure that we can't have a problem anymore ?

LGTM otherwise. A small naming remark.

Crash--:
> Is there a need to add a test to be sure that we can't have a problem anymore ?

I'm tempted to say yes, but in fact there is still dead code and I really want to make a consequent refacto in this app. I think we should clean up this stuff and then writing tests.

Crash--:
In fact, I can add simple test to check if the right modal is open, will do that tomorrow (off for the day)

PR #2317, opened 2021-04-22
refactor: Use RealTimeQueries from CozyClient

Crash--:
The component is now available in CozyClient. Let's use it


src/drive/mobile/modules/settings/Settings.jsx
```diff
@@ -3,6 +3,8 @@
 import React, { Component } from 'react'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
+import { RealTimeQueries } from 'cozy-client'
```
JF-Cozy, line 6: nit: est-ce qu'on ne passerait pas maintenant toujourspar des liens profonds pour toutes nos libs ?

JF-Cozy (approved):

src/drive/mobile/modules/settings/Settings.jsx
Crash--: With Cozy-Client I don't see any args of doing it. The lib is pretty small. Specially when we're already in a react context. 

I don't like to use deep path because to me, internal structure is not an "public api".

JF-Cozy: Pourquoi faire des liens profonds avec `Q` dans ce cas ?

PR #2327, opened 2021-05-06
OnlyOffice file renamer

JF-Cozy:
Add possibility to rename onyloffice file when consulting it.
Not active on mobile

### DIFF
--- a/src/drive/web/modules/filelist/FilenameInput.jsx
+++ b/src/drive/web/modules/filelist/FilenameInput.jsx
@@ -1,11 +1,13 @@
 import React, { Component } from 'react'
-import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import { CozyFile } from 'cozy-doctypes'
-import styles from 'drive/styles/filenameinput.styl'
+import cx from 'classnames'
 
+import { CozyFile } from 'models'
+import Spinner from 'cozy-ui/transpiled/react/Spinner'
 import { Dialog } from 'cozy-ui/transpiled/react/CozyDialogs'
 import Button from 'cozy-ui/transpiled/react/Button'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
+
+import styles from 'drive/styles/filenameinput.styl'
 
 const ENTER_KEY = 13
 const ESC_KEY = 27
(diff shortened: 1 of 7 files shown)


src/drive/web/modules/filelist/FilenameInput.jsx
```diff
@@ -1,12 +1,14 @@
 import React, { Component } from 'react'
-import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import { CozyFile } from 'cozy-doctypes'
-import styles from 'drive/styles/filenameinput.styl'
+import cx from 'classnames'
 
+import { CozyFile } from 'cozy-doctypes'
```
Crash--, line 4: here it's imported from models https://github.com/cozy/cozy-drive/pull/2327/commits/541d762343b065a55f0cc5d3dcbe5f65ceea5129#diff-dc47bb533ec304621e4f8f11f3068a149cd759b50b2ea40e08186c1b51bcb045R7

Can we harmonize?

src/drive/web/modules/views/OnlyOffice/Toolbar/FileName.jsx
```diff
@@ -27,8 +56,23 @@ const FileName = ({ fileWithPath }) => {
...
+const mapStateToProps = (state, ownProps) => ({
+  isRenaming:
+    isRenaming(state) &&
+    get(getRenamingFile(state), 'id') === ownProps.fileWithPath.id
+})
+
+const FileName = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(DumbFileName)
```
Crash--, line 76: I'm a little bothered by this. We want to avoid the use of redux for this kind of thing. It is not easy to take this component out of drive because of the redux action. 

I understand you have done that way because the other component is written like that... So if you have time to rewrite this one without redux it can be cool, otherwise it doesn't matter, we'll do it later.

src/drive/web/modules/filelist/FilenameInput.jsx
JF-Cozy: Oui ok je vais changer partout

src/drive/web/modules/views/OnlyOffice/Toolbar/FileName.jsx
JF-Cozy: Pas de souci je modifie 👍

PR #2359, opened 2021-06-02
Fix only office file creation

JF-Cozy:
see commit messages


Crash-- (commented):
Can you update the second commit to say how did you get the template file?

Crash-- (approved):

PR #2383, opened 2021-06-16
fix: Typo in createFileOpeningHandler props

JF-Cozy:
We added a destructuring on the props, which caused a typo problem for the prop availableOffline


Merkur39:
Good for me 👍

PR #2399, opened 2021-07-21
Migrate PouchDB adapter

paultranvan:
This asks the user having an old Pouch adapter (e.g. idb, cordova-sqlite)
if he wants to migrate it. We detect such need by an entry in the
localstorage set by cozy-pouch-link.

If the user refuses, we do not migrate and the user can use its Drive
normally. We ask the user each time it loads Drive until he accepts.
Once he does, we run the migration; at the end he can use again its
Drive, with better performances.

Indeed, the  indexedDB adapter greatly improves performances
for queries, replication and so on.
For example we measured the following on a Cozy with 10K docs:

* idb adapter
  * ~8s to index docs
  * ~260ms to query 1K docs
* indexeddb adapter
  * ~2.8s to index docs
  * ~160ms to query 1K docs

See also https://github.com/cozy/cozy-client/pull/989


package.json
```diff
@@ -145,6 +145,8 @@
     "piwik-react-router": "0.8.2",
     "popper.js": "1.16.1",
     "pouchdb-adapter-cordova-sqlite": "^2.0.7",
+    "pouchdb-adapter-idb": "7.2.2",
+    "pouchdb-adapter-indexeddb": "7.2.2",
```
Crash--, line 149: Do we really need this package? Since https://github.com/cozy/cozy-client/commit/94ce6df5797ec09a436b50723aef11bdd1e9dce6#diff-a8d46e44d0bdbf56926cd82f0cd633678572de99cc8e8eeb1d56c205b3dd031eR17-R18 we updated Pouch. This adapter doesn't come by default?

src/drive/mobile/lib/cozy-helper.js
```diff
@@ -31,8 +33,50 @@ export const getOauthOptions = () => {
...
+    return require('pouchdb-adapter-idb')
+  }
+  return require('pouchdb-adapter-indexeddb').default
+}
+
+export const shouldMigrateAdapter = async () => {
+  const alreadyOnboarded = await localForage.getItem(ONBOARDED_ITEM)
+  const adapterName = window.localStorage.getItem(
+    'cozy-client-pouch-link-adaptername'
+  )
```
Crash--, line 54: nit: Maybe we should use a method from Cozyclient to get the item. This is not the responsability of the app to know the key

Crash-- (approved):
Looks good!

src/drive/mobile/lib/cozy-helper.js
paultranvan: Right, I assigned an issue to the guy who did that: https://github.com/cozy/cozy-client/issues/1005

package.json
paultranvan: I tried to build with this version of cozy-client but I get `Error: Can't resolve 'pouchdb-adapter-indexeddb'`, so it seems necessary

PR #2408, opened 2021-08-11
fix: Few fixes on public page

Crash--:
- We hide the CreateNote button 
- We don't display the MoreMenu if the is no file in the folder (since the only action within the menu is "Select file" and this action is not display when there is no file...) 
- We hide the download button if there is no file to download

### DIFF
--- a/src/drive/web/modules/drive/AddMenu/AddMenu.spec.jsx
+++ b/src/drive/web/modules/drive/AddMenu/AddMenu.spec.jsx
@@ -1,19 +1,15 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
+import { act } from '@testing-library/react'
 
 import { isMobileApp } from 'cozy-device-helper'
 
 import { setupFolderContent, mockCozyClientRequestQuery } from 'test/setup'
-
+import { useAppLinkWithStoreFallback } from 'cozy-client'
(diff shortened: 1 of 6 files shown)


ptbrowne:
Maybe tests could cover this case ?

src/drive/web/modules/drive/AddMenu/AddMenu.spec.jsx
```diff
@@ -1,19 +1,15 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
+import { act } from 'react-dom/test-utils'
```
JF-Cozy, line 3: nit: on ne devrait pas plutôt utiliser le `act` de `@testing-library/react` ? Il me semble que c'est conseillé dans la doc de `@testing-library/react`

JF-Cozy (approved):

src/drive/web/modules/drive/AddMenu/AddMenu.spec.jsx
Crash--: > This is a light wrapper around the react-dom/test-utils act function. All it does is forward all arguments to the act function if your version of react supports act. It is recommended to use the import from @testing-library/react over react-dom/test-utils for consistency reasons.

https://testing-library.com/docs/react-testing-library/api/#act


Right, fixed.

PR #2409, opened 2021-08-16
fix: Update hammerjs API usage

acezard:
https://trello.com/c/80mmMpTV/1497-%F0%9F%93%81-drive-mobile-le-clic-long-sur-le-mobile-nenclenche-plus-une-s%C3%A9lection-de-l%C3%A9l%C3%A9ment

The "onpress" event never existed, see: https://hammerjs.github.io/recognizer-press/
It used to work in cozy-drive because "onpress" was a custom event. With the new API, we can simply use the native 'press' event.

### DIFF
--- a/src/drive/web/modules/filelist/FileOpener.jsx
+++ b/src/drive/web/modules/filelist/FileOpener.jsx
@@ -72,11 +72,11 @@ const FileOpener = ({
 
       const gesturesHandler = propagating(new Hammer(rowRef.current))
 
-      gesturesHandler.on('tap onpress singletap', ev => {
+      gesturesHandler.on('tap press singletap', ev => {
         if (actionMenuVisible || disabled) return
         if (enableTouchEvents(ev)) {
           ev.preventDefault() // prevent a ghost click
-          if (ev.type === 'onpress' || selectionModeActive) {
+          if (ev.type === 'press' || selectionModeActive) {
             ev.srcEvent.stopImmediatePropagation()
             toggle(ev.srcEvent)
           } else {


Crash--:
It worked before https://github.com/cozy/cozy-drive/commit/776bb63a5cc8dbc3d4f5b3f7d8aacfafc6a87f46#diff-66f571a15fdeccd6293e111e9483d731a1d8a1bddf527123ece478a9605e12bfL27 so it seems to me like a weird fix. 

I just checked, and we didn't upgrade hammer on anything since https://github.com/cozy/cozy-drive/commit/7deaa72f0dcb1d09791e38f09421c0a2c8710640 so I'm not very confident with this fix. 

Doest it works when you enable touch device in your browser?

acezard:
> It worked before [776bb63#diff-66f571a15fdeccd6293e111e9483d731a1d8a1bddf527123ece478a9605e12bfL27](https://github.com/cozy/cozy-drive/commit/776bb63a5cc8dbc3d4f5b3f7d8aacfafc6a87f46#diff-66f571a15fdeccd6293e111e9483d731a1d8a1bddf527123ece478a9605e12bfL27) so it seems to me like a weird fix.
> 
> I just checked, and we didn't upgrade hammer on anything since [7deaa72](https://github.com/cozy/cozy-drive/commit/7deaa72f0dcb1d09791e38f09421c0a2c8710640) so I'm not very confident with this fix.
> 
> Doest it works when you enable touch device on your browser?

Yes it does work with "press" but not with "onpress"

Crash-- (approved):

JF-Cozy:
Complètement oublié de le passer celui-là. Merci @acezard !

PR #2411, opened 2021-08-19
feat: Update cozy-ui to latest for MidEllipsis

acezard:
Related to https://trello.com/c/AcrjAGzm/1498-vue-recent-chemin-mal-d%C3%A9coup%C3%A9

Currently, the truncated (or even non-truncated) directory trees under recent files appear broken under certain circumstances (notably when a directory name ends with punctuation or emoticons).

The fix had to be done in the consumed component that handled that issue: MidEllipsis from Cozy-Ui. Thus, this pull request only aims to update Cozy-Drive with a fixed version of MidEllipsis.

Also add a CHANGELOD.md file.


.nvmrc
```diff
@@ -1 +1 @@
-8.17.0
+10
```
acezard, line 1: That is the version used by the CLI

src/drive/web/modules/views/testUtils.jsx
```diff
@@ -24,3 +24,7 @@ export const getByTextWithMarkup = (getByText, text) => {
     return hasText(node) && childrenDontHaveText
   })
 }
+
+// We need this because MidEllipsis uses control characters (invisible formatting characters)
+// For the purpose of testing it is simpler to strip them from the computed paths
+export const removeNonASCII = str => str.replace(/[^\x20-\x7E]/g, '')
```
acezard, line 30: Looked the simplest to me , if you have a better idea don't hesitate. I tried a few different ways but it's difficult because of the nature of control characters and how they are outputted by the tests

Crash-- (approved):

PR #2412, opened 2021-08-19
fix: Remove double saving when creating directory

acezard:
Related to https://trello.com/c/W6b1eknd/1505-%F0%9F%93%81-drive-r%C3%A9gression-1-erreur-%C3%A0-la-cr%C3%A9ation-dun-dossier

Issue: creating a directory triggered a "this folder already exists error"

Bug: FilenameInput launches two save() action if a file has no name => directory never has a name when creating one, so it will always be saved twice creating dupe error

Resolution: return the save action if the component has no name

### DIFF
--- a/src/drive/web/modules/filelist/FilenameInput.jsx
+++ b/src/drive/web/modules/filelist/FilenameInput.jsx
@@ -67,7 +67,7 @@ class FilenameInput extends Component {
     const { value } = this.state
     const { file } = this.props
     this.setState({ working: true, error: false })
-    if (!this.fileNameOnMount) this.save()
+    if (!this.fileNameOnMount) return this.save()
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/filelist/FilenameInput.jsx
```diff
@@ -67,7 +67,7 @@ class FilenameInput extends Component {
     const { value } = this.state
     const { file } = this.props
     this.setState({ working: true, error: false })
-    if (!this.fileNameOnMount) this.save()
+    if (!this.fileNameOnMount) return this.save()
```
acezard, line 70: I did some tests but couldn't file an issue with this with non directory files. Do you think it can create one?

Crash-- (approved):

PR #2413, opened 2021-08-19
fix: Add busy state to delete confirm button

acezard:
Related to https://trello.com/c/Oon3REW1/1500-%F0%9F%93%81-drive-ajouter-un-spinner-dans-la-modale-de-suppression

Bug: no UX feedback when deleting batch of files

Fix: useState in the confirm modal and inject busy state in the button when starting delete process


Crash-- (approved):
⚡

PR #2415, opened 2021-08-20
fix: Update browser navigation on directory links

acezard:
Related to: https://trello.com/c/BdnfOXYA/1510-%F0%9F%93%81-drive-le-lien-sur-le-chemin-du-dossier-ne-fonctionne-pas-bien

Issues:
- note directory opens a note instead of its dir
- other files open the dir, but the preview tool when going back

Bugs:

- conflict with FileName onClick() and Link in Dir Name (link inside a link => double routing)

Resolution:
- add a guard on file opening method

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,7 +159,7 @@
     "react-measure": "2.5.0",
     "react-redux": "7.2.0",
     "react-remove-scroll": "^2.4.1",
-    "react-router": "3.2.4",
+    "react-router": "3.2.6",
     "react-tooltip": "3.11.1",
     "redux": "3.7.2",
     "redux-logger": "3.0.6",
--- a/src/drive/web/modules/filelist/FileOpener.jsx
+++ b/src/drive/web/modules/filelist/FileOpener.jsx
@@ -42,14 +42,9 @@ const enableTouchEvents = ev => {
     return false
   }
 
-  // remove events when they are on the file's path, because it's a different behavior
-  const parentLink = getParentLink(ev.target)
-  if (
-    parentLink &&
-    parentLink.className.indexOf(styles['fil-file-path']) >= 0
-  ) {
-    return false
-  }
+  // We need to filter here, is the event target the FileOpener or a link INSIDE the FileOpener?
+  // Assuming the FileOpener itself will never be inside an anchor, we can use the following guard
+  if (ev.srcEvent.target.closest('a')) return false
 
   return true
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "semi": false,
+  "singleQuote": true
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fixed an error in directory tree names appearing under filenames where sometimes, the path appeared scrambled
 * Fixed an error where creating a directory sent two save actions instead of one
 * Added a missing loading status on delete confirm modal button
+* Fixed issues related to recent view not going where it should when navigating back and forth in directory paths
 
 ## 🔧 Tech
 * Add CodeQL in order to scan the code🚫
--- a/src/drive/web/modules/views/Recent/index.spec.jsx
+++ b/src/drive/web/modules/views/Recent/index.spec.jsx
@@ -33,6 +33,8 @@ jest.mock('drive/web/modules/views/hooks', () => ({
   useFilesQueryWithPath: jest.fn()
 }))
 
+const Folder = () => <div>Folder</div>
+
 const setup = () => {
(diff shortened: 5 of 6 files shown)


src/drive/web/modules/filelist/FileOpener.jsx
```diff
@@ -80,6 +80,10 @@ const FileOpener = ({
             ev.srcEvent.stopImmediatePropagation()
             toggle(ev.srcEvent)
           } else {
+            // We need to filter here, is the event target the FileOpener or a link INSIDE the FileOpener?
+            // Assuming the FileOpener itself will never be inside an anchor, we can use the following guard
+            if (ev.srcEvent.target.closest('a')) return
```
Crash--, line 85: Should we put this logic in enableTouchEvents() method?

acezard: hum yeah I guess so, we can put it in the first guard (L32)

acezard: actually line 45 was trying to do that it seems, I guess it's not working anymore for some reason

acezard: updated

package.json
```diff
@@ -159,7 +159,7 @@
     "react-measure": "2.5.0",
     "react-redux": "7.2.0",
     "react-remove-scroll": "^2.4.1",
-    "react-router": "3.2.4",
+    "react-router": "3.2.6",
```
Crash--, line 162: should be done in another commit.

src/drive/web/modules/filelist/FileOpener.jsx
```diff
@@ -42,14 +42,9 @@ const enableTouchEvents = ev => {
...
-  const parentLink = getParentLink(ev.target)
-  if (
-    parentLink &&
-    parentLink.className.indexOf(styles['fil-file-path']) >= 0
-  ) {
-    return false
-  }
+  // We need to filter here, is the event target the FileOpener or a link INSIDE the FileOpener?
+  // Assuming the FileOpener itself will never be inside an anchor, we can use the following guard
+  if (ev.srcEvent.target.closest('a')) return false
```
Crash--, line 47: I know that there is no test for this method, but I think this is something we should add. We have too many regressions on this. What do you think?

acezard: I don't know it's hard to test, would basically needs to render a whole DOM tree up to window. I'm basing myself on the assumption that it's impossible to find an anchor parent in the event unless it was launched by a child node, ie router link in this case. I tried to not be specific but I could also use a class on the folder link. But then the bug could happen elsewhere if some link is added

package.json
acezard: updated

src/drive/web/modules/filelist/FileOpener.jsx
acezard: I guess FileOpener having anchor as a parent should be a bug anyway, it would not be valid HTML

Crash--: > I don't know it's hard to test, would basically need to render a whole DOM tree up to window. I'm basing myself on the assumption that it's impossible to find an anchor parent in the event unless it was launched by a child node, ie router link in this case. I tried to not be specific but I could also use a class on the folder link. But then the bug could happen elsewhere if some link is added

We already have a DOM test for /recent view. We already do some click on a few actions. I don't think it would be difficult to test with that, no?

acezard: I guess it's possible sure, maybe verifying react-router is only invoked once. Or creating stubs and see where react-router sends me on different clicks. I'm worrying about delivery time though, my tests are failing in local and I don't know why. I need to fix this and then write the tests. I'd propose to do this in another PR if at all possible

Crash--: >  I need to fix this and then write the tests. I'd propose to do this in another PR if at all possible

Since the test file already exist, already display a File within the view with a path displayed I think, we should be able to click on the path and check the new url. I think this is mandatory.

Crash--: You can also only test this method by creating an event with different kind of targets / html .

acezard: I added two expects (I would prefer to have multiple tests but this test is already using a pattern with a lot of assert in a single test). 

I also add a small prettier file to match the eslint config for IDE purposes, not just mine but everyone having Prettier plugin + VSCODE for instance. Is that ok?

Crash-- (approved):
Thanks for the effort

PR #2418, opened 2021-08-23
feat: Do not enable redux-logger by default

paultranvan:
redux-logger was enabled by default, even in production. This is useful
to debug, but is also quite verbose. Therefore, we disable it by
default. Note one can still use the redux brower extension to get useful
insights.

### DIFF
--- a/src/drive/store/configureStore.js
+++ b/src/drive/store/configureStore.js
@@ -35,7 +35,7 @@ const configureStore = options => {
     middlewares.push(createTrackerMiddleware())
   }
 
-  if (options.logger !== false) {
+  if (options.logger === true) {
     middlewares.push(createLogger(loggerOptions()))
   }
 
--- a/src/drive/store/configureStore.js
+++ b/src/drive/store/configureStore.js
@@ -1,5 +1,6 @@
 /* global __DEVELOPMENT__, __TARGET__ */
 import { compose, createStore, applyMiddleware } from 'redux'
+import flag from 'cozy-flags'
 import { createLogger } from 'redux-logger'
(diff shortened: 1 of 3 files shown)


src/drive/store/configureStore.js
```diff
@@ -35,7 +35,7 @@ const configureStore = options => {
     middlewares.push(createTrackerMiddleware())
   }
 
-  if (options.logger !== false) {
+  if (options.logger === true) {
```
Crash--, line 38: We should add a flag in order to be able to use it without having to change the code. WDYT?

paultranvan: Sounds good! Something like 5ff40be03ca279d1b7c67346a3b12362fa1867d0 ?

src/drive/store/configureStore.js
```diff
@@ -35,7 +36,7 @@ const configureStore = options => {
     middlewares.push(createTrackerMiddleware())
   }
 
-  if (options.logger !== false) {
+  if (flag('drive.logger') === true) {
```
acezard, line 39: Isn't the equality comparison to `true` redundant? If it's a truthy value and we only want strict true, maybe we can convert the result with `Boolean()` or double NOT operator `!!`. This is just a style concern. Though from looking at the flag it would be strange to me to have truthy values that I don't want (but then it's a cozy-flags issue)

acezard (approved):

src/drive/store/configureStore.js
paultranvan: We can have flags that are not boolean, but it will be in this case, so `if (flag('drive.logger')` should be enough

Crash-- (approved):

PR #2426, opened 2021-09-17
Add menu entry to add encrypted folder 

paultranvan:
If the flag 'drive.encryption' is enabled, one can create an encrypted
folder.
An encrypted folder does the following things:

- Ask the user to unlock the vault, if it is locked
- Generate an encryption key, which is encrypted and stored in a
io.cozy.files.encryption doc
- Create a relationship between the folder doc and the encryption doc

For now, an encrypted folder does nothing but asking the user to unlock
the vault to enter the folder.


src/drive/targets/public/index.jsx
```diff
@@ -92,7 +92,6 @@ const init = async () => {
     cozyURL: cozyUrl,
     token: sharecode
   })
-
   const polyglot = initTranslation(dataset.locale, lang =>
```
Crash--, line 95: we can't share an encrypted folder for now?

Crash--: No target mobile update ?

src/drive/lib/doctypes.js
```diff
@@ -15,6 +16,10 @@ export const schema = {
       old_versions: {
         type: 'has-many',
         doctype: 'io.cozy.files.versions'
+      },
+      encryption: {
+        type: 'io.cozy.files:has-many',
```
Crash--, line 21: do we really need `io.cozy.files:` here? It's not just `has-many`?

src/drive/web/modules/drive/Toolbar/components/AddEncryptedFolderItem.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+import { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu'
+import EncryptedFolderIcon from 'drive/web/modules/views/Folder/EncryptedFolderIcon'
+import { useVaultUnlockContext } from 'cozy-keys-lib'
+import {
+  showNewFolderInput,
+  encryptedFolder
+} from 'drive/web/modules/filelist/duck'
+
+const AddEncryptedFolderItem = translate()(({ t, addEncryptedFolder }) => {
```
Crash--, line 13: you can remove the `translate()` HOC and use the `use18n()` hook inside your component.

src/drive/web/modules/drive/Toolbar/components/AddEncryptedFolderItem.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const { showUnlockForm } = useVaultUnlockContext()
+
+  console.log('add encrypted')
```
Crash--, line 16: to remove

src/drive/web/modules/filelist/AddFolder.jsx
```diff
@@ -76,13 +77,24 @@ const mapStateToProps = state => ({
...
+  const isEncrypted = isEncryptedFolder(state)
+  return dispatch(
+    createFolder(ownProps.client, ownProps.vaultClient, name, {
+      isEncryptedFolder: isEncrypted
     })
-  },
+  ).then(() => {
+    if (ownProps.refreshFolderContent) ownProps.refreshFolderContent()
+    return dispatch(hideNewFolderInput())
+  })
```
Crash--, line 93: async / await

src/drive/web/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -98,6 +101,7 @@ const FolderViewBody = ({
   const isSharingContextEmpty = Object.keys(sharingsValue).length <= 0
 
   const { syncingFakeFile } = useSyncingFakeFile({ isEmpty, queryResults })
+  const isEncFolder = isEncryptedFolder(client.store.getState())
```
Crash--, line 104: Why do we need to read it from the store? Don't we already have the `io.cozy.files` here? And then read its metadata without having to store it in the store? If you can avoid to manipulate / store in redux you should do.

src/drive/lib/doctypes.js
paultranvan: No, `io.cozy.files:has-many` is the expected name for HasManyFiles: https://github.com/cozy/cozy-client/blob/master/docs/relationships.md#files-relations

src/drive/targets/public/index.jsx
paultranvan: No we can't for sharing. Mobile should be added indeed

src/drive/web/modules/views/Folder/FolderViewBody.jsx
paultranvan: For now in the component we only have the dirID, but we could pass the whole `io.cozy.files` in props. But why is it better than reading the store? Is it a costful operation? Or is it because we want to less rely on the store?

Crash--: Having to read it from the store, makes the "externalization" harder. We tend to remove the use of redux in our app's land code. Why? Because of its verbosity and because we (mainly) duplicate the informations. Redux's store should only be used to store DATA from your cozy, not for the app's state. And since the communication between the app and your "cozy" is done by CozyClient, only cozyclient should store data in redux. Elsewhere, context should be used. 

Here, I think the UI can handle that directly by reading if the current folder is encrypted or not. Like that, no need to dispatch action to set the attribute or else. Only one path to read the info, and one path to set it.

paultranvan: Ok, I changed the behaviour to directly read from the `displayedFolder` props

src/drive/lib/encryption.js
```diff
@@ -0,0 +1,36 @@
...
+  const docId = `${DOCTYPE_FILES}/${dir._id}`
+  const key = await vaultClient.generateEncryptionKey()
+  const { data: encryption } = await client.create(DOCTYPE_FILES_ENCRYPTION, {
+    _id: docId,
+    key: key.encryptedKey.encryptedString
+  })
+
+  // Add the relationship
+  const hydratedDir = client.hydrateDocument(dir)
+  hydratedDir.encryption.addById(encryption._id)
```
Crash--, line 35: since addById is doing a save, should we await it?

paultranvan: Certainly, done with 95fcb1e

Crash-- (approved):

PR #2433, opened 2021-10-07
fix: Do not update files in parallel

paultranvan:
Trying to update all files in parallel might cause issues with nsjail
because of internal restrictions, if we try to udpate too many files at
once: we experienced failures in production in some cases.
As performances are not critical for this service, we simply remove the
parallelism.


paultranvan:
> Est-ce que tu peux rajouter une entrée dans le fichier CHANGELOG stp ?

Done!

PR #2437, opened 2021-11-15
feat: Upload encrypted files and open them, on mobile

paultranvan:
From the mobile app, we are now able to upload encrypted files. It is
also now possible to visualize them through the cordova `open`.
Note the direct display through the Viewer is not possible at this
moment because it requires changes in cozy-ui.

This also uses the mobile upload methods moved in cozy-client in https://github.com/cozy/cozy-client/pull/1051 instead of using the one in cozy-scanner.


Crash-- (approved):
let's fix the tests, but code looks good :)

paultranvan:
@Crash-- tests fixed and I added one in the meantime ;)

Crash-- (approved):

PR #2442, opened 2021-12-07
fix(drive): add client validation of illegal characters

trollepierre:
[Trello Issue](https://trello.com/c/66o9Dm5f/1148-drive-gestion-du-message-derreur-si-le-dossier-fichier-contient-un-caract%C3%A8re-interdit) 

Drive - Gestion du message d'erreur si le dossier / fichier contient un caractère interdit

- Afficher un Alert.error en indiquant que le nom du fichier contient un caractère interdit. Si le dit caractère n'est pas retourné dans la requête HTTP, demander à nono si y a une liste accessible quelque part pour afficher un message pertinent à l'utilisateur. 

[equivalent MR in cozy-client to merge before this one in cozy-drive](https://github.com/cozy/cozy-client/pull/1085)

Capture d’écran 2021-12-10 à 18 06 48

Capture d’écran 2021-12-10 à 18 07 34


README.md
```diff
@@ -44,12 +44,28 @@ $ cd cozy-drive
...
+
+Run in build mode:
+```sh
+$ yarn start:drive:browser
+```
+
+Build the app and run in cozy-stack:
+```sh
+$ yarn build:drive
+$ cd ./<path_de_ton_dossier_de_dev>/cozy-stack  && cozy-stack serve --appdir drive:/<path_de_ton_dossier_de_dev>/cozy-drive/build/drive
```
trollepierre, line 66: to rewrite better

src/drive/web/modules/drive/RenameInput.jsx
```diff
@@ -37,8 +50,12 @@ export const RenameInput = ({
       onSubmit={async newValue => {
         const newName = withoutExtension ? newValue + extension : newValue
         try {
-          await updateFileNameQuery(client, file, newName)
-          if (refreshFolderContent) refreshFolderContent()
+          if(isSanitized(newValue)) {
+            await updateFileNameQuery(client, file, newName)
+            if (refreshFolderContent) refreshFolderContent()
+          } else {
+            Alerter.error('alert.file_name_not_sanitized', { fileName: newName })
```
trollepierre, line 57: to test

src/drive/web/modules/drive/RenameInput.jsx
```diff
@@ -37,8 +50,12 @@ export const RenameInput = ({
...
+          }
         } catch {
```
trollepierre, line 59:
```suggestion
        } catch (error) {
          if (error.message.includes('Invalid filename: empty or contains an illegal character')) { 
             Alerter.error('alert.file_name_not_sanitized', { fileName: newName })
          } else {
```

trollepierre: in order to make frontend more resistant if backend updates the illegal characters list

src/drive/locales/en.json
```diff
@@ -321,6 +321,7 @@
     "empty_trash_success": "The trash has been emptied.",
     "folder_name": "The element %{folderName} already exists, please choose a new name.",
     "file_name": "The element %{fileName} already exists, please choose a new name.",
+    "file_name_not_sanitized": "The element %{fileName} contains invalid characters, please choose a new name.",
```
trollepierre, line 324: translations to fill in FR as well, but where?

README.md
```diff
@@ -44,12 +44,28 @@ $ cd cozy-drive
...
+:warning: During its early ages, _cozy-drive_ uses beta versions of [cozy-ui] and [cozy-client-js], take a look at the ["living on the edge" note](#living-on-the-edge) below to know how to install and configure the latest available versions.
 
 Cozy's apps use a standard set of _npm scripts_ to run common tasks, like watch, lint, test, build…
 
+Run in dev mode:
+```sh
+$ yarn watch:drive:browser
+```
+
+Run in build mode:
```
Ldoppea, line 58: maybe using "build" is a bit confusing here, as we have `yarn build` for that.

README.md
Ldoppea, line 66: If installed correctly, you shouldn't have to `cd` on cozy-stack dir to execute it.
```suggestion
$ cozy-stack serve --appdir drive:/<project_absolute_path>/cozy-drive/build/drive
```

src/drive/web/modules/drive/RenameInput.jsx
Ldoppea: This suggestion would stop working the day we decide to change the error message on stack side. But I suppose the error code won't be enough to guess the error.

What is the risk of having this error message change in the future @nono ?

src/drive/web/modules/drive/RenameInput.jsx
```diff
@@ -18,6 +18,19 @@ const updateFileNameQuery = async (client, file, newName) => {
   })
 }
 
+const isSanitized = fileName => {
+  if (fileName === '.' || fileName === '..') {
+    return false;
+  }
+  let result = true;
+  '/\x00\n\r'.split('').forEach(forbiddenCharacter => {
```
Ldoppea, line 26: More forbidden characters exist. You can find a (_should be complete_) list here : https://github.com/cozy-labs/cozy-desktop/blob/master/core/incompatibilities/platform.js#L110-L160

Ldoppea (changes requested):
Thanks for the PR and Kudos for the updated doc 👍 

I have a few changes requests (see comments). Also can you split the commit in multiple ones? I see 3 commits : one for the `.nvmrc` edit, one for the doc update and one for the feature implementation.

Last but not least : this problematic is handled differently on multiple projects (drive, desktop, mespapiers etc). Maybe we can find a way to mutualise those codes somewhere? 

@Crash-- is that something that we can handle by using schema attribute validation? https://docs.cozy.io/en/cozy-client/schema/

src/drive/web/modules/drive/RenameInput.jsx
Crash--: @Ldoppea This list is about Desktop. Not Cozy "in general". We've specific forbidden characters for different OS, but we do not add all these forbidden characters to Cozy (yet?). This is a discussion we had multiple times, but we can launch it again to see if something has changed. 

Cozy-Stack will throw an error only for this small list.

Crash--:
> @Crash-- is that something that we can handle by using schema attribute validation? https://docs.cozy.io/en/cozy-client/schema/

This is something we've in mind for several months now: Schema validation. We can do that on client's side, but also on stack's side. There are pros and cons for each options and we haven't decided yet. 

For today, I think that having the list of forbidden char in this model https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js will be way better. Also `FileCollection` (https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L388-L406) should handle this error, like that, every app will have the gain.

src/drive/web/modules/drive/RenameInput.jsx
Crash--: This error message `The element %{fileName} contains invalid characters, please choose a new name.` is not helping the user to choose a new one. We should point out the forbidden character he used in the error's message in order to help the correction.

nono: > What are the chances of having this error message changed in the future @nono ?

Very low, we don't change error messages, except if we have a very good reason to do that.

Ldoppea:
> I have a few changes requests

I just realized the PR was marked as WIP, maybe my comments were too early, sorry for that.

src/drive/web/modules/drive/RenameInput.jsx
Ldoppea: What would be the benefits of maintaining a different list on OS and on Cozy? Wouldn't it be easier to have the more restrictive list everywhere?

> This is a discussion we had multiple times

Do we have a SEC somewhere? If not maybe we should write one next time we will discuss this. This is the kind of question that would come over and over every time new devs will encounter those codes 😅

nono: We want people to be able to synchronize all of their files from their computer to Cozy. For example, we have seen macOS users to have office document with a ? in the filename. And cozy-desktop should be able to upload them to the Cozy in background, with no user interaction.

README.md
trollepierre: I need a little more explanation in order to fill this line. For now, I don't see the point of using yarn start:drive:browser. But maybe you can suggest me a better way to rewrite the line 🙏

src/drive/locales/en.json
```diff
@@ -321,6 +321,7 @@
     "empty_trash_success": "The trash has been emptied.",
     "folder_name": "The element %{folderName} already exists, please choose a new name.",
     "file_name": "The element %{fileName} already exists, please choose a new name.",
+    "file_name_illegal_characters": "The element %{fileName} contains invalid characters: %{characters}",
```
trollepierre, line 324: Don't forget to update translation for other locale

src/drive/web/modules/drive/RenameInput.jsx
trollepierre: the check has been moved into cozy-client in this MR: https://github.com/cozy/cozy-client/pull/1085

I resolve this thread that is outdated now

src/drive/web/modules/drive/RenameInput.jsx
trollepierre: this thread is outdated, I resolve this conversation

src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.jsx
```diff
@@ -54,8 +54,25 @@ const ShortcutCreationModal = ({ onClose, onCreated, displayedFolder }) => {
         Alerter.success(t('Shortcut.created'))
         if (onCreated) onCreated()
         onClose()
-      } catch (e) {
-        Alerter.error(t('Shortcut.errored'))
+      } catch (error) {
+        if (
+          error.message.includes(
+            'Invalid filename containing illegal character(s):'
```
Ldoppea, line 60: As we maintain CozyClient, I think it would be great to type our errors. For example here we could have
```js
class CozyInvalidCharacterError extends Error {
  constructor(errorMessage, invalidChars) {
    // store custom properties
  }
}
```
And it would be consumed like this
```js
try {
} catch (e) {
  if (e instance of CozyInvalidCharacterError) {
    // our specific code
  }
}
```

Also this would reduce the number of `Invalid filename containing illegal character(s)` string declaration on cozy-client side.

But this seems not to be used a lot at Cozy. @Crash-- do you have any opinion on this?

Crash--: This is not the way we do it ATM. I don't know if I'm more convinced by creating custom classes for each of our errors or if we add something like "app error code". 

So let's move forward with the current implementation based on the string manipulation, and let's discuss about errors in an CozyClient's issue. 👍

Ldoppea: Here it is : https://github.com/cozy/cozy-drive/issues/2445

Crash--: I think we can move this issue to CozyClient, it's more relevant there ;). 

Also, take a look at FetchError from CozyClient

Crash-- (approved):

src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.jsx
Ldoppea: Holy potato, It was supposed to be on client, how am I able to do this kind of mistake with so few opened tabs?? 😅 

Issue transfered https://github.com/cozy/cozy-client/issues/1088

src/drive/lib/migration/qualification.js
```diff
@@ -100,15 +100,15 @@ export const extractFilesToMigrate = files => {
...
+ * @param {string} oldLabel - The old qualification label
  * @returns {string} The new qualification label
  */
 const getNewLabelSetFromCozyScanner = oldLabel => {
   if (oldLabel === 'registration') {
     return 'vehicle_registration'
   }
   if (oldLabel === 'insurance_card') {
-    return 'national_insurance_card'
+    return 'national_health_insurance_card'
```
Ldoppea, line 111: ⚠️  Those edits shouldn't be in this PR

src/drive/lib/migration/qualification.spec.js
```diff
@@ -62,6 +62,38 @@ describe('qualification migration', () => {
...
+    const qualif = getFileRequalification(file)
+    expect(qualif).toEqual({
+      label: 'national_health_insurance_card',
+      purpose: 'attestation',
+      sourceCategory: 'gov',
+      sourceSubCategory: 'health',
+      subjects: ['identity', 'insurance']
+    })
+  })
+
```
Ldoppea, line 96: ⚠️ those edits shouldn't be in this PR

Ldoppea (approved):

src/drive/lib/migration/qualification.js
Ldoppea: Issue resolved: those edits are related to the cozy-client upgrade.

src/drive/lib/migration/qualification.spec.js
Ldoppea: Issue resolved: those edits are related to the cozy-client upgrade.

Crash-- (approved):

PR #2448, opened 2021-12-21
fix: Wrong cozy-keys-lib mock 

paultranvan:
The c3bd4aa commit broke the app
because of cozy-keys-lib wrongly mocked into the AppRoute component.
Instead, we mock it in the revelant files to make the tests run.

### DIFF
--- a/src/drive/web/modules/navigation/Index.spec.js
+++ b/src/drive/web/modules/navigation/Index.spec.js
@@ -5,6 +5,10 @@ import { SHAREDWITHME_DIR_ID } from 'drive/constants/config'
 
 import { fetchSharing } from './Index'
 
+jest.mock('cozy-keys-lib', () => ({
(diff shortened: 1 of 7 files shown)


src/drive/web/modules/navigation/Index.spec.js
```diff
@@ -1,3 +1,7 @@
+jest.mock('cozy-keys-lib', () => ({
+  withVaultClient: jest.fn().mockReturnValue({}),
+  useVaultClient: jest.fn()
+}))
```
trollepierre, line 4: Can we keep the jest.mock below the import/require 🙏 ?

trollepierre (approved):

src/drive/web/modules/navigation/Index.spec.js
paultranvan: Sure!

PR #2447, opened 2021-12-17
fix(warning): remove warnings in test

trollepierre:
- Adds ConsoleUsageReporter that verify console is not used in test
- Quick fixes easy test to prevent spying console for real errors
- Spying console only when tests could not be easily fixed
- throw Node unhandledRejection errors during test to detect them

PR #2451, opened 2021-12-23
fix(warning): Remove warnings in test, follow recommandations, add ConsoleUsageReporter

trollepierre:
- Adds ConsoleUsageReporter that verify console is not used in test
- Quick fixes easy test to prevent spying console for real errors
- Spying console only when tests could not be easily fixed
- Throw Node unhandledRejection errors during test to detect them
- Upgrade cozy-client to remove fetchOwnPermission warning


trollepierre:
Apparently, some of Cozy App were implementing an equivalent system that would make pipeline fail on console error:
https://github.com/cozy/cozy-banks/blob/master/test/jest.setup.js#L80

However I found some console error in the Travis CI on the last success branch:
https://app.travis-ci.com/github/cozy/cozy-banks/jobs/553363462#L618

I have created a [branch with console warn and console error](https://github.com/cozy/cozy-drive/pull/2452/commits) , if you check the [Travis pipeline, it's failing](https://app.travis-ci.com/github/cozy/cozy-drive/builds/243919761)

Ldoppea:
The error you mention on [cozy-banks](https://app.travis-ci.com/github/cozy/cozy-banks/jobs/553363462#L618) is a log from React. React displays a lot of warning as errors, we have a lot of same in our project. For example `Typography` can generates a `<p> inside of <p>` error, and we have a lot of them.

Whatever your code do, please care of not breaking tests due to those side effect we can't control.

I did not read the code in details yet, but before I wan't to ask:
- Why don't we want console.log to happen in tests?
- Will this prevent the dev to add temporary debug logs in their tests?

trollepierre:
First let me explain why warnings why running tests are bad.
When developers add dangerous code in their test and new warnings messages are added in a test case that already shut a lot of warnings, developers tend not to pay attention to the warning messages. However warning messages are precious, and should be taken into consideration. Preventing console methods of shutting regularly is a good way to ensure quality of our application.

> The error you mention on [cozy-banks](https://app.travis-ci.com/github/cozy/cozy-banks/jobs/553363462#L618) is a log from React. React displays a lot of warning as errors (but they are considered as warnings), we have a lot of those in our projects. For example `Typography` can generates a `<p> inside of <p>` error, and we have a lot of them.

If you have a good reason, I recommend to disable the warning for the "justified error" in this case exclusively. But I recommend adding a link to the Github issue mentioning the "justification"

> Whatever your code do, please care of not breaking tests due to those side effect we can't control.
> 
> I did not read the code in details yet, but before I wan't to ask:
> 
>     * Why don't we want console.log to happen in tests?

Because all the warnings in the test are not justified warnings.
=> Most of the warnings explicit breaking changes that needed to be taken in account, when upgrading the dependencies (moreover, the warnings were triggered also in the production.)
See this PR to see which changes were required https://github.com/cozy/cozy-drive/pull/2447/files

> 
>     * Will this prevent the dev to add temporary debug logs in their tests?

=> It's allowed to add temporary debug logs, but not to merge them in master, thanks to the failing jobs `yarn test`.
=> If ever the console.log is justified, developers can spy the console.

Crash-- (approved):

PR #2456, opened 2021-12-24
fix(sharing): Upgrade cozy-sharing to fix typo in French

trollepierre:
https://github.com/cozy/cozy-libs/pull/1401

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "cozy-realtime": "3.13.0",
     "cozy-scanner": "2.0.1",
     "cozy-scripts": "5.1.1",
-    "cozy-sharing": "^3.10.0",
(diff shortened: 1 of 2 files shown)


Ldoppea (approved):
Looks good to me, but as this upgrade cozy-sharing I would a double check from one of the project leads.

https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/CHANGELOG.md

Many of the changes between `3.10.0` and `3.12.7` are from me when I reworked the sharing modal (see `3.11.0`, `3.12.0` and `3.12.1`).
I did some local tests in the past, but I not sure I checked every scenario supported by cozy-drive.

Crash--:
can be closed in favor of #2457

PR #2458, opened 2022-01-04
feat(rename): Alert correctly when renaming offline failed

trollepierre:


src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.jsx
```diff
@@ -50,12 +52,22 @@ const ShortcutCreationModal = ({ onClose, onCreated, displayedFolder }) => {
         url: makedURL
       }
       try {
-        await client.collection('io.cozy.files.shortcuts').create(data)
-        Alerter.success(t('Shortcut.created'))
-        if (onCreated) onCreated()
+        if (isOffline) {
+          Alerter.error('alert.offline')
+        } else {
+          await client.collection('io.cozy.files.shortcuts').create(data)
```
Crash--, line 58: We should avoid using the collection directly. Can't we use save() with the _type / doctype set to io.cozy.files.shortcuts ?

Crash-- (approved):

src/drive/web/modules/drive/Toolbar/components/ShortcutCreationModal.jsx
trollepierre: it should be done

trollepierre: I will ask another review as I am not very familiar with cozy-client.

Crash-- (approved):

PR #2459, opened 2022-01-05
feat: Do not display banner in flagship

acezard:
https://trello.com/c/UA4mUUko/148-e113-supprimer-lencart-proposant-dinstaller-drive-05-jh

Simply return null if flagship env is detected

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "cozy-ci": "0.4.1",
     "cozy-client": "27.6.3",
     "cozy-client-js": "0.19.0",
-    "cozy-device-helper": "1.12.0",
+    "cozy-device-helper": "^1.14.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.6.0",
     "cozy-harvest-lib": "5.3.1",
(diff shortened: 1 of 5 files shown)


src/components/pushClient/Banner.jsx
```diff
@@ -53,9 +54,13 @@ class BannerClient extends Component {
     track(element)
   }
 
+  isDisabled = () =>
+    Config.promoteDesktop.isActivated !== true ||
+    !this.state.mustShow ||
+    isFlagshipApp()
+
```
Crash--, line 61: We also have https://github.com/cozy/cozy-drive/blob/master/src/components/pushClient/Button.jsx

I think the right fix should be in Main.jsx https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/layout/Main.jsx

Also, I'm wondering if we don't have this pushClient in Photo too. Can you take a look?

acezard: will look

acezard: The conditionals look a bit wild

Crash-- (approved):

PR #2463, opened 2022-01-07
feat(deps): Upgrade cozy UI

Crash--:
Let's upgrade Cozy-UI. 

We had many late versions, let's upgrade


jestHelpers/setup.js
```diff
@@ -1,5 +1,15 @@
 import Enzyme from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
+
```
trollepierre, line 3: can I squash everything @Crash-- ?

Crash--: do want you it Mr!

trollepierre (approved):

jestHelpers/setup.js
trollepierre: who merge this?

PR #2466, opened 2022-01-11
chore(warnings): Mute warnings during build

trollepierre:
While building the app, a warning is displayed mentioning webpack cannot require conditionnally. 
This MR fixed that, adds a test on this fix and keep preventing error attached below.

### DIFF
--- a/src/drive/web/modules/services/components/iconsContext.js
+++ b/src/drive/web/modules/services/components/iconsContext.js
@@ -1,0 +1,9 @@
+const iconsContext = require.context(
+  'drive/assets/icons/',
+  false,
+  /icon-type-.*.svg$/
+)
+
+export default {
+  iconsContext
+}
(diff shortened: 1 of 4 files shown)


src/drive/web/modules/services/components/iconsContext.js
```diff
@@ -0,0 +1,9 @@
+const iconsContext = require.context(
```
trollepierre, line 1: oups:
Capture d’écran 2022-01-11 à 17 50 39

Crash--: I don't understand if I still need to read this PR or not with this screenshot?

trollepierre: The PR was in Draft when you saw it. Now the error above is fixed in the PR and the PR is ready for review.

trollepierre:
close for https://github.com/cozy/cozy-drive/pull/2476

PR #2467, opened 2022-01-12
test(jest-dom): add jest-dom matcher

trollepierre:


Crash-- (approved):
You should use `test:`  instead of chore, it's more relevant IMO.

PR #2469, opened 2022-01-12
chore(eslint): prevent warning - no-param-reassign + exhaustive-deps

trollepierre:


Crash-- (approved):
Great! 👍 

But this is not a `chore` commit since you updated code used in production. I think that `refactor` will do the trick ;)

Crash-- (approved):

PR #2473, opened 2022-01-17
test: use alias import while calling test folder

trollepierre:


Crash-- (approved):
chore(test) should not exist. You should use `test:`directly ;)

PR #2475, opened 2022-01-19
fix(dropzone): invalid prop were shouting errors instead of uploading file

trollepierre:
When uploading file, we used to have
Capture d’écran 2022-01-19 à 16 00 20


paultranvan (approved):
Thanks!

PR #2477, opened 2022-01-20
feat(add): Log sentry exception on click on add menu when offline

trollepierre:
Plenty of users complains because AddMenu button appears offline, but they are not.

We want to log exception to keep eyes on that bug.

### DIFF
--- a/src/drive/web/modules/drive/AddMenu/AddMenuProvider.jsx
+++ b/src/drive/web/modules/drive/AddMenu/AddMenuProvider.jsx
@@ -6,6 +6,7 @@ import React, {
   createContext
 } from 'react'
 import { useSelector } from 'react-redux'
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/drive/AddMenu/AddMenuProvider.jsx
```diff
@@ -50,6 +51,7 @@ const AddMenuProvider = ({
   const handleOfflineClick = useCallback(e => {
     e.stopPropagation()
     Alerter.error('alert.offline')
+    logException('Offline click on AddMenu button detected')
```
Crash--, line 54:
```suggestion
    logException(`Offline click on AddMenu button detected. Here is the value of window....:${window....onLine}`)
```

Crash-- (approved):

PR #2481, opened 2022-01-26
chore(stylint): Stylint with Cozy config each .styl files in src

trollepierre:
Previously, stylint checks only a folder in src/drive/styles .
Moreover it was not very strict, that is why I follow the recommended configuration of [stylint](https://www.npmjs.com/package/stylint)


trollepierre:
@Ldoppea @Crash-- @JF-Cozy : I asked you review, because I think we should use one config of stylus in all the projects of Cozy, and I would like your opinion.

Ldoppea (approved):
Looks good to me

Crash--:
Hmm it seems we already have a shared `.stylintrc`: https://github.com/cozy/create-cozy-app/blob/master/packages/cozy-scripts/template/app/.stylintrc & https://github.com/cozy/create-cozy-app/blob/master/packages/cozy-scripts/config/.stylintrc

trollepierre:
Thanks to Crash, I discover how to use directly Cozy shared config. Fixed done!

trollepierre:
@Crash-- : do you mean the changelog of Drive adding a
```
# 1.39.0
## 🔧 Tech
* Stylint with Cozy config each .styl files in src 
```

isn't it weird to prepublish a changelog of a release not published?

Crash--:
> @Crash-- : do you mean the changelog of Drive adding a

# 1.39.0
## 🔧 Tech
* Stylint with Cozy config each .styl files in src 

Exactly. 

> isn't it weird to prepublish a changelog of a release not published?

We do not publish. It's only on master. So if you take the master, you have all the changes. It makes sens to me.

PR #2488, opened 2022-01-31
feat: Add DACC service, to send anonymized measures.

paultranvan:
This service aggregates the file size by grouping them by the app/konnector
that created them.
It then sends all the measures to the DACC, an external service used to
aggregate anonymized measures, which require new remote doctype
permissions.
Note this feature is yet restricted by a flag.

### DIFF
--- a/src/drive/lib/dacc/query.js
+++ b/src/drive/lib/dacc/query.js
@@ -1,0 +1,31 @@
+import { DOCTYPE_FILES } from 'drive/lib/doctypes'
+
+/**
+ * Query files by created_at
+ *
+ * @param {object} client - The CozyClient instance
+ * @param {string} endDate - The max file date to query
+ * @param {string} bookmark - The query bookmark
+ * @returns {Array} The files sorted by (createdByApp, created_at)
+ */
+export const queryFilesByDate = async (client, endDate, bookmark = '') => {
+  const selector = {
+    'cozyMetadata.createdByApp': {
+      $gt: null
+    },
+    created_at: { $lte: endDate }
+  }
+  const options = {
+    partialFilter: {
+      type: 'file',
+      trashed: false
+    },
+    fields: ['cozyMetadata.createdByApp', 'size'],
+    indexedFields: ['cozyMetadata.createdByApp', 'created_at'],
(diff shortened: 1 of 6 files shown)


src/drive/targets/services/dacc.js
```diff
@@ -0,0 +1,63 @@
...
+    )
+    sizesBySlug[nonExcludedGroupLabel] = totalNonExcluded
+  }
+
+  const startDateMeasure = format(startOfMonth(aggregationDate), 'YYYY-MM-DD')
+
+  await sendToRemoteDoctype(client, remoteDoctype, sizesBySlug, {
+    measureName,
+    startDate: startDateMeasure
+  })
```
trollepierre, line 57: nit: if services are so complicated to test, I recommend to extract the logic out of the file, in order to test / document the feature, and keep only one single call of a method inside a service file.

src/drive/lib/dacc/query.js
```diff
@@ -0,0 +1,31 @@
...
+      trashed: false
+    },
+    fields: ['cozyMetadata.createdByApp', 'size'],
+    indexedFields: ['cozyMetadata.createdByApp', 'created_at'],
+    sort: [{ 'cozyMetadata.createdByApp': 'asc' }, { created_at: 'asc' }],
+    limit: 1000,
+    bookmark
+  }
+  // We directly use the collection to avoid using the store for nothing.
+  return client.collection(DOCTYPE_FILES).find(selector, options)
```
trollepierre, line 30: nit: in my opinion, query can be tested as any other file.

it should call client.collection with the correct params

trollepierre (approved):

src/drive/targets/services/dacc.js
paultranvan: Good point, I actually did this for a [konnector](https://github.com/konnectors/tracemob/blob/master/src/index.js)
This service is simpler and I lack of time to refactor, but it might be envisioned if this service evolves

trollepierre (commented):
could you add in the changelog.md the description of the feature added?  🙏

paultranvan:
> could you add in the changelog.md the description of the feature added? pray

Right, I forgot we do this now: 2d4fde2

PR #2492, opened 2022-02-02
fix(photos): Build failed because of incorrect require

trollepierre:
https://github.com/cozy/cozy-drive/pull/2481 introduces a bug while building Photos :

https://app.travis-ci.com/github/cozy/cozy-drive/jobs/558010052

### DIFF
--- a/src/photos/styles/topbar.styl
+++ b/src/photos/styles/topbar.styl
@@ -1,7 +1,8 @@
 @require 'components/forms.styl'
 @require 'settings/breakpoints.styl'
 @require 'settings/z-index.styl'
-@require 'styles/coz-bar-size.styl'
+
+$coz-bar-size = 3rem // @stylint ignore
 
 .pho-content-header
     display       flex


JF-Cozy (commented):
me semble qu'on a une variable global pour ça, à vérifier dans cozy-ui, non ? Genre un var(--xxx)

Crash-- (approved):

trollepierre:
@JF-Cozy : 

What do you suggest exactly?

- I think it's not a good idea to use UploadQueue specific style.
https://github.com/cozy/cozy-ui/blob/55b0995d955f5108245cedfa057b74e0cd4eb90e/react/UploadQueue/styles.styl#L9

- We can maybe use this value / not related to coz-bar : what do you  think?
https://github.com/cozy/cozy-ui/blob/0c6c0e93845b44c541a527e52f63e8ae39e7fd94/stylus/settings/spaces.styl#L20

PR #2496, opened 2022-02-04
feat: Implement cozy-bar AA navigation

acezard:
In order to make all **Cozy Apps** compatible with **Amirale**, we have to wrap them in the `cozy-intent` provider.

To do so, we have to:
- Add `cozy-intent`
- Upgrade `cozy-device-helper` to the latest version (necessary for `cozy-intent` as its peer dependency)
- Wrap the application at the highest possible level with the `<WebviewIntentProvider />`


Crash--:
Drive have severals target I don't think you have done all of them. 

Search for `CozyProvider` you'll find other places (like that you'll do Photos at the same time).

Also, there is a `DriveProvider`. Can't you use it?

package.json
```diff
@@ -118,11 +118,11 @@
     "cozy-ci": "0.4.1",
     "cozy-client": "27.15.0",
     "cozy-client-js": "0.19.0",
-    "cozy-device-helper": "1.16.1",
+    "cozy-device-helper": "^1.17.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.6.0",
     "cozy-harvest-lib": "5.10.5",
-    "cozy-intent": "1.5.0",
+    "cozy-intent": "^1.8.0",
```
trollepierre, line 125: let's use no ^ before deps here please

trollepierre (commented):
it's ok on my side, but I think @Crash-- 's approval is required after his comment

Crash-- (commented):
does drive public target uses this `App.jsx` component? If yes then 👍

acezard:
> does drive public target uses this `App.jsx` component? If yes then 👍

Yes it is used in tests

> does drive public target uses this `App.jsx` component? If yes then 👍

it does indeed https://github.com/cozy/cozy-drive/blob/master/src/drive/targets/public/index.jsx#L31

PR #2499, opened 2022-02-09
feat: Add date attribute to flag

paultranvan:
This will be useful to set up a custom date for the files query, so the
indicator can be more easily tested.

### DIFF
--- a/src/drive/targets/services/dacc.js
+++ b/src/drive/targets/services/dacc.js
@@ -1,67 +1,8 @@
 import log from 'cozy-logger'
-import CozyClient from 'cozy-client'
-import { schema } from 'drive/lib/doctypes'
-import flag from 'cozy-flags'
-import { endOfMonth, startOfMonth, format, subMonths } from 'date-fns'
 import fetch from 'node-fetch'
+import { run } from './dacc-run'
+
 global.fetch = fetch
-
-import {
-  aggregateFilesSize,
-  aggregateNonExcludedSlugs,
-  sendToRemoteDoctype
-} from 'drive/lib/dacc/dacc'
-
-/**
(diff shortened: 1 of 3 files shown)


src/drive/targets/services/dacc.js
```diff
@@ -35,11 +35,17 @@ export const run = async () => {
...
+    maxFileDateQuery
   } = daccFileSizeFlag
-  const aggregationDate = endOfMonth(subMonths(new Date(), 1))
+
+  const aggregationDate =
+    maxFileDateQuery || endOfMonth(subMonths(new Date(), 1))
 
   const sizesBySlug = await aggregateFilesSize(client, aggregationDate)
+  if (Object.keys(sizesBySlug).length < 1) {
+    log('info', `No files found to aggregate with date ${aggregationDate}`)
```
trollepierre, line 47: I would expect 2 tests for this kind of PRs, but I wonder if there are drawbacks testing service. I asked a question to @Crash-- , and according to his answer, I will approve or suggest the 2 tests

trollepierre: => https://github.com/cozy/cozy-drive/commit/ced3151db01869145235133e2a1bd74b972c0918

you can cherry pick this commit (and modify to comply with the business need) :-)

paultranvan: That's very nice, I slightly adapted it with [25d28d6](https://github.com/cozy/cozy-drive/pull/2499/commits/25d28d6b355233d0315a1029ca674df006785f8d) :)

trollepierre (approved):
test is failing 🙏

paultranvan:
Acutally, the tests were failing both because of master and a small change in the PR. It's fixed :)

PR #2506, opened 2022-02-14
chore(deps): remove cozy-jobs-cli useless devDependencies

trollepierre:
If needed, developers can install temporarily

fix #2304


JF-Cozy (approved):

trollepierre:
it may be used in test. To go deeper...

package.json
```diff
@@ -81,11 +81,11 @@
     "@testing-library/react-hooks": "3.7.0",
     "@welldone-software/why-did-you-render": "^6.1.4",
     "babel-core": "7.0.0-bridge.0",
+    "babel-runtime": "^6.26.0",
```
Crash--, line 84: do we really need this `babel-runtime`?

trollepierre: yes 😢 during yarn test. (this was the reason given by test explaining why they failed)

PR #2507, opened 2022-02-14
chore(bundlemon): add bundlesize

trollepierre:


Crash--:
I don't know why the previous size is not kept. Can you take a look?

PR #2508, opened 2022-02-15
test: add clearMocks in jest.config

trollepierre:
Automatically clear mock calls, instances and results before every test.
Equivalent to calling jest.clearAllMocks() before each test.
This does not remove any mock implementation that may have been provided.

First step before using resetMocks in jest.config

Reply to request from #2502


trollepierre:
@JF-Cozy  :pray:

Crash-- (approved):

PR #2521, opened 2022-02-23
File openings inside new tab

trollepierre:
This PR handles most of the case I found:
- Drive view
- Recent view
- Sharing view/index  (click on "/" path below a folder can have unexpected behaviour)
- Sharing view/ inside a folder
- Trash (not verified a last time)

Handle:
- img/mp3/txt file (so most common file)
- notes
- shortcut
- folder

But has not been cleant / tested at all.

⚠️ Won't fixed => to add in another story:
- in RecentView, right click on folder filepath MidEllipsis (because this span is located inside FileName inside FileOpener, that is an anchor opening the recent file - not the folder)

comment about this gif: 
- the link below the anchor is wrong (it's the same than the file of the table cell) => ❌ you can see on the below left corner on my browser
- when Cmd + Click, we have the expected feature (folder opened in a new tab) => ✅ I arrived inside the test folder

- when right click, opening, we open the file instead of the folder => ❌ I arrived on the Recent gif, instead of the folder test

- on click on Notes / Shortcut, this needs the big refactor "Don't use on click, use only anchor - prefetch (query?)"

### DIFF
--- a/src/drive/web/modules/filelist/File.jsx
+++ b/src/drive/web/modules/filelist/File.jsx
@@ -104,7 +104,9 @@ const File = props => {
     refreshFolderContent,
     isInSyncFromSharing,
     extraColumns,
-    breakpoints: { isExtraLarge, isMobile }
+    breakpoints: { isExtraLarge, isMobile },
+    fileUrlToNavigate,
+    folderUrlToNavigate
   } = props
 
   const isImage = attributes.class === 'image'
@@ -144,6 +146,8 @@ const File = props => {
         open={open}
         toggle={toggle}
         isRenaming={isRenaming}
+        fileUrlToNavigate={fileUrlToNavigate}
+        folderUrlToNavigate={folderUrlToNavigate}
       >
         <FileThumbnail
           file={attributes}
@@ -160,6 +164,7 @@ const File = props => {
           formattedUpdatedAt={formattedUpdatedAt}
           refreshFolderContent={refreshFolderContent}
           isInSyncFromSharing={isInSyncFromSharing}
+          folderUrlToNavigate={folderUrlToNavigate}
         />
         <LastUpdate
           date={updatedAt}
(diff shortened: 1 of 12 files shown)


src/drive/web/modules/filelist/File.jsx
```diff
@@ -62,10 +62,26 @@ const File = props => {
...
+      onFileOpen,
+      isAvailableOffline,
+      folderUrlToNavigate
+    } = props
     event.stopPropagation()
     if (isDirectory(attributes)) {
-      onFolderOpen(attributes.id)
+      console.log('is directory')
+
+      if (event.ctrlKey || event.metaKey || event.shiftKey) {
```
Crash--, line 77: it seems that you build an href for a folder https://github.com/cozy/cozy-drive/pull/2521/files#diff-66f571a15fdeccd6293e111e9483d731a1d8a1bddf527123ece478a9605e12bfR121 why do you need to check the event key?

src/drive/web/modules/filelist/FileOpener.jsx
```diff
@@ -110,15 +109,47 @@ const FileOpener = ({
...
+    file.attributes && file.attributes.type === 'directory'
+  const isFile = file =>
+    (file.attributes && file.attributes.type === 'file') ||
+    file._type === 'io.cozy.files'
+  const isShortcut = file => file.class === 'shortcut'
+  const isNote = file => file.name.endsWith('.cozy-note')
+  let buildHref = ''
+  if (isFolder(file)) {
+    buildHref = `/#${folderUrlToNavigate(file.id)}`
+  } else if (isNote(file)) {
```
Crash--, line 122: All these helpers method should already been in https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L36-L50 I think

Crash--:
The thing I dislike here is that we still have several ways to handle navigation. Sometimes we will have an href, so we'll be able to make a right click and use browser action, sometimes not since the href will be empty. Right?

zatteo:
Done in another PR by @cballevre

PR #2527, opened 2022-03-02
fix(deps): Upgrade cozy-sharing to stop shutting justifyContent grid warnings

trollepierre:


Crash--:
cozy-sharing has something to do with justifycontent ? Sharing should not have a direct dep to cozy-ui. peerDep & devDep. 

So I don't understand why we need to upgrade cozy-sharing to get off of this warning. We should only upgrade cozy-ui, no?

trollepierre:
Close, because Crash is totally right. My bad

PR #2535, opened 2022-03-08
fix(flags): Use kebab case in Drive feature flags

trollepierre:
To be compliant with the wiki page about "utilisation de flags"

⚠️ Some users may be impacted if the flags have already been set

Maybe we should activate/deactivate some flags?
Bottom Sheet custom feature flags has been created in March 2021
Only Office custom feature flags has been created in June 2021


src/lib/flags.js
```diff
@@ -17,11 +17,12 @@ export const initFlags = () => {
...
   flag('switcher', true)
-  flag('viewer-bottomSheet-tension')
-  flag('viewer-bottomSheet-friction')
-  flag('viewer-bottomSheet-clamp')
+  flag('viewer-bottom-sheet-tension')
+  flag('viewer-bottom-sheet-friction')
+  flag('viewer-bottom-sheet-clamp')
   flag('drive.onlyoffice.forceReadOnlyOnDesktop')
   flag('drive.onlyoffice.editorToolbarHeight')
   flag('drive.logger')
```
Crash--, line 29: this one we can remove

src/lib/flags.js
```diff
@@ -23,7 +23,7 @@ const flagsList = () => {
   flag('viewer-bottom-sheet-tension')
   flag('viewer-bottom-sheet-friction')
   flag('viewer-bottom-sheet-clamp')
-  flag('drive.onlyoffice.forceReadOnlyOnDesktop')
-  flag('drive.onlyoffice.editorToolbarHeight')
+  flag('drive.onlyoffice.force-read-only-on-desktop')
+  flag('drive.onlyoffice.editor-toolbar-height')
```
Crash--, line 28: we can't touch at this one. There is a log of cozy relying on this flag.

src/drive/web/modules/views/Public/index.jsx
```diff
@@ -188,7 +188,7 @@ const PublicFolderView = ({
   }, [setIsFabDisplayed, isMobile, hasWritePermissions])
 
   const showNewBreadcrumbFlag = flag(
-    'drive.breadcrumb.showCompleteBreadcrumbOnPublicPage'
+    'drive.breadcrumb.show-complete-breadcrumb-on-public-page'
```
Crash--, line 191: Same this is already used. Let's keep this one and remove it pretty soon

src/lib/flags.js
JF-Cozy: Tout ce qui concerne le bottomsheet peut être supprimé non ? C'est maintenant un composant dans cozy-ui, le comportement n'est plus trop à challenger dans une app je pense

trollepierre: @JF-Cozy : do you mean I can change this import? 
https://github.com/cozy/cozy-drive/blob/c2d06570eed6bfa2869842d3e34a1fb3c68b5432/src/drive/web/modules/viewer/Footer/FooterContent.jsx#L11
by
`import { BottomSheet } from 'cozy-ui/transpiled/react'` ?

src/drive/web/modules/views/Public/index.jsx
trollepierre: Suggestion: why not specifying an end date for each feature flags? (or at least, a date that flag is outdated)

src/lib/flags.js
trollepierre: @JF-Cozy ⬆️  🙏

trollepierre:
@Crash-- or @JF-Cozy : could you approve this PR or specify comments to help me to merge this 🙏

src/lib/flags.js
Crash--: Yep. And then you can remove the concerned flag.

trollepierre: @Crash-- : https://github.com/cozy/cozy-drive/blob/c3bd4aa1c00492d9671c9ec81da5293e9e3ef29d/src/drive/store/configureStore.js#L39 what should I do with this code, if I remove the drive.logger flag?

trollepierre: @JF-Cozy : 

I think there is some confusion in the Viewer / FooterContent :  
https://github.com/cozy/cozy-drive/blob/d0729f358224497f8bd08a8b33f65577e4cfba77/src/drive/web/modules/viewer/FilesViewer.jsx#L186 <= here drive FilesViewer tries to set its own FooterContent

but Cozy-UI viewer does not pass the Props to ViewerInformationsWrapper
https://github.com/cozy/cozy-ui/blob/5df4e2b2c8da396996b0be8be64b24ce03de999f/react/Viewer/index.jsx#L223

so this component ViewerInformationsWrapper uses directly FooterContent from cozy-ui : 
https://github.com/cozy/cozy-ui/blob/5df4e2b2c8da396996b0be8be64b24ce03de999f/react/Viewer/index.jsx#L171 

however I see a ViewerWithCustomPanelAndFooter  https://github.com/cozy/cozy-ui/blob/5df4e2b2c8da396996b0be8be64b24ce03de999f/react/Viewer/index.jsx#L270 and I have no idea, if Drive choose use it or not.

That makes me not able to test the deletion of the feature.

=> I will let the code like this, and create a new issue to fix this.
 @JF-Cozy : maybe you have a better knowledge than me on what should be done here.

trollepierre: => https://github.com/cozy/cozy-drive/issues/2569

JF-Cozy: Pour le Viewer, dans cozy-ui il est maintenant "intelligent" et devrait plutôt s'appeler FileViewer car s'occupe d'afficher les bons blocs en fonction du file passé. En effet on ne peut plus lui passer un content depuis l'app, mais en théorie ce n'est plus nécessaire. A priori tu peux donc utiliser la dernier version du Viewer dans Drive, beaucoup plus "simplement" qu'avant. Ici il y a un peu de refacto à faire je pense, d'où les guillemets...
Non tu ne devrais pas avoir à utiliser `ViewerWithCustomPanelAndFooter`.
C'est @Merkur39 qui a fait cette implémentation, si besoin de plus d'info.

JF-Cozy (approved):

trollepierre:
without answer from @Crash--  to this question https://github.com/cozy/cozy-drive/pull/2535#discussion_r838568697 , I am going to merge. This change will be implemented later, I guess

PR #2539, opened 2022-03-09
feat: Do not look for encryption when flag is not set

paultranvan:
The query is useless if the encryption flag is not set

### DIFF
--- a/src/drive/lib/encryption.js
+++ b/src/drive/lib/encryption.js
@@ -1,4 +1,5 @@
 import get from 'lodash/get'
+import flag from 'cozy-flags'
 import { DOCTYPE_FILES, DOCTYPE_FILES_ENCRYPTION } from 'drive/lib/doctypes'
 import { ENCRYPTION_MIME_TYPE } from 'drive/constants/config'
 import { buildEncryptionByIdQuery } from 'drive/web/modules/queries'
@@ -17,6 +18,9 @@ export const isEncryptedFile = file => {
 }
 
 export const getEncryptionKeyFromDirId = async (client, dirId) => {
+  if (!flag('drive.enable-encryption')) {
+    return null
+  }
   const docId = `${DOCTYPE_FILES}/${dirId}`
   const query = buildEncryptionByIdQuery(docId)
   const res = await client.query(query.definition, { options: query.options })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
  * Compute sizes in MB instead of MiB in dacc service.
  * Query files based on their uploaded date in dacc service.
  * Upload drop zone can correctly upload file with name containing ampersand "&" and hash "#"
+ * Do not query encryption files when flag is not set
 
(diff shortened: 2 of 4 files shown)


src/drive/lib/encryption.js
```diff
@@ -17,6 +18,9 @@ export const isEncryptedFile = file => {
 }
 
 export const getEncryptionKeyFromDirId = async (client, dirId) => {
+  if (flag('drive.enable-encryption')) {
```
trollepierre, line 21: I think we should use the opposite
`if (!flag('drive.enable-encryption')) {`

src/drive/lib/encryption.js
```diff
@@ -1,4 +1,5 @@
 import get from 'lodash/get'
+import flag from 'cozy-flags'
```
trollepierre, line 2: - please add the flag  in the list of the development flag: https://github.com/cozy/cozy-drive/blob/master/src/lib/flags.js
- enable the feature in prod' for cozy internal here: https://bender.cozycloud.cc/contexts/prod/cozy_internal

src/drive/lib/encryption.js
```diff
@@ -17,6 +18,9 @@ export const isEncryptedFile = file => {
 }
 
 export const getEncryptionKeyFromDirId = async (client, dirId) => {
+  if (flag('drive.enable-encryption')) {
+    return null
+  }
```
trollepierre, line 23: let's update the changelog also 🙏

trollepierre (commented):
Thanks for the feature!

src/drive/lib/encryption.js
paultranvan: * Ah right, I forgot the existence of this list
* We don't want to enable this feature yet

trollepierre: even for Cozy internal ? (it means the employee of cozy)

src/drive/lib/encryption.js
paultranvan: Sorry, I will eventually think of this... Maybe an automatic reminder in the PR would be useful?

trollepierre: https://github.com/cozy/cozy-drive/pull/2541

trollepierre: so bad memory, or so happy with automatisation ^^

trollepierre (approved):

src/drive/lib/encryption.js
```diff
@@ -17,6 +18,9 @@ export const isEncryptedFile = file => {
 }
 
 export const getEncryptionKeyFromDirId = async (client, dirId) => {
+  if (!flag('drive.enable-encryption')) {
```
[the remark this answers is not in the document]
paultranvan: Just for your pleasure: [70f2ee5](https://github.com/cozy/cozy-drive/pull/2539/commits/70f2ee5b1a963f7bda24a64f9411d28d57c1e751)

trollepierre (approved):

PR #2557, opened 2022-03-22
chore(deps): remove react-autosuggest as not used

trollepierre:
Autosuggest from ShareAutocomplete moved in cozy-sharing

fix #1748

- [x] Changelog updated if needed


Crash-- (approved):
Top !

PR #2576, opened 2022-04-06
Feat/file image loader

Crash--:
See commit message


src/lib/flags.js
```diff
@@ -30,4 +30,5 @@ const flagsList = () => {
   flag('drive.enable-encryption')
   flag('drive.dacc-files-size-by-slug')
   flag('drive.breadcrumb.showCompleteBreadcrumbOnPublicPage') // flagName should use kebab case
+  flag('drive.thumbnails-pdf.enabled')
```
trollepierre, line 33: Did you set it in Bender for Cozy Internal?

src/drive/web/modules/filelist/FileIcon.jsx
```diff
@@ -1,18 +1,19 @@
...
+import FileImageLoader from 'cozy-ui/transpiled/react/FileImageLoader'
 import styles from 'drive/styles/filelist.styl'
 
 const FileIcon = ({ file, size, isEncrypted }) => {
   const isImage = file.class === 'image'
   const isShortcut = file.class === 'shortcut'
-
-  if (isImage)
+  const shoudUseThumbnailsForPDF = flag('drive.thumbnails-pdf.enabled')
+  if (isImage || (shoudUseThumbnailsForPDF && file.class === 'pdf'))
```
trollepierre, line 12: I think this feature needs a little time to be tested. If you don't have it, tell me :-)

Crash--: What do you want to test here?

trollepierre:
```
import React from 'react'
import { render } from '@testing-library/react'
import FileIcon from './FileIcon'

jest.mock('cozy-flags', () => () => true)
jest.mock('cozy-ui/transpiled/react/FileImageLoader', () => () => (
  <div data-testid="FileImageLoader" />
))

describe('FileIcon', () => {
  it('should return file image loader when file is image', () => {
    // Given
    const file = { class: 'image' }

    // When
    const { getByTestId } = render(<FileIcon file={file} />)

    // Then
    expect(getByTestId('FileImageLoader')).toBeInTheDocument()
  })

  it('should return file image loader when file is pdf', () => {
    // Given
    const file = { class: 'pdf' }

    // When
    const { getByTestId } = render(<FileIcon file={file} />)

    // Then
    expect(getByTestId('FileImageLoader')).toBeInTheDocument()
  })
})
```

trollepierre: or https://github.com/cozy/cozy-drive/commit/9d449f4a9a78c177eb4a3a0d9f4336f3866bf86e

Crash--: Typically what I call "testing for testing sake" but hey you're in charge here now so I cherry-picked your commit.

trollepierre (approved):

PR #2577, opened 2022-04-06
fix: Fix pagination on Photos

Crash--:
We need to check if lastFetch is defined to know
if we're in a fetchMore context or in the first
render.


Ldoppea (approved):
I would edit the commit message to explain that all this is done to prevent the spinner to rerender on every "load more". For know you explain what we need, but not why we need it.

PR #2588, opened 2022-04-22
feat: Change Sentry url

JF-Cozy:
same url for prod/dev env, but environment var already passed
to senty https://github.com/cozy/cozy-drive/blob/master/src/drive/lib/reporter.js#L38


trollepierre (approved):
thank you very much 👍

PR #2603, opened 2022-05-16
fix: Request encryption key only when encryption flag is set

paultranvan:
In a shared folder, the upload was broken because one does not have the
rights to GET on the `io.cozy.files.encryption` doctype.
This is a quick fix to solve the issue found in prod


Crash-- (approved):
Can you edit the changelog please?

paultranvan:
> Can you edit the changelog please?

I did, but forgot to add it :disappointed:

PR #2609, opened 2022-05-25
feat: Support move for encrypted files

paultranvan:
This adds the possiilbity to move files from/to an encrypted folder.

3 scenarios are supported:
- From a non-encrypted folder to an encrypted folder
- From an encrypted folder to a non-encrypted folder
- From an encrypted folder to another encrypted folder

Note we do not support the moving of non-encrypted folder to an
encrypted one, because of the potential cost if it has a deep hierarchy
and/or many files. However, the moving of an encrypted folder to is
supported for both non-encrypted and encrypted folder, as the files remain
encrypted with the same encryption key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 # 1.42.0
 
+## ✨ Features
+
+* Support moving files from/to encrypted folder
+
 ## 🐛 Bug Fixes
 
 * Disable sharing on public file viewer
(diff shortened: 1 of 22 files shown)


JF-Cozy:
@paultranvan CI red ?

paultranvan:
@JF-Cozy Yes, I had many issues because of the cozy-keys-lib mock :( I thought it was fixed, but not everywhere obviously, I have to check that.

PR #2663, opened 2022-08-05
perf: improve perf inside search

trollepierre:
Needs: 
feat(stack-client): Add findAll method : https://github.com/cozy/cozy-client/pull/1219
feat(SearchBar): call onSelect when it's provided : https://github.com/cozy/cozy-bar/pull/770
feat(SearchBar): call onSelect when it's provided : https://github.com/cozy/cozy-bar/pull/771

- [x] Changelog updated if needed
- [x] Benchmark to do


src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -1,10 +1,12 @@
 /* global cozy */
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
-import { withClient } from 'cozy-client'
+import { withClient, Q } from 'cozy-client'
 
+import { DOCTYPE_FILES } from 'drive/lib/doctypes'
 import { TYPE_DIRECTORY, makeNormalizedFile } from './helpers'
 import { getIconUrl } from './iconContext'
+import { TRASH_DIR_ID } from '../../../../constants/config'
```
Crash--: import { TRASH_DIR_ID } from 'drive/constants/config'

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -46,21 +48,37 @@ class SuggestionProvider extends React.Component {
...
+                }
+              ]
+            }
+          })
+          .select(['_id', 'trashed', 'dir_id', 'name', 'path'])
+          .where({
+            _id: {
+              $gt: null
+            }
+          })
```
Crash--: can you link this to https://github.com/cozy/cozy-client/issues/1216 ?

trollepierre: done, in PR description, inside first commit description and third commit description. Thanks.

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -1,12 +1,11 @@
...
 import React from 'react'
 import FuzzyPathSearch from '../FuzzyPathSearch'
-import { withClient, Q } from 'cozy-client'
+import { withClient } from 'cozy-client'
 
-import { DOCTYPE_FILES } from 'drive/lib/doctypes'
 import { TYPE_DIRECTORY, makeNormalizedFile } from './helpers'
 import { getIconUrl } from './iconContext'
-import { TRASH_DIR_ID } from 'drive/constants/config'
+import { buildSuggestionsQuery } from '../../queries'
```
Crash--, line 8: By doing this, you include the `queries` file inside the bundled intent. So you can have increase the bundle size without notifying it. 

Today, bundlemon doesn't take a look at `/targets/intents/` bundle size.

src/drive/web/modules/queries.js
```diff
@@ -325,6 +325,36 @@ export const buildEncryptionByIdQuery = id => ({
...
+        $or: [
+          {
+            $exists: false
+          },
+          {
+            $eq: false
+          }
+        ]
+      }
+    })
```
Crash--, line 348: I think we can also add something like: 

```
path: {
$or: [{
$exists: false
},{
$regex": "^(?!/.cozy_trash)
}
]
}
```

to remove all the folder having their path starting with .cozy_trash/

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -46,44 +47,30 @@ class SuggestionProvider extends React.Component {
...
-        `/data/io.cozy.files/_all_docs?include_docs=true`
-      )
-      const files = resp.rows
-        // TODO: fix me
-        // eslint-disable-next-line no-prototype-builtins
-        .filter(row => !row.doc.hasOwnProperty('views'))
-        .map(row => ({ id: row.id, ...row.doc }))
 
-      const folders = files.filter(file => file.type === TYPE_DIRECTORY)
+    const files = await client.queryAll(buildSuggestionsQuery().definition)
```
Crash--, line 54: I'm wondering if this is not going to make the browser unresponsive. `client.query*` dispatch redux action, store result in a redux store, within the RAM and other. I'm not sure we need it here. 

What about calling `fetchJSON()` with the right path (the one you have inside the browser console) manually? Is it decrease the RAM consomption?

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -46,44 +47,30 @@ class SuggestionProvider extends React.Component {
...
+
+    const folders = files.filter(file => file.type === TYPE_DIRECTORY)
 
-      const notInTrash = file =>
-        !file.trashed && !/^\/\.cozy_trash/.test(file.path)
-      const notOrphans = file =>
-        folders.find(folder => folder._id === file.dir_id) !== undefined
+    const notInTrash = file =>
+      !file.trashed && !/^\/\.cozy_trash/.test(file.path)
```
Crash--, line 59: If you succeed to create the right partialIndex here https://github.com/cozy/cozy-drive/pull/2663/files#r938665394 this map will not be needed

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -46,44 +47,30 @@ class SuggestionProvider extends React.Component {
...
 
-      const normalizedFiles = await Promise.all(
-        normalizedFilesPrevious.map(
-          async file =>
-            await makeNormalizedFile(client, folders, file, getIconUrl)
-        )
+    const normalizedFiles = await Promise.all(
+      normalizedFilesPrevious.map(
+        async file =>
+          await makeNormalizedFile(client, folders, file, getIconUrl)
```
Crash--, line 68: This method can be pretty slow if we have a lot of Notes since it makes one call per note to get the url. Maybe we should fix that behavior too

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: > Today, bundlemon doesn't take a look at /targets/intents/ bundle size.
https://github.com/cozy/cozy-drive/pull/2664

trollepierre: > By doing this, you include the queries file inside the bundled intent. So you can have increase the bundle size without notifying it.

Well done! 

Capture d’écran 2022-08-05 à 16 15 50
Capture d’écran 2022-08-05 à 16 15 56

trollepierre: I remove the refacto(queries) commit then.

src/drive/web/modules/queries.js
trollepierre: Ok. I tried:
Capture d’écran 2022-08-05 à 16 27 10

I think I will have a look with @paultranvan

src/drive/web/modules/services/components/SuggestionProvider.jsx
Crash--: I'm reading bad? We pass from 170KB to 169KB. I think we can let this commit, no?

src/drive/web/modules/queries.js
paultranvan: @Crash-- I'd assume we entirely rely on the `trashed` attribute, but we have issues with it, right? I'd prefer to fix those potential issues rather than complexify the index because of it

Crash--: We don't know how to fix this particular issue. We assume this is caused by the sharing protocole or by cozy-desktop. but we're not sure about it. This is tracked by our `data scrubber` and you can see that this is one of our top issues. 

Capture d’écran 2022-08-08 à 10 20 03

We fix issues when we find them (for instance https://github.com/cozy/cozy-stack/pull/3442) but we're not clean yet.

trollepierre: changes made in https://github.com/cozy/cozy-drive/pull/2663/commits/59c86d5524666d92c6f711891919cb19813e0adc while @paultranvan is looking at the warning in cozy-client

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: changes made in https://github.com/cozy/cozy-drive/pull/2663/commits/59c86d5524666d92c6f711891919cb19813e0adc while @paultranvan is looking at the warning in cozy-client

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: In local,  without the refacto:
```
 [PASS] drive/intents-drive.(hash).min.css: 36.11KB < 40KB
  [PASS] drive/intents/drive.(hash).js: 169.84KB < 171KB
  [PASS] drive/intents/drive.(hash).min.css: 36.11KB < 37KB
```
we have the same value.

So yes, we can let the commit

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: - do you mean keeping the fetchJSON mechanism, already used in production? https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/services/components/SuggestionProvider.jsx#L55

I am going to benchmark this.

About performance, 
I am following `about:performance` in Firefox and the `Task Manager` in Brave, without seeing a high consumption. 

I am going to work on the benchmark with lots of data to figure out, that problem.

Crash--: How many io.cozy.files do you have on your test env?

trollepierre: 100 000 files  now.

| Méthode         | Nb requêtes     | Data |Temps |
|--------------|-----------|------------|------------|
| fetchJSON | 8      | 12 Mo       | 9.7 s |
| queryAll      | **120**  | 8 Mo       | 3.2 s |

FetchJSON : les 8 requêtes (dont 6 liées à des notes) :
Capture d’écran 2022-08-08 à 17 21 01


client.queryAll 2 cycles sur les 10 (exécutant à chaque fois 12 requêtes)
Capture d’écran 2022-08-08 à 17 15 03

Crash--: Tu donnes le fetchJSON ici sans l'optimisation que tu viens de faire à savoir de ne pas passer par all_docs. Ça ne fait pas sens. 

un truc du genre à tester : 
```
const req = await client.getStackClient.fetchJSON('POST', '/data/io.cozy.files', {le même selecteur que ton query})
while(req.next){
req.fetchJSON(...) //avec manipulation du bookmark à la main (comme le queryAll quoi)
}
```

A la fin tu dois avoir le même nombre de requête entre queryAll() et fetchJSON() car à la fin queryAll() fait un fetchJSON ! Et donc même nombre d'appels et même nombre de Mo transférés etc. 

Là où il va y avoir une différence c'est sur la RAM utilisée par l'application ! Car queryAll() va setter tout un store redux alors que fetchJSON ne va rien faire.

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: I think this is possible.
https://github.com/cozy/cozy-bar/blob/a61360a8d40f844102d9d12f4dd0ef9e5ae6bdba/src/components/SearchBar.jsx#L234
In order to make that, we could use onSelect as a function that will be called directly.

src/drive/web/modules/services/components/SuggestionProvider.jsx
Crash--: Aussi, c'est via l'onglet Performances de tes dev tools qu'on verra le pic de CPU et de RAM ainsi que l'occupation du thread JS.

trollepierre: after optims with paul:
Capture d’écran 2022-08-08 à 18 43 24

trollepierre: another fetchJSON:
Capture d’écran 2022-08-08 à 18 45 03

trollepierre: Idea: using a *fetchingAllJSONAndCreateIndex* method to handle a dynamic search + return result to display directly to the user. (what's about UserExperience)

Crash--: > Idea: using a fetchingAllJSONAndCreateIndex method to handle a dynamic search + return result to display directly to the user.

Can't we do something like:
```
const req = await client.collection('io.cozy.files').find(...)
```

to handle indexCreation and next attribute. 

We can create a DocumentCollection.findAll() method to handle the "full paginated result". @taratatach wants one too ;)

src/drive/web/modules/queries.js
```diff
@@ -325,6 +325,38 @@ export const buildEncryptionByIdQuery = id => ({
...
+      $gt: null
+    }
+  }
+  const options = {
+    partialFilter: {
+      _id: {
+        $ne: TRASH_DIR_ID
+      },
+      path: {
+        $or: [{ $exists: false }, { $regex: '^(?!/.cozy_trash)' }]
```
paultranvan, line 346: nit: as this predicate on the `path` shouldn't be necessary if the trash was more reliable, it would be worth addding a comment here to mention it

src/drive/web/modules/queries.js
```diff
@@ -325,6 +325,38 @@ export const buildEncryptionByIdQuery = id => ({
...
+      },
+      trashed: {
+        $or: [{ $exists: false }, { $eq: false }]
+      }
+    },
+    fields: ['_id', 'trashed', 'dir_id', 'name', 'path'],
```
paultranvan, line 352: No need for the file mime, or class? And type, to know when it's a directory?

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -46,44 +47,41 @@ class SuggestionProvider extends React.Component {
...
-            await makeNormalizedFile(client, folders, file, getIconUrl)
-        )
+    this.hasIndexFilesBeenLaunched = true
+
+    let next = true
+    let files = []
+    let bookmark = ''
+
+    while (next) {
+      const { selector, options } = prepareSuggestionQuery(bookmark)
```
paultranvan, line 60: nit: since the selector does not change, it is not ideal to recompute it at each while iteration, only the bookmark needs to change

src/drive/web/modules/services/components/SuggestionProvider.jsx
paultranvan: On my instance (with a lot of notes coming from the team sharing) , I confirm the time taken by this method is huge and worst than the query itself: a search takes ~22s to complete, most of this time took by the notes queries. So it makes definitely sense to fix it

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: https://github.com/cozy/cozy-client/pull/1219

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: This method is not called several time.

src/drive/web/modules/queries.js
trollepierre: type has been added as a fields

mime or class are not needed.

src/drive/web/modules/services/components/helpers.js
```diff
@@ -2,20 +2,33 @@ import { models } from 'cozy-client'
...
+  let path, url, onSelect
   if (isDir) {
     path = file.path
     url = urlToFolder
   } else {
     const parentDir = folders.find(folder => folder._id === file.dir_id)
     path = parentDir && parentDir.path ? parentDir.path : ''
     if (models.file.isNote(file)) {
-      url = await models.note.fetchURL(client, file)
+      onSelect = () => models.note.fetchURL(client, file)
```
trollepierre, line 31: @acezard : here is the call of the function (unique, in cozy)

src/drive/web/modules/services/components/SuggestionProvider.jsx
trollepierre: done

src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -38,52 +56,50 @@ class SuggestionProvider extends React.Component {
...
+    const { selector, options } = prepareSuggestionQuery()
+    const files = await client
+      .collection(DOCTYPE_FILES)
+      .findAll(selector, options)
 
-      const notInTrash = file =>
-        !file.trashed && !/^\/\.cozy_trash/.test(file.path)
-      const notOrphans = file =>
-        folders.find(folder => folder._id === file.dir_id) !== undefined
+    const folders = files.filter(file => file.type === TYPE_DIRECTORY)
```
Crash--, line 91: Since we can benefits from http2 multiplexing, I'm wondering if we should not make two requests: 
- one for the files 
- one for the folders 

We'll have the almost the same number of round trips (1 or 2) but we'll have a few of them parallelized. @paultranvan what do you think?

src/drive/web/modules/services/components/helpers.js
Crash--: What do we do with the return of this fetchURL() method? 

Because if we click on the suggestion, it'll call onSelect: https://github.com/cozy/cozy-bar/commit/7b7491dea69d75e8edc15aece2327972f1aae8bf#diff-bd7d5c5cdb7c2b6d78ffe2d7d7045e90beba76f929eff8695969c06a4d1accc7R237 

Then onSelect make the call to fetchURL() but we do we do with the result? There is no window.location or else, how does it works to open the note?

test/dummies/dummyFile.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Create a dummy file, with overridden value of given param
+ *
+ * @param file
+ * @returns {*&{path: string, name: string, icon: string, id: string, _id: string, dir_id: string, type: string}}  a dummy file
```
Crash--, line 5: What kind of jsdoc this is?

src/drive/web/modules/services/components/SuggestionProvider.jsx
paultranvan: I assume that, *most of the time*, the number of files will be order of magnitudes higher than the number of folders. So, the folder query might end up sooner than the files one, which is good, as we want to display the folders first. The downside is that we have now 2 indexes to compute instead of one, which might slow down the search when we make a query with non-indexed docs. 
I think what you propose is still relevant either way, as it should allow to get partial (i.e. folders) results faster and I think faster user response is more important than the overall performance. 
That being said, I also think paginated mango queries is sub-optimal for this use-case :/

test/dummies/dummyFile.js
trollepierre: this is generated by my IDE.

* means any type. (because i didn't set the type of the file)

& means intersection

I will reformulate.

Crash--: I don't see anything related to this syntax on https://jsdoc.app/ . Can you point me the doc please?

src/drive/web/modules/queries.js
```diff
@@ -325,6 +325,37 @@ export const buildEncryptionByIdQuery = id => ({
...
+ * Provide selector and options to fetch files
+ *
+ * @returns {{options: {MangoQueryOptions}, selector: object}} A minimalist file list
+ */
+export const prepareSuggestionQuery = () => {
+  const selector = {
+    _id: {
+      $gt: null
+    }
+  }
```
Crash--, line 338: @trollepierre after the merge of https://github.com/cozy/cozy-client/pull/1225 you should be able to remove this selector

test/dummies/dummyFile.js
trollepierre: I don't know exactly this syntax. My IDE generates it.

About: Allows any type
```
/**
 * @param {*} somebody - Whatever you want.
 */
```

I guess about intersection:
https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types

src/drive/web/modules/queries.js
trollepierre: done

src/drive/web/modules/services/components/helpers.js
trollepierre: ps: 
PRs de fix de cozy-bar en draft:
- https://github.com/cozy/cozy-bar/pull/773
- https://github.com/cozy/cozy-bar/pull/774

trollepierre: j'ai ajouté un nouveau commit pour fixer ceci.

paultranvan: `metadata` peut être assez volumineux, ça contient les qualifications, les infos EXIF, etc : https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files_metadata.md
C'est un peu dommage de devoir rapatrier tout le bloc juste pour savoir si c'est une note. Un `file.mime === 'text/vnd.cozy.note+markdown' n'est pas suffisant @Crash-- ?

Et sur un autre sujet, ça m'embête de voir que : 
1. les metadata des notes (content, schema, title, version) sont à la racine de metadata. J'aurais trouvé mieux de grouper tout ça dans un objet `metadata.notes`, ça devient un sacré bazard les metadata.
2. Je ne vois nul part documenté ces attributs : https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files_metadata.md

trollepierre: et un deuxième, car le postMessage serialize la fonction onSelect, je suis donc obligé de réaliser le fetchUrl directement depuis la cozy bar

src/drive/web/modules/queries.js
```diff
@@ -343,7 +343,16 @@ export const prepareSuggestionQuery = () => ({
...
-  fields: ['_id', 'trashed', 'dir_id', 'name', 'path', 'type', 'mime'],
+  fields: [
+    '_id',
+    'trashed',
+    'dir_id',
+    'name',
+    'path',
+    'type',
+    'mime',
+    'metadata'
```
Crash--, line 354: Une chose est sûre, c'est qu'on peut pas garder `metadata` au complet. Soit on arrive à aller lire quelques champs particuliers, soit on change la façon de checker si c'est une note. Car `metadata` sur les notes, ça peut vite peser des centaines de ko par note.

trollepierre: j'ai recréé un helper isNote, car les metadata peuvent être volumineuses.

Crash--: I don't know if I like this idea. 

You can add to the select(['metadata.title', 'metadata.version']) . title + version are not big. So the helper can be if(metata.title && metadata.version...) 

See a test query with metadata.datetime to see the result. 

Capture d’écran 2022-08-22 à 11 31 33

What do you think?

trollepierre: I have only one note in my local Drive.
When adding:
```
    'metadata.content',
    'metadata.schema',
    'metadata.title',
    'metadata.version'
```
as fields, the size of request increase 
from 650ko / 39,87 ko transferred in 538ms 
to 657,97 ko / 41,61ko transferred in 578ms.

Capture d’écran 2022-08-22 à 12 03 37
Capture d’écran 2022-08-22 à 12 05 08

That is quite huge, I think.

Crash--: This is not what I suggested. I suggested to only add version + title. Not schema + content.

trollepierre: perfect 
Capture d’écran 2022-08-22 à 14 11 58
 PR updated

src/drive/web/modules/services/components/helpers.js
```diff
@@ -1,21 +1,53 @@
...
+ *
+ * @param {IOCozyFile} file io.cozy.files
+ * @returns {boolean}
+ */
+export const isNote = file => {
+  if (
+    file &&
+    file.name &&
+    file.name.endsWith('.cozy-note') &&
+    file.type === FILE_TYPE
```
Crash--, line 18: you can edit this helper since you have more informations now (metadata.title & metadata.version)

Crash--: I'm wondering if we should not change the helper from cozy-client directly instead of having this one in Drive.

trollepierre: 6 usages in Cozy:

=> 5 in Drive, 1 in Cozy UI :

https://cs.github.com/?scope=org%3Acozy&scopeName=cozy&q=models.file.isNote%28

Crash--: I know and this is exactly my point... We learnt that requesting metadata.content it's big. So we should not request it anymore. And we should not relying on it to know if the files is a note or not a note. No?

Crash--:
@trollepierre Can you split this PR in two PR? One that remove the request for the note, and the other that change the all_docs to /data/io.cozy.files/_find ?

I think that the first one can be merged without the second one.

PR #2672, opened 2022-08-22
fix: Do not display Download actions on iOS Flagship app

Ldoppea:
Current implementation of iOS Flagship app doesn't handle files
download correctly

So we want to temporarily disable Download button on this platform

This commit is meant to be reverted when files download is fixed on iOS
Flagship app

---

Related PRs:

- https://github.com/cozy/cozy-ui/pull/2215

---

TODO:

- [ ] Upgrade cozy-ui after cozy/cozy-ui#2215 is merged (current upgrade commit doesn't have correct version and yarn.lock)

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "cozy-scripts": "^6.3.8",
     "cozy-sharing": "4.1.5",
     "cozy-stack-client": "^32.2.5",
-    "cozy-ui": "^68.4.0",
+    "cozy-ui": "^70.6.2",
     "date-fns": "1.30.1",
     "diacritics": "1.3.0",
     "fastclick": "1.0.6",
--- a/src/drive/web/modules/actions/index.jsx
+++ b/src/drive/web/modules/actions/index.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 
 import { isFile } from 'cozy-client/dist/models/file'
 import { ShareModal } from 'cozy-sharing'
-import { isIOSApp, isMobileApp } from 'cozy-device-helper'
+import { isFlagshipApp, isIOS, isIOSApp, isMobileApp } from 'cozy-device-helper'
 import { EditDocumentQualification } from 'cozy-scanner'
 import { getTracker } from 'cozy-ui/transpiled/react/helpers/tracker'
 import { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu'
@@ -92,6 +92,11 @@ export const download = ({ client, vaultClient }) => {
     : {
         icon: 'download',
         displayCondition: files => {
+          // Temporarily disable Download button on iOS Flagship app until
+          // the download feature is fixed on this platform
+          // When fixed on this platform, revert this commit from PR #2672
+          if (isFlagshipApp() && isIOS()) return null
+
           // We cannot generate archive for encrypted files, for now.
           // Then, we do not display the download button when the selection
           // includes an encrypted folder or several encrypted files
--- a/.bundlemonrc
+++ b/.bundlemonrc
@@ -32,7 +32,7 @@
     },
     {
       "path": "services/qualificationMigration/drive.js",
-      "maxSize": "242 KB"
+      "maxSize": "244 KB"
     },
     {
       "path": "vendors/drive.<hash>.<hash>.min.css",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2477,6 +2477,24 @@
   resolved "https://registry.yarnpkg.com/@juggle/resize-observer/-/resize-observer-3.3.0.tgz#2e7a2935adbfae46f9efbd1e1455254ee7ea2219"
   integrity sha512-P1v2nvK7z2gOLVM/bveIRLG9L99uEahTGgTltyF03zixZAjI9YmKLj5Z9MpS9wBIUt5WDoQORT2lXvLOIF89iA==
 
+"@material-ui/core@4.12.3":
+  version "4.12.3"
+  resolved "https://registry.yarnpkg.com/@material-ui/core/-/core-4.12.3.tgz#80d665caf0f1f034e52355c5450c0e38b099d3ca"
+  integrity sha512-sdpgI/PL56QVsEJldwEe4FFaFTLUqN+rd7sSZiRCdx2E/C7z5yK0y/khAWVBH24tXwto7I1hCzNWfJGZIYJKnw==
+  dependencies:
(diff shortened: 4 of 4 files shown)


src/drive/web/modules/actions/useActions.jsx
```diff
@@ -19,15 +19,21 @@ const withHideSelectionBarDispatch = (action, dispatch) => {
...
-
-    actions.push({ [name]: enhancedAction })
-  })
+  actionCreators
+    .map(createAction => {
+      const action = createAction(actionOptions)
+
+      // Temporarily disable Download button on iOS Flagship app until
+      // the download feature is fixed on this platform
+      // When fixed on this platform, revert PR #2672
```
trollepierre, line 28: nit: it's not revert PR, it's revert commit https://github.com/cozy/cozy-drive/pull/2672/commits/1d6666ae0c14348e5630afe0f76373e28ac281e8
```suggestion
      // When fixed on this platform, revert PR #2672
```

package.json
```diff
@@ -131,7 +131,7 @@
     "cozy-scripts": "^6.3.8",
     "cozy-sharing": "4.1.5",
     "cozy-stack-client": "^32.2.5",
-    "cozy-ui": "^68.4.0",
+    "cozy-ui": "^70.6.2",
```
trollepierre, line 134: it's not 71, it's 70

did you look at the breaking change of cozy-ui?
I think we should apply the codemods + remove @material-ui/core as dependency

src/drive/web/modules/actions/useActions.jsx
```diff
@@ -19,15 +19,21 @@ const withHideSelectionBarDispatch = (action, dispatch) => {
...
+      if (action === null) return null
+
+      const enhancedAction = withHideSelectionBarDispatch(action, dispatch)
+      const name = enhancedAction.label || enhancedAction.icon
+
+      actions.push({ [name]: enhancedAction })
+    })
+    .filter(x => x)
```
trollepierre, line 36: i think this line is useless

trollepierre:
I am quite afraid by the delta in the bundlemon report (2x + 13kb)

Crash--:
> I am quite afraid by the delta in the bundlemon report (2x + 13kb)

Mainly because we add a lot of icons in UI. And Drive still uses the Sprite.

package.json
Ldoppea: Yep this is mentioned in the TODO list from the PR, i cannot put the final commit here yet as cozy-ui's PR is still not merged

The codemods have been applied already, not Text component is used in the project. Regarding material-ui, I don't get it, the BC is not that we MUST remove it, but it is about them not being mandatory anymore. If we must remove them, maybe we should rewrite the BC?

src/drive/web/modules/actions/useActions.jsx
Ldoppea: Good catch, I put it before realizing that it was a push inside of a map. I'll remove it

src/drive/web/modules/actions/useActions.jsx
Ldoppea: What about `// When fixed on this platform, revert this commit from PR #2672` ?

I cannot link to the commit because its SHA may change after the rebase/merge.

trollepierre: Smart solution! Yes!

package.json
trollepierre: Yes it's not mandatory. => nit

But the maintainer of Drive will appreciate, not to have this useless dependency in the application.  😄

src/drive/web/modules/actions/index.jsx
```diff
@@ -63,6 +63,10 @@ export const share = ({ hasWriteAccess, pushModal, popModal }) => {
 }
 
 export const download = ({ client, vaultClient }) => {
+  // Temporarily disable Download button on iOS Flagship app until
+  // the download feature is fixed on this platform
+  // When fixed on this platform, revert this commit from PR #2672
+  if (isFlagshipApp() && isIOS()) return null
```
Crash--, line 69: You can do that differently and if you do that, you'll not have to adit the useActions hook. 

something like that should be enough: 
```
export const download = ({ client, vaultClient }) => {
  return isMobileApp()
    ? {
        icon: 'download',
        displayCondition: files => {
          if (isIOSApp()) return files.length === 1 && isFile(files[0])
          return files.reduce(
            (onlyFiles, file) => onlyFiles && isFile(file),
            true
          )
        },
        action: files => exportFilesNative(client, files, { vaultClient }),
        label: 'forwardTo',
        Component: function Download(props) {
          const { t } = useI18n()
          return (
            <ActionMenuItem
              onClick={() => {
                return exportFilesNative(client, props.files, { vaultClient })
              }}
              left={<Icon icon={isIOSApp() ? ShareIosIcon : ReplyIcon} />}
            >
              {t('SelectionBar.forwardTo')}
            </ActionMenuItem>
          )
        }
      }
    : {
        icon: 'download',
        displayCondition: files => {
        // ******* HERE is the change ****/ 
        if(isFlagshipApp() && isOS()) return false
          // We cannot generate archive for encrypted files, for now.
          // Then, we do not display the download button when the selection
          // includes an encrypted folder or several encrypted files
          return (
            !files.some(file => isEncryptedFolder(file)) &&
            !(files.length > 1 && files.some(file => isEncryptedFile(file)))
          )
        },
        action: files => downloadFiles(client, files, { vaultClient }),
        Component: function Download(props) {
          const { t } = useI18n()
          return (
            <ActionMenuItem
              onClick={() => {
                return downloadFiles(client, props.files, { vaultClient })
              }}
              left={<Icon icon={DownloadIcon} />}
            >
              {t('SelectionBar.download')}
            </ActionMenuItem>
          )
        }
      }
}
```

Crash-- (changes requested):
see comment 🙏

package.json
trollepierre: As we decide changes on that import during last front meeting, I think it's better to let it go.

src/drive/web/modules/actions/index.jsx
Ldoppea: How did I miss such an obvious method name? 😐 

Fixed in: https://github.com/cozy/cozy-drive/compare/cc0953ded5753e2e12e4172adbf11cfa67898622..5ae32366af08c739dcb09ecd56b524abd5b4b49b

Crash-- (approved):

PR #2675, opened 2022-08-24
fix(search): Improve speed of search suggestions

trollepierre:
By preventing every notes fetching url, loading only once a note is clicked

Last question: should we update isNote in cozy-client? (I think yes, but I will confirm)

after https://github.com/cozy/cozy-drive/pull/2663#issuecomment-1224092333

### DIFF
--- a/src/drive/web/modules/services/components/helpers.js
+++ b/src/drive/web/modules/services/components/helpers.js
@@ -1,13 +1,26 @@
 import { models } from 'cozy-client'
+import { getIconUrl } from './iconContext'
 
 export const TYPE_DIRECTORY = 'directory'
 
-export const makeNormalizedFile = async (client, folders, file, getIconUrl) => {
+/**
+ * Normalize file for Front usage in <AutoSuggestion> component inside <SearchBar>
+ *
+ * To reduce API call, the fetching of Note URL has been delayed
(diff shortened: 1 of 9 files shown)


src/drive/web/modules/services/components/helpers.js
```diff
@@ -1,21 +1,53 @@
...
+/**
+ * Is file param a correct note
+ *
+ * Inspired by https://github.com/cozy/cozy-client/blob/b7ffdd97d017bbd52226e0b61099da3090d8253f/packages/cozy-client/src/models/file.js#L52
+ * but don't check metadata (not provided while fetching Suggestions)
+ *
+ * @param {IOCozyFile} file io.cozy.files
+ * @returns {boolean}
+ */
+export const isNote = file => {
```
trollepierre, line 13: to be removed in favor of https://github.com/cozy/cozy-client/pull/1230

Crash--:
For this PR we don't need to change the behavior of the isNote method since we don't change the way we fetch the data ;)

Ldoppea (approved):
The code part looks good to me. However I'm not confident enough to review the logic part and all implications, so consider my approval only for the code structure part

Crash-- (approved):

PR #2679, opened 2022-08-29
feat(search): fetch only required params for search

trollepierre:
- [ ] Changelog updated if needed

In local, benchmark:

with 11 432 files with 10000 in my trashes. (there are 1059 remaining files in my cozy)

Without optims: 120ms - 9.66Mo
Capture d’écran 2022-08-29 à 15 37 14

With optimisation:  2s - 2.46Mo
Capture d’écran 2022-08-29 à 15 38 51

### DIFF
--- a/src/drive/web/modules/services/components/SuggestionProvider.jsx
+++ b/src/drive/web/modules/services/components/SuggestionProvider.jsx
@@ -81,12 +81,7 @@ class SuggestionProvider extends React.Component {
       'GET',
       '/data/io.cozy.files/_all_docs?Fields=_id,trashed,dir_id,name,path,type,mime,metadata.title,metadata.version&DesignDocs=false'
     )
-    const files = resp.rows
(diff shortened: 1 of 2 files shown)


Crash--:
Can we have the global time for the user: When I write the word I'm looking for until the results are displayed? 

Having this time for: 
- current status (so with the note fix)
- this PR 
- the previous implementation based on mango query 

Also, can you test on a more realistic cozy, aka : few files in the trash (let's say 100) for 10k docs in total?

trollepierre:
Master:  (all_docs)
master

PR 2679: (all_docs Fields=)
PR 2679

PR 2663: (mango / findAll)
PR 2663

Crash-- (approved):
I would have preferred numbers, but let's go 👍

Crash--:
I think this PR needs https://github.com/cozy/cozy-client/pull/1230 to work correctly.

PR #2693, opened 2022-09-08
fix: Make Search backward compatible

Crash--:
Like that, we don't have a dep between the stack
deployed and the drive version.

If the stack handles the Fields attribute then
fetch is optimized, if not, then we still got
the docs.

### DIFF
--- a/src/drive/web/modules/services/components/SuggestionProvider.jsx
+++ b/src/drive/web/modules/services/components/SuggestionProvider.jsx
@@ -81,7 +81,7 @@ class SuggestionProvider extends React.Component {
     const { client } = this.props
     const resp = await cozy.client.fetchJSON(
       'GET',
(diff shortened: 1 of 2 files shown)


src/drive/web/modules/services/components/SuggestionProvider.jsx
```diff
@@ -81,7 +81,7 @@ class SuggestionProvider extends React.Component {
     const { client } = this.props
     const resp = await cozy.client.fetchJSON(
       'GET',
-      '/data/io.cozy.files/_all_docs?Fields=_id,trashed,dir_id,name,path,type,mime,metadata.title,metadata.version&DesignDocs=false'
+      '/data/io.cozy.files/_all_docs?Fields=_id,trashed,dir_id,name,path,type,mime,metadata.title,metadata.version&DesignDocs=false&DesignDocs=false&include_docs=true'
```
JF-Cozy, line 84: attention y'a 2 fois `&DesignDocs=false` non ?

JF-Cozy (approved):

PR #2698, opened 2022-09-16
Duplicate file

trollepierre:
- [ ] Changelog updated if needed


src/drive/web/modules/drive/duplicate.js
```diff
@@ -0,0 +1,457 @@
...
+    const doc = {
+      ...file,
+      name: newName,
+      attributes: { ...file.attributes, name: newName, created_at: undefined },
+      relationships: undefined,
+      created_at: undefined,
+      id: undefined,
+      _id: undefined
+    }
+    const response = await client.create(file._type, doc, reference)
```
trollepierre, line 49: to create duplicated file, is it the right method?

PR #2700, opened 2022-09-20
Use new DACC API

paultranvan:
- Use new DACC remote-doctype
- Use new groups API

### DIFF
--- a/src/drive/targets/manifest.webapp
+++ b/src/drive/targets/manifest.webapp
@@ -176,12 +176,12 @@
       "verbs": ["GET"]
     },
     "dacc": {
-      "type": "cc.cozycloud.dacc",
+      "type": "cc.cozycloud.dacc_v2",
       "verbs": ["POST"],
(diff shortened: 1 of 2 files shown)


src/drive/targets/manifest.webapp
```diff
@@ -101,7 +101,7 @@
     "dacc": {
       "type": "node",
       "file": "services/dacc/drive.js",
-      "trigger": "@every 720h"
+      "trigger": "@monthly on the 1-3 before 5am"
```
trollepierre, line 104: 1-3 is not clear for me

Let's try, something like this?

```suggestion
      "trigger": "@monthly on the 1st to 3rd of the month before 5am"
```

paultranvan: I don't know if it's clearer, but it definitely won't work ;) https://github.com/cozy/cozy-stack/blob/master/docs/jobs.md#monthly-syntax

trollepierre: oops, big confusion. No worry. I resolve

trollepierre (approved):

PR #2721, opened 2022-10-05
chore: Make version:drive:androidbuild POSIX compliant

sblaisot:
Make android build (i.e. `yarn version:drive:androidbuild`) POSIX compliant so it doesn't requires explicitely instructing node to use bash or wrongly assume `sh` shell is `bash`.


JF-Cozy (approved):

Crash--:
After a try: I got a generated `src/drive/targets/mobile/config.xml-E` file

PR #2725, opened 2022-10-21
chore: Webpack build css in public folder too [skip-ci]

Merkur39:
We need to change the configuration
of the `MiniCssExtractPlugin` plugin,
so that it copies the app's style file to the `public` folder.

- [x] Override base config file of webpack
- [ ] The CSS file should only be copied to the Public folder


Merkur39:
Je ferme cette PR, voir celle-ci plus pertinente! https://github.com/cozy/create-cozy-app/pull/1477

PR #2727, opened 2022-10-25
feat: Add duplicate file button

zatteo:
### ✨ Features

* Users can now copy a file

### 🐛 Bug Fixes

*

### 🔧 Tech

* The bundlemon size limit and percentage size increase limit of _services_ files has been increased


src/drive/web/modules/actions/index.jsx
```diff
@@ -229,6 +230,32 @@ export const move = ({ canMove, pushModal, popModal }) => {
...
 }
 
+export const copy = ({ refresh, client, hasWriteAccess }) => {
+  return {
+    icon: 'copy',
+    displayCondition: selection =>
+      selection.length === 1 && isFile(selection[0]) && hasWriteAccess,
+    action: async files => {
+      await client.collection('io.cozy.files').copy(files[0].id)
+      refresh()
```
Crash--, line 240: I'm not sure that we want this refresh call? 

If I'm not mistaken this function comes from cozy-sharings and its goal is to refresh all the sharings by making a few http requests. 

When copying a new file, I don't think we need this call. Don't we?

Crash--: Or is is needed for the public view? 

If all those assumptions are true, maybe we should add a new `actionsOptions` : `isPublic` and then only call the method if it's public?

Crash--: (in production, we have a few cozy with a lot of sharings, so we should avoid to make too much requests)

zatteo: Exactly, it is used when the view is public. I make the change.

zatteo: Done!

Crash--: For 3/ I've created https://github.com/cozy/cozy-libs/pull/1837 but since we need to rewrite the tests... If you want to take it, feel free ;). 

Let's only call refresh() on public view. 

For 2/ OK, we didn't find the answer, let's move on, maybe one day we'll have it ;)

zatteo: Great, I updated it.

For the 2/, the answer is in the 4/. It is called in the `DeleteConfirm` component directly.

.bundlemonrc
```diff
@@ -137,7 +137,7 @@
     },
     {
       "path": "drive/services/**",
-      "maxPercentIncrease": 0.1
+      "maxPercentIncrease": 0.5
```
Crash--, line 140: I would have preferred a separated commit for this.

Crash-- (approved):

Crash-- (approved):

PR #2731, opened 2022-10-27
Update docs and add jsconfig.json to assist navigation

zatteo:
- Add changelog PR template
- Add jsconfig file for assist navigation in editor
- Add again mails in dev docs


Crash-- (approved):
<3

PR #2733, opened 2022-10-29
fix: Remove invalid EditDocumentQualification prop

Peltoche:
The current props for EditDocumentQualification are:
```
{
  document: PropTypes.object.isRequired,
  documents: PropTypes.arrayOf(PropTypes.object),
  onClose: PropTypes.func.isRequired,
  t: PropTypes.func.isRequired,
  client: PropTypes.object,
  onDescribed: PropTypes.func,
  isOffline: PropTypes.bool.isRequired
}
```
The onQualified method is not listed and never calls by the component. I think we can safely remove this method as the same action is done the line below with the `onClose` prop which is valid.


Peltoche:
Wrong diff :facepalm:

PR #2734, opened 2022-10-30
[WIP] Refactor the web drive actions

Peltoche:
The drive actions are used in two place: 
- in the selection bar
- in the action menu on the right side of a file line

Those two component are described by the same action object, contains the same informations but have two completely different API. The goal of this PR is to try to bring those two together in order to reduce code duplication and to ensure the same content for the components.

### DIFF
--- a/src/drive/web/modules/actions/index.jsx
+++ b/src/drive/web/modules/actions/index.jsx
@@ -45,6 +45,7 @@ import { useI18n } from 'cozy-ui/transpiled/react'
 
 export const share = ({ hasWriteAccess, pushModal, popModal }) => {
   return {
+    name: 'share',
     icon: ShareIcon,
     displayCondition: selection =>
       hasWriteAccess &&
@@ -68,6 +69,7 @@ export const share = ({ hasWriteAccess, pushModal, popModal }) => {
 export const download = ({ client, vaultClient }) => {
   return isMobileApp()
     ? {
+        name: 'download',
         icon: isIOSApp() ? ShareIosIcon : ReplyIcon,
         label: 'forwardTo',
         displayCondition: files => {
@@ -91,6 +93,7 @@ export const download = ({ client, vaultClient }) => {
         }
       }
     : {
+        name: 'download',
         icon: DownloadIcon,
         label: 'download',
         displayCondition: files => {
@@ -128,6 +131,7 @@ export const hr = () => {
 
 export const trash = ({ pushModal, popModal, hasWriteAccess, refresh }) => {
   return {
+    name: 'trash',
     icon: TrashIcon,
     label: 'trash',
     displayCondition: () => hasWriteAccess,
@@ -157,6 +161,7 @@ export const trash = ({ pushModal, popModal, hasWriteAccess, refresh }) => {
 }
 export const open = ({ client, vaultClient }) => {
   return {
+    name: 'openWith',
     icon: isIOSApp() ? EyeIcon : LinkOutIcon,
     label: isIOSApp() ? 'applePreview' : 'openWith',
     displayCondition: selection =>
@@ -178,6 +183,7 @@ export const open = ({ client, vaultClient }) => {
 
 export const rename = ({ hasWriteAccess, dispatch }) => {
   return {
+    name: 'rename',
     icon: RenameIcon,
     label: 'rename',
     displayCondition: selection => hasWriteAccess && selection.length === 1,
@@ -198,6 +204,7 @@ export const rename = ({ hasWriteAccess, dispatch }) => {
 
 export const move = ({ canMove, pushModal, popModal }) => {
   return {
+    name: 'moveto',
     icon: MovetoIcon,
     label: 'moveto',
     displayCondition: () => canMove,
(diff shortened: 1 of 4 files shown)


Crash--:
In fact... `onQualified` has been renamed to `onDescribed` during the export of this module from Drive. It seems that we (I..!) forgot to rename the props here 🤦 

I think that the right fix should not be the removal, but the renaming of the prop.

Crash--:
Also, there is lint issue with your PR, do not hesitate to run `yarn eslint '{config,src,examples,test}/**/*.{js,jsx}' --fix`

Peltoche:
The `onDescribe` prop is called juste before the `onClose` (https://github.com/cozy/cozy-libs/blob/master/packages/cozy-scanner/src/EditDocumentQualification.jsx#L78). As the `onDescribe` is only used to call close I think we should remove it in favor of the `onClose` method. It would avoid to call the method twice.

src/drive/web/modules/actions/index.jsx
```diff
@@ -80,9 +80,7 @@ export const download = ({ client, vaultClient }) => {
           const { t } = useI18n()
           return (
             <ActionMenuItem
-              onClick={() => {
-                return exportFilesNative(client, props.files, { vaultClient })
-              }}
+              onClick={props.onClick}
```
Crash--, line 87: Where does this onClick props come from?

src/drive/web/modules/actions/index.jsx
```diff
@@ -75,21 +76,21 @@ export const download = ({ client, vaultClient }) => {
...
-        label: 'forwardTo',
         Component: function Download(props) {
           const { t } = useI18n()
           return (
             <ActionMenuItem
               onClick={props.onClick}
               left={<Icon icon={isIOSApp() ? ShareIosIcon : ReplyIcon} />}
             >
-              {t('SelectionBar.forwardTo')}
+              {t('SelectionBar.' + props.label)}
```
Crash--, line 90: We try to respect the grep test https://jamie-wong.com/2013/07/12/grep-test/

This is more easier to respect it in order to be able to remove unused translation string and so on.

Crash--: So maybe having: 
```
label: t('SelectionBar.' + props.label)
```
?

src/drive/web/modules/actions/index.jsx
Crash--: Does it come from https://github.com/cozy/cozy-drive/blob/a731531b6d2077fd82e5edb3b751bd41530bb120/src/drive/web/modules/actionmenu/ActionsItems.jsx#L79 ? 

ActionsItems are used for the selection bar but also for the "..." ?

src/drive/web/modules/actions/index.jsx
```diff
@@ -81,15 +83,15 @@ export const download = ({ client, vaultClient }) => {
           return (
             <ActionMenuItem
               onClick={props.onClick}
-              left={<Icon icon={isIOSApp() ? ShareIosIcon : ReplyIcon} />}
+              left={<Icon icon={props.icon} />}
```
Crash--, line 88: Without changing the caller, we change the result here, no?

src/drive/web/modules/actions/index.jsx
Peltoche: Erf, I was looking for a line of code inside `SelectionBar`  showing that this component use the `props.label` the same way and I was forced the comply with it but I just found out that I have missread the code...

In fact it appeared that the SelectionBar "label" use `t('SelectionBar.' + actionName)`. This means that the `label` field have been never used before.

In any case I was going to say that I'm agree with you and I would like to change the way the label is generated inside the `SelectionBar` in order to make the two coherent. So let's go for you proposition and maybe in a second step we could change [this line](https://github.com/cozy/cozy-ui/blob/master/react/SelectionBar/index.jsx#L126) to use directly the `label` field. That way we have several benefices:
- We ensure that we have the same wording on both components
- We use `I18n` only once
- We make the label explicite
- We make this code grep test compliante.

Peltoche: If we don't want to deduplicate the label between `Component` and `SelectionBar` we can remove the label completly.

Peltoche: ``` React Hook "useI18n" is called in function "move" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"      react-hooks/rules-of-hooks```

I guess I will need to use something like:
```
label: 'SelectionBar.forwardTo', 
```
and translate inside the component

src/drive/web/modules/actions/index.jsx
Peltoche: > Does it come from ... ?

Yep that's it:
```js
    const onClick = action
      ? () => {
          action([file])
          onClose()
        }
      : null```
https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/actionmenu/ActionsItems.jsx#L62.

You can check the `onClick` I remove is the exact duplicate of the `action` fields setup in the same object. So instead of duplicating you can use this `onClick` function which was already given in props.

src/drive/web/modules/actions/index.jsx
Peltoche: I have only moved the logique around `isIOSApp` inside the `actions[...].icon` field so this component will have the exact same behavior. The main difference is that the icon inside the `SelectionBar` will now have the same behaviour.

PR #2735, opened 2022-10-31
fix: Shared OO document displayed a blank page due to infinite loop

zatteo:
The presence of a sharecode was used to determine if the document should be opened on other instance, but it was not sufficient. It only occured now because I updated recently the condition.

```
### ✨ Features

*

### 🐛 Bug Fixes

* Shared OO document displayed a blank page due to infinite loop

### 🔧 Tech

*
```


Crash-- (approved):
This is nit, but I prefer passing the all `client` object instead of just the string, but it works, so let's test that ;)

PR #2736, opened 2022-10-31
feat: favicon change based on file type

cballevre:
```
### ✨ Features

* Favicon change based on file type opened

### 🐛 Bug Fixes

* Update title after closing a file viewer

### 🔧 Tech

* Allow 0.4 max increase for bundle

```


src/drive/web/modules/views/OnlyOffice/Toolbar/index.jsx
```diff
@@ -16,12 +16,16 @@ import FileName from 'drive/web/modules/views/OnlyOffice/Toolbar/FileName'
 import ReadOnly from 'drive/web/modules/views/OnlyOffice/Toolbar/ReadOnly'
 import Sharing from 'drive/web/modules/views/OnlyOffice/Toolbar/Sharing'
 import useUpdateDocumentTitle from 'drive/web/modules/views/useUpdateDocumentTitle'
+import useUpdateDocumentFavicon from '../../useUpdateDocumentFavicon'
```
zatteo, line 19: Why relative `../..` path ?

cballevre:
[📁 DRIVE - Avoir une favicon en fonction du fichier ouvert](https://trello.com/c/JE7tDECr/1985-%F0%9F%93%81-drive-avoir-une-favicon-en-fonction-du-fichier-ouvert)

src/drive/web/modules/views/OnlyOffice/Toolbar/index.jsx
cballevre: I need to fix my auto import in VS Code. I'll correct that :)

src/drive/web/modules/views/useUpdateDocumentFavicon.js
```diff
@@ -0,0 +1,41 @@
...
+]
+
+const defaultFaviconHref = '/public/app-icon.svg'
+
+const makeFaviconHref = type => {
+  return getIconUrlFromKey(acceptedTypes.includes(type) ? type : 'files')
+}
+
+const useUpdateDocumentFavicon = docId => {
+  const { data: fileWithPath, fetchStatus } = useFileWithPath(docId)
```
Crash--, line 26: Do we really need the path? Can’t a simple useQuery should be enough?

Crash--: I mean: 
- Do we really need to get the path in order to change the favico? If not, then there is no need to request the parent, right? https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/views/hooks.js#L57-L62 

Also, I know that I said that each component should be standalone, so having inside this `useUpdateDocumentFavicon` something like: 

```jsx
const useUpdateDocumentFavicon = docId => {
const fileQuery = buildFileByIdQuery(docId)
const fileResult = useQuery(fileQuery.definition, fileQuery.options)
const resultData = fileResult.data
```

Will work, but I'll not be happy with that, because we'll make a request to get data for the file even if we already have those data inside the cozy-client's store. 

Why?
Because we list the content of the folder before clicking on the file. So by doing a request on `/files/dirId` we fetch the content of the folder, so we got the data for each files inside this folder. Then, if we make a request to `files/fileId` CozyClient will check if there is already this request inside its cache, but it'll respond no, so CC will make the http request. 

To avoid that we can do something like: 
- call getDocumentFromState from cozyclient (https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1318) 
- if the result is there then it's done, you have the data 
- If there is nothing, then you have to make the query (It should be the case if you go to the fileId url directly without listing the folder content) 

Banks have added an hook called `useDocument` that do something very similar https://github.com/cozy/cozy-banks/blob/master/src/components/useDocument.js . Maybe we should add it to cozy-client and then use it in Drive. What do you think about that?

src/drive/web/modules/services/components/iconContext.js
```diff
@@ -27,8 +27,12 @@ export function getIconUrl(file) {
       ? 'note'
       : getFileMimetype(icons)(file.mime, file.name) || 'files'
   }
-  const icon = icons[keyIcon].default
 
+  return getIconUrlFromKey(keyIcon)
+}
+
+export function getIconUrlFromKey(keyIcon) {
+  const icon = icons[keyIcon].default
```
Crash--, line 31: I think you can revert this edit.

src/drive/web/modules/views/OnlyOffice/Toolbar/index.jsx
Crash--: If you've fixed that later, you should squash the fix commit with this one.

src/drive/web/modules/views/useUpdateDocumentFavicon.js
Crash--: can you do something about your commits? 🙏

Crash-- (approved):

PR #2737, opened 2022-10-31
feat: Adapt OO header in responsive for iOS and Android

zatteo:
The size of the OO header change between iOS and Android, so we need to adapt the size of our header for iOS and Android.

```
### ✨ Features

*

### 🐛 Bug Fixes

* OnlyOffice header is correctly displayed on smartphone.

### 🔧 Tech

*
```


zatteo:
I updated the PR to change the default header size to the Android one and not the iOS one.

Crash-- (approved):

PR #2739, opened 2022-11-03
fix: Delete warning only when referenced by manual album

zatteo:
Previously, when you wanted to delete a photo, a delete warning was displayed every time the photo was referenced by an album. Even if the album was a cluster album which does not make sense.

```
### ✨ Features

*

### 🐛 Bug Fixes

* Show delete photo from album warning only when necessary

### 🔧 Tech

*
```


Crash-- (approved):
Solid! 👏

PR #2746, opened 2022-11-07
fix(useUpdateDocumentFavicon): Show the correct favicon for Only Office spreadsheet

cballevre:
The class attribute of Only Office spreadsheet document is a spreadsheet. The icon type is only named «sheet». To match the two, I had a normalisation function to spread with future exception.

```
### 🐛 Bug Fixes

* Show the correct favicon for Only Office spreadsheet
```

### DIFF
--- a/src/drive/web/modules/views/useUpdateDocumentFavicon.jsx
+++ b/src/drive/web/modules/views/useUpdateDocumentFavicon.jsx
@@ -19,7 +19,15 @@ const acceptedTypes = [
 
 const defaultFaviconHref = '/public/app-icon.svg'
 
-const makeFaviconHref = type => {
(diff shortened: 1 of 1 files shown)


src/drive/web/modules/views/useUpdateDocumentFavicon.jsx
```diff
@@ -19,7 +19,16 @@ const acceptedTypes = [
 
 const defaultFaviconHref = '/public/app-icon.svg'
 
-const makeFaviconHref = type => {
+const normalizeClass = docClass => {
+  console.log(docClass)
```
Crash--, line 23: to remove

Crash-- (approved):

PR #2758, opened 2022-11-17
fix: Open directly an OO file from the searchbar

zatteo:
URL sent to cozy-bar is now the direct OO editor URL and not the preview URL

```
### 🐛 Bug Fixes

* Open directly an OO file from the searchbar
```


Crash-- (approved):
Nice! (just a test to update)

PR #2764, opened 2022-11-24
fix: Align buttons when dezooming

zatteo:
```
### 🐛 Bug Fixes

* Align buttons when dezooming
```


Crash--:
Still works on mobile and desktop resolution?

zatteo:
> Still works on mobile and desktop resolution?

Yes!

Crash-- (approved):

PR #2768, opened 2022-11-28
fix: Set correct returnUrl for note creation from Flagship app

Ldoppea:
When on a Flagship app, the navigation should not rely on Universal links but on cozy-intent's `openApp` method

So we want to send the cozy-drive's web url as `returnUrl` instead of an universal link. Then on cozy-notes, the AppLinker component will use this url to call the `openApp` method

```
### ✨ Features

*

### 🐛 Bug Fixes

* When creating and opening a cozy-note from the Flagship app, the back button now correctly go back to cozy-drive

### 🔧 Tech

*
```


Crash-- (approved):
Don't like the current fix because webapp should not handle that but it'll do the work ATM.

PR #2770, opened 2022-11-29
fix: Upgrade cozy-ui to fix Viewer BottomSheet Size

Crash--:
Need https://github.com/cozy/cozy-ui/commit/884f89d89126e6c04f5b3b13c19556df368e47f2 to fix the bottomSheet size when there is a small content.

```

### 🐛 Bug Fixes

* Bottomsheet of the viewer should take less place when there is not so much content

```


Crash--:
See https://github.com/cozy/cozy-ui/pull/2292

cballevre (approved):

PR #2799, opened 2022-12-23
docs: Add info about transifex config file version

zatteo:


README.md
```diff
@@ -177,6 +177,8 @@ As a _translator_, you can login to [Transifex][tx-signin] (using your Github ac
 
 As a _developer_, you must [configure the transifex client][tx-client], and claim access as _maintainer_ to the [app repository][tx-app]. Then please **only update** the source locale file (usually `en.json` in client and/or server parts), and push it to Transifex repository using the `tx push -s` command.
 
+Be careful, the transifex configuration file is still in an old version. See [https://github.com/transifex/cli#migrating-from-older-versions-of-the-client](https://github.com/transifex/cli#migrating-from-older-versions-of-the-client).
```
paultranvan, line 180: I assume that the simpler method is probably to use an old client version?

Ldoppea: Can we have more precision than "Be careful" here? I'm not sure if the suggested action is to do the migration as described in the linked transifex documentation or if the action is to ensure to use the old client instead.

Crash--: ping @zatteo

zatteo: I updated the message and added a link to the previous client. If we want to update to next CLI one day : 
- we need to update every configuration file with `tx migrate` from new CLI
- we need to update cozy-ci https://github.com/cozy/cozy-libs/blob/master/packages/cozy-ci/transifex.sh

Next CLI is harder, better, faster and stronger (of course) but on another side it is just the translation CLI which is currently working well.

zatteo:
```
 Important Announcement! The deprecation period of API 2.0/2.5 and old Python Transifex CLI came to an end. Calls to these APIs and any scripts that depend on them may not work after Nov 30, 2022. We highly recommend you migrate to newer versions if you observe any issues. 
```

PR #2807, opened 2022-12-30
Fix : click on suggestion path

cballevre:
```
### 🐛 Bug Fixes
* Click on parent path into search bar suggestions 
```

### DIFF
--- a/src/drive/web/modules/search/components/SuggestionItemTextSecondary.jsx
+++ b/src/drive/web/modules/search/components/SuggestionItemTextSecondary.jsx
@@ -1,0 +1,35 @@
+import React from 'react'
+
+import styles from 'drive/web/modules/search/components/styles.styl'
+import SuggestionItemTextHighlighted from 'drive/web/modules/search/components/SuggestionItemTextHighlighted'
+
+const SuggestionItemTextSecondary = ({
+  text,
+  query,
+  url,
+  onOpened,
+  isMobile
+}) => {
+  const handleClick = e => {
+    e.stopPropagation()
+    if (typeof onOpened == 'function') {
+      onOpened(e)
+    }
+  }
+
+  if (isMobile) {
+    return <SuggestionItemTextHighlighted text={text} query={query} />
+  }
+
+  return (
+    <a
+      className={styles['suggestion-item-parent-link']}
+      href={url}
+      onClick={handleClick}
+    >
+      <SuggestionItemTextHighlighted text={text} query={query} />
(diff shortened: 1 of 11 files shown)


src/drive/web/modules/search/components/helpers.js
```diff
@@ -33,13 +33,18 @@ export const makeNormalizedFile = (client, folders, file) => {
...
   if (isDir) {
     path = file.path
     url = urlToFolder
+    parentUrl = urlToFolder
   } else {
     const parentDir = folders.find(folder => folder._id === file.dir_id)
     path = parentDir && parentDir.path ? parentDir.path : ''
+    parentUrl =
+      parentDir && parentDir._id
+        ? `${window.location.origin}/#/folder/${parentDir._id}`
```
zatteo, line 46: @Crash-- Can we use window.location in front application ? I do not remember if we should always use URL from client or if we can use this.

Crash--: We should avoid using window.location for a lot of stuff. 

If you need to create a link to an app, you should use the `AppLinker` component from cozy-ui. 

If you need to get the `location` for an app, you should rely on cozyclient and generateWebLink for the slug.  

There is one case that it ATM difficult to handle: if you need to make a redirection without a user interaction. 

In that case, you need to test if you are in the flagship app or not and call the right method by yourself. ex: https://github.com/cozy/cozy-store/blob/6b9549a6aaac8cc388ad015bcb0034768e384230/src/ducks/apps/components/ApplicationPage/Header.jsx#L40-L56

(cc @Ldoppea)

src/drive/web/modules/search/components/SuggestionItemTextSecondary.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  if (isMobile) {
+    return <SuggestionItemTextHighlighted text={text} query={query} />
+  }
+
+  const app = {
+    slug: 'drive'
+  }
+
+  const { subdomain: subDomainType } = client.getInstanceOptions()
```
cballevre, line 30: @Crash-- is it a correct way to access subdomain or I should prefer the code below ?
```
const { capabilities } = useCapabilities(client)
  const isFlatDomain = useMemo(() => {
    return get(capabilities, 'flat_subdomains')
  }, [capabilities])
```

Crash-- (approved):

zatteo (approved):

PR #2809, opened 2023-01-02
fix: Translation for duplicate in selection bar

cballevre:
```
### 🐛 Bug Fixes

* Translation for duplicate in selection bar
```


Crash-- (approved):

src/drive/web/modules/actions/index.jsx
```diff
@@ -226,16 +226,17 @@ export const move = ({ canMove, pushModal, popModal }) => {
   }
 }
 
-export const copy = ({ client, hasWriteAccess, refresh, isPublic }) => {
+export const duplicate = ({ client, hasWriteAccess, refresh, isPublic }) => {
   return {
-    icon: 'copy',
+    label: 'duplicate',
```
zatteo, line 231: Okk it works because you added label at the first place ? Still tricky but at least it works.

Ldoppea:
It took me a bit of time to understand what is happening here, as your PR title is about translation but no file contains any translation related changes.

When doing those kind of PR, don't forget to link the PR you are fixing 😉 : https://github.com/cozy/cozy-drive/commit/077747dfdf12ce1e5eb05c57b9ccf9a307177fdf

Also your PR is changing the icon, you should document that in the PR's description (or better in the commit's one)

src/drive/web/modules/actions/index.jsx
cballevre: @zatteo I made some change to add systematically name to be more explicit see : https://github.com/cozy/cozy-drive/pull/2809/commits/6d66c2c56b6ed33920ac6ce2ef852f01ed74af5b

zatteo (approved):

src/drive/web/modules/actions/index.jsx
zatteo: Nice! Next developer will understand this part faster than we did 😄

PR #2815, opened 2023-01-06
Fix: Use cross circle outline icon

cballevre:
```
### 🐛 Bug Fixes

* Use cross circle outline icon

### 🔧 Tech

* Update cozy-ui from 77.9.2 to 79.4.0
```


zatteo (approved):

Crash--:
@cballevre As the bundle increases quite a bit, can you create a separate PR : one for UI one for cozy-client? In order to see which one is responsible for this increment

PR #2822, opened 2023-01-09
fix: File name took not all width in mobile file list

zatteo:
It was due to action button taking too much width in mobile file list.

```
### 🐛 Bug Fixes

* File name took not all width in mobile file list
```

### DIFF
--- a/src/drive/styles/filelist.styl
+++ b/src/drive/styles/filelist.styl
@@ -379,6 +379,12 @@ column-width-thumbnail-bigger = 7rem
         padding-right .2rem
         margin-right 1em
 
+    .fil-content-header-action
+    .fil-content-file-action
+        flex 0
+        box-sizing content-box
+        width 100%
+
     .fil-content-cell.fil-content-file
     .fil-content-row-bigger .fil-content-cell.fil-content-file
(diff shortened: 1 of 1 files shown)


src/drive/styles/filelist.styl
```diff
@@ -372,10 +372,14 @@ column-width-thumbnail-bigger = 7rem
         flex 1 1 auto
 
     .fil-content-header-action
+        flex 0
+        box-sizing content-box
```
cballevre, line 376: You can add `width 100%` so container and icon take the same place because there is an `width 1.5rem` set by default for both header and file.

src/drive/styles/filelist.styl
```diff
@@ -372,10 +372,14 @@ column-width-thumbnail-bigger = 7rem
...
         padding 0
         margin-right 1em
 
     .fil-content-file-action
+        flex 0
+        box-sizing content-box
```
cballevre, line 382: Same has above. Nit : Create a selector for both 
```
.fil-content-header-action, .fil-content-file-action
        flex 0
        box-sizing content-box
        width 100%
```

cballevre (approved):

PR #2834, opened 2023-01-23
fix(Search): Use shortcut icon instead IconFiles

cballevre:
```
### 🐛 Bug Fixes

*  Use shortcut icon instead IconFiles in search
```


Crash-- (approved):
Next time, maybe you can add a few tests to avoid the regression ;)

PR #2858, opened 2023-02-22
feat: Harmonize OnlyOffice FAB to primary color

cballevre:
```
### 🐛 Bug Fixes

*  Fab is now the same color on mobile & desktop
```


Crash-- (approved):
This is clearly not a "tech" stuff since it has impact on the user. But this is not really a feat neither. Let's make that as a "fix: Fab is now the same color on mobile & desktop" in the changelog.

PR #2872, opened 2023-03-09
feat: Add OnlyOffice paywall

cballevre:
```
### ✨ Features

* Add OnlyOffice paywall
```

**Notes :** This PR comes with a new flag calls `drive.office` which replace `drive.onlyoffice.enabled` and extends it's features. You can set a default mode for the onlyoffice editor and give write permission. For example: 
```
{
  "enabled": true,
  "engine": "onlyOffice",
  "write": true,
  "onlyOffice": {
    "defaultMode": "edit"
  }
}
```

**Related PRs:**
* https://github.com/cozy/cozy-ui/pull/2366
* https://github.com/cozy/cozy-libs/pull/2058

### DIFF
--- a/src/drive/web/modules/drive/AddMenu/AddMenu.jsx
+++ b/src/drive/web/modules/drive/AddMenu/AddMenu.jsx
@@ -13,7 +13,7 @@ import CreateShortcut from 'drive/web/modules/drive/Toolbar/components/CreateSho
 import UploadItem from 'drive/web/modules/drive/Toolbar/components/UploadItem'
 import StartScanner from 'drive/web/modules/drive/Toolbar/components/StartScanner'
 import CreateOnlyOfficeItem from 'drive/web/modules/drive/Toolbar/components/CreateOnlyOfficeItem'
-import { isOnlyOfficeEnabled } from 'drive/web/modules/views/OnlyOffice/helpers'
+import { isOfficeEnabled } from 'drive/web/modules/views/OnlyOffice/helpers'
 import flag from 'cozy-flags'
 
 export const ActionMenuContent = ({
@@ -42,7 +42,7 @@ export const ActionMenuContent = ({
         <AddEncryptedFolderItem />
       )}
       {!isPublic && !isEncryptedFolder && <CreateNoteItem />}
-      {canUpload && isOnlyOfficeEnabled() && !isEncryptedFolder && (
+      {canUpload && isOfficeEnabled() && !isEncryptedFolder && (
         <>
           <CreateOnlyOfficeItem fileClass="text" />
           <CreateOnlyOfficeItem fileClass="spreadsheet" />
--- a/src/drive/web/modules/drive/Toolbar/components/CreateOnlyOfficeItem.jsx
+++ b/src/drive/web/modules/drive/Toolbar/components/CreateOnlyOfficeItem.jsx
@@ -7,7 +7,10 @@ import Icon from 'cozy-ui/transpiled/react/Icon'
 
 import { ROOT_DIR_ID } from 'drive/constants/config'
 import { useRouter } from 'drive/lib/RouterContext'
-import { makeOnlyOfficeIconByClass } from 'drive/web/modules/views/OnlyOffice/helpers'
+import {
+  makeOnlyOfficeIconByClass,
+  canWriteOfficeDocument
+} from 'drive/web/modules/views/OnlyOffice/helpers'
 
 const CreateOnlyOfficeItem = ({ fileClass }) => {
   const { t } = useI18n()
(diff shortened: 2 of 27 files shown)


src/drive/web/modules/drive/Toolbar/components/CreateOnlyOfficeItem.jsx
```diff
@@ -18,10 +21,13 @@ const CreateOnlyOfficeItem = ({ fileClass }) => {
...
-    () => router.push(`/onlyoffice/create/${folderId}/${fileClass}`),
-    [router, fileClass, folderId]
-  )
+  const handleClick = useCallback(() => {
+    if (canWriteOfficeDocument()) {
+      router.push(`/onlyoffice/create/${folderId}/${fileClass}`)
+    } else {
+      router.push(`/folder/${folderId}/paywall`)
+    }
+  }, [router, fileClass, folderId])
```
Crash--, line 30: canWriteOfficeDocument should be in the dep array, no?

src/drive/web/modules/drive/AddMenu/AddMenu.jsx
```diff
@@ -42,7 +42,7 @@ export const ActionMenuContent = ({
         <AddEncryptedFolderItem />
       )}
       {!isPublic && !isEncryptedFolder && <CreateNoteItem />}
-      {canUpload && isOnlyOfficeEnabled() && !isEncryptedFolder && (
+      {canUpload && isOfficeEnabled() && !isEncryptedFolder && (
```
Crash--, line 45: not sure about this renaming. Everywhere we still use OnlyOffice. The links that this condition display are related to OO and so on

src/drive/web/modules/views/OnlyOffice/helpers.js
```diff
@@ -3,7 +3,29 @@ import FileTypeSheetIcon from 'cozy-ui/transpiled/react/Icons/FileTypeSheet'
...
 import FileTypeTextIcon from 'cozy-ui/transpiled/react/Icons/FileTypeText'
 
-export const isOnlyOfficeEnabled = () => flag('drive.onlyoffice.enabled')
+export const isOfficeEnabled = () => {
+  const office = flag('drive.office')
+  if (office) {
+    return office.enabled
+  }
+  return false
+}
```
Crash--, line 12: you should keep the backward compatibility too 🙏

src/drive/web/modules/views/OnlyOffice/useCreateFile.jsx
```diff
@@ -9,7 +9,7 @@ import {
   makeMimeByClass
 } from 'drive/web/modules/views/OnlyOffice/helpers'
 
-const useCreateFile = (folderId, fileClass) => {
+const useCreateFile = (folderId, fileClass, { enable } = { enable: true }) => {
```
Crash--, line 12: Did you just add an arg to the hook because you can't make a condition into the component that call this hook? 

You should refactor / split the component, not adding complexity to the hook

src/drive/web/modules/drive/Toolbar/components/CreateOnlyOfficeItem.jsx
cballevre: I don't think so because canWriteOfficeDocument evaluated each time useEffect render

src/drive/web/modules/views/OnlyOffice/helpers.js
```diff
@@ -14,6 +14,22 @@ export const isOfficeEnabled = () => {
...
+export function canWriteOfficeDocument() {
+  const office = flag('drive.office')
+  if (office) {
+    return office.write
+  }
+  return false
+}
+
+export function redirectToOnlyOfficePaywall(nextState, replace) {
+  if (!canWriteOfficeDocument()) {
```
Crash--, line 26: I don't have clear mind about the flags and how we will handle the production's deployment of this new Drive version. 

We'll need to add this `drive.office` flag to a few contexts thats all? 

Don't we need to change the stack code in order to be able to use the new `drive.office.write` flag to return the right editorMode value (used in useConfig)

cballevre: As I have elaborated, we will need the `drive.office` flag in all contexts after this PR as well as `drive.office.write` properly set. We could modify the stack so that it already returns `drive.office` so that the contexts in the users would already be correct before updating the version of Drive with this PR.

src/drive/web/modules/views/OnlyOffice/useCreateFile.jsx
cballevre: Now, I'm using the `onEnter` property of the route to check permissions rather than conditionally calling the hook. If the user doesn't have write permissions, his redirect to root folder with the paywall

Crash-- (approved):

PR #2892, opened 2023-04-06
fix: Error on query result

Crash--:
isClientAlreadyInstalled is trigger several times during the fetchPolicy when switching for instance from recent to trash view.

We know that we've an issue with client.query() not returning the correct data during the fetchPolicy
https://github.com/cozy/cozy-client/issues/931

We have to change it for fetchQueryAndGetFromState


cballevre:
https://github.com/cozy/cozy-client/issues/931

cballevre:
cozy/cozy-client#931

cballevre (approved):

PR #2893, opened 2023-04-07
fix: Spacing on empty share page

cballevre:
```
### 🐛 Bug Fixes

* Spacing on empty share page
```

Before : 

After :

### DIFF
--- a/src/drive/styles/main.styl
+++ b/src/drive/styles/main.styl
@@ -10,3 +10,10 @@ div:focus
     +small-screen()
         padding 0
 
+.center-layout
+    display flex
+    width 100%
+
+    +medium-screen()
+        min-height 100vh
+        padding 0 2rem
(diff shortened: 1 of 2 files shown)


src/components/Error/empty.styl
```diff
@@ -1,6 +1,15 @@
...
+
 +small-screen() // @stylint ignore
     .empty
         margin-top calc(50vh - 6rem)
         transform translateY(-50%)
+
+    .empty-share
+        height 100vh
+        padding 0 2rem
+        width fit-content
```
Crash--, line 15: Le composant Empty de UI ne gère pas ça tout seul ? Est-ce que ça ne devrait pas être le cas ? 

https://caniuse.com/?search=fit-content ne semble pas être bien supporté sur iOS 15, tu as pu tester ?

cballevre: The Empty component of `cozy-ui` is integrated in different layout types. After thinking about your question, I came to the conclusion that the problem was with Drive's public layout for errors. I applied a fix there rather than on ErrorShare which is also used in photos. `cozy-ui` would probably need a layout with content centered in for all error states, or empty states. To be discussed with more people to ensure that it benefits all apps

PR #2896, opened 2023-04-11
fix: Scroll to top

Crash--:
Since we are not able to restore the scroll correctly,
 and force the scroll to top every time we change the
 current folder. This is to avoid this kind of weird
 behavior:
 - If I go to a sub-folder, if this subfolder has a lot of data and I scrolled down until the bottom. If I go back, then my folder will also be scrolled down.

 This is an ugly hack, yeah.

```
### ✨ Features

*

### 🐛 Bug Fixes

* 

### 🔧 Tech

*
```


cballevre (commented):
The scroll top didn't work on mobile. From what I understood on tablet & mobile, the layout change and the `TableBody` is displayed completely. As it doesn't scroll anymore, the window does. We could make this code to adapt to the situation: 

```
if (isDesktop) {
  const scrollable = document.querySelectorAll(
    '[data-testid=fil-content-body]'
  )[0]
  if (scrollable) {
    scrollable.scroll({ top: 0 })
  }
} else {
  window.scroll({ top: 0 })
}
```

PR #2912, opened 2023-05-12
feat: Update cozy-ui from 83.3.0 to 84.2.0

cballevre:
```
### 🔧 Tech

* Update cozy-ui from 83.3.0 to 84.2.0

```
This update gets a fix on selection bar. 

The version 84 of cozy-ui have breaking changes on Menu and deprecated ActionMenu. I took advantage of having to replace the MenuItem by ActionMenuItem to update the design in MobileSortMenu. I also needed to replace Menu into photos.

Before : 

After :


Crash-- (approved):
So much little changes that I don't understand. But I trust your changes. Overall it looks good to me

PR #2917, opened 2023-05-31
fix: Blank screen when onlyoffice is disabled

cballevre:
```
### 🐛 Bug Fixes

* Redirect to the viewer when accessing OnlyOffice although it is deactivated
```

### DIFF
--- a/src/drive/targets/public/components/AppRouter.jsx
+++ b/src/drive/targets/public/components/AppRouter.jsx
@@ -28,7 +28,7 @@ const AppRouter = ({
   return (
     <Router history={hashHistory}>
       <Route component={PublicLayout}>
-        {isOfficeEnabled(isDesktop) && (
+        {isOfficeEnabled(isDesktop) ? (
           <>
             <Route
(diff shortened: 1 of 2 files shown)


src/drive/targets/public/components/AppRouter.jsx
```diff
@@ -70,6 +70,8 @@ const AppRouter = ({
               <Redirect from="/" to={`onlyoffice/${data.id}`} />
             )}
           </>
+        ) : (
+          <Redirect from="onlyoffice/*" to="/" />
```
Crash--, line 74: Maybe we should add a test?

Crash-- (approved):

src/drive/targets/public/components/AppRouter.jsx
cballevre: Yes of course, I've added tests to check that the right component has been called depending on the input parameters (route, file, onlyoffice activation). I mock the components to limit problems with the redux store

Crash-- (approved):

PR #2931, opened 2023-07-11
fix: Breadcrumb endless loop for not found folder

cballevre:
In case of an error (eg. Not found), the breadcrumb displayed the root path (eg. Trash)

```
### 🐛 Bug Fixes

*  Remove the endless loop when accessing a folder not found
```


Crash--:
Can't we add a test?

cballevre:
I haven't managed to set up tests to reproduce the endless loop. I have added a test to display the root path in the case of an error

Crash-- (approved):

PR #2941, opened 2023-07-24
Manage quota exceeded during backup

zatteo:
Display quota paywall when quota exceeded during backup

### DIFF
--- a/src/photos/ducks/backup/components/BackupError.jsx
+++ b/src/photos/ducks/backup/components/BackupError.jsx
@@ -1,9 +1,14 @@
 import React from 'react'
 
+import { QuotaPaywall } from 'cozy-ui/transpiled/react/Paywall'
 import Snackbar from 'cozy-ui/transpiled/react/Snackbar'
 import Alert from 'cozy-ui/transpiled/react/Alert'
 
 import { useBackupActions } from '../hooks/useBackupActions'
+
+const shouldDisplayQuotaPaywall = backupError => {
+  return backupError.statusCode === 413
+}
 
 export const BackupError = () => {
   const { backupError, setBackupError } = useBackupActions()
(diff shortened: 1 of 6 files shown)


src/photos/ducks/backup/index.jsx
```diff
@@ -33,7 +33,7 @@ const BackupPage = () => {
             <BackupInfo />
             {!isFlagshipApp() ? <InstallAppAlert /> : null}
             <BackupActions />
-            <BackupDescription />
+            {isMobile ? <BackupDescription /> : null}
```
cballevre, line 36: nit: You can also use `&&` to be less verbose
```
{isMobile && <BackupDescription />}
```

src/photos/ducks/backup/components/BackupError.jsx
```diff
@@ -1,10 +1,15 @@
...
 
+import { QuotaPaywall } from 'cozy-ui/transpiled/react/Paywall'
 import Snackbar from 'cozy-ui/transpiled/react/Snackbar'
 import Alert from 'cozy-ui/transpiled/react/Alert'
 
 import { useBackupActions } from '../hooks/useBackupActions'
 
+const shouldDisplayQuotaPaywall = backupError => {
+  return backupError.statusCode === 413
+}
```
cballevre, line 11: nit: Maybe extract into a helper file with other business logic app like `parseBackupError`

cballevre (approved):
Happy to see new usage of the `QuotaPaywall` 🙂

src/photos/ducks/backup/index.jsx
zatteo: Agree about less verbose but trying this way for the next weeks https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx (still not sure about the value but trying 😄)

zatteo:
> Happy to see new usage of the `QuotaPaywall` 🙂

A pleasure to use!

src/photos/ducks/backup/index.jsx
cballevre: Thanks for the article 🙂 I use this way also now to avoid having `0` or `NaN` popping randomly

PR #2946, opened 2023-07-25
fix(backup): Flags were initialized too late

zatteo:
We use flags directly in AppRouter to determine default router. It was too late when they were loaded before.


Crash-- (approved):

Crash--:
Tt least with the implemented solution, you don't block the React Work. I'm more happy with that, than awaiting at the top level.

Crash--:
> For example, how do we prove that upper components like SharingProvider, PushBannerProvider or BackupDataProvider will never access flags before them initialized in the AppRouter?

You're right, in that case, we should have this component, awaiting for the flag init to be on top of other component. But we should not block the render() call IMO.

PR #2949, opened 2023-07-26
refactor(backup): Create a component that wait flags

zatteo:
What do you think about it?

### DIFF
--- a/src/photos/components/AppRouter.jsx
+++ b/src/photos/components/AppRouter.jsx
@@ -1,8 +1,7 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import { createHashRouter, Navigate, RouterProvider } from 'react-router-dom'
 import { isFlagshipApp } from 'cozy-device-helper'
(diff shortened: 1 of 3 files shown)


cballevre:
This will undoubtedly make Time to Interactive longer. Do you have any idea how much initialization takes ? If it takes a long, I might want to put a spinner in the meantime or something else to warn the user that it's not just a blank page. 

Another thing: we could bring in the same component the initialization of flags to this component using the `initFlags` function instead of `Layout`

Ldoppea:
> This will undoubtedly make Time to Interactive longer. Do you have any idea how much initialization takes ? If it takes a long, I might want to put a spinner in the meantime or something else to warn the user that it's not just a blank page.

In web apps (and so in cozy-drive), flags are initialized from the DOM `data-cozy` values, so it should be nearly instant. However on OAuth clients, they would be initialized from an HTTP query, so it would take longer.

Ldoppea (approved):

cballevre (approved):

PR #2951, opened 2023-07-27
fix(backup): Add upload illus for backup and use it

zatteo:


src/photos/ducks/backup/components/BackupHeader.jsx
```diff
@@ -3,14 +3,14 @@ import React from 'react'
 import Typography from 'cozy-ui/transpiled/react/Typography'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
-import PhotoSync from 'photos/assets/illustrations/cozy-photo-sync.svg'
```
Crash--, line 6: `cozy-photo-sync.svg` is still used? Or can we remove it?

Crash-- (approved):

src/photos/ducks/backup/components/BackupHeader.jsx
zatteo: Still used for permissions modal

PR #2954, opened 2023-07-28
fix(Toolbar): Render for each properties

cballevre:
Since the migration to react-router v6, the component has not re-rendered every time, which created inconsistencies in the sharing modal

```
### 🐛 Bug Fixes

* The sharing modal display the correct folder on mobile
```


src/drive/web/modules/drive/Toolbar/index.jsx
```diff
@@ -131,32 +117,20 @@ class Toolbar extends Component {
...
-    displayedFolder,
-    folderId,
-    breakpoints,
-    showSelectionBar,
-    isSelectionBarVisible
-  }
 
   return !folderId ? (
-    <Toolbar {...toolbarProps} />
+    <Toolbar folderId={folderId} {...props} />
```
Crash--, line 122:
```
if(!folderId) then <Toolbar folderId={folderId}
```

I found this line confusing. Why passing the folderId props then?

PR #2967, opened 2023-08-18
feat: Update manifest to add acceptFromFlagship

acezard:
### ✨ Features

* Update manifest to add acceptFromFlagship
 
* Also add placeholder route that will handle mobile uploads later

### DIFF
--- a/src/drive/targets/manifest.webapp
+++ b/src/drive/targets/manifest.webapp
@@ -175,5 +175,12 @@
       "verbs": ["POST"],
       "description": "Remote-doctype required to send anonymized measures to the DACC shared among mycozy.eu's Cozy."
     }
+  },
+  "accept_from_flagship": true,
+  "accept_documents_from_flagship": {
+    "accepted_mime_types": ["*/*"],
+    "max_number_of_files": 1,
+    "max_size_per_file_in_MB": 10,
+    "route_to_upload": "/#/create?fromFlagshipUpload=true"
   }
 }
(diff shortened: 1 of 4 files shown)


src/drive/web/modules/views/Create/CreateFromFlagship.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
Crash--, line 1: should be more in `upload` than `Create`.

src/drive/targets/manifest.webapp
```diff
@@ -175,5 +175,12 @@
       "verbs": ["POST"],
       "description": "Remote-doctype required to send anonymized measures to the DACC shared among mycozy.eu's Cozy."
     }
+  },
+  "accept_from_flagship": true,
+  "accept_documents_from_flagship": {
+    "accepted_mime_types": ["application/octet-stream"],
```
Crash--, line 181: this will not be the only accepted mime. In fact, with Drive, you can go with : `[*]`.

acezard: Updated with "*/*" according to https://www.rfc-editor.org/rfc/rfc7231#section-5.3.2

src/drive/web/modules/views/Create/CreateFromFlagship.jsx
acezard: updated, will replace 
```javascript
      {__TARGET__ === 'mobile' && (
        <Route path="uploadfrommobile" element={<UploadFromMobile />} />
      )}
 ```
 in subsequent pull request

acezard:
closing this pull request as it is now redundant with https://github.com/cozy/cozy-drive/pull/2969

PR #2969, opened 2023-08-23
feat: Handle file upload from Flagship

acezard:
- [x] Upload one file
- [x] Upload multiple files
- [x] Handle upload progression without leaving the app with one file
- [x] Handle upload progression without leaving the app with multiple files
- [x] Handle upload progression after leaving and opening the app again with one file 
- [x] Handle upload progression after leaving and opening the app again with multiple files


src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+      size="large"
+      classes={{ paper: 'u-h-100' }}
+      title={
+        <>
+          <Header
+            entries={items}
+            title={t('ImportToDrive.title', { smart_count: items.length })}
+            subTitle={t('ImportToDrive.to')}
+          />
+          <Query
```
Crash--, line 140: Query component is deprecated. You should have refactor to useQuery(). I still thin that this kind of refactor to be up to date with the cozy latest standard should be done BEFORE converting to TS...

Crash--:
I have to delete the cache on travis to be able to build this branch. And locally I have to rm -rf node_modules/ ....

src/drive/targets/manifest.webapp
```diff
@@ -175,5 +175,12 @@
       "verbs": ["POST"],
       "description": "Remote-doctype required to send anonymized measures to the DACC shared among mycozy.eu's Cozy."
     }
+  },
+  "accept_from_flagship": true,
+  "accept_documents_from_flagship": {
+    "accepted_mime_types": ["*/*"],
+    "max_number_of_files": 1,
+    "max_size_per_file_in_MB": 10,
+    "route_to_upload": "/#/create?fromFlagshipUpload=true"
```
Crash--: @Ldoppea did we update the documentation of the manifest.webapp ? https://docs.cozy.io/en/cozy-stack/apps/#the-manifest

src/drive/web/modules/views/Upload/UploadFromFlagship.jsx
```diff
@@ -0,0 +1,13 @@
...
+export const UploadFromFlagship = () => {
+  const { loading } = useUploadFromFlagship()
+
+  return (
+    <div>
+      <h1>Upload From Flagship Page</h1>
+      <p>{loading ? 'loading' : 'Ready to call hasFilesToHandle'}</p>
+    </div>
+  )
+}
```
Crash--: is it still needed?

package.json
```diff
@@ -80,13 +80,15 @@
     "@testing-library/jest-dom": "5.16.4",
     "@testing-library/react": "11.2.7",
     "@testing-library/react-hooks": "8.0.1",
+    "@types/react-redux": "7.1.26",
     "@welldone-software/why-did-you-render": "^6.1.4",
-    "babel-core": "7.0.0-bridge.0",
+    "babel-preset-cozy-app": "2.1.0",
     "babel-runtime": "^6.26.0",
```
Crash--, line 86: we should certainly remove `"babel-runtime": "^6.26.0",

src/components/App/App.jsx
```diff
@@ -13,7 +13,15 @@ import cozyBar from 'lib/cozyBar'
 
 const App = ({ store, client, lang, polyglot, children }) => {
   return (
-    <WebviewIntentProvider setBarContext={cozyBar.setWebviewContext}>
+    <WebviewIntentProvider
+      setBarContext={cozyBar.setWebviewContext}
+      methods={{
+        onFileUploaded: (file, isSuccess, isLast) => {
+          window.postMessage({ file, isSuccess, isLast })
```
Crash--, line 20: is it wanted to use postMessage instead of post-me?

src/drive/web/modules/views/Upload/UploaderComponent.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+}))
+
+const tSpy = jest.fn()
+const uploadFilesFromNativeSpy = jest.fn()
+
+describe('DumbUpload component', () => {
+  const defaultItems = [{ fileName: 'File1.pdf' }]
+
+  const setupComponent = (): RenderResult<
+    typeof import('/home/anc/cozy/cozy-drive/node_modules/@testing-library/dom/types/queries'),
```
Crash--, line 18: to remove ?

src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+} from 'drive/web/modules/upload'
+import { ROOT_DIR_ID } from 'drive/constants/config'
+import Header from 'drive/web/modules/move/Header'
+import Explorer from 'drive/web/modules/move/Explorer'
+import FileList from 'drive/web/modules/move/FileList'
+import Loader from 'drive/web/modules/move/Loader'
+import LoadMore from 'drive/web/modules/move/LoadMore'
+import Footer from 'drive/web/modules/move/Footer'
+import Topbar from 'drive/web/modules/move/Topbar'
+import { buildMoveOrImportQuery, buildOnlyFolderQuery } from '../../queries'
```
Crash--, line 23: alias ?

src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+    }): void => {
+      if (!file) return
+
+      if (!isSuccess) {
+        dispatch({ type: RECEIVE_UPLOAD_ERROR, file })
+        return
+      }
+
+      removeFileToUploadQueue({ name: file.fileName })
+        .then(() => {
```
Crash--, line 71: does this method is really async?

src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+            return TypedAlerter.success(t('ImportToDrive.success'))
+          }
+          return
+        })
+        .catch(e => {
+          console.log('removeFileToUploadQueue error', e)
+        })
+    }
+
+    window.addEventListener('message', onFileUpload)
```
Crash--, line 82: no postme here?

src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }, [dispatch, removeFileToUploadQueue, t])
+
+  const onClose = useCallback(async () => {
+    await webviewIntent?.call('resetFilesToHandle').then(() => {
```
Crash--, line 86: I still really don't like this way of writing code... Can we do it differently to be the same as other place?

src/drive/web/modules/views/Upload/UploaderComponent.tsx
```diff
@@ -0,0 +1,236 @@
...
+      uploadFilesFromFlagship
+    )
+      .then(() => {
+        return console.log('uploadFilesFromNative done')
+      })
+      .catch(e => {
+        console.log('uploadFilesFromNative error', e)
+      })
+
+    setTimeout(() => navigate(`/folder/${folder._id}`), 50)
```
Crash--, line 110: I don't think this timeout is needed anymore? It was there for Cordova I think.

src/drive/targets/manifest.webapp
```diff
@@ -180,7 +180,7 @@
   "accept_documents_from_flagship": {
     "accepted_mime_types": ["*/*"],
     "max_number_of_files": 1,
-    "max_size_per_file_in_MB": 10,
+    "max_size_per_file_in_MB": 100,
```
Crash--, line 183: No way. We can't handle 100MB.

acezard: > No way. We can't handle 100MB.

what max value should we use? 10MB seems too low for me as even pictures can be above that

src/drive/web/modules/upload/index.js
```diff
@@ -401,12 +407,21 @@ export const uploadFilesFromNative =
...
+            })
+          } else {
+            await doMobileUpload(client, file.file.fileUrl, {
+              ...fileOpts,
+              contentType: file.file.type
+            })
+          }
         }
-        dispatch(removeFileToUploadQueue(file.file))
+        if (!alternateUploader) dispatch(removeFileToUploadQueue(file.file))
```
Crash--, line 424: Can't we get the same mechanism between alternateUploader and other?

Because : 
- We add to the UploadQueue here, no matter if this is an alternateUploader or not 
- But we remove only if this is not an alternate. This a strange. We should ave the same behavior.

acezard: I did so because there is no real way to know when the file will need to be removed (since it comes from a post-me message). It was too complex to couple post-me with the upload module. Instead we dispatch a removeFile to the current state whenever we receive the instruction

src/drive/targets/manifest.webapp
Crash--: We have speced 10MB. No?

src/drive/web/modules/upload/index.js
Crash--: In that case, can the alternateUploader send by itself the add to queue?

acezard: > In that case, can the alternateUploader send by itself the add to queue?

yes that would make things more aligned

src/drive/targets/manifest.webapp
acezard: > We have speced 10MB. No?

I believe that was for mespapiers only, could not find a hard number for Drive

Crash--: Use 10MB too. We'll see later.

src/drive/web/modules/views/Upload/UploaderComponent.tsx
acezard: refactored

src/drive/targets/manifest.webapp
```diff
@@ -175,5 +175,12 @@
       "verbs": ["POST"],
       "description": "Remote-doctype required to send anonymized measures to the DACC shared among mycozy.eu's Cozy."
     }
+  },
+  "accept_from_flagship": true,
+  "accept_documents_from_flagship": {
+    "accepted_mime_types": ["*/*"],
+    "max_number_of_files": 10,
```
acezard, line 182: @Crash-- same here, we don't have any info about what max number of files would be good. We want as many as possible, but I'm unsure what can be handled. I tested with 10 max on my end

src/drive/web/modules/views/Upload/UploaderComponent.spec.tsx
Ldoppea: Agree, this shouldn't be merged until we fix this line

src/drive/web/modules/views/Upload/UploadUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  } catch {
+    // fallback in case there's an error stringifying the maybeError
+    // like with circular references for example.
+    return new Error(String(maybeError))
+  }
+}
+
+export const getErrorMessage = (error: unknown): string => {
+  return toErrorWithMessage(error).message
+}
```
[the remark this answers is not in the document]
acezard: > nit: this should be in its own file as we expect to use it on other features in the future

@Ldoppea I moved it to src/drive/web/modules/drive/helpers.ts

src/drive/web/modules/views/Upload/UploadUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+  webviewIntent: WebviewService,
+  folder: { _id: string }
+): Promise<void> => {
+  for (const file of filesForQueue) {
+    if (!file.file) throw new Error('No file to upload')
+
+    const fileOptions = {
+      name: file.file.name,
+      dirId: folder._id,
+      conflictStrategy: 'rename'
```
Ldoppea, line 78: Also it seems like you does nothing with the `conflictStrategy` on Flagship side. Am I missing something?

src/drive/targets/manifest.webapp
Ldoppea: Done here : https://github.com/cozy/cozy-stack/pull/4179

src/drive/web/modules/views/Upload/UploadUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    logger('info', 'uploadFilesFromFlagship called')
+
+    await webviewIntent?.call('uploadFiles', JSON.stringify({ fileOptions }))
```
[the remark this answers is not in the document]
acezard: Yes it's one by one, it was made like that to mimick the old upload to mobile feature. In this case renaming seems the best approach

PR #2978, opened 2023-08-29
feat: Open modals with route instead pushModal

cballevre:
This aims to make easier to navigate into the Flagship application

```
### ✨ Features

* Use a route instead pushModal to open MoveModal
* Use a route instead pushModal to open qualification modal
* Use a route instead pushModal to open a ShareModal

### 🐛 Bug Fixes

* Center the spinner before OnlyOffice editor is ready
```


src/drive/styles/actionmenu.styl
```diff
@@ -1,40 +1,4 @@
...
-
-.fil-mobileactionmenu
-    color      var(--charcoalGrey)
-    hr
-        margin-top 0
-
-.fil-mobileactionmenu-header
-    font-weight bold
-    max-width 100%
-    display flex
```
Crash--, line 37: popopop \o/ 🎉

src/drive/web/modules/actions/share.jsx
```diff
@@ -0,0 +1,59 @@
...
+      const { t } = useI18n()
+      const { isMobile } = useBreakpoints()
+
+      return (
+        <ActionsMenuItem {...props} ref={ref}>
+          <ListItemIcon>
+            <Icon icon={ShareIcon} />
+          </ListItemIcon>
+          <ListItemText primary={t('Files.share.cta')} />
+          {isMobile && props.doc ? (
```
Crash--, line 35: why did you add `props.doc`?

src/drive/web/modules/actions/index.jsx
```diff
@@ -350,19 +335,25 @@ export const destroy = ({ pushModal, popModal }) => {
...
+          <ListItemText
+            className="u-error"
+            primary={t('SelectionBar.destroy')}
+          />
+        </ActionsMenuItem>
       )
-    }
+    })
   }
 }
```
Crash--, line 347: I didn't understand why you have to check isArray() everywhere and then [files]. Can you explain?

Crash-- (approved):
Nice work. 

A remark for later since I don't have a strong opinion on it yet: 

You added a route for every possible action on every level. Now there is a lof of declared routes. And each time we'll add an action, we'll have to add 3 routes. I love the "declarative" approach, but I don't know if it's suitable. 

Maybe we can create 3 routes : 
`/files/:id/action/:action`

And then, the `action` component linked to these route will do the job to check the `action` param to display the right modal. In that case, if we had an action, we'll not have to add new routes, but just editing this `action` component

src/drive/web/modules/actions/share.jsx
cballevre: Previously, we obtained a files property but only used the first element. With MenuActions, only the document linked to the action is shared with the doc property. I use it to get the identifier to display the recipients

src/drive/web/modules/actions/index.jsx
cballevre: Your comment makes me think that I haven't documented this point. 

Actions are used in two places in a file's drop-down menu and in the selection bar. Before the migration to ActionsMenu, we always received an array in both case. The ActionsMenu now sends the document directly to the action, while the SelectionBar sends an array of documents. This code seeks to manage the two cases.

I'm going to add a comment in the commit for the future me

cballevre:
This would really make sense to simplify the routes and share the file request logic. 

We could do something similar with another route for actions related to multiple files, such as moving or deleting.

I'll ask myself this question again when I migrate the last remaining modals using pushModal

PR #2999, opened 2023-10-06
fix: Make working back button for a file at another Cozy

cballevre:
```
### 🐛 Bug Fixes

* Make working back button for a file at another Cozy
```

**Context:**
Historically, we were going backwards in the browser's history for 3 reasons: 
- We cannot use the relative route because it is not a sub-route of where the user is coming from
- The OnlyOffice shared file opens in the owner's cozy, to return you must reopen the previous cozy 
- The creation process has redirect process that we need to avoid if we don't want to fall into an endless lopp

Why did we decide to use redirectLink instead of jumping to the browser history?
- It doesn't matter how many steps before if you have redirect link you can bring the user back to the starting point
- It's a lot more predictable where the user will end up 

To get around the problem of whether or not the url is shared in the search parameters, we use the `/permission/self` request in the stack as it is included with who the file is shared with. Each person in a share has a unique link, so there is only one person included.

### DIFF
--- a/src/drive/targets/public/components/AppRouter.jsx
+++ b/src/drive/targets/public/components/AppRouter.jsx
@@ -47,16 +47,6 @@ const AppRouter = ({
               />
             </Route>
             <Route
-              path="onlyoffice/:fileId/fromCreate"
-              element={
-                <OnlyOfficeView
-                  isPublic={true}
-                  isReadOnly={isReadOnly}
-                  isInSharedFolder={!isFile}
-                />
-              }
-            />
-            <Route
               path="onlyoffice/create/:folderId/:fileClass"
               element={<OnlyOfficeCreateView />}
             />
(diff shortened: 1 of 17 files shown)


src/drive/targets/public/components/AppRouter.jsx
Crash--, line 1: Why this change? Can we add a bit more of context please?

src/drive/web/modules/views/OnlyOffice/Toolbar/BackButton.jsx
```diff
@@ -3,15 +3,19 @@ import React from 'react'
...
   // TODO: remove u-ml-half-s when https://github.com/cozy/cozy-ui/issues/1808 is fixed
+  const { t } = useI18n()
+
   return (
     <IconButton
       data-testid="onlyoffice-backButton"
       className="u-ml-half-s"
       onClick={onClick}
       size="medium"
+      aria-label={t('button.back')}
```
Crash--, line 18: <3

Crash-- (approved):

PR #3001, opened 2023-10-10
fix(OnlyOffice): Regression from migration react-route to v6

cballevre:
In react-router v6, we need an `<outlet />` instead of a children otherwise the subroute like paywall would no longer be displayed

```
### 🐛 Bug Fixes

* Display the OnlyOffice paywall when there are no editing rights
```


Crash-- (approved):
If you say so

PR #3002, opened 2023-10-12
feat: When photos are trashed, delete the reference to albums

cballevre:
When you put an image in the trash even though it was part of an album, it is not removed from the album. You can therefore access the photo in the trash via album. To avoid bugs and inconsistencies, we remove the reference to the album via a service after each deletion of a photo

```
### ✨ Features

*  When photos are trashed, delete the reference to albums
```

**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1401

**Remaning work:**
- ~~Check if we can throttle the service execution~~


package.json
```diff
@@ -50,7 +50,9 @@
     "version:drive:androidbuild": "perl -pi -e 's,android-versionCode=\"[0-9]+\",android-versionCode=\"'`echo ${npm_package_version} | tr -d .`'0000\",g' ./src/drive/targets/mobile/config.xml",
     "version:drive:useragent": "replace 'value=\"(.+)\\d+\\.\\d+\\.\\d+\"' 'value=\"$1'$npm_package_version'\"' ./src/drive/targets/mobile/config.xml",
     "version:drive": "npm-run-all --parallel 'version:drive:*'",
-    "version:photos": "replace '\\d+\\.\\d+\\.\\d+' $npm_package_version ./src/photos/targets/manifest.webapp"
+    "version:photos": "replace '\\d+\\.\\d+\\.\\d+' $npm_package_version ./src/photos/targets/manifest.webapp",
+    "service:drive": "yarn cozy-konnector-dev -m src/photos/targets/manifest.webapp",
+    "service:photos": "yarn cozy-konnector-dev -m src/photos/targets/manifest.webapp"
```
Crash--, line 55: since your commit is "adding tooling", can you explain how to use them within the commit message?

src/photos/lib/onPhotoTrashed.js
```diff
@@ -0,0 +1,52 @@
+import { getReferencedBy } from 'cozy-client'
+import log from 'cozy-logger'
+
+import { buildPhotosTrashed } from 'photos/queries/queries'
+import { DOCTYPE_ALBUMS, DOCTYPE_FILES } from 'drive/lib/doctypes'
+
+const onPhotoTrashed = async client => {
+  log('info', `Service called with COZY_URL: ${process.env.COZY_URL}`)
```
Crash--, line 8: to remove :)

src/photos/lib/onPhotoTrashed.js
```diff
@@ -0,0 +1,52 @@
...
+      photosTrashedQuery.definition,
+      photosTrashedQuery.options
+    )
+
+    if (photos.length > 0) {
+      log(
+        'info',
+        `Start deleting album references on ${photos.length} trashed photos`
+      )
+      photos.forEach(async photo => {
```
Crash--, line 22: Be careful, in that case, forEach will be call in //. So we can make a lot of requests to the stack and in that case, we open a lot of handlers (handler can be : file opening or... network request). 

We have a very low limit on the number of open handlers in prod to avoid crazy memory consumption.

So you should do that differently? 

Or do I've a bad understanding of how forEach with async works?

Crash--: Yep or p-limit (https://www.npmjs.com/package/p-limit we use it in a few konnectors i think)

src/photos/lib/onPhotoTrashed.js
cballevre: Is this for data protection reasons? Or to keep the console clean?

Crash--: Both:
- Why do you need to log the cozy_url? 
- We try to avoid to log data within the log. You can do it, you can even log with debug() if needed.

But here, I really don't see the need to log that information.

cballevre: It wasn't just a mimicking of the other service. I was asking about applying a correction to this one too, if it was a matter of privacy. I'm going to correct mine so that I don't need it any more 👍

package.json
cballevre: Good point, I added a text in the commit message

src/photos/lib/onPhotoTrashed.js
Crash--: In fact, to be more precise: 
- forEach is way more performant because we do actions in // 
- `for of` do actions sync 

So in fact, forEach will be more performant in a lof of cases. Except when you have too much work in //. You can fully fill the RAM or... you can reach the opened handlers limit. 

My favorite solution for that, is to keep async, but with concurrency. With something like 20 async actions, it's ok. Like that, we still do 20 actions in //, so we have good perf, without killing the servers. 

(in our case, you make call from nsjail to the stack, no "user" network concerned or else, so the gain is not as good as when running on the client side). 

So if you want to use p-limit go for it, if you want to keep that way, go for it too ;). Your call

src/photos/queries/queries.js
```diff
@@ -44,6 +41,19 @@ export const buildTimelineQuery = () => ({
   }
 })
 
+export const buildPhotosTrashed = () => ({
+  definition: Q(DOCTYPE_FILES)
+    .where({
+      class: 'image',
+      trashed: true
+    })
+    .indexFields(['class', 'trashed']),
```
Crash--, line 50: since the predicate is static, should we use a partial index @paultranvan ?

src/photos/lib/onPhotoTrashed.js
```diff
@@ -0,0 +1,42 @@
+import { getReferencedBy } from 'cozy-client'
```
Crash--, line 1: no cozy-client update?

cballevre: If, of course, the CI turns overnight. I updated it this morning

src/photos/lib/onPhotoTrashed.js
cballevre: Thanks for the additional explanation. I think I'll stick with `for...of` because I don't want to add a new dependency to the project and I think the syntax is pretty clean

src/photos/queries/queries.js
paultranvan: Absolutely: using
```
.partialIndex({
      class: 'image',
      trashed: true
    })
 ```
 will pre-filter documents entering the index, so here, only the trashed images. While using the `.where` will index all the docs, making it unnecessarily bigger, and slower.

cballevre: Thanks for confirming Quentin's explanation. In the case of referenced_by, it doesn't seem to work because this field isn't part of the `io.cozy.files` document, but I may be wrong.

Crash-- (approved):
Nice improvement at the end!

PR #3005, opened 2023-10-24
fix: do not dispatch in resume hook catch block

acezard:
### 🐛 Bug Fixes

*  Removed the purgeUpload, as it's not clear if it would really be useful
in case of an unknown error
*  Explicitely handle post-me not implemented error
In this case doing nothing is what's needed

### DIFF
--- a/src/drive/web/modules/views/Upload/useResumeFromFlagship.ts
+++ b/src/drive/web/modules/views/Upload/useResumeFromFlagship.ts
@@ -5,11 +5,7 @@ import { useClient } from 'cozy-client'
 import { useWebviewIntent } from 'cozy-intent'
 import logger from 'cozy-logger'
 
-import {
-  getUploadQueue,
-  ADD_TO_UPLOAD_QUEUE,
-  purgeUploadQueue
-} from 'drive/web/modules/upload'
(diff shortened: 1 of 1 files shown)


zatteo:
Should work 👍

I did not read the cide for this feature so I do not know if it is relevant but do not forget we have the [isAvailable](https://github.com/cozy/cozy-flagship-app/blob/8763c11779cf57de5e05bb5c035bea1679bc4f4b/src/libs/intents/localMethods.ts#L84C11-L84C11) local methods for these issues now.

zatteo (approved):

acezard:
> Should work 👍
> 
> I did not read the code for this feature so I do not know if it is relevant but do not forget we have the [isAvailable](https://github.com/cozy/cozy-flagship-app/blob/8763c11779cf57de5e05bb5c035bea1679bc4f4b/src/libs/intents/localMethods.ts#L84C11-L84C11) local methods for these issues now.

Yes, that reminded me of that feature. I think here as it's not even necessary to verify if the method is available (the catch block only logs an error but does nothing), it should be fine at this point, but I'll see if I can improve the flow at a later point

PR #3006, opened 2023-10-25
fix: Harmonize flagship upload view

acezard:
### 🐛 Bug Fixes

- file icon
- no back arrow
- fix paddings
- fix breadcrumb overlay

image

### DIFF
--- a/src/declarations.d.ts
+++ b/src/declarations.d.ts
@@ -18,7 +18,8 @@ declare module 'cozy-ui/transpiled/react/providers/I18n' {
 
 declare module 'cozy-ui/transpiled/react/deprecated/Alerter' {
   const Alerter: {
-    error: (message: string) => void
+    error: (message: string, options?: Record<string, unknown>) => void
+    success: (message: string, options?: Record<string, unknown>) => void
   }
 
   export default Alerter
--- a/src/drive/locales/en.json
+++ b/src/drive/locales/en.json
@@ -533,6 +533,7 @@
     "header": "Uploading %{smart_count} photo to Cozy Drive |||| Uploading %{smart_count} photos to Cozy Drive",
     "header_mobile": "Uploading %{done} of %{total}",
     "header_done": "Uploaded %{done} out of %{total} successfully",
+    "success_flagship": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
     "close": "close",
     "item": {
       "pending": "Pending"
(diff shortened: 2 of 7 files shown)


src/drive/web/modules/views/Upload/useResumeFromFlagship.ts
```diff
@@ -46,4 +54,24 @@ export const useResumeUploadFromFlagship = (): void => {
 
     void doResumeCheck()
   }, [client, dispatch, webviewIntent, uploadQueue])
+
+  useEffect(() => {
+    // The upload can only happen once as the webapp is closed before the upload process can start again
+    if (didFlow.current) return
+
+    const processed = (getProcessed(state) as () => unknown[]).length
```
Ldoppea, line 62: This typing seems weird. Shouldn't this be `getProcesses(state) as unknown[]`? Otherwise you seems to call `.length` on a method pointer.

src/declarations.d.ts
```diff
@@ -18,7 +18,8 @@ declare module 'cozy-ui/transpiled/react/providers/I18n' {
 
 declare module 'cozy-ui/transpiled/react/deprecated/Alerter' {
   const Alerter: {
-    error: (message: string) => void
+    error: (message: string, options?: Record<string, unknown>) => void
+    success: (message: string, options?: Record<string, unknown>) => void
```
Ldoppea, line 22: Should we modify cozy-ui to expose those new definitions?

src/drive/locales/en.json
```diff
@@ -533,6 +533,7 @@
     "header": "Uploading %{smart_count} photo to Cozy Drive |||| Uploading %{smart_count} photos to Cozy Drive",
     "header_mobile": "Uploading %{done} of %{total}",
     "header_done": "Uploaded %{done} out of %{total} successfully",
+    "success_flagship": "%{smart_count} file uploaded with success. |||| %{smart_count} files uploaded with success.",
```
Ldoppea, line 536: Don't forget to add other languages translations

Ldoppea (approved):

src/drive/web/modules/views/Upload/useResumeFromFlagship.ts
acezard: that's because I'm casting the actual getProcessed selector instead of the result, in any case it doesn't look great when using untyped code in strictly typed code

src/declarations.d.ts
acezard: yes ideally, in this case I took the fastest way, I think it can be done after publication in test flight (for flagship too)

src/drive/locales/en.json
acezard: Going to do that in Transifex 👍

PR #3012, opened 2023-11-15
fix: Display shortcut icon for links from applications in the same cozy

cballevre:
```
### ✨ Features
 
* Display shortcut icon for links from applications in the same cozy
```

I changed `useFetchShortcut` to manage icon for links from applications in the same cozy. This change can return images from the registry, I display the fallback if the image is not found.

**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1422


src/drive/web/modules/filelist/FileIconShortcut.jsx
```diff
@@ -8,20 +8,22 @@ import GlobeIcon from 'cozy-ui/transpiled/react/Icons/Globe'
...
+
+  if (shortcutImg && !isBroken) {
+    return (
+      <img
+        src={shortcutImg}
+        width={size}
+        height={size}
+        onError={() => {
+          setBroken(true)
         }}
```
Crash--, line 21: why did you remove all the `div style`? Are you sure that was not needed? 

How an image coming from the registry can be broken?

cballevre: I removed the `div style` during my tests and left it in my final version because it was consistent with the way react works. Did I miss an optimisation to avoid reloading the image?

I don't know, I was inspired by the way AppIcon works in cozy-ui (cf: [code](https://github.com/cozy/cozy-ui/blob/6166abd6346fbfb77da8f058124e65340410282c/react/AppIcon/index.jsx#L114)).

cballevre: "how react works" I was saying more declarative. I don't think I changed the display. If there is an image it displays it else I display the fallback 🤔 From my point of view, the change is that the component always return what it displays without manipulating it with CSS

Crash-- (approved):

PR #3016, opened 2023-11-24
Polish photos backup interface

zatteo:
```
### ✨ Features

* Show a nice alert on error or success on backup page

```

Screenshot 2023-11-24 at 12 03 46
Screenshot 2023-11-24 at 12 03 55


src/photos/ducks/backup/components/BackupInfo.jsx
```diff
@@ -32,6 +54,33 @@ const BackupInfo = () => {
         </Alert>
       </div>
     )
+  } else if (status === 'done') {
```
cballevre, line 57: really nit: I was wondering whether removing a level of nesting wouldn't make it easier to read 🤔 
```js
} else if (status === 'done' && lastBackup.status === 'success') {
  return <BackupInfoSuccess />
} else if (status === 'done') {
  return <BackupInfoError />
}

return null
```

PR #3022, opened 2023-12-05
feat(move): Allow moving files only shared by link into shared folder

cballevre:
```
### ✨ Features

* Allow moving files only shared by link into shared folder
```


src/drive/web/modules/move/MoveModal.spec.jsx
```diff
@@ -229,6 +229,32 @@ describe('MoveModal component', () => {
   })
 
   describe('move shared folder inside another', () => {
+    it('should move the shared folder inside another if its by ling', async () => {
```
Crash--, line 232: nit: 
```suggestion
    it('should move the shared folder inside another if its by link', async () => {
```

Crash-- (approved):

src/drive/web/modules/move/MoveModal.spec.jsx
cballevre: Fix 👍

src/drive/web/modules/move/helpers.js
```diff
@@ -100,3 +100,24 @@ export const getEntriesType = entries => {
     return current.type
   }, null)
 }
+
+/**
+ * Returns whether one of the entries that is shared not only by link
+ * @param {object[]} entries
+ * @param {object} byDocId
```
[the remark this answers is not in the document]
cballevre: I added the type from cozy-client to be more specific

PR #3024, opened 2023-12-07
fix: Back button in shared OnlyOffice

cballevre:
When the user access OnlyOffice file in shared folder but the owner of the file isn't the owner of the share, it becomes impossible to know instance from where the user coming. 

This pull request aim to resolve this issue by implementing  by using browser history to redirection the user where he come from. However, it's essential to note a limitation of this solution: while it successfully addresses the scenario where someone navigates from another Cozy instance, it might not distinguish users accessing the link directly.

```
### 🐛 Bug Fixes

* Back button in shared OnlyOffice
```


src/drive/hooks/useRedirectLink.jsx
```diff
@@ -25,8 +27,16 @@ const useRedirectLink = () => {
         const permissions = await client
           .collection('io.cozy.permissions')
           .fetchOwnPermissions()
+
         if (permissions.included.length > 0) {
-          setInstance(permissions.included[0].attributes.instance)
+          if (permissions.included[0].attributes.instance) {
+            setInstance(permissions.included[0].attributes.instance)
+          } else {
+            // This case appears when the user where the owner of onlyoffice file
```
zatteo, line 35: This first line is strange ?

zatteo (approved):

src/drive/hooks/useRedirectLink.spec.jsx
```diff
@@ -88,44 +99,78 @@ describe('useRedirectLink', () => {
...
     await act(async () => {
-      render = renderHook(() => useRedirectLink())
+      render = renderHook(() => useRedirectLink({ isPublic: true }))
     })
 
-    expect(render.result.current.redirectWebLink).toBe(
+    render.result.current.redirectBack()
+
+    expect(mockNavigate).toHaveBeenCalledTimes(0)
+    expect(mockAssign).toHaveBeenCalledWith(
```
cballevre, line 113: @Crash-- do you know how to spy a `window.location` with jest? I tried several technique but I can't found a way to make it work

PR #3036, opened 2023-12-14
chore: Handle node 16/node 20 at the same time

Crash--:
Since we're stuck on webpack 4 ATM because of cozy-scripts, we need to fallback to --openssl-legacy-provider for crypto stuff.

But this node option is only available for node 17+. So we've to wrap our cozy-scripts call to a script that check the node version and add or not the option

```
### 🔧 Tech

* Can develop on node 16 or 20 
```


Crash--:
Not needed. It's already handled by cozy-script. https://github.com/cozy/cozy-drive/pull/3037

PR #3037, opened 2023-12-14
feat: Upgrade cozy-script

Crash--:
To be able to support node 16 and node 20 at the same time


cballevre:
This update has been made into https://github.com/cozy/cozy-drive/pull/3064 so we can develop into Node 20 now on Drive

PR #3047, opened 2023-12-26
feat: Warn user when he moves files into shared folder

cballevre:
```
### ✨ Features

* Warn user when he moves files into shared folder
```

### DIFF
--- a/src/drive/locales/fr.json
+++ b/src/drive/locales/fr.json
@@ -587,6 +587,12 @@
       "cancel": "Annuler",
       "confirm": "J'ai compris"
     },
+    "insideSharedFolder": {
+      "title": "Déplacer vers un dossier partagé ?",
+      "content": "Tous les membres ayant accès à %{destination} auront également accès à %{source}. |||| Tous les membres ayant accès à %{destination} auront également accès aux %{type} sélectionnés.",
+      "cancel": "Annuler",
+      "confirm": "Ok"
+    },
     "sharedFolderInsideAnother": {
       "title": "Déplacement impossible",
       "content_1": "Vous souhaitez déplacer un élément partagé dans un dossier lui-même partagé. Ce type déplacement n'est pas autorisé.",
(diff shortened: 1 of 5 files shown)


src/drive/locales/fr.json
```diff
@@ -592,6 +592,12 @@
       "cancel": "Annuler",
       "confirm": "J'ai compris"
     },
+    "insideSharedFolder": {
+      "title": "Déplacer vers un dossier partagé ?",
+      "content": "Tous les membres ayant accès à %{destination} auront également accès à %{source}. |||| Tous les membres ayant accès à %{destination} auront également accès aux %{type} sélectionnés.",
```
Ldoppea, line 597: Seems like `type` is always singular? So this would print `accès aux dossier sélectionnés` instead of `accès aux dossierS sélectionnés`?

src/drive/web/modules/move/MoveModal.jsx
```diff
@@ -106,22 +109,33 @@ const MoveModal = ({ onClose, entries, classes }) => {
...
-        moveEntries()
-      }
+
+    const areMovedFilesShared = hasOneOfEntriesShared(entries, byDocId)
+    const isOriginParentShared = hasSharedParent(entries[0].path)
+    const isTargetShared = hasSharedParent(targetPath)
+    const isInsideSameSharedFolder = targetPath.startsWith(sharedParentPath)
+
+    if (isInsideSameSharedFolder) {
+      moveEntries()
```
Ldoppea, line 119: As this is the same effect as the default path (line 138), does this means that this if/return condition is necessary to be called before other ifs?

Ldoppea (approved):

src/drive/locales/fr.json
Ldoppea: Never mind, this would be fixed by https://github.com/cozy/cozy-drive/pull/3045/files

src/drive/web/modules/move/MoveModal.jsx
cballevre: Yes this condition should be called before every others. If we are moving a file into the same shared folder we don't need to warn the user. @Ldoppea you see this list of conditions differently ?

Ldoppea: No, I just wanted to be sure to understand this correctly

PR #3048, opened 2023-12-27
[VO-266] feat: Update eslint-config-cozy-app from 4.0.0 to 6.1.0

cballevre:
The major benefit of this update is automatic import ordering. eslint-config-cozy-app doesn't handle import/order for aliases, so I put a custom configuration in Drive. @acezard if you have a better idea for the override I'd be interested

```
### 🔧 Tech

* Update eslint-config-cozy-app from 4.0.0 to 6.1.0
```

**Remaining work :**
- Make test pass again 😢


acezard (commented):
@cballevre is this still being worked on? I was also thinking lately that an eslint update was starting to be needed

cballevre:
@acezard I worked on it again at the start of the week but I'm getting the same errors. I'll let you have a look if you have a bit of time

```
> TrashFolderView › should contain breadcrumb with root path
    This test called the console which is forbidden.
    Here is what have been logged:
    console.error
      Error: Uncaught [TypeError: (0 , _cozyKeysLib.useVaultClient) is not a function]
```

acezard (approved):

PR #3053, opened 2023-12-28
[VO-10] feat: Allow the user to create folder inside move modal

cballevre:
```
### ✨ Features

* Allow the user to create folder inside move modal
```
Capture d’écran 2023-12-29 à 11 18 51

**Remaining work :**
- ~~Add a test to check whether the input folder display~~

### DIFF
--- a/src/drive/web/modules/filelist/AddFolder.jsx
+++ b/src/drive/web/modules/filelist/AddFolder.jsx
@@ -1,88 +1,38 @@
 import React from 'react'
-import { connect } from 'react-redux'
-import cx from 'classnames'
+import { connect, useDispatch } from 'react-redux'
 import compose from 'lodash/flowRight'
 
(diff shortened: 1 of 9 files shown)


src/drive/web/modules/filelist/AddFolder.jsx
```diff
@@ -1,88 +1,38 @@
...
-        </>
-      )}
-      <TableCell
-        className={cx(
-          styles['fil-content-cell'],
-          styles['fil-content-file-action']
-        )}
+  extraColumns
+}) => {
+  if (visible) {
```
Crash--, line 23: nit: we prefer the rule `return as quick as possible` to enhance the readability 

```
if(!visible) return null
return <AddFolder...
```

Crash-- (approved):
Nice!

PR #3057, opened 2024-01-03
[VO-144] feat: Add a 404 page when a folder does not exist

cballevre:
```
### ✨ Features

* Add a 404 page when a folder does not exist
```

drive alice cozy localhost_8080_


src/hooks/useDisplayedFolder.jsx
```diff
@@ -1,15 +1,26 @@
...
 
 const useDisplayedFolder = () => {
-  const client = useClient()
   const folderId = useCurrentFolderId()
 
+  const folderQuery = buildOnlyFolderQuery(folderId)
+  const folderResult = useQuery(folderQuery.definition, {
+    ...folderQuery.options,
+    enabled: !!folderId
+  })
```
Crash--, line 13: are we sure that we already requested `buildOnlyFolderQuery` ? Because we used `getDocumentFromState` because we knew that the folder was already be loaded. But we don't know by who. Here, we make the assumption that it was by `buildOnlyFolderQuery`. Is it true?

Crash--: FYI, we use `useDocument` a lot in banks (I removed a lot of call to this hook by replacing by a request and a fetch policy but still). 

https://github.com/search?q=repo%3Acozy%2Fcozy-banks%20useDocument&type=code

This hook should be moved to cozy-client one day.

Crash-- (approved):

src/hooks/useDisplayedFolder.jsx
cballevre: That's a very good point, I hadn't thought about the file if this hook is used in a route other than a folder. I'll look into this before merging

PR #3058, opened 2024-01-04
refactor: Remove Cordova app

cballevre:
```
### 🔧 Tech

* Remove Cordova app
```

**Why remove all this code?**
We are no longer publishing the Drive application via Cordova but through the [flagship application](https://github.com/cozy/cozy-flagship-app). As it was dead code, we've decided to remove it to facilitate maintenance and the development of new functionality.

**What does the PR do?** 
- It remove e2e test done with TestCafe as there are no longer in use
- Remove dependencies used into Cordova app
- Remove all component specific to mobile
- Remove offline and open action as there only used into the Cordova app

**Remaining work:**
- ~~Add some extra description of this PR~~
- ~~Make more test to make sure thats everything works~~

**Work for a next PR:**
- Extract every folder into 
- Clean and set new key for Travis to remove decrypt script


zatteo (approved):

zatteo:
Seems that we will also be able to do some cleaning on cozy-client ? https://github.com/cozy/cozy-client/blob/2ded95151510d0d9eddbdb585d8f860dce9240be/packages/cozy-client/src/models/fsnative.js#L145C5-L145C5

cballevre:
Yep its planned but we need to remove Cordova into cozy-banks before

Crash--:
> Yep its planned but we need to remove Cordova into cozy-banks before

I don't think that banks use cozy-client's cordova stuff?

PR #3059, opened 2024-01-05
fix: Add AlertProvider to AddMenuProvider

acezard:
### 🐛 Bug Fixes

* App will not crash in flagship anymore because AddMenuProvider doesn't seem
able to find the appwide AlertProvider

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -8,6 +8,7 @@ import { useClient } from 'cozy-client'
 import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
 import { useWebviewIntent } from 'cozy-intent'
 import useBreakpoints from 'cozy-ui/transpiled/react/providers/Breakpoints'
+import AlertProvider from 'cozy-ui/transpiled/react/providers/Alert'
 
 const wrap = (Component, className) => {
   const WrappedBarComponent = ({ children }) => {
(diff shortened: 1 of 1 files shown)


cballevre:
I suggest you add it [here](https://github.com/cozy/cozy-drive/blob/089497f3b6df4bc2c723bc9ef51f53ffdea3a152/src/components/Bar.jsx#L49) so that it will be available for all other add-ons in the future rather than specifically for the add menu.

acezard:
> I suggest you add it [here](https://github.com/cozy/cozy-drive/blob/089497f3b6df4bc2c723bc9ef51f53ffdea3a152/src/components/Bar.jsx#L49) so that it will be available for all other add-ons in the future rather than specifically for the add menu.

Not sure, don't you mean Toolbar ? https://github.com/cozy/cozy-drive/blob/fix--Add-AlertProvider-to-AddMenuProvider/src/drive/web/modules/drive/Toolbar/index.jsx

cballevre:
No I mean the Bar component. The problem is that the bar does not include the same React tree as Drive. You therefore need to provide it with all the providers you use in Drive. That's why I think it's more appropriate to add into the `BarRigthWithProvider` component.

acezard:
> No I mean the Bar component. The problem is that the bar does not include the same React tree as Drive. You therefore need to provide it with all the providers you use in Drive. That's why I think it's more appropriate to add into the `BarRigthWithProvider` component.

updated 👍

zatteo (approved):

PR #3070, opened 2024-01-29
feat: Remove cozy-authentification

cballevre:
This dependency was only necessary for Cordova

```
### 🔧 Tech

* Remove cozy-authentification
```


Crash-- (approved):
Nice! You can now archive the lib on cozy-libs ? :D

cballevre:
I think so it has been remove on cozy-drive and cozy-banks. My only question, is how do we archive a package into mono-repository in Github ?

PR #3071, opened 2024-01-31
fix: Make sure that the files to be moved have a path

cballevre:
After duplicating a file, it has no path. This makes it impossible to know if it is shared. For a move, we check some conditions around the sharing so we make sure that the moved file always has a path.

```
### 🐛 Bug Fixes

* Make sure that the files to be moved have a path
```

### DIFF
--- a/src/components/FilesRealTimeQueries.jsx
+++ b/src/components/FilesRealTimeQueries.jsx
@@ -1,0 +1,130 @@
+import { memo, useEffect } from 'react'
+
+import { useClient, Mutations } from 'cozy-client'
+import { ensureFilePath } from 'cozy-client/dist/models/file'
+import { receiveMutationResult } from 'cozy-client/dist/store'
+
+import { buildFileByIdQuery } from 'modules/queries'
+
+/**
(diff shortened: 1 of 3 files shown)


acezard (approved):

src/components/FilesRealTimeQueries.jsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+    const dispatchUpdate = async couchDBDoc => {
+      const doc = await computeDocBeforeDispatchUpdate(couchDBDoc, client)
+      dispatchChange(client, doctype, doc, Mutations.updateDocument)
+    }
+    const dispatchDelete = async couchDBDoc => {
+      const doc = await computeDocBeforeDispatchDelete(couchDBDoc, client)
+      dispatchChange(client, doctype, doc, Mutations.deleteDocument)
+    }
+
```
Crash--, line 106: Nice!

nit: I think in order to easily migrate later, you should have pass the method `computeDocBeforeDispatchXXX` to the dispatchChange. But we can start with this one 👍

PR #3087, opened 2024-02-22
feat: Add dropzone into SharingsFolderView (VO-323)

acezard:
### ✨ Features

* Allow users to drag & drop files into the sharings view inside a directory

### DIFF
--- a/src/modules/views/Sharings/SharingsFolderView.jsx
+++ b/src/modules/views/Sharings/SharingsFolderView.jsx
@@ -33,6 +33,7 @@ import {
   buildDriveQuery,
   buildFileWithSpecificMetadataAttributeQuery
 } from 'modules/queries'
+import Dropzone from 'modules/upload/Dropzone'
 
 const desktopExtraColumnsNames = ['carbonCopy', 'electronicSafe']
 const mobileExtraColumnsNames = []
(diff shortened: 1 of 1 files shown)


src/hooks/useCurrentFolderId.jsx
```diff
@@ -12,6 +12,8 @@ const useCurrentFolderId = () => {
     return ROOT_DIR_ID
   } else if (pathname === '/trash') {
     return TRASH_DIR_ID
+  } else if (pathname === '/sharings') {
```
cballevre, line 15: Why do you need these lines? Normally you should always use folderId

acezard: without it, how can we determine the writeAccess? https://github.com/cozy/cozy-drive/pull/3087/files#diff-46a2a380a427ee4574a6020ae85eb31d41f1ebb96e8a3bfddcb6b2df65e62469R52

(same implementation as root folder view)

Crash--: But `root folder view` is really ROOT DIR. Here this is not the case. 

So instead of faking the currentFolder, can't you just do 

```
canWriteToCurrentFolder={currentFolderId && hasWriteAccess(currentFolderId)} 
```

or something similar?

acezard: yes I guess I should also use https://github.com/cozy/cozy-drive/blob/master/src/modules/views/Sharings/SharingsFolderView.jsx#L98, updating

cballevre: We can determine it with the param into the path here : https://github.com/cozy/cozy-drive/blob/9e72873f2e82177a9b30acee98e6407b55d5adee/src/hooks/useCurrentFolderId.jsx#L10
As we access : `/sharings/:folderId`
The trick for `ROOT_DIR_ID` was made for when you are accessing `/folder` which is not anymore available because we redirect to `/folder/io.cozy.files.root-dir` so its dead code I think

acezard: removed it and use `hasWrite` instead

cballevre (approved):

PR #3106, opened 2024-03-08
fix: Refactor ShareFileView component (VO-370)

acezard:
### 🐛 Bug Fixes

* Fix a bug when infinite loader could show up when removing self from a sharing

### DIFF
--- a/src/modules/views/Modal/ShareFileView.jsx
+++ b/src/modules/views/Modal/ShareFileView.jsx
@@ -14,19 +14,31 @@ const ShareFileView = () => {
   const fileQuery = buildFileByIdQuery(fileId)
   const fileResult = useQuery(fileQuery.definition, fileQuery.options)
 
-  if (hasQueryBeenLoaded(fileResult) && fileResult.data) {
-    const onClose = () => {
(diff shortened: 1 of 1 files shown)


src/modules/views/Modal/ShareFileView.jsx
```diff
@@ -14,21 +14,33 @@ const ShareFileView = () => {
...
+  // After successfully removing self from a shared file, the file is not found anymore but the query is considered loaded
+  // We check if the data is null, meaning the sharing has been removed
+  if (hasQueryBeenLoaded(fileResult) && !fileResult.data) {
+    handleExit()
+  }
+
+  // Accessing the URL of a file that doesn't exist anymore (or never existed)
+  // e.g. /folder/io.cozy.files.shared-with-me-dir/file/someidresolvingto404/share
+  if (fileResult.fetchStatus === 'failed') {
+    handleExit()
```
cballevre, line 41: nit: Add an alert to notify there was a problem with his action

cballevre (approved):
Thanks, this looks much more robust 👍

PR #3115, opened 2024-03-21
fix: Can open shortcut even on preview page

Crash--:
We had a bug that caused an issue where if we wanted to open a shortcut from the preview page, it wouldn't work.

Indeed, since we were only testing the public case and forgetting the preview case, it couldn't work.

So now, we rely on the props to be able to know if we are public or not instead of just checking the url.

```
### ✨ Features

*

### 🐛 Bug Fixes

* We had a bug that caused an issue where if we wanted to open a shortcut from the preview page, it wouldn't work.

### 🔧 Tech

*
```


cballevre (approved):
LGTM 👍

PR #3122, opened 2024-03-26
feat: Display external drives mobile and desktop (VO-529)

acezard:
### External Drive Data Fetching Module Documentation

This module provides functionalities to fetch and build queries for external drive data within the application, leveraging the `cozy-client` library for interaction with Cozy's backend services.

#### Imports

- **CozyClient, QueryDefinition**: Imported from `cozy-client`. `CozyClient` is used to initialize and manage the interaction with Cozy's stack, while `QueryDefinition` is utilized for constructing queries to fetch data.
- **QueryOptions**: Imported from `cozy-client/types/types`, specifying options for queries made to the Cozy stack.
- **logger**: A custom logging utility from `lib/logger` for error logging.
- **buildDriveQuery**: A function from `modules/queries` that assists in building queries for drive data based on specified parameters.

##### buildExternalDriveFilesQuery

```typescript
function buildExternalDriveFilesQuery(
  sortOrder: SortOrder,
  externalDriveId: string
): { definition: QueryDefinition; options: QueryOptions }
```

Builds a query for fetching files from an external drive with sorting.

**Parameters:**
- `sortOrder`: An object of type `SortOrder` specifying how the results should be sorted.
- `externalDriveId`: The unique identifier for the external drive.

**Returns:** An object containing the `definition` of the query and the `options` for executing it.

**Usage:**
```typescript
const { definition, options } = buildExternalDriveFilesQuery({ attribute: 'name', order: 'asc' }, 'driveId');
```

##### buildExternalDriveQuery

```typescript
function buildExternalDriveQuery(): {
  definition: QueryDefinition;
  options: QueryOptions;
}
```

Convenience function to build a query for fetching external drive data, preconfigured with sorting by name in ascending order and targeting the shared drives directory.

**Returns:** An object with the `definition` of the query and the `options` for its execution.

**Usage:**
```typescript
const { definition, options } = buildExternalDriveQuery();
```

* External drives will be displayed in the sidebar in desktop view and in a new tab in mobile view in drive root

image 
image


package.json
```diff
@@ -95,6 +94,7 @@
     "cozy-intent": "^2.18.0",
     "cozy-keys-lib": "6.0.0",
     "cozy-logger": "1.9.1",
+    "cozy-minilog": "3.3.1",
```
cballevre, line 97: nit: Next time maybe split this update into its own commit

src/modules/navigation/NavItem.jsx
```diff
@@ -49,9 +53,9 @@ const NavItem = ({
 }
 
 NavItem.propTypes = {
-  to: PropTypes.string.isRequired,
+  to: PropTypes.string,
   icon: PropTypes.string,
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
```
cballevre, line 58: nit: you can add `forcedLabel` in propTypes

src/modules/views/Folder/ExtraDrive.tsx
```diff
@@ -52,12 +62,15 @@ export const ExtraDrive = (): JSX.Element | null => {
       >
         {files.map(file => (
           <NavItem
-            key={file._id}
+            key={file.id}
```
cballevre, line 65: Shouldn't we always use _id ? 🤔

src/modules/views/Folder/hooks/useExtraDrive.ts
```diff
@@ -1,41 +1,30 @@
...
+  // Execute the query using cozy-client's useQuery hook
+  // The result of the query is cast to ensure it's treated as an array of ExtraDriveFile objects
+  const { data } = useQuery(query.definition, query.options) as
+    | {
+        data: ExtraDriveFile[]
+      }
+    | { data: null }
+
+  // Return the fetched data
+  return data ?? []
```
cballevre, line 29: nit: Maybe return the data into an object to be able in future to add a loading state, for example, but this may be premature optimisation

src/modules/views/Folder/queries/fetchExtraDrive.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Asynchronously fetches data from external drives.
+ * @param client - An instance of CozyClient to interact with Cozy's stack.
+ * @returns A Promise resolving to the fetched data as FetchedExtraDriveData.
+ *
+ * Usage:
+ * const data = await fetchExternalDriveData(cozyClient);
+ */
+export const fetchExternalDriveData = async (
```
cballevre, line 20: I have the impression that this function is not used anywhere, perhaps a residual from development that needs to be removed

src/modules/views/Folder/ExtraDrive.tsx
acezard: well, I had an item without `_id`, not sure how that happened though? But looking in the project, `file.id` is used 40 times, and `file._id` 11 times. Is this something to investigate?

src/modules/views/Folder/queries/fetchExtraDrive.ts
acezard: Yes it was replaced with the query definition, removing it

cballevre (approved):

src/modules/views/Folder/hooks/useExtraDrive.ts
Crash--: What is the point to have this custom hook instead of just using the build method and the query directly within the component? You are creating a new API, in fact a subset API, that will bring more complexity over the time than just using the hook from cozy-client directly.

src/modules/views/Folder/ExtraDrive.tsx
Crash--: I don't know how you can have a file without an `_id` since this is what we store to the DB. In fact cozy-client is there to let you use `_id` or `id` so you should be able to use both.

src/modules/views/Folder/hooks/useExtraDrive.ts
acezard: @Crash-- it's because I used it at two different places and wanted to centralise the query at one place only. Indeed it has the side effect of hiding cozy-client here, so you would useQuery with the same definition/options at the two different components then?

Crash--: This is what we do yes. We write the buildQueryXXX, and then we call the useQuery with this query in several place.

acezard: removed the file and did that instead 👍

PR #3135, opened 2024-04-10
fix: Update err max file size handling (VO-637)

acezard:
### 🐛 Bug Fixes

* Removed an issue where uploading a file above quota would trigger max file size error wrongly


cballevre (approved):
LGTM 👍

PR #3136, opened 2024-04-11
feat: Hide shared drives dir from view (VO-647)

acezard:
This dir will be hidden from view in the following:
* search
* drive view
* move view
* upload view (uses move query)

Using db filter when possible
Javascript if not

### DIFF
--- a/src/modules/search/components/helpers.js
+++ b/src/modules/search/components/helpers.js
@@ -1,6 +1,6 @@
 import { models } from 'cozy-client'
 
-import { ROOT_DIR_ID } from 'constants/config'
+import { ROOT_DIR_ID, SHARED_DRIVES_DIR_ID } from 'constants/config'
 import FuzzyPathSearch from 'lib/FuzzyPathSearch.js'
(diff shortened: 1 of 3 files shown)


src/modules/search/components/helpers.js
```diff
@@ -94,9 +94,16 @@ export const indexFiles = async client => {
   const notOrphans = file =>
     folders.find(folder => folder._id === file.dir_id) !== undefined
   const notRoot = file => file._id !== ROOT_DIR_ID
+  // Shared drives folder to be hidden in search.
+  // The files inside it though must appear. Thus only the file with the folder ID is filtered out.
+  const notSharedDrivesDir = file => file.id !== SHARED_DRIVES_DIR_ID
```
paultranvan, line 99: You should test on `_id` rather than `id`.  `id` might work thanks to cozy-client support, but we better not rely on this much

acezard: yes good catch that was my initial intent

paultranvan (approved):

PR #3138, opened 2024-04-11
fix: improve shared drives UI/UX (VO-651) (VO-652) (VO-653)

acezard:
remove url suffix
add ellipsis

image

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -8,7 +8,7 @@
     "item_trash": "Trash",
     "item_settings": "Settings",
     "item_collect": "Administrative",
-    "item_external_drives": "Shared Drives",
+    "item_shared_drives": "Shared Drives",
     "item_my_drive": "My Drive",
     "btn-client": "Get Cozy Drive for desktop",
     "support-us": "View offers",
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -9,6 +9,7 @@
     "item_settings": "Settings",
     "item_collect": "Administrative",
     "item_external_drives": "Shared Drives",
+    "item_my_drive": "My Drive",
     "btn-client": "Get Cozy Drive for desktop",
     "support-us": "View offers",
     "support-us-description": "Would you like to benefit from more space or simply support cozy?",
--- a/src/modules/filelist/File.jsx
+++ b/src/modules/filelist/File.jsx
@@ -47,7 +47,7 @@ const File = ({
   onFolderOpen,
   onFileOpen,
   disableSelection = false,
-  isSharedDrives
+  isExternalDrive
 }) => {
   const [actionMenuVisible, setActionMenuVisible] = useState(false)
   const filerowMenuToggleRef = useRef()
@@ -143,7 +143,7 @@ const File = ({
             file={attributes}
             size={isLargeRow ? 96 : undefined}
             isInSyncFromSharing={isInSyncFromSharing}
-            isSharedDrives={isSharedDrives}
+            isExternalDrive={isExternalDrive}
           />
         </TableCell>
         <FileName
@@ -156,7 +156,7 @@ const File = ({
           formattedUpdatedAt={formattedUpdatedAt}
           refreshFolderContent={refreshFolderContent}
           isInSyncFromSharing={isInSyncFromSharing}
-          isSharedDrives={isSharedDrives}
+          isExternalDrive={isExternalDrive}
         />
         <LastUpdate
           date={updatedAt}
--- a/src/modules/filelist/File.jsx
+++ b/src/modules/filelist/File.jsx
@@ -47,7 +47,7 @@ const File = ({
   onFolderOpen,
   onFileOpen,
   disableSelection = false,
-  isExternalDrive
+  isSharedDrives
 }) => {
   const [actionMenuVisible, setActionMenuVisible] = useState(false)
   const filerowMenuToggleRef = useRef()
@@ -143,7 +143,7 @@ const File = ({
             file={attributes}
             size={isLargeRow ? 96 : undefined}
             isInSyncFromSharing={isInSyncFromSharing}
-            isExternalDrive={isExternalDrive}
+            isSharedDrives={isSharedDrives}
           />
         </TableCell>
         <FileName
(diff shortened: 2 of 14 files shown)


src/modules/filelist/File.jsx
```diff
@@ -47,7 +47,7 @@ const File = ({
   onFolderOpen,
   onFileOpen,
   disableSelection = false,
-  isExternalDrive
+  isSharedDrives
```
cballevre, line 50: Why rename external by shared? As far as I'm concerned, the two concepts are similar but different. For example, we'll have a different icon for displaying shared drives in the tabs view

acezard: it was to match the doctype `io.cozy.files.shared-drives-dir`, what do you think?

cballevre: I understand your change better, but I'd put more nuance into it. I would keep the external naming for everything related to the external link. For the moment, we only retrieve external links via the magic folder but soon we'll have folders for shared drives with a special design.

src/locales/en.json
```diff
@@ -9,6 +9,7 @@
     "item_settings": "Settings",
     "item_collect": "Administrative",
     "item_external_drives": "Shared Drives",
```
cballevre, line 11: I think this should be `item_shared_drives` to be consistent

src/modules/filelist/File.jsx
cballevre: For example, I will still use the external here

src/modules/navigation/NavContent.jsx
```diff
@@ -6,9 +6,10 @@ import Circle from 'cozy-ui/transpiled/react/Circle'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import OpenWithIcon from 'cozy-ui/transpiled/react/Icons/Openwith'
 import { NavIcon, NavText } from 'cozy-ui/transpiled/react/Nav'
+import Typography from 'cozy-ui/transpiled/react/Typography'
 import useBreakpoints from 'cozy-ui/transpiled/react/providers/Breakpoints'
 
-const NavContent = ({ icon, badgeContent, label, external }) => {
+const NavContent = ({ icon, badgeContent, label, sharedDrives }) => {
```
cballevre, line 12: Here I will also still use external as this determines whether the external icon appears. But I prefer this if to separate cases 👍

src/modules/navigation/NavItem.jsx
```diff
@@ -31,21 +30,20 @@ const NavItem = ({
   rx,
   clickState,
   badgeContent,
-  external,
+  sharedDrives,
```
cballevre, line 33: I would keep `external` also here

src/modules/views/Folder/SharedDrives.tsx
```diff
@@ -51,7 +48,7 @@ export const ExtraDrive = ({
                 onFileOpen={handleFileOpen}
                 isInSyncFromSharing={false}
                 disableSelection={true}
-                isExternalDrive={true}
+                isSharedDrives={true}
```
cballevre, line 51: This is where the subtlety comes in. This component knows that it is displaying the contents of shared drives. But we specify the display cases for the children. Here an external drive and in the future a shared drive.

src/locales/en.json
acezard: indeed, seems I missed it

src/modules/filelist/File.jsx
acezard: agreed

src/modules/navigation/NavContent.jsx
acezard: agreed

src/modules/views/Folder/SharedDrives.tsx
acezard: yes, instead I will check if it's of type shortcut. I assume shared drives will be of type file or directory?

cballevre: From my understanding it would be a directory

src/modules/navigation/NavItem.jsx
```diff
@@ -31,21 +30,20 @@ const NavItem = ({
...
   return (
-    <UINavItem secondary={secondary} onClick={onClick}>
+    <UINavItem secondary={secondary}>
       <NavLink to={to} rx={rx} clickState={clickState}>
         <NavContent
           icon={icon}
           label={forcedLabel ?? t(`Nav.item_${label}`)}
           badgeContent={badgeContent}
-          external={external}
+          sharedDrives={externalDrive}
```
cballevre, line 46: I think there's a renaming error here

src/modules/filelist/File.jsx
```diff
@@ -47,7 +47,7 @@ const File = ({
   onFolderOpen,
   onFileOpen,
   disableSelection = false,
-  isExternalDrive
+  externalDrive
```
cballevre, line 50: Usual I prefer to prefix with an `is` the boolean but the props just below don't follow this convention so its only a nit

src/modules/navigation/NavContent.jsx
```diff
@@ -6,9 +6,10 @@ import Circle from 'cozy-ui/transpiled/react/Circle'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import OpenWithIcon from 'cozy-ui/transpiled/react/Icons/Openwith'
 import { NavIcon, NavText } from 'cozy-ui/transpiled/react/Nav'
+import Typography from 'cozy-ui/transpiled/react/Typography'
 import useBreakpoints from 'cozy-ui/transpiled/react/providers/Breakpoints'
 
-const NavContent = ({ icon, badgeContent, label, external }) => {
+const NavContent = ({ icon, badgeContent, label, externalDrive }) => {
```
cballevre, line 12: nit: `external` is more general maybe we can use it for other purpose in the futur

cballevre (approved):
LTGM 👍  after fixing the renaming error

src/modules/navigation/NavContent.jsx
Crash--: Capture d’écran 2024-04-16 à 15 37 07
from https://github.com/cozy/cozy-guidelines/?tab=readme-ov-file#naming-of-functions ;)

PR #3146, opened 2024-04-22
feat: Hide shared drives files from recent/search (VO-647)

acezard:
Do not show files in shared drives dir in recent or search


Ldoppea (approved):
Commit title seems to be misleading. It suggests that shared drives are hidden in both recent and search sections, but the PR description and the code suggest that they are not hidden in search section.

The PR content is good anyway.

PR #3153, opened 2024-05-07
[VO-711 & VO-741 & VO-755 & VO-707] feat: Show the content of Nextcloud folder

cballevre:
```
### ✨ Features

* Show the content of Nextcloud folder
```

All this work is behind the flag `drive.show-nextcloud-dev`

**Note on this PR :** 
This is the first part of my work on Nextcloud so that it can be tested. Some parts will be removed or merged with others in the next version.

Especially the actions, the ones that don't have Nextcloud in the naming correspond to the ones I intend to make common with the `io.cozy.files`. I plan to use 2 strategies : 
1. For action with input by user, I would use `FolderBody` as the minimal part and to expose two new methods (onFileRename, onFileCreate). They will be implemented specifically for `io.cozy.files` and `io.cozy.remote.nextcloud.files` in a wrapper component like I begin in `NextcloudFolderBody`.
2. For actions using a modal, I want to redirect use to a specific routes like `:parent/delete` where we can have different implementation

The two slightly more hybrid cases would be the move modal and uploading a file, which I need to look at in more detail.

This version still produces a lot of 404s because it tries to download the thumbnail of a Nextcloud file using the `io.cozy.files` api.

**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1471


src/locales/en.json
```diff
@@ -787,5 +788,11 @@
   "NotFound": {
     "title": "The element cannot be found",
     "text": "We have not found anything at this address. This may be a typing error."
+  },
+  "NextcloudBreadcrumb": {
+    "root": "Drive partagés"
```
Ldoppea, line 793: Wrong language here

src/modules/views/Folder/SharedDrives.tsx
```diff
@@ -1,42 +1,83 @@
...
+  const handleFileOpen = ({ file }: { file: IOCozyFile }): void => {
+    if (
+      file.cozyMetadata?.createdByApp === 'nextcloud' &&
+      flag('drive.show-nextcloud-dev')
+    ) {
+      navigate(`/nextcloud/${file._id}`)
+    } else {
+      const { subdomain: subDomainType } = (client?.getInstanceOptions() ?? {
+        subdomain: 'flat'
+      }) as { subdomain: string }
```
Ldoppea, line 58: I'm not sure you should call `getInstanceOptions()` here.

As you are in a web app, you can access `client.instanceOptions` as it is already initialized from DOM here: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1677-L1689

But even better, you can call `client.capabilities.flat_subdomains` that is initialized at the same place and directly contains the `boolean` you are looking for. Also `client.capabilities` is typed [here](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L188-L191) so you dont need to use any `as XXX` notation

Last but not least, you can also use the [useCapabilities](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useCapabilities.jsx) hook if needed (but it is not typed, we should fix that someday)

src/modules/folder/hooks/useNeedsToWait.jsx
```diff
@@ -0,0 +1,30 @@
...
+ * more simpler component and then the files
+ */
+const useNeedsToWait = ({ isLoading }) => {
+  const [needsToWait, setNeedsToWait] = useState(true)
+  useEffect(() => {
+    let timeout = null
+    if (!isLoading) {
+      timeout = setTimeout(() => {
+        setNeedsToWait(false)
+      }, 50)
```
Ldoppea, line 23: Here are we waiting for 50ms in order to be in a new render loop? If yes we can reduce it to 1ms can't we?

Or maybe this is to display the skeleton long enough to avoid any flickering?

Ldoppea (approved):
Maybe the 2 `make future action visible` commits messages need to be more explicit. I'm not sure to understand them.

But anyway, only a few suggestion but everything is fine.

src/modules/navigation/DrivesNavItem.jsx
```diff
@@ -1,32 +1,86 @@
...
+    sortOrder: 'asc'
+  })
   const { data } = useQuery(
     sharedDrivesQuery.definition,
     sharedDrivesQuery.options
   )
 
-  const label = data?.length > 0 ? 'drives' : 'drive'
+  const hasMultipleDrive = data?.length > 0
+  const showNextcloud = flag('drive.show-nextcloud-dev') == true
```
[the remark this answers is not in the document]
cballevre: Yes if you have shortcut into the shared drives folder it will display them there. In the future, it will have also folder as the real version of shared drives

src/modules/folder/hooks/useNeedsToWait.jsx
cballevre: I thinks this for avoid flickering but I only copy it from FolderViewBody trying to extract the essential piece. My commit message are enough explicite ☹️

PR #3174, opened 2024-06-12
feat: Simplify css and theme integration

JF-Cozy:
and load CozyTheme before AlertProvider, and so fix a style problem

```

### 🔧 Tech

* Simplify css and theme integration
```


Merkur39 (approved):
Miss sur le premier commit ?

PR #3183, opened 2024-07-08
feat: Regroup queries

cballevre:
The purpose of this is to have an access point to all the queries to facilitate optimisation process (cc: @paultranvan). 

This brings them all together in two files :
* `lib/queries.js` : that contains utility functions for queries
* `queries/index.ts` : that contains the query builder function

Now QueryBuilder returns the definition encapsulated in a function to unify the way it works. However, I've had to update their use in `.query({` and `.fetchAndGetFromState({` because it doesn't handle functions like `useQuery` does.


acezard (approved):

paultranvan:
To quote a great and wise brand : I'm lovin it.

PR #3187, opened 2024-07-15
[VO-740] feat(nextcloud): Show trash

cballevre:
```
### ✨ Features

*  Show trash inside nextcloud
```

**Related PRs :**
- https://github.com/cozy/cozy-client/pull/1499
- https://github.com/cozy/cozy-client/pull/1501
- https://github.com/cozy/cozy-client/pull/1502


src/modules/trash/components/DestroyConfirm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onConfirm
+}) => {
+  const { t } = useI18n()
+  const [isBusy, setBusy] = useState(false)
+
+  const handleDestroy = async (): Promise<void> => {
+    try {
+      setBusy(true)
+      await onConfirm()
+    } finally {
```
acezard, line 36: we don't handle the error scenario with some UI feedback here?

src/modules/views/Trash/TrashDestroyView.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  if (!hasFileIds) {
+    return <Navigate to=".." replace={true} />
+  }
+
+  const handleConfirm = async (): Promise<void> => {
+    if (!fileResult.data) return
+    for (const file of fileResult.data) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
acezard, line 41: I've seen multiple of these maybe it's easier to type cozy-client in d.ts for this method

src/modules/filelist/File.jsx
```diff
@@ -118,7 +117,7 @@ const File = ({
   // because they are magic folder created by the stack
   const canInteractWithFile =
     attributes._id !== 'io.cozy.files.shared-drives-dir' &&
-    attributes._id !== TRASH_DIR_ID
+    !attributes._id.endsWith('.trash-dir')
```
acezard, line 120: why use a literal string instead of a constant here?

src/modules/nextcloud/components/NextcloudTrashFolderBody.tsx
```diff
@@ -32,12 +36,21 @@ const NextcloudTrashFolderBody: FC<NextcloudTrashFolderBodyProps> = ({
     window.open(file.links.self, '_blank')
   }
 
-  const fileActions = makeActions([], {
+  const handleRefresh = async (files: NextcloudFile[]): Promise<void> => {
```
acezard, line 47: what is supposed to happen in case of throw during this refresh?

src/modules/views/Nextcloud/NextcloudDestroyView.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  if (
+    hasDataLoaded(nextcloudResult as UseQueryReturnValue) &&
+    nextcloudResult.data
+  ) {
+    const entries = nextcloudResult.data.filter(({ _id }) =>
+      state.fileIds?.includes(_id)
+    )
+
+    const handleConfirm = async (): Promise<void> => {
```
acezard, line 58: same here, what happens in case we can't handleConfirm? on ui side particularly (maybe it's irrelevant, not sure)

src/modules/filelist/File.jsx
cballevre: This allows me to manage all the trash from both doctype `io.cozy.files` and `io.cozy.remote.nextcloud.files` in one line. You would have prefer something like that : 
```
const canInteractWithFile =
    attributes._id !== SHARED_DRIVES_DIR_ID &&
    attributes._id !== TRASH_DIR_ID &&
    attributes._id !== NEXTCLOUD_TRASH_DIR_ID
```
Maybe this constant should be provided by cozy-client or cozy-stack-client as they are already set as type [here](https://github.com/cozy/cozy-client/blob/556aef794fd9cdb577b7c22860f932187c355856/packages/cozy-client/types/types.d.ts#L8). 

Note:  I tend to put the doctype directly because I find them quicker to search globally in a project.

src/modules/views/Trash/TrashDestroyView.tsx
cballevre: I don't really want to modify `CozyClient` typing directly in Drive. We'd have to see how we could integrate this directly into the package, but we'd probably have to consider typing cozy-stack-client at the same time as `.collection` is direct access to their classes.

src/modules/trash/components/DestroyConfirm.tsx
cballevre: Good point, I've added alerts in case of success or error

src/modules/views/Nextcloud/NextcloudDestroyView.tsx
cballevre: The UI side is handle inside the `DestroyConfirm` because the message are the same between `io.cozy.files` and `io.cozy.remote.nextcloud.files`

acezard (approved):

PR #3196, opened 2024-07-22
[VO-924] fix: Reset the query to update the view even if it doesn't exist

cballevre:
**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1503

### DIFF
--- a/docs/nextcloud.md
+++ b/docs/nextcloud.md
@@ -1,0 +1,13 @@
+# Nextcloud
+
+The integration of Nextcloud within cozy-drive relies heavily on cozy-client and the proxy made by cozy-stack ([doc](https://docs.cozy.io/en/cozy-stack/nextcloud/)). This implies 2 main constraints that are worth mentioning if you want to understand the code better.
+
+**1. Obtain a folder itself**
+
+The query to `io.cozy.remote.nextcloud.files` can only retrieve the contents of one folder. To avoid this problem, we query its parent and filter by id to get data about.
+
+**2. Reload data after mutation**
+
+The mutations have any effect on the store because there are not back by real-time or update with the request response like for `io.cozy.files`. To avoid this problem, we reset the affected queries with `client.reset(queryId)`.
+
+When the query involves moving a file, the destination query is privileged. The cache of the source query will be updated when cozy-client receives the reset query return. This avoids an additional network request. If the query does not exist, then we reset the source query.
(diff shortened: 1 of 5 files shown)


docs/nextcloud.md
```diff
@@ -0,0 +1,13 @@
...
+
+The integration of Nextcloud within cozy-drive relies heavily on cozy-client and the proxy made by cozy-stack ([doc](https://docs.cozy.io/en/cozy-stack/nextcloud/)). This implies 2 main constraints that are worth mentioning if you want to understand the code better.
+
+**1. Obtain a folder itself**
+
+The query to `io.cozy.remote.nextcloud.files` can only retrieve the contents of one folder. To avoid this problem, we query its parent and filter by id to get data about.
+
+**2. Reload data after mutation**
+
+The mutations have any effect on the store because there are not back by real-time or update with the request response like for `io.cozy.files`. To avoid this problem, we reset the affected queries with `client.reset(queryId)`.
```
Ldoppea, line 11:
```suggestion
The mutations doesn't have any effect on the local store because we cannot update it with the request's response like for `io.cozy.files` and there are no real-time event that would update. To avoid this problem, we reset the affected queries with `client.reset(queryId)`.
```

docs/nextcloud.md
```diff
@@ -0,0 +1,13 @@
...
+
+When the query involves moving a file, the destination query is privileged. The cache of the source query will be updated when cozy-client receives the reset query return. This avoids an additional network request. If the query does not exist, then we reset the source query.
```
Ldoppea, line 13:
```suggestion
When the query involves moving a file, the destination query is privileged. The cache of the source query will be updated when cozy-client receives the reset query answer. This avoids an additional network request. If the query does not exist, then we reset the source query.
```

src/modules/nextcloud/components/actions/restoreNextcloudFile.tsx
```diff
@@ -33,27 +33,46 @@ export const restoreNextcloudFile = ({
...
+        )
+
+        // If the query for the folder containing the restored files does not exist,
+        // we need to reset the query of the current folder to refresh the view
+        if (resetResults.some(query => query === null)) {
+          const queryId =
+            computeNextcloudFolderQueryId({
+              sourceAccount,
+              path: parentPath
+            }) + '/trashed'
```
Ldoppea, line 72: This parts needs more comment to explain why the `/trashed` path is added here but not in the previous call on line 55

Ldoppea (approved):

PR #3237, opened 2024-10-22
[VO-942] feat: Open File in New Tab (Cmd+Click or Right-Click)

cballevre:
```
### ✨ Features

* Open File in New Tab (Cmd+Click or Right-Click)

### 🔧 Tech
 
* Update cozy-ui from 111.12.0 to 111.21.0
```

By creating this new feature, I've completely redefined the way we trigger navigation in Drive. Now we
 rely heavily on `react-router-dom` and its relative route mechanism. We're also benefiting from advances in the amiral application, which is capable of interpreting open links and redirecting them to the right applications. This allows us to make the code generic while adapting to the opening context. For example, to open a folder in the Sharing tab : 
- Before, we would share callback function to file so it could navigate to folder with the prefix `sharing`
- After, we use a relative route `../${file.id}` so react-router-dom can replace the current folder with the new id
This limit the props drilling and reduce code. 

The main commit is d70ba45b218e7b0176d393ccded6143c5df81fda. The commit before are the preparation and adaptation of code after. Some noteworthy changes are : 
- The public viewer to open file from folder is in separate route 
- The breadcrumb is also using relative route
- `useFileLink` is used to open favorite inside the navigation sidebar

**Remaining works:**
- ~~Test all navigation cases~~
- ~~Make unit tests on `useFileLink` to avoid repetitive testing~~
- ~~Add more extensive message to some commit~~


src/modules/navigation/hooks/helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  isOfficeEnabled?: boolean
+}
+
+interface ComputePathOptions {
+  type: string
+  pathname: string
+  isPublic: boolean
+}
+
+export const computeFileType = (
```
cballevre, line 23: @paultranvan I'd like to take this opportunity to point you to this function where you'll find a good number of paths to files on drive

paultranvan: Nice! This might be worth moving it to cozy-client file's models, as we might have the use elsewhere

paultranvan:
Behold the mighty long-expected feature! :heart:

zatteo (approved):
Looks very nice and clean. Did you try on the flagship app ?

cballevre:
@zatteo I did a quick test of the notes on the flagship app. It opens the Notes app in a new webview and when you come back it returns to Drive. So nice that flagship app interprets opened link so we can remove custom code inside others applications

PR #3248, opened 2024-11-19
fix: Hide back icon in public viewer

Merkur39:
In the Public view, no return is possible, as there is none.


cballevre (approved):
LGTM 👍

PR #3250, opened 2024-11-21
feat: Show share synchronization as a minor action (Part 1)

Merkur39:
Multiple updates, to facilitate future functionalities!
- Replacement of the latest ActionMenu
- Updated cozy-ui & cozy-harvest-lib, and added cozy-viewer
- Various adjustments


src/modules/views/OnlyOffice/Error.jsx
```diff
@@ -3,12 +3,13 @@ import { RemoveScroll } from 'react-remove-scroll'
...
-import ForwardOrDownloadButton from 'cozy-ui/transpiled/react/Viewer/Footer/ForwardOrDownloadButton'
-import SharingButton from 'cozy-ui/transpiled/react/Viewer/Footer/Sharing'
-import ToolbarButtons from 'cozy-ui/transpiled/react/Viewer/components/ToolbarButtons'
 import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
+import Viewer, {
+  FooterActionButtons,
+  ForwardOrDownloadButton,
+  ToolbarButtons
+} from 'cozy-viewer'
+import SharingButton from 'cozy-viewer/dist/Footer/Sharing'
```
cballevre, line 12: Same here import it from root 🙏

src/modules/drive/Toolbar/selectable/SelectableItem.jsx
```diff
@@ -9,11 +9,11 @@ import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
 /**
  * Action to show the selection bar
  */
-const SelectableItem = ({ showSelectionBar }) => {
+const SelectableItem = ({ onClick }) => {
   const { t } = useI18n()
 
   return (
-    <ActionMenuItem
+    <ActionMenuItem left={<Icon icon={CheckSquareIcon} />} onClick={onClick}>
```
cballevre: Maybe there is a problem here the props are only duplicated 🤔

src/modules/drive/Toolbar/delete/DeleteItem.jsx
```diff
@@ -9,32 +9,20 @@ import { translate } from 'cozy-ui/transpiled/react/providers/I18n'
...
-      data-testid="fil-action-delete"
-      left={<Icon icon={TrashIcon} color="var(--errorColor)" />}
-      onClick={() =>
-        onLeave(displayedFolder).then(() => trashFolder(displayedFolder))
-      }
-    >
-      <span className="u-pomegranate">{t('toolbar.leave')}</span>
-    </ActionMenuItem>
-  ) : (
+const DeleteItem = ({ t, isSharedWithMe, trashFolder, displayedFolder }) => {
```
cballevre, line 14: I love this clearer refactor 💜

src/modules/actions/index.jsx
```diff
@@ -67,13 +67,19 @@ export const hr = () => {
...
+  hasWriteAccess,
+  refresh,
+  byDocId,
+  isOwner
+}) => {
   const icon = TrashIcon
 
   return {
     name: 'trash',
-    label,
```
cballevre, line 76: I think you should keep the `label` has this action menu is also used into the SelectionBar

src/modules/views/Drive/DriveFolderView.jsx
```diff
@@ -156,7 +157,9 @@ const DriveFolderView = () => {
     canMove: true,
     isPublic: false,
     allLoaded,
-    showAlert
+    showAlert,
```
cballevre, line 160: This props should also be added inside `PublicFolderView`, `RecentView` and `SharingsFolderView` because they are using also the action `trash`

src/modules/drive/Toolbar/components/AddEncryptedFolderItem.jsx
```diff
@@ -2,24 +2,30 @@ import React from 'react'
...
-    <ActionMenuItem
+    <ActionsMenuItem
       data-testid="add-encrypted-folder-link"
-      onClick={() => showUnlockForm({ onUnlock: addEncryptedFolder })}
-      left={<EncryptedFolderIcon width={16} height={16} />}
+      onClick={handleClick}
     >
-      {t('toolbar.menu_new_encrypted_folder')}
-    </ActionMenuItem>
+      <EncryptedFolderIcon width={16} height={16} />
```
cballevre, line 26: Shouldn't it be wrapped in a component `ListItemIcon`?

src/modules/drive/AddMenu/AddMenu.jsx
```diff
@@ -23,25 +25,27 @@ export const ActionMenuContent = ({
...
-        </>
+      <ActionsMenuMobileHeader>
+        <ListItemText
+          primary={t('toolbar.menu_add')}
+          primaryTypographyProps={{ variant: 'h6' }}
+        />
+      </ActionsMenuMobileHeader>
+
+      {canCreateFolder && !isEncryptedFolder && (
+        <AddFolderItem onClick={onClick} />
```
cballevre, line 45: I would rather prefer something more explicit than a `onClick` because its always the same action like `onClose` or `onDismiss`

src/modules/actions/index.jsx
Merkur39: The `label` property no longer seems to be used in either `ActionsMenu` or `ActionsBar`.

src/modules/drive/AddMenu/AddMenu.jsx
Merkur39: Seen together, this is indeed generic, but necessary for the proper functioning of the overload below in the `PublicToolbarMoreMenu.jsx` file.

src/modules/drive/Toolbar/components/AddEncryptedFolderItem.jsx
Merkur39: Good catch!

cballevre (approved):
LGTM 👍

PR #3253, opened 2024-11-26
feat: Hide sharing banner for the synchronize action 

Merkur39:
We no longer wish to highlight this action, but it remains in the 3-point menu.


cballevre (approved):
LGTM 👍

PR #3283, opened 2024-12-24
feat: Files doctype should appear first in search

paultranvan:
When searching through the dataproxy, the default doctype search order is apps, contacts, files, mainly because of their respective volume. It is assumed to have very few apps, and less contacts than files. Therefore, since there is a limit to displayed results, we prefer to display files in last.
This works in the Home search contexte, but not so much in Drive, where it makes sense to have the files displayed first.


paultranvan:
It's christmas: no need to review! :santa:

PR #3295, opened 2025-01-15
Remove hammerjs

zatteo:
Remove the hammerjs dep which was used only for one behavior. Not necessary to add two packages which what it implies for this behavior that can easily be managed manually.

To do : 
- [x] test on Chrome
- [x] test on Firefox
- [x] test on Safari
- [x] test on iOS AA
- [x] test on iOS mobile
- [x] test on Android AA
- [x] test on Android mobile


src/modules/filelist/FileOpener.jsx
```diff
@@ -26,103 +18,76 @@ export const getParentLink = element => {
...
+  // Check if the clicked element is an input to avoid disturbing events while renaming
   if (['INPUT', 'BUTTON'].indexOf(ev.target.nodeName) !== -1) {
     return false
   }
 
-  const parentDiv = getParentDiv(ev.target)
-  // remove event when it's the checkbox or the more button
-  if (
-    parentDiv.className.indexOf(styles['fil-content-file-select']) !== -1 ||
-    parentDiv.className.indexOf(styles['fil-content-file-action']) !== -1
```
zatteo, line 38: This condition is not necessary anymore. These buttons does not trigger the event at all.

JF-Cozy (approved):
👏

PR #3320, opened 2025-02-17
Remove old search for mobile

paultranvan:
This is not useful anymore, as the new search is handled by the
dataproxy in a much more efficient way.
Note that it should remove some unpleasant VFS lock that we noticed on instances with many files


shepilov:
@paultranvan why it was not merged? is there any blockers?

PR #3345, opened 2025-04-01
refactor(UI): Replace trash icon with custom SVG asset

lenhanphung:
Replace the default trash icon from cozy-ui with a custom SVG asset in the Empty trash component to improve visual appearance.

This change improves the user experience when viewing empty trash folder.


src/components/Error/Empty.jsx
```diff
@@ -5,41 +5,44 @@ import Empty from 'cozy-ui/transpiled/react/Empty'
...
-        icon: (
-          <div className="u-w-100">
-            <Icon icon={FolderEmptyIllu} size={160} />
-          </div>
-        ),
-        iconSize: isMobile ? 'large' : 'medium'
-      }
+
+  let iconToShow = EmptyIcon[type]
+  if (isTwakeTheme() && type !== 'trash') {
```
JF-Cozy, line 30: if we can use only `isTwakeTheme()` in the condition, it will be easier to simplify the code when we no longer need this condition (and then have to remove uesless code in few weeks)

JF-Cozy (approved):

src/components/Error/Empty.jsx
lenhanphung: updated

src/components/Error/Empty.jsx
```diff
@@ -27,7 +27,7 @@ const EmptyCanvas = ({ type, canUpload, localeKey, hasTextMobileVersion }) => {
   const { isMobile } = useBreakpoints()
 
   let iconToShow = EmptyIcon[type]
-  if (isTwakeTheme() && type !== 'trash') {
+  if (isTwakeTheme()) {
```
JF-Cozy, line 30: I meant to use `isTwakeTheme()` only **somehow** 🤔 so be changing the code if necessary... we have the same result here than before?

PR #3353, opened 2025-04-04
Fix/add button behavior

lenhanphung:
Update the create button:
- Show create button on all folder screens
- When in virtual folders (Favorites, Recent, etc.), allow creating folders and uploading files to root Files folder
- Implement NavContext to manage this behavior across different views( Desktop and Mobile)

### DIFF
--- a/src/modules/drive/FabWithAddMenuContext.jsx
+++ b/src/modules/drive/FabWithAddMenuContext.jsx
@@ -20,7 +20,7 @@ const FabWithAddMenuContext = ({ noSidebar }) => {
   } = useContext(AddMenuContext)
 
   const styles = useFabStyles({
-    bottom: noSidebar ? '1rem' : 'calc(var(--sidebarHeight) + 2rem)'
+    bottom: noSidebar ? '4rem' : 'calc(var(--sidebarHeight) + 2rem)'
   })
 
   return (
--- a/src/modules/layout/Layout.jsx
+++ b/src/modules/layout/Layout.jsx
@@ -1,11 +1,11 @@
-import React from 'react'
-import { Outlet } from 'react-router-dom'
+import React, { useEffect } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import { Outlet, useNavigate } from 'react-router-dom'
 
 import { BarComponent } from 'cozy-bar'
 import CozyDevtools from 'cozy-devtools'
 import flag from 'cozy-flags'
 import FlagSwitcher from 'cozy-flags/dist/FlagSwitcher'
-import { SharedDocument } from 'cozy-sharing'
 import Sprite from 'cozy-ui/transpiled/react/Icon/Sprite'
 import { Layout as LayoutUI } from 'cozy-ui/transpiled/react/Layout'
 import Sidebar from 'cozy-ui/transpiled/react/Sidebar'
@@ -16,21 +16,36 @@ import StorageButton from '@/components/Storage/StorageButton'
 import StorageProgress from '@/components/Storage/StorageProgress'
 import ButtonClient from '@/components/pushClient/Button'
 import SupportUs from '@/components/pushClient/SupportUs'
-import { useDisplayedFolder, useCurrentFolderId } from '@/hooks'
+import { ROOT_DIR_ID } from '@/constants/config'
+import { useDisplayedFolder } from '@/hooks'
 import { initFlags } from '@/lib/flags'
 import AddMenuProvider from '@/modules/drive/AddMenu/AddMenuProvider'
 import AddButton from '@/modules/drive/Toolbar/components/AddButton'
 import Nav from '@/modules/navigation/Nav'
+import {
+  wasOperationRedirected,
+  RESET_OPERATION_REDIRECTED
+} from '@/modules/navigation/duck/reducer'
 import { SelectionProvider } from '@/modules/selection/SelectionProvider'
 import UploadQueue from '@/modules/upload/UploadQueue'
 
 initFlags()
 
 const Layout = () => {
+  const navigate = useNavigate()
+  const dispatch = useDispatch()
   const { isMobile } = useBreakpoints()
-  const folderId = useCurrentFolderId()
   const { displayedFolder } = useDisplayedFolder()
   const { isDesktop } = useBreakpoints()
+
+  const shouldRedirect = useSelector(wasOperationRedirected)
+
+  useEffect(() => {
+    if (shouldRedirect) {
+      navigate(`/folder/${ROOT_DIR_ID}`)
+      dispatch({ type: RESET_OPERATION_REDIRECTED })
+    }
+  }, [shouldRedirect, navigate, dispatch])
 
   return (
     <LayoutUI>
@@ -41,25 +56,16 @@ const Layout = () => {
       <FlagSwitcher />
       <Sidebar className="u-flex-justify-between">
         <div>
-          {folderId && isDesktop ? (
-            <SharedDocument docId={folderId}>
-              {sharingProps => {
-                const { hasWriteAccess } = sharingProps
-                return (
-                  hasWriteAccess && (
-                    <AddMenuProvider
-                      canCreateFolder={true}
-                      canUpload={true}
-                      disabled={false}
-                      displayedFolder={displayedFolder}
-                      isSelectionBarVisible={false}
-                    >
-                      <AddButton className="u-mh-1-half u-mt-1-half u-miw-4 u-bdrs-6" />
-                    </AddMenuProvider>
-                  )
-                )
-              }}
-            </SharedDocument>
+          {isDesktop ? (
+            <AddMenuProvider
+              canCreateFolder={true}
+              canUpload={true}
+              disabled={false}
+              displayedFolder={displayedFolder}
+              isSelectionBarVisible={false}
+            >
+              <AddButton className="u-mh-1-half u-mt-1-half u-miw-4 u-bdrs-6" />
+            </AddMenuProvider>
           ) : null}
           <Nav />
         </div>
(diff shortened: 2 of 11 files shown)


src/modules/navigation/duck/reducer.js
```diff
@@ -24,4 +41,8 @@ export default combineReducers({
  * @param {Object} state - The state object containing the view property.
  * @returns {string} The sort value.
  */
-export const getSort = ({ view }) => view.sort
+// Selector needs to point to the correct state slice (`view`)
+export const getSort = state => state.view.sort
+
+// Selector for the new state (`view`)
```
JF-Cozy, line 47: We should probably change this comment, the new state is new now, but won't be tomorrow

src/modules/navigation/duck/reducer.js
```diff
@@ -14,8 +18,21 @@ const sort = (state = null, action) => {
...
+      return false
+    default:
+      return state
+  }
+}
+
 export default combineReducers({
-  sort
+  sort,
+  operationRedirected: operationRedirectedReducer // Add the new reducer
```
JF-Cozy, line 35: We should probably change this comment, the new state is new now, but won't be tomorrow

src/modules/navigation/duck/actions.jsx
```diff
@@ -185,7 +211,7 @@ const uploadQueueProcessed =
  * negatives.
  */
 const doesFolderExistByName = (state, parentFolderId, name) => {
-  const filesInCurrentView = getFolderContent(state, parentFolderId) || [] // TODO in the public view we don't use a query, so getFolderContent returns null. We could look inside the cozy-client store with a predicate to find folders with a matching dir_id.
+  const filesInCurrentView = getFolderContent(state, parentFolderId) || []
```
JF-Cozy, line 237: why removing the comment, is it fixed?

src/modules/drive/FabWithAddMenuContext.jsx
```diff
@@ -20,7 +20,7 @@ const FabWithAddMenuContext = ({ noSidebar }) => {
   } = useContext(AddMenuContext)
 
   const styles = useFabStyles({
-    bottom: noSidebar ? '1rem' : 'calc(var(--sidebarHeight) + 2rem)'
+    bottom: noSidebar ? '4rem' : 'calc(var(--sidebarHeight) + 2rem)'
```
JF-Cozy, line 23: is it the right height on Cozy and Twake theme?

src/modules/views/Favorites/FavoritesView.tsx
```diff
@@ -140,6 +144,21 @@ const FavoritesView: FC = () => {
...
+      {isMobile && (
+        <AddMenuProvider
+          canCreateFolder={true}
+          canUpload={true}
+          disabled={false}
+          displayedFolder={null}
+          isSelectionBarVisible={isSelectionBarVisible}
+          isPublic={false}
+          refreshFolderContent={(): void => {
+            // No action needed in virtual folders
```
JF-Cozy, line 156: `refreshFolderContent` should not be present though, isn't it? Is it required?

src/modules/navigation/duck/actions.jsx
```diff
@@ -189,17 +211,18 @@ const uploadQueueProcessed =
...
-      errors.length === 0 &&
-      networkErrors.length === 0 &&
-      quotas.length === 0 &&
-      fileTooLargeErrors.length === 0
-    if (
-      navigateAfterUpload &&
-      isSuccess &&
-      (created.length > 0 || updated.length > 0)
-    ) {
+    // Check if there are any successful uploads (created or updated files)
```
JF-Cozy, line 214: we can remove the comment, the var name is self-explanatory 👍

src/modules/navigation/duck/actions.jsx
```diff
@@ -189,17 +211,18 @@ const uploadQueueProcessed =
...
+    const hasSuccessfulUploads = created.length > 0 || updated.length > 0
+
+    // If we need to navigate after upload and there were successful uploads, redirect
```
JF-Cozy, line 217: same here

src/modules/navigation/NavContext.jsx
JF-Cozy, line 1: you can remove the comment, if they say the same thing as the code, we don't need them :)

src/modules/layout/Layout.jsx
```diff
@@ -31,21 +32,24 @@ import UploadQueue from '@/modules/upload/UploadQueue'
...
   const { displayedFolder } = useDisplayedFolder()
   const { isDesktop } = useBreakpoints()
 
   const shouldRedirect = useSelector(wasOperationRedirected)
+  const [, setLastClicked] = useNavContext()
 
   useEffect(() => {
     if (shouldRedirect) {
+      // Update lastClicked state to ensure sidebar shows the correct active item
+      setLastClicked(`/folder/${ROOT_DIR_ID}`)
```
JF-Cozy, line 48: is it working on every cases? We used `setLastClicked` on different places in the code, specifically, we are sure there is no conflict with them, or double trigger?

src/modules/layout/Layout.jsx
```diff
@@ -41,25 +56,16 @@ const Layout = () => {
...
-                      isSelectionBarVisible={false}
-                    >
-                      <AddButton className="u-mh-1-half u-mt-1-half u-miw-4 u-bdrs-6" />
-                    </AddMenuProvider>
-                  )
-                )
-              }}
-            </SharedDocument>
+          {isDesktop ? (
+            <AddMenuProvider
```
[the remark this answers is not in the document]
zatteo: Because even if you are in a directory where do not have write access, you can create a file with this PR, and in this case it will be created in root folder. So no need to know if you have write access you can always create a file.

src/modules/navigation/duck/actions.jsx
```diff
@@ -217,26 +256,41 @@ export const createFolder = (
...
+        dispatch(operationRedirected())
       }
     } catch (err) {
       if (err.response && err.response.status === HTTP_CODE_CONFLICT) {
         showAlert({
           message: t('alert.folder_name', { folderName: name }),
           severity: 'error'
         })
-      } else {
+      } else if (!err.message?.includes('Cannot create encrypted folder')) {
```
zatteo, line 316: I do not understand why you need this condition. If you do not already call `showAlert` line 277, you can remove the condition no?

zatteo:
Did you try in sharing context? It is a important thing to do regularly when working on layout and core elements in cozy-drive. For example : 
- with sharing link : in Cozy A, create a folder, share it by link with read only permissions, open the in private window
- with cozy to cozy sharing : in Cozy A, create a folder, share it to Cozy B with read only permissions, in Cozy B check the shared folder

Sometime also to test with read write permissions. I don't say that you need to try every case here but it is an important thing to have in mind.

But looks good overall!

src/modules/layout/Layout.jsx
```diff
@@ -41,25 +56,16 @@ const Layout = () => {
...
+              canCreateFolder={true}
+              canUpload={true}
+              disabled={false}
+              displayedFolder={displayedFolder}
+              isSelectionBarVisible={false}
+            >
+              <AddButton className="u-mh-1-half u-mt-1-half u-miw-4 u-bdrs-6" />
```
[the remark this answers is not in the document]
lenhanphung: yes. updated

src/modules/layout/Layout.jsx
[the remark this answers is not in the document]
lenhanphung: thanks @zatteo :)

src/modules/views/Favorites/FavoritesView.tsx
lenhanphung: you're correct that FavoritesView itself, being a virtual folder driven by a query, doesn't require a manual refresh action after creating items.
the refreshFolderContent prop is necessary here because the `AddMenuProvider` component passes this prop down to its children, specifically `ActionMenuContent`. `ActionMenuContent`.
let me know if this makes sense

src/modules/navigation/duck/actions.jsx
lenhanphung: yes

src/modules/navigation/NavContext.jsx
lenhanphung: yes. updated

src/modules/layout/Layout.jsx
lenhanphung: setLastClicked call specifically handles the scenario where the user is automatically redirected to the root folder after certain operations (like move or copy, managed by the `shouldRedirect` flag). Its purpose is to ensure the sidebar's active item correctly reflects this programmatic navigation to the root, synchronizing the UI with the current location.
The other usage  is triggered by direct user clicks on navigation items. I think it wont be conflict.

src/modules/navigation/duck/actions.jsx
lenhanphung: This condition was originally there to prevent showing the generic error alert (alert.folder_generic) if the specific 'Cannot create encrypted folder...' error occurred, as that error already triggered the same alert inside the try block before being caught, I also refactor it.

src/modules/drive/FabWithAddMenuContext.jsx
JF-Cozy: Nevermind, we don't maintain Cozy theme anymore, you just can make changes for Twake theme purposes 👍

src/modules/views/Favorites/FavoritesView.tsx
JF-Cozy: Ok so having `refreshFolderContent` undefined will throw an error, where `refreshFolderContent={() => {}}` will not?

JF-Cozy:
@zatteo My review is ✅ , I let you validate when you're ok too.

src/helpers/permissions.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Checks if a folder is a shared folder for which the current user has read-only access.
+ * @param {object} folder - The folder object (presumably io.cozy.files type).
+ * @returns {boolean} True if the folder is shared and the user cannot write to it, false otherwise.
+ */
+export const isReadOnlySharedFolder = folder => {
```
zatteo, line 6: You do not use this methods ?

lenhanphung: yes. Deleted

zatteo (approved):

PR #3366, opened 2025-04-23
Update wording to Twake

zatteo:


public/screenshots/fr/screenshot04.png
Ldoppea, line 1: Seems like you put the wrong file here, it is the same as screenshot03

zatteo: Nice catch, fix!

PR #3370, opened 2025-05-07
Replace useFileWithPath with simple query

JF-Cozy:
`useFileWithPath` and `useFilesQueryWithPath` were hooks that allowed, by querying the parent folder via the file's `dir_id`, to retrieve the parent's path and add it to an attribute called `file.displayedFolder`. We wanted to modify them because there is the queryDef `buildParentsByIdsQuery([])` in them, which is suboptimal and will request all docs (because empty array passed to getByIds query).

Here, we rely on a simple query with a `where` clause that allows, through the stack, to directly retrieve a file with its path. From there, we create a simple method based on this path to determine the parent's path. Additionally, we avoid adding a 'virtual' attribute (`file.displayedFolder`) that is not on io.cozy.files.


src/modules/filelist/helpers.ts
```diff
@@ -20,22 +24,22 @@ export const getFileNameAndExtension = (
...
     }
   }
 
-  if (file._id === 'io.cozy.files.shared-drives-dir') {
+  if (file._id === SHARED_DRIVES_DIR_ID || file._id === ROOT_DIR_ID) {
     return {
       title: t('FileName.sharedDrive'),
       filename: t('FileName.sharedDrive')
     }
   }
```
zatteo, line 39: Why if `file._id === ROOT_DIR_ID` we return this ?

zatteo (approved):
Good work to remove the `file.displayedFolder` that was non standard.

src/modules/views/OnlyOffice/Toolbar/index.jsx
```diff
@@ -37,6 +37,9 @@ const Toolbar = ({ sharingInfos }) => {
 
   const fileQuery = buildFileWhereByIdQuery(fileId)
   const { data: files } = useQuery(fileQuery.definition, fileQuery.options)
+  // console.info(' ')
+  // console.info('files :', JSON.stringify(files))
+  // console.info(' ')
```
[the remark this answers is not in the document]
JF-Cozy: oops

src/modules/filelist/helpers.ts
JF-Cozy: yes good question. Because in sharing view, with a hard refresh, we fetch root folder too so we have to do the same exception as for shared_drives_dir_id

zatteo: Ok its clearly not self explanatory

src/queries/index.ts
```diff
@@ -142,6 +142,10 @@ export const buildSharingsQuery: QueryBuilder<buildSharingsQueryParams> = ({
   definition: () =>
     Q('io.cozy.files')
       .getByIds(ids)
+      .partialIndex({
+        trashed: false,
+        dir_id: { $ne: TRASH_DIR_ID }
+      })
```
paultranvan, line 148: Excellent use of partial index 👍

PR #3371, opened 2025-05-12
fix: Upgrade cozy-sharing to visualize sharings with pouch

paultranvan:
When using PouchDB, displaying sharing was crashing because we were relying on JSON-API `attributes`, which does not exist with PouchDB. So we use this fix: https://github.com/cozy/cozy-libs/pull/2771


paultranvan:
Upgrade already done elsewhere

PR #3379, opened 2025-05-20
fix: Favorith path not working on mobile

zatteo:
To support routing in tabs that are not based on a folder (for example /sharings, /recent, /favorites), we used to duplicate every routes so that we can browse files and folders in /sharings or /recent.

Here, we take advantage of /favorites routes to be broken to try another approach. When in the /favorites tab, we set every link to default folders (folder/:id) and files (folder/:id/file/:id) links. So we do not browse files and folders in /favorites/xxx route anymore.

When clicking on a file or a folder :
- browser back button goes back to /favorites
- Drive navigation back button goes to folder higher in folder hierarchy


src/modules/navigation/hooks/helpers.ts
```diff
@@ -100,6 +100,11 @@ export const computePath = (
     case 'shortcut':
       return `/external/${file._id}`
     case 'directory':
+      // If we are in /favorites tab, we do not want it to appears in computed path
```
JF-Cozy, line 103: `On mobile, if we are in /favorites tab...` I think will be better to understand next time what we are talking about here.

JF-Cozy (approved):

src/modules/navigation/hooks/helpers.ts
zatteo: Updated

PR #3395, opened 2025-06-11
feat: Add basic Docs support

paultranvan:
Add support to Docs notes: https://github.com/suitenumerique/docs

This allows to:
- Create a Docs note, that will make a redirection to the Docs app
- Display the icon of a Docs note when listed in a directory

A Docs note is represented as a io.cozy.files, with a special extension,
`.docs-note`, and its Docs id, stored in `metadata.externalId`.
The mimeType is `text/markdown` as it is indeed a markdown file,
exported from Docs, but the extension is not `.md` to avoid any local
edition in an external editor, that would be eventually overwritten
by Docs.

A flag `drive-lasuitedocs.enabled` is required to enable this feature.


src/modules/drive/Toolbar/components/CreateDocsItem.jsx
```diff
@@ -0,0 +1,62 @@
...
+      subDomainType: isFlatDomain ? 'flat' : 'nested',
+      pathname: '',
+      hash: `/bridge/docs/new/${_displayedFolder._id}`
+    })
+
+    window.location.href = url
+  }
+
+  return (
+    <ActionsMenuItem data-testid="create-a-note" onClick={handleClick}>
```
zatteo: nit: data-testid is not correct (should be removed)

src/modules/drive/Toolbar/components/CreateDocsItem.jsx
```diff
@@ -0,0 +1,62 @@
...
+      <ListItemIcon>
+        <Icon icon={IconDocs} />
+      </ListItemIcon>
+      <ListItemText primary={t('toolbar.menu_create_docs')} />
+    </ActionsMenuItem>
+  )
+}
+
+export default translate()(withClient(CreateDocsItem))
```
zatteo: nit: Better to use useClient and useI18n instead of these wrappers (I know its copy paste)

src/modules/navigation/hooks/helpers.ts
```diff
@@ -68,6 +75,8 @@ export const computeApp = (type: string): string => {
     case 'note':
     case 'public-note-same-instance':
       return 'notes'
+    case 'docs':
+      return 'docs'
```
zatteo, line 84: What is this one ?

paultranvan: This function is used to determine the slug of the app to open the file. So the `return docs` is the slug, and the `case docs` is the value computed by the [computeFileType](https://github.com/cozy/cozy-drive/blob/88de004bfaf57d87c2d84f696d4b1c1a83053ff4/src/modules/navigation/hooks/helpers.ts#L26) method that returns a string describing the "type" of the io.cozy.files, i.e. if it's 'file', 'directory', 'onlyoffice', etc.
Here it is simply seen as "docs"

src/modules/drive/Toolbar/components/CreateDocsItem.jsx
paultranvan: Fair enough

src/modules/navigation/hooks/helpers.ts
```diff
@@ -23,6 +24,13 @@ interface ComputePathOptions {
   isPublic: boolean
 }
 
+const isDocs = (file: File): boolean => {
+  return (
+    file.name?.endsWith('.docs-note') &&
+    !!(file as IOCozyFile).metadata.externalId
+  )
+}
+
```
zatteo, line 34: You can now use the one from cozy-client

zatteo (approved):

PR #3396, opened 2025-06-12
fix(ErrorShare): Reduce icon height from 8rem to 2rem

lenhanphung:


src/components/Error/ErrorShare.jsx
JF-Cozy, line 1: did you try `icon={<Icon icon={CloudBrokenIcon} size="XXX" or height="XXX" />}`? Is it working? Otherwise ok for your approach 👍

lenhanphung: @JF-Cozy it works as expected. Im using componentProps because parent component is designed to handle icon props this way, but your way seems more flexible and direct to the way of rendering icons 🤔
I will keep your solution for same case in future

JF-Cozy: Ok, so I think when you have 5min without rush, you can modify and push without review :)

lenhanphung: next push will include this :D

PR #3404, opened 2025-06-18
Update shared drive creation UI

zatteo:
**Update UI following new Twake UI**
- New creation modal
- Shared drive button in "Create" menu
- Shared drive list in sidebar
- Some of UI updates are already outdated when seeing latest Figma, but it is a least consistent and it allows to play with shared drives 

**Foundation of shared drive logic**
- POC of browsing in a shared drive from the sidebar


src/modules/drive/AddMenu/AddMenuContent.jsx
```diff
@@ -63,6 +64,9 @@ const AddMenuContent = forwardRef(
         {canCreateFolder && !isEncryptedFolder && (
           <AddFolderItem onClick={onClick} isReadOnly={isReadOnly} />
         )}
+        {canCreateFolder && !isPublic && flag('drive.shared-drive.enabled') && (
```
JF-Cozy, line 67: nit: we probably should refactor this and use actions with ActionsMenu instead of "manual" ActionsMenuItems 🤔

PR #3412, opened 2025-07-01
Improve performances of mutliple updates

paultranvan:
When receiving a batch of realtime events, we used to treat them
sequentially and do the following:
- Query the parent folder if it's a file to compute the path
- Dispatch the mutation

This was highly inefficient, because the dispatch is a costful
operation, especially if the store is filled with documents and queries.
Indeed, each dispatch will trigger the full reevaluation of the existing
queries against the documents in the store.

So, to improve this behaviour we now introduce a debounce mechanism for
relatime events: we wait a fixed amount of time before processing the
events all at once.
To do this, each new realtime event is temporarily stored in a buffer,
and removed once treated.

This impact all treatments involving multiple realtime events, such as
trash, restore, or deletion from trash.

Note the debounce does not negatively affect single realtime events, as
we force the first event to be treated immediatly (and debounce only the
next ones if they occur less than 500ms after).

To show how impactful this is, we measured how long takes a folder to be
put in the trash. The folder contains 1K files.

- Before : ~61000 ms :turtle: 
- After : ~700 ms :rabbit2: 

=> This is a x87 improvements :rocket:

### DIFF
--- a/src/components/FilesRealTimeQueries.jsx
+++ b/src/components/FilesRealTimeQueries.jsx
@@ -1,3 +1,4 @@
+import debounce from 'lodash/debounce'
 import { memo, useEffect } from 'react'
 
 import { useClient, Mutations } from 'cozy-client'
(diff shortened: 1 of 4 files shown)


src/components/FilesRealTimeQueries.jsx
```diff
@@ -6,68 +7,109 @@ import { receiveMutationResult } from 'cozy-client/dist/store'
...
+        files.length > 1
+          ? Mutations.updateDocuments(files)
+          : Mutations.updateDocument(files[0])
+    }
+    if (mutationType === 'deleted') {
+      mutation =
+        files.length > 1
+          ? Mutations.deleteDocuments(files)
+          : Mutations.deleteDocument(files[0])
+    }
```
Crash--: I think I miss something, but: 
- bufferFiles is "global" 

So if I received an update for id: 1, and a delete for id: 2 (let's say at the same time), then in bufferFiles I'll have 1 and 2. 

And then, we'll call debouncedDispatchEvents twice. One with "updated", one with "deleted". 
And then, when we execute `debouncedDispatchEvents` we will see "updated" but with 1 and 2, and so we'll try to update 1 and 2, instead of just 1.

No?

paultranvan: You didn't miss anything, I simply forgot to deal with different buffers :face_in_clouds: 
I updated the commit, good catch

rezk2ll (approved):

PR #3413, opened 2025-07-02
fix: Add drag and drop in public sharing page :bug:

lethemanh:


src/modules/views/Public/PublicFolderView.jsx
```diff
@@ -176,70 +177,76 @@ const PublicFolderView = () => {
...
-      <Content className={isMobile ? '' : 'u-ml-1 u-pt-1'}>
-        <FolderViewHeader>
-          {currentFolderId && (
-            <>
-              {isOldBreadcrumb ? (
-                <OldFolderViewBreadcrumb
-                  currentFolderId={currentFolderId}
-                  getBreadcrumbPath={geTranslatedBreadcrumbPath}
+    <Dropzone
+      role="main"
```
JF-Cozy: `role` is not spread anymore, you can remove it. Why did you add it? (I'm asking because maybe you picked an old part of code as an example and we need to update it also)

JF-Cozy (commented):
It's more a `feat` than a `fix` in my opinion 🤔 

No need to use DropzoneDnD? I don't remember if we have dnd feature on public pages...

lethemanh:
> It's more a `feat` than a `fix` in my opinion 🤔
> 
> No need to use DropzoneDnD? I don't remember if we have dnd feature on public pages...

@JF-Cozy there is a ticket that require to add drag and drop in public page https://www.notion.so/linagora/Public-shared-page-add-drag-and-drop-22362718bad180e98f13d504d4f68868?source=copy_link

src/modules/views/Public/PublicFolderView.jsx
lethemanh: I picked from `src/modules/views/Sharings/SharingsFolderView.jsx`

JF-Cozy: ok, so you can remove the prop there as well (take a look to the code of Dropzone, it doesn't get any `role` prop)

JF-Cozy:
@lethemanh don't forget to modify `src/modules/views/Sharings/SharingsFolderView.jsx` please and then I think we are good to go

src/modules/views/Public/PublicFolderView.jsx
```diff
@@ -181,62 +182,67 @@ const PublicFolderView = () => {
       <ModalManager />
       {isSharingBannerPluginDisplayed && <SharingBannerPlugin />}
       <Content className={isMobile ? '' : 'u-ml-1 u-pt-1'}>
```
JF-Cozy, line 184: please take a deep look to Dropzone component and implementation elsewhere, <Content> is embedded into it :)

rezk2ll:
LGTM, but there is a small issue:

I just tested and the drag&drop works, just a small issue with the folder view not being updated after the upload is complete ( it refreshes if you upload via the create button ).
I think there is an issue with the `refreshFolderContent` not being triggered after the drag-and-drop action.


https://github.com/user-attachments/assets/2531d9e5-5ecd-4e0f-a5c0-e919c3cea4d6

lethemanh:
> LGTM, but there is a small issue:
> 
> I just tested and the drag&drop works, just a small issue with the folder view not being updated after the upload is complete ( it refreshes if you upload via the create button ). I think there is an issue with the `refreshFolderContent` not being triggered after the drag-and-drop action.
> 
> i made a small video:
> 
>  dragdrop.mp4 
> edit: after some checking, the dropzone component dispatches the uploadFiles call with fileUploadedCallback = () => null
> 
> and this approach relies on the cozy-realtime to handle the update and this won't happen because the publicly shared folder view can't subscribe to the events.
> 
> one possible solution is to update the Dropzone component to support this case:
> 
> ```ts
> export const Dropzone = ({
>   displayedFolder,
>   disabled,
>   refreshFolderContent = null,
>   children,
> }) => {
>   const client = useClient()
>   const { t } = useI18n()
>   const { isMobile } = useBreakpoints()
>   const { showAlert } = useAlert()
>   const sharingState = useSharingContext()
>   const vaultClient = useVaultClient()
>   const dispatch = useDispatch()
>   const fileUploadCallback = refreshFolderContent
>     ? refreshFolderContent
>     : () => null
> 
>   const onDrop = async (files, _, evt) => {
>     if (!canDrop(evt)) return
> 
>     const filesToUpload = canHandleFolders(evt) ? evt.dataTransfer.items : files
>     dispatch(
>       uploadFiles(
>         filesToUpload,
>         displayedFolder.id,
>         sharingState,
>         fileUploadCallback,
>         {
>           client,
>           vaultClient,
>           showAlert,
>           t,
>         }
>       )
>     )
>   }
> ```
> 
> and in the `src/modules/views/Public/PublicFolderView.jsx`
> 
> ```tsx
> <Dropzone
>   disabled={!hasWritePermissions}
>   displayedFolder={displayedFolder}
>   refreshFolderContent={refreshFolderContent}
> >
> ```

@rezk2ll I've updated

rezk2ll (approved):

PR #3416, opened 2025-07-03
Remove duplicate call to action & redirect StorageButton to premium link instead of hardcoded link

zatteo:


src/components/Storage/StorageButton.jsx
```diff
@@ -19,7 +31,8 @@ const StorageButton = ({ className }) => {
       size="small"
       height="auto"
       fullWidth
-      onClick={() => window.open(`https://cozy.io/${_lang}/pricing/`)}
+      component="a"
```
JF-Cozy, line 34: `target="_blank"` to keep same behavior maybe?

PR #3421, opened 2025-07-10
feat: Implement grid layout :sparkles:

lethemanh:
CleanShot 2025-07-10 at 14 53 13@2x
CleanShot 2025-07-10 at 14 55 03@2x


src/components/App/App.jsx
```diff
@@ -36,11 +37,13 @@ const Providers = ({ children }) => {
       <BarProvider>
         <PushBannerProvider>
           <AcceptingSharingProvider>
-            <ThumbnailSizeContextProvider>
-              <ModalContextProvider>
-                <DnDProvider {...dnDProviderProps}>{children}</DnDProvider>
-              </ModalContextProvider>
-            </ThumbnailSizeContextProvider>
+            <ViewSwitcherContextProvider>
+              <ThumbnailSizeContextProvider>
```
Crash--: Since you replace this thumbnail size, you should remove it from everywhere.

src/modules/filelist/File.jsx
```diff
@@ -84,6 +88,13 @@ const File = ({
     [styles['fil-content-row-bigger']]: thumbnailSizeBig
   })
 
+  const filContentColumnSelected = cx(styles['fil-content-column'], {
+    [styles['fil-content-column-selected']]: selected,
+    [styles['fil-content-column-actioned']]: actionMenuVisible,
+    [styles['fil-content-column-disabled']]: styleDisabled,
+    [styles['fil-content-column-bigger']]: thumbnailSizeBig
```
Crash--: can we still have `thumbnailSizeBig` ?

src/modules/filelist/File.jsx
```diff
@@ -107,93 +118,175 @@ const File = ({
...
+              file={attributes}
+              anchorElRef={filerowMenuToggleRef}
+              actions={actions}
+              onClose={hideActionMenu}
+            />
+          )}
+        </Card>
       )}
-    </TableRow>
+    </>
```
Crash--: do we really need to have this full copy / past between list & grid? Can't we make a few if? Because it seems that there is a lot of stuff that are just duplicated

src/modules/filelist/FileListHeaderMobile.jsx
```diff
@@ -76,11 +76,11 @@ const FileListHeaderMobile = ({
           <Button
             variant="text"
             onClick={() => {
-              toggleThumbnailSize()
+              toggleViewType(viewType === 'list' ? 'grid' : 'list')
```
Crash--: Why can't we just `toggleViewType()`? For me, toggle can only have two values and we change from one to another. So we should not test the previous value every time we want to change. This should be done within the provider.

Otherwise, rename this method to switchViewType and in that case you can do what you did, but I still prefer the first solution.

src/modules/filelist/cells/SelectBox.jsx
```diff
@@ -6,23 +6,30 @@ import { TableCell } from 'cozy-ui/transpiled/react/deprecated/Table'
...
-        onChange={() => {
-          // handled by onClick on the <TableCell>
-        }}
-      />
-    )}
-  </TableCell>
-)
+import { useViewSwitcherContext } from '@/lib/ViewSwitcherContext'
+
+const SelectBox = ({ withSelectionCheckbox, selected, onClick, disabled }) => {
```
Crash--: This component should be dump, and should not rely on the hook. Pass a `viewType` props directly.

src/modules/filelist/cells/ShareContent.jsx
```diff
@@ -33,7 +35,11 @@ const ShareContent = ({ file, disabled, isInSyncFromSharing }) => {
       })}
     >
       {isInSyncFromSharing || !isShared ? (
-        <span data-testid="fil-content-sharestatus--noAvatar">—</span>
+        viewType === 'list' ? (
+          <span data-testid="fil-content-sharestatus--noAvatar">—</span>
+        ) : (
+          <></>
```
Crash--: let's return `null` instead of an empty fragment

src/modules/filelist/icons/FileThumbnail.tsx
```diff
@@ -90,14 +93,22 @@ const FileThumbnail: React.FC<FileThumbnailProps> = ({
...
           height={size}
-          bgcolor="var(--contrastBackgroundColor)"
+          bgcolor={viewType === 'list' ? 'var(--contrastBackgroundColor)' : ''}
         >
           {isKonnectorFolder ? (
             <BadgeKonnector file={file}>
-              <FileIcon file={file} size={48} isEncrypted={isEncrypted} />
+              <FileIcon
+                file={file}
+                size={viewType === 'grid' ? 96 : 48}
```
Crash--: why don't you pass `96` when you call `FileThumbnail`? Like that, no need to rely on the hook.

src/styles/filelist.styl
```diff
@@ -27,6 +27,60 @@
...
+        .fil-content-file-select
+            opacity 1
+
+        .fil-content-file-action
+            opacity 1
+
+    .fil-content-file-select
+        position absolute
+        top -4px
+        left -3px
```
Crash--: what are those magic numbers? Are you sure it still works if you change your font size? Should we use `rem` or something here?

src/styles/filelist.styl
```diff
@@ -27,6 +27,60 @@
...
+        z-index 1
```
Crash--: Can you `@require 'settings/z-index.styl'` and then use the zIndex variable you need?

src/styles/filelist.styl
```diff
@@ -113,6 +167,24 @@ column-width-thumbnail-bigger = 7rem
     .fil-content-file-action
         margin-right 0
 
+.fil-content-grid-view.fil-content-cell
+    white-space normal
+    height auto
+    padding-right 0
+    margin-top 10px
```
Crash--: don't use px, you rem please

test/components/AppLike.jsx
```diff
@@ -68,23 +69,27 @@ const AppLike = ({
...
-                                  <Layout>{children}</Layout>
-                                </RightClickProvider>
-                              </FabProvider>
-                            </ModalContext.Provider>
-                          </PushBannerProvider>
-                        </AlertProvider>
-                      </BreakpointsProvider>
-                    </ThumbnailSizeContextProvider>
+                    <ViewSwitcherContextProvider>
+                      <ThumbnailSizeContextProvider>
```
Crash--: remove ThumbnailSizeContextProvider

Crash--:
Can you split your PR into at least 2 commits:
- One that add this "grid view" 
- One that add the "select all" 

Also, what about `FolderViewBodyVz`? You only change what is not "virtualized", but it seems that this is not what we'll be used anymore?

src/modules/views/Drive/DriveFolderView.jsx
```diff
@@ -165,10 +166,12 @@ const DriveFolderView = () => {
     byDocId,
     isMobile,
     isNativeFileSharingAvailable,
-    shareFilesNative
+    shareFilesNative,
+    selectAll: () => selectAll(allResults.map(query => query.data).flat())
   }
   const actions = makeActions(
     [
+      selectAllItems,
```
rezk2ll, line 174: Adding selectAllItems here will make it appear everywhere there are actions involved, for example, when you right-click on an item:

For example, here:
```tsx
  <RightClickFileMenu
    key={file._id}
    doc={file}
    actions={actions}
  >
```

also

```jsx
<SelectionBar actions={actions} />
```

and 
```jsx
  <File
      key={file._id}
      attributes={file}
      withSelectionCheckbox
      withFilePath={withFilePath}
      thumbnailSizeBig={isBigThumbnail}
      actions={actions}
      refreshFolderContent={refreshFolderContent}
```

The better approach can be:
- give a context to the displayCondition action method
- or to separate the selectionBar actions from the file right click menu + file actions ( duplicates actions etc )

lethemanh:
> Can you split your PR into at least 2 commits:
> 
> * One that add this "grid view"
> * One that add the "select all"
> 
> Also, what about `FolderViewBodyVz`? You only change what is not "virtualized", but it seems that this is not what we'll be used anymore?

@Crash-- the FolderViewBodyVz is using virtualized table from cozy-ui, I think this is a big change if we want to apply grid layout 
cc @rezk2ll

src/modules/views/Drive/DriveFolderView.jsx
lethemanh: @rezk2ll I've filtered the select all option from the prop of RightClickFileMenu and ActionMenuWithHeader

src/styles/filelist.styl
```diff
@@ -113,6 +168,24 @@ column-width-thumbnail-bigger = 7rem
...
+    .fil-file-filename-and-ext
+        display -webkit-box
+        -webkit-line-clamp 2
+        -webkit-box-orient vertical
+
+        overflow hidden
+        text-overflow ellipsis
+        white-space normal
+
+        max-width 96px
```
rezk2ll: This will unnecessarily truncate the file/folder names, and it's in px.

use max-width 100% instead

src/styles/folder-view.styl
```diff
@@ -0,0 +1,4 @@
+.fil-folder-body-grid
+    display grid
+    grid-template-columns repeat(auto-fill, minmax(140px, 1fr))
```
rezk2ll:
```suggestion
    grid-template-columns repeat(auto-fill, minmax(8.4rem, 1fr))
```

rezk2ll: In the Figma design, we have each item having a 136px width ( say 8.5rem ) and the row should have 12 items

image

If we use the current implementation values, we will end up with 11 items max per row

src/styles/filelist.styl
lethemanh: If use max-width 100%, we cannot limit the long text, so I changed to 6rem instead

rezk2ll (approved):

PR #3424, opened 2025-07-11
Adding CZ to index.js

amlercz:


zatteo:
Hi @amlercz, thank you for your contribution!

Any important reason for adding `cs` and `cs_CZ`? Is there huge differences?

Our i18n mecanism is not very efficient for the moment so we prefer staying small in number of locales.

PR #3430, opened 2025-07-18
feat: implement virtualized grid view :sparkles:

lethemanh:
requires: https://github.com/cozy/cozy-ui/pull/2822


src/modules/views/Folder/virtualized/FolderViewBody.jsx
```diff
@@ -141,32 +144,40 @@ const FolderViewBody = ({
...
+            currentFolderId={currentFolderId}
+            withFilePath={withFilePath}
+            actions={actions}
+          />
+        ) : (
+          <Grid
+            items={rows}
+            withFilePath={withFilePath}
+            actions={actions}
+            fetchMore={fetchMore}
```
rezk2ll, line 178: Does the "select all" in the bar work in the virtualized grid view if the Grid component does not require the `selectedItems` or `SelectAll`?
also the onDrop prop

lethemanh: The select box is inside `GridFile` component so we don't need `selectedItems` or `SelectAll` as the props of `Grid` component

rezk2ll (approved):

PR #3432, opened 2025-07-24
feat: Add support for offline navigation in web

paultranvan:
We now support offline navigation thanks to the `DataProxyLink`.  By
default, all request go through the `StackLink`, but it now detects if
the user is offline, and then forward data queries to the DataProxy,
which will run them on PouchDB.

Note this only work for read queries that rely on `io.cozy.files`: we
typically do not support offline download, or write operations such as
renaming yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "classnames": "2.3.1",
     "cozy-bar": "^22.1.3",
     "cozy-client": "^60.1.1",
-    "cozy-dataproxy-lib": "^4.1.0",
+    "cozy-dataproxy-lib": "^4.6.0",
     "cozy-device-helper": "^3.7.1",
     "cozy-devtools": "^1.2.1",
     "cozy-doctypes": "1.85.4",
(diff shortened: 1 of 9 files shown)


rezk2ll (dismissed):

paultranvan:
> Also, why was the DataProxyProvider removed from the main App.jsx ?

I specifically remember doing this with @Ldoppea but I have to admit I don't recall why we did that... 
I initially thought it was to move it below the CozyProvider, but it was actually already the case (I mixed up `Provider` and `Providers`...) , so my guess is that it actually changes nothing. 
If you're ok with that, I'll keep it like this

paultranvan:
> Did you try the app? And the search in search bar? I think DataProxyProvider must be before BarProvider

Yes I tested with no problem, including the search online and offline. And the `DataProxyProvider` is indeed before the BarProvider:
```
<DriveProvider> // include DataProxyProvider
  <Providers> // include BarProvider
...
```

PR #3433, opened 2025-07-29
feat: Change delete shared text :speech_balloon:

lethemanh:


src/modules/drive/Toolbar/delete/DeleteItem.jsx
```diff
@@ -16,14 +16,16 @@ const DeleteItem = ({ t, isSharedWithMe, trashFolder, displayedFolder }) => {
     trashFolder(displayedFolder)
   }
 
-  const label = isSharedWithMe ? t('toolbar.leave') : t('toolbar.trash')
+  const label = isSharedWithMe
+    ? t('toolbar.leave')
+    : t('toolbar.delete_shared_drive')
```
zatteo: We need to keep `t('toolbar.trash')` for everything that is not a sharing. So I think this change is "too much" no?

src/modules/drive/Toolbar/delete/DeleteItem.jsx
```diff
@@ -11,19 +11,30 @@ import { translate } from 'cozy-ui/transpiled/react/providers/I18n'
...
 const DeleteItem = ({ t, isSharedWithMe, trashFolder, displayedFolder }) => {
   const handleClick = () => {
     trashFolder(displayedFolder)
   }
 
-  const label = isSharedWithMe ? t('toolbar.leave') : t('toolbar.trash')
+  const isSharedDrive = useMemo(
+    () => displayedFolder.dir_id === SHARED_DRIVES_DIR_ID,
+    [displayedFolder]
+  )
```
zatteo: Why did you use `useMemo` ? It is necessary ? Looks good otherwise 👍

lethemanh: you're right, we don't need useMemo here

zatteo (approved):
🎉 Just remove totally the `useMemo` to pass linting.

PR #3435, opened 2025-07-31
feat: Display shared drive as tab in sharings section ::sparkles:

lethemanh:


src/styles/topbar.styl
```diff
@@ -20,3 +20,19 @@
...
+
+        .fil-tab-icon
+            path
+                fill var(--primaryColor)
+
+    .fil-tab-icon
+        margin-right .5rem
+
+        path
+            fill: var(--shadow7)
```
zatteo, line 34: When we can, we now try to use utilitary classes.
 
 So for example : 
-  `.u-ml-0` to set margin-left to 0
- maybe your `margin-right .5rem` can be replaced by `u-mr-1` or `u-mr-2`

About icon color, you can pass the color directly as a prop. Example : `<Icon icon={MyIcon} width={16} height={16} color="var(--primaryColor)"/>`

src/modules/views/Sharings/index.jsx
```diff
@@ -84,6 +88,22 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
   )
   const result = useQuery(query.definition, query.options)
 
+  const filteredResult = useMemo(
+    () => ({
+      ...result,
```
zatteo, line 97: Why do you need to keep `results` in this object?

src/modules/views/Sharings/index.jsx
```diff
@@ -84,6 +88,22 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
...
+      data:
+        tab === 1
+          ? result.data.filter(item => item.dir_id === SHARED_DRIVES_DIR_ID)
+          : result.data,
+      count:
+        tab === 1
```
zatteo: Can you find a way to make the condition more explicit, so avoiding `tab === 1` but something like `TAB === SHARED_DRIVE_TAB` or `isShareDriveTab(tab)`. Same for `const [tab, setTab] = useState(0)`.

src/modules/views/Sharings/index.jsx
lethemanh: In `FolderViewBody` component, it uses some fields in result so I think it's better to keep it

src/styles/topbar.styl
lethemanh: The color prop does not apply the color to correct element

zatteo: Can you send a picture of what is expected (Figma) and what is happening ? Maybe there is an issue with the icon or cozy-ui. But let's not wait this to merge.

zatteo (approved):
🎉 Can you just remove the package.json update before merging?

PR #3436, opened 2025-08-18
feat: Allow read-only actions in shared drive :sparkles:

lethemanh:
### Need:

- [cozy/cozy-libs#2807](https://github.com/cozy/cozy-libs/pull/2807)
- [cozy/cozy-ui#2828](https://github.com/cozy/cozy-ui/pull/2828)
- [cozy/cozy-ui#2838](https://github.com/cozy/cozy-ui/pull/2838)

### What has been changed?

Add context menu to recipient's shared drive 

https://github.com/user-attachments/assets/1fa04581-731f-4b9d-a32a-5e5c4ecdbb48


src/modules/shareddrives/components/SharedDriveFolderBody.jsx
```diff
@@ -1,13 +1,71 @@
...
+    isPublic: false,
+    allLoaded,
+    showAlert,
+    isOwner,
+    byDocId,
+    isMobile,
+    isNativeFileSharingAvailable,
+    shareFilesNative,
+    sharedDriveId: driveId
+  }
```
zatteo, line 36: Are all these attributes necessary here? Or is it a copy past of another folder body component ?

zatteo:
I expected that we would need a lot more code like we had to do for [nextcloud actions](https://github.com/cozy/cozy-drive/tree/master/src/modules/nextcloud/components/actions) but for the moment it seems to work pretty well like this thanks to cozy-client 👍

PR #3440, opened 2025-08-19
feat: Remove sharing users in grid view :sparkles:

lethemanh:
### Before
CleanShot 2025-08-19 at 13 52 39@2x

### After
CleanShot 2025-08-19 at 13 51 53@2x

### DIFF
--- a/src/modules/filelist/File.jsx
+++ b/src/modules/filelist/File.jsx
@@ -182,13 +182,6 @@ const File = ({
               }
             }}
           />
-          {viewType === 'grid' && (
-            <Status
-              file={attributes}
-              disabled={isRowDisabledOrInSyncFromSharing}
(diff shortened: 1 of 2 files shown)


JF-Cozy (approved):
is `Status` component still used elsewhere? (to know if we can remove it also)

zatteo:
> we did we remove this? i don't recall we discussed this change.
> 
> IMO it should be an alternative design if it looks ugly, not remove it. and having to switch between views just to see who you shared your folders with is bad UX

This was asked by the designer 3 weeks ago. Maybe you can talk to him to challenge this.

I agree with him because the grid view : 
- is considered as a "view where you can see content quickly"
- already have less information than list view

rezk2ll:
Closing this PR is because the change was not approved by the Product owner

PR #3443, opened 2025-08-22
feat: Add shared drive list into sidebar :sparkles:

lethemanh:
### Evidence:

CleanShot 2025-08-22 at 17 11 04@2x

### DIFF
--- a/src/modules/views/Folder/FolderViewBody.jsx
+++ b/src/modules/views/Folder/FolderViewBody.jsx
@@ -125,12 +125,11 @@ const FolderViewBody = ({
     navigate('/folder')
   }, [navigate])
 
-  const FileListBodyWrapper = ({ viewType, isDesktop, children }) => {
+  const FileListBodyWrapper = ({ viewType, children }) => {
     return (
       <div
         className={cx(
-          viewType === 'grid' ? styles['fil-folder-body-grid'] : '',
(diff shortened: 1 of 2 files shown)


src/modules/navigation/components/FileLink.tsx
```diff
@@ -13,6 +13,14 @@ const FileLink = forwardRef<HTMLAnchorElement, FileLinkProps>(
   function FileLinkComponent({ link, children, ...props }, ref) {
     const openInNewTab = link.openInNewTab ? { target: '_blank' } : {}
 
+    if (link.isSharedDrive) {
```
rezk2ll: I believe this change is unrelated to the task, just add back the code removed from this commit https://github.com/cozy/cozy-drive/commit/79281c8a62bc7b858c0ec58d40b3e02571421908

if there are changes to be made to the file link or opening shared drives from the list it should be done in a separate PR

src/modules/navigation/hooks/useFileLink.tsx
```diff
@@ -99,11 +101,24 @@ const useFileLink = (
     ...Object.fromEntries(new URLSearchParams(to.search).entries())
   })
 
+  const generateSharedDrivePath = (
```
rezk2ll: same

rezk2ll (approved):

PR #3444, opened 2025-08-22
fix: not being able to scroll on mobile 🐛

rezk2ll:
fixes scroll on mobile for both grid and list view.

root issue:
`overflow: hidden` was added when on mobile to the folder body, which clips the content, disables scrolling, touch dragging, etc

## demo

https://github.com/user-attachments/assets/56f3a095-658f-49ba-90ed-2f12a49c2dc0

### DIFF
--- a/src/modules/views/Folder/FolderViewBody.jsx
+++ b/src/modules/views/Folder/FolderViewBody.jsx
@@ -125,12 +125,11 @@ const FolderViewBody = ({
     navigate('/folder')
   }, [navigate])
 
-  const FileListBodyWrapper = ({ viewType, isDesktop, children }) => {
+  const FileListBodyWrapper = ({ viewType, children }) => {
     return (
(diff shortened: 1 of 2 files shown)


lethemanh (approved):

JF-Cozy:
is it working for both virtualized and not virtualized tables? BTW I think we should now remove the old not virtualized table to avoid maintaining both code, and so maybe prioritize minor issue of the virtualized table.

rezk2ll:
> is it working for both virtualized and not virtualized tables? BTW I think we should now remove the old not virtualized table to avoid maintaining both code, and so maybe prioritize minor issue of the virtualized table.

Yes it works on both views.
The issue was an `overflow: hidden !important` was added to the folder view body when in mobile. There was no reason for it to exist, and i suspect it was added by some AI autogenerate.

PR #3445, opened 2025-08-25
feat: Open shared drive from sharing section :sparkles:

lethemanh:
### What has been changed?
Shared drives in the Sharing section only appear as shortcuts and don’t include a root folder id, making them unusable. Since `io.cozy.sharings` provides the needed driveId and root folder id but in an incompatible format, we transform them into directory-like objects with the correct structure. This allows them to be displayed properly in the UI and opened via `shareddrive/:driveId/:rootFolderId`.

#### Before:

https://github.com/user-attachments/assets/0f2a7e2f-ea57-447d-96ae-542092aaf731

#### After:

https://github.com/user-attachments/assets/6406e962-909c-44b6-b975-9279cb2ca5bf

### Note:
~~- In recipient's sharing section, the shared drive ID is not correct as shared drive list in sidebar, maybe there was an error from `/files/_all_docs?include_docs=true` endpoint~~

### DIFF
--- a/src/modules/navigation/hooks/helpers.ts
+++ b/src/modules/navigation/hooks/helpers.ts
@@ -8,7 +8,7 @@ import {
 import { IOCozyFile } from 'cozy-client/types/types'
 
 import type { File } from '@/components/FolderPicker/types'
-import { TRASH_DIR_ID } from '@/constants/config'
+import { TRASH_DIR_ID, SHARED_DRIVES_DIR_ID } from '@/constants/config'
 import { joinPath } from '@/lib/path'
 import { isNextcloudShortcut } from '@/modules/nextcloud/helpers'
 import { makeOnlyOfficeFileRoute } from '@/modules/views/OnlyOffice/helpers'
@@ -25,6 +25,13 @@ interface ComputePathOptions {
   isPublic: boolean
 }
 
+interface FileAttribute {
+  type: string
+  name: string
+  dir_id: string
+  driveId: string
+}
+
 export const computeFileType = (
   file: File,
   {
@@ -37,6 +44,8 @@ export const computeFileType = (
     return 'trash'
   } else if (file._id === 'io.cozy.remote.nextcloud.files.trash-dir') {
     return 'nextcloud-trash'
+  } else if (file.dir_id === SHARED_DRIVES_DIR_ID) {
+    return 'shared-drive'
   } else if (file._type === 'io.cozy.remote.nextcloud.files') {
     return isDirectory(file) ? 'nextcloud-directory' : 'nextcloud-file'
   } else if (isNote(file)) {
@@ -123,6 +132,10 @@ export const computePath = (
         fromPathname: pathname,
         fromPublicFolder: isPublic
       })
+    case 'shared-drive':
+      return `/shareddrive/${(file.attributes as FileAttribute).driveId}/${
+        file._id
+      }`
     default:
       // On mobile, if we are in /favorites tab, we do not want it to appears in computed path
       // so we redirect to root route for files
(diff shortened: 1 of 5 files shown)


zatteo:
> ### Note:
> 
>     * In recipient's sharing section, the shared drive ID is not correct as shared drive list in sidebar, maybe there was an error from `/files/_all_docs?include_docs=true` endpoint

When you share a file on Cozy, you have two documents : the file (`io.cozy.file`) and the sharing (`io.cozy.sharing`).

In sidebar, we get the `io.cozy.sharing` document directly. The shared drive ID is the id of the `io.cozy.sharing`.

Here we get `io.cozy.files` document. That's why it's not the "good" id. 

You can learn more here : https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.sharings/ & https://docs.cozy.io/en/cozy-client/relationships/

src/modules/navigation/hooks/useFileLink.tsx
```diff
@@ -99,11 +101,24 @@ const useFileLink = (
...
   // so we open their link directly
   const href =
     type === 'nextcloud-file'
       ? path
+      : file.dir_id === SHARED_DRIVES_DIR_ID
+      ? generateSharedDrivePath(
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          file.relationships?.referenced_by?.data?.[0]?.id as string,
+          file._id
+        )
```
zatteo: Did you try to implement this (which is correct) in the `computeFileType` and `computePath` method. It's not a special case as nextcloud.

src/modules/navigation/hooks/helpers.ts
```diff
@@ -85,6 +87,7 @@ export const computePath = (
   { type, pathname, isPublic }: ComputePathOptions
 ): string => {
   const paths = pathname.split('/').slice(1)
+  console.log('type: ', type)
```
rezk2ll:
```suggestion
```

src/modules/navigation/hooks/helpers.ts
```diff
@@ -37,6 +37,8 @@ export const computeFileType = (
     return 'trash'
   } else if (file._id === 'io.cozy.remote.nextcloud.files.trash-dir') {
     return 'nextcloud-trash'
+  } else if (file.dir_id === SHARED_DRIVES_DIR_ID) {
+    return 'shared-dirve'
```
zatteo: Careful you wrote **dirve** instead of **drive**

zatteo (approved):
Maybe you can add a comment when you transform shared drive because it is not obvious _why_ we do this and it is kind of hacky.

src/modules/navigation/hooks/helpers.ts
```diff
@@ -123,6 +125,9 @@ export const computePath = (
         fromPathname: pathname,
         fromPublicFolder: isPublic
       })
+    case 'shared-drive':
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
rezk2ll: we can't have this

rezk2ll (commented):
this feels like treating the symptoms instead of fixing the root issue.

it is not clear what this PR is trying to achieve:
- would be better to have a description in the PR to know what changed and why.
- a demo video to see the result

package.json
```diff
@@ -133,5 +133,6 @@
     "redux-mock-store": "1.5.4",
     "redux-thunk": "2.4.2",
     "whatwg-fetch": "3.0.0"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
rezk2ll: unrelated

PR #3446, opened 2025-08-27
feat: update the default sort order 🔧

rezk2ll:
- changes the default sort order to `update_at` instead of by name
- Use the config DEFAULT_SORT instead of a component-specific default value


src/config/sort.js
```diff
@@ -1,7 +1,11 @@
...
 export const SORTABLE_ATTRIBUTES = [
   { label: 'name', attr: 'name', css: 'file', defaultOrder: 'asc' },
   { label: 'update', attr: 'updated_at', css: 'date', defaultOrder: 'desc' }
   // TODO: activate sorting by size when it's ready on the back side
   // { label: 'size', attr: 'size', css: 'size', defaultOrder: 'desc' }
 ]
-export const DEFAULT_SORT = { attribute: 'name', order: 'asc' }
+export const DEFAULT_SORT = flag('drive.default-updated-at-sort.enabled')
+  ? { attribute: 'updated_at', order: 'desc' }
+  : { attribute: 'name', order: 'asc' }
```
zatteo: Did your try with the flag enabled and disabled? We already had issues with flag at root level of a file because sometimes they are not initialized when code is executed.

rezk2ll: yes after testing, the flag is not set until you do two refreshes, so moving the flag check in the components

PR #3453, opened 2025-09-01
feat: Shared drive recipients can leave a shared drive

doubleface:
When a user displays a shared drive, the menu is now available with the "Leave shared drive" command which revokes the current user from the selected shared drive.

image

When done, a success message is displayed and the user is redirected to the list of sharings

image

- **feat: Add header to shared folder view**
- **feat: Add "leave shared drive" to toolbar**
- **feat: Hide shared recipients and share buttons**

### DIFF
--- a/src/modules/drive/Toolbar/components/LeaveSharedDriveButtonItem.jsx
+++ b/src/modules/drive/Toolbar/components/LeaveSharedDriveButtonItem.jsx
@@ -10,6 +10,8 @@ import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
 import { useAlert } from 'cozy-ui/transpiled/react/providers/Alert'
 import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
 
+import { getSharingIdFromRelationships } from '@/modules/shareddrives/helpers'
+
 const LeaveSharedDriveButtonItem = ({ files }) => {
   const { t } = useI18n()
   const client = useClient()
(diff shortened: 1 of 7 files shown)


rezk2ll (approved):

src/modules/drive/Toolbar/components/LeaveSharedDriveButtonItem.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default LeaveSharedDriveButtonItem
+
+function findSharingId(folder) {
+  const references = folder.relationships.referenced_by.data
+  const sharingId = references.find(
+    (ref) => ref.type === 'io.cozy.sharings'
+  )?.id
+  return sharingId
+}
```
zatteo: nit: Maybe this method should be somewhere else? Or maybe we already have something like this?

zatteo (approved):
Looks good!

src/modules/drive/Toolbar/components/LeaveSharedDriveButtonItem.jsx
doubleface: Now I moved it in shared drives helpers

src/modules/shareddrives/helpers.ts
```diff
@@ -11,6 +11,20 @@ export interface Rule {
...
+ */
+export const getSharingIdFromRelationships = (doc: {
+  relationships?: {
+    referenced_by?: { data?: { id: string; type: string }[] }
+  }
+}): string | undefined => {
+  const references = doc.relationships?.referenced_by?.data
+  if (!references || references.length === 0) return undefined
+  return references.find(ref => ref.type === 'io.cozy.sharings')?.id
+}
```
rezk2ll:
```suggestion
}): string | undefined =>
  doc.relationships?.referenced_by?.data?.find(ref => ref.type === SHARING_TYPE)?.id
```

PR #3454, opened 2025-09-02
feat: Added breadcrumb for shared drives

doubleface:
Now the breadcrumb is display with shared drive for sender and recipient

image

And when the user clicks on the shared drive image, he navigates to the list of shared drives


doubleface:
> Does it work if you go inside a folder inside the shared drive for the recipient?

Yes I tested it. I did not need specific code for it @zatteo

zatteo (approved):

PR #3455, opened 2025-09-03
feat: Add accordion behavior to shared drives menu

doubleface:
Now the shared drive menu can be expanded/collapsed when there are more than 5 elements.

Also added this behavior the list of favourites as shown in the mockup.

[Capture vidéo du 2025-09-03 17-10-46.webm](https://github.com/user-attachments/assets/d32ea25f-b71f-46d7-aee1-87ed600d9630)

### DIFF
--- a/src/components/SideBarAccordion.jsx
+++ b/src/components/SideBarAccordion.jsx
@@ -1,0 +1,61 @@
+import React, { useState } from 'react'
+
+import Accordion from 'cozy-ui/transpiled/react/Accordion'
+import AccordionDetails from 'cozy-ui/transpiled/react/AccordionDetails'
+import AccordionSummary from 'cozy-ui/transpiled/react/AccordionSummary'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import BottomIcon from 'cozy-ui/transpiled/react/Icons/Bottom'
+import { makeStyles } from 'cozy-ui/transpiled/react/styles'
+
+const useStyles = makeStyles({
+  accordion: {
+    boxShadow: 'none',
+    border: 'none',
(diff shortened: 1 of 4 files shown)


src/components/SideBarAccordion.jsx
```diff
@@ -0,0 +1,61 @@
...
+const useStyles = makeStyles({
+  accordion: {
+    boxShadow: 'none',
+    border: 'none',
+    backgroundColor: 'transparent',
+    marginTop: '0 !important'
+  },
+  summary: {
+    textTransform: 'none',
+    color: 'var(--coolGrey)',
```
zatteo, line 19: Is it still a good practice to use this color @JF-Cozy ?

JF-Cozy: nope

zatteo (approved):
Just waiting an answer about style but looks good otherwise!

src/components/SideBarAccordion.jsx
doubleface: Then, what color should I use ?

JF-Cozy: sorry message was cut, internet very laggy...

We already overrides accordion in cozy-ui, I don't think it's a good idea to override it again inside the app. We have already a "show more" "show less" system on this sidebar with `NavDesktopLimiter` component, maybe we can use it or modify it someway 🤔 

Be careful about creating new Ui component, we already have a lot of them in cozy-ui, please take a look to cozy-ui first (tell swann to do so as well ;) ). If we don't have the exact behavior, maybe we can adapt the design.

We also have `DropdownButton` and `DropdownText`, maybe useful?

PR #3458, opened 2025-09-05
Open shared drive note

zatteo:
To do before merging: 
- [x]  update cozy-client https://github.com/cozy/cozy-client/pull/1628

### DIFF
--- a/src/modules/navigation/hooks/helpers.ts
+++ b/src/modules/navigation/hooks/helpers.ts
@@ -109,7 +109,7 @@ export const computePath = (
     case 'public-note-same-instance':
       return `/?id=${file._id}`
     case 'public-note':
-      return `/note/${file._id}`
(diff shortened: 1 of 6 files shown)


src/modules/navigation/hooks/helpers.ts
```diff
@@ -133,9 +127,7 @@ export const computePath = (
         fromPublicFolder: isPublic
       })
     case 'shared-drive':
-      return `/shareddrive/${(file.attributes as FileAttribute).driveId}/${
-        file._id
-      }`
+      return `/shareddrive/${driveId!}/${file._id}`
```
rezk2ll:
```suggestion
      return `/shareddrive/${driveId}/${file._id}`
```

and handle the case where driveId is undefined before or hack a fallback route

zatteo: Fixed

PR #3466, opened 2025-09-11
Selection bar now disappear if no more files are selected

Coussecousse:
Ticket : https://www.notion.so/linagora/Selection-bar-does-not-disappear-after-all-selected-items-are-deleted-23162718bad180748157e16ff77366f2

[Capture vidéo du 2025-09-11 13-54-43.webm](https://github.com/user-attachments/assets/aeaf250c-6689-42a7-9e42-08441b0d903e)

Explanation: Update the list of the selectedItems when deleting (other actions work as expected). The update of this list close automatically the selection bar

### DIFF
--- a/src/modules/drive/DeleteConfirm.jsx
+++ b/src/modules/drive/DeleteConfirm.jsx
@@ -15,6 +15,7 @@ import { DOCTYPE_ALBUMS } from '@/lib/doctypes'
 import { getEntriesTypeTranslated } from '@/lib/entries'
 import { trashFiles } from '@/modules/actions/utils'
 import { buildAlbumByIdQuery } from '@/queries'
+import { useSelectionContext } from '../selection/SelectionProvider'
 
 const Message = ({ type, fileCount }) => {
   const icon =
@@ -48,6 +49,7 @@ export const DeleteConfirm = ({
   const [isDeleting, setDeleting] = useState(false)
   const [isReferencedByManualAlbum, setIsReferencedByManualAlbum] =
     useState(false)
+  const { setSelectedItems } = useSelectionContext()
 
   useEffect(() => {
     const fetchAlbums = async () => {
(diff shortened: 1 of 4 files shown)


src/modules/drive/DeleteConfirm.jsx
```diff
@@ -84,8 +86,23 @@ export const DeleteConfirm = ({
     setDeleting(true)
     await trashFiles(client, files, { showAlert, t })
     afterConfirmation()
+    setSelectedItems(prevSelectedItems => {
+      const updated = { ...prevSelectedItems }
+      files.forEach(file => {
+        delete updated[file.id]
+      })
+      return updated
+    })
```
rezk2ll:
```suggestion
    setSelectedItems((prevSelectedItems) =>
      prevSelectedItems.filter(
        (selectedItem) => !files.some((file) => file.id === selectedItem.id)
      )
    )
```

src/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -37,7 +37,6 @@ import { FolderUnlocker } from '@/modules/folder/components/FolderUnlocker'
 import { useFolderSort } from '@/modules/navigation/duck'
 import SelectionBar from '@/modules/selection/SelectionBar'
 import { isReferencedByShareInSharingContext } from '@/modules/views/Folder/syncHelpers'
-
```
rezk2ll, line 40: this change is not needed

src/modules/drive/DeleteConfirm.jsx
Coussecousse: prevSelectedItems in not an array but an object. So i can use filter() or Object.fromEntries 
```js 
    setSelectedItems(prevSelectedItems => {
      const fileIdsToRemove = files.map(file => file.id)
      return Object.fromEntries(
        Object.entries(prevSelectedItems).filter(
          ([id]) => !fileIdsToRemove.includes(id)
        )
      )
    })
 ``` 
 And nop doing this isn't working : 
 ```js
 setSelectedItems(prevSelectedItems => {
      files.forEach(file => {
        delete prevSelectedItems[file.id]
      })
      return updated
    })
 ```

rezk2ll: yes, good catch

PR #3467, opened 2025-09-12
feat: Implement create, upload files or folders in shared drive :sparkles:

lethemanh:
### What has been changed?

For now, we can create an empty folder, upload files/folders, and drop to upload files/folders in shared drives.

### Result:

https://github.com/user-attachments/assets/df6f2d38-7e6a-4aa8-81c3-f07c6d6c9a8c

### DIFF
--- a/src/lib/encryption.js
+++ b/src/lib/encryption.js
@@ -37,14 +37,15 @@ export const getEncryptionKeyFromDirId = async (client, dirId) => {
 export const createEncryptedDir = async (
   client,
   vaultClient,
-  { name, dirID }
+  { name, dirID, driveId }
 ) => {
-  // Create the directory
-  const { data: dir } = await client.create(DOCTYPE_FILES, {
-    name,
-    dirId: dirID,
-    type: 'directory'
-  })
+  const { data: dir } = await client
+    .collection(DOCTYPE_FILES, { driveId })
+    .create({
+      name,
+      dirId: dirID,
+      type: 'directory'
+    })
 
   // Create the encryption key
   const docId = `${DOCTYPE_FILES}/${dir._id}`
@@ -64,12 +65,12 @@ export const createEncryptedDir = async (
 export const encryptAndUploadNewFile = async (
   client,
   vaultClient,
-  { file, encryptionKey, fileOptions }
+  { file, encryptionKey, fileOptions, driveId }
 ) => {
   const { name, dirID, onUploadProgress } = fileOptions
   const encryptedFile = await vaultClient.encryptFile(file, encryptionKey)
   const resp = await client
-    .collection(DOCTYPE_FILES)
+    .collection(DOCTYPE_FILES, { driveId })
     .createFile(encryptedFile, {
       name,
       dirId: dirID,
--- a/src/modules/navigation/duck/actions.jsx
+++ b/src/modules/navigation/duck/actions.jsx
@@ -42,7 +42,8 @@ export const uploadFiles =
     dirId,
     sharingState,
     fileUploadedCallback = () => null,
-    { client, vaultClient, showAlert, t }
+    { client, vaultClient, showAlert, t },
+    driveId
   ) =>
   dispatch => {
     let targetDirId = dirId
@@ -82,7 +83,8 @@ export const uploadFiles =
               navigateAfterUpload
             )
           ),
-        { client, vaultClient }
+        { client, vaultClient },
+        driveId
       )
     )
   }
@@ -249,7 +251,8 @@ export const createFolder = (
   vaultClient,
   name,
   currentFolderId,
-  { isEncryptedFolder = false, showAlert, t } = {}
+  { isEncryptedFolder = false, showAlert, t } = {},
+  driveId
 ) => {
   return async (dispatch, getState) => {
     const state = getState()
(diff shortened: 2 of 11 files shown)


src/modules/navigation/duck/actions.jsx
```diff
@@ -280,16 +283,19 @@ export const createFolder = (
     let createdFolder = null
     try {
       if (!isTargetEncrypted) {
-        createdFolder = await client.create('io.cozy.files', {
-          name: name,
-          dirId: targetFolderId,
-          type: 'directory'
-        })
+        createdFolder = await client
```
rezk2ll, line 286: This behaviour should be tested in both cases, shared drives and normal folder

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
```diff
@@ -52,24 +53,34 @@ const useSharedDriveFolder = ({
...
       void fetchSharedDriveFolder()
     }
-  }, [client, sharedDriveQuery])
+  }, [client, fetchSharedDriveFolder])
 
   return {
     sharedDriveQuery,
-    sharedDriveResult
+    sharedDriveResult,
+    refresh: () => fetchSharedDriveFolder(true)
```
zatteo: What is the purpose of this refresh method? To avoid the need to do F5 to see the folder uploaded with a drag n drop?

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
```diff
@@ -52,24 +53,34 @@ const useSharedDriveFolder = ({
...
   )
 
-  useEffect(() => {
-    const fetchSharedDriveFolder = async (): Promise<void> => {
+  const fetchSharedDriveFolder = useCallback(
+    async (isRefresh?: boolean): Promise<void> => {
       const newSharedDriveResult = (await client?.query(
         sharedDriveQuery.definition(),
-        sharedDriveQuery.options
+        // when refreshing the fetchPolicy needs to be unset to avoid client.query block the request in cache time
```
rezk2ll: Do you think this is really needed?
Having a different fetch behaviour on initial load and on refresh.

src/modules/navigation/duck/actions.jsx
```diff
@@ -42,7 +42,8 @@ export const uploadFiles =
     dirId,
     sharingState,
     fileUploadedCallback = () => null,
-    { client, vaultClient, showAlert, t }
+    { client, vaultClient, showAlert, t },
+    driveId
```
rezk2ll, line 46: Does this break the normal behaviour when not in the shared drive context? ( driveId is undefined )

src/lib/encryption.js
```diff
@@ -64,12 +65,12 @@ export const createEncryptedDir = async (
...
   client,
   vaultClient,
-  { file, encryptionKey, fileOptions }
+  { file, encryptionKey, fileOptions, driveId }
 ) => {
   const { name, dirID, onUploadProgress } = fileOptions
   const encryptedFile = await vaultClient.encryptFile(file, encryptionKey)
   const resp = await client
-    .collection(DOCTYPE_FILES)
+    .collection(DOCTYPE_FILES, { driveId })
```
rezk2ll, line 73: We are chaining adding driveId without testing the behaviour when it's undefined.

Does the cozy-client handle this value being undefined?

src/modules/views/SharedDrive/SharedDriveFolderView.jsx
```diff
@@ -24,19 +31,32 @@ const SharedDriveFolderView = () => {
     ? [{ fetchStatus: 'loaded', data: sharedDriveResult.included }]
     : []
 
+  const canWriteToCurrentFolder = hasWriteAccess(folderId)
+
+  const DropzoneComp =
+    flag('drive.virtualization.enabled') && !isMobile ? DropzoneDnD : Dropzone
```
rezk2ll:
```suggestion
    !isMobile ? DropzoneDnD : Dropzone
```

in desktop, we stopped supporting the old view mode, and virtualization is the default mode, so no need to juggle between dropzone components here.

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
lethemanh: yes, after uploading / creating folder we need to refresh the shared drive

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
lethemanh: without it we cannot refresh in cache time and the list will be empty

src/modules/navigation/duck/actions.jsx
lethemanh: in case `driveId` is undefined, this function works as normal, cozy-client already handle this case

src/lib/encryption.js
lethemanh: yes, cozy-client already handles this case, without `driveId` it will use the normal endpoint

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
zatteo: This will be solved, I think, when we implement the realtime next week. So I don't know if we need to keep it for 1 week or not.

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
rezk2ll: i mean unsetting the fetchPolicy

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
lethemanh: ok, I will remove it

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
lethemanh: In cozy-client fetchPolicy uses to check if query can be executed or not and currently, the fetchPolicy in this query just about how long should the query be cached.
But realtime shared will be implemented in next week (as @zatteo said) I think I will remove the refresh function.

lethemanh: it has been removed

rezk2ll (approved):

PR #3469, opened 2025-09-12
fix: Redirects to shared drives list

doubleface:
Now redirect to the list of sharings in the "Shared drives" tab

### DIFF
--- a/src/modules/navigation/AppRoute.jsx
+++ b/src/modules/navigation/AppRoute.jsx
@@ -21,14 +21,18 @@ import FilesViewerTrash from '../views/Trash/FilesViewerTrash'
 import TrashFolderView from '../views/Trash/TrashFolderView'
 
 import FileHistory from '@/components/FileHistory'
-import { ROOT_DIR_ID, TRASH_DIR_ID } from '@/constants/config'
+import {
+  ROOT_DIR_ID,
+  TRASH_DIR_ID,
+  SHARED_DRIVES_DIR_ID,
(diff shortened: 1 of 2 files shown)


src/modules/drive/Toolbar/delete/delete.jsx
```diff
@@ -4,15 +4,20 @@ import { useNavigate } from 'react-router-dom'
...
-    [navigate, folder]
-  )
+  const navigateToParentFolder = useCallback(() => {
+    const isSharedDrive = folder.dir_id === SHARED_DRIVES_DIR_ID
+    if (isSharedDrive) {
+      navigate(`/sharings?tab=${SHARING_TAB_DRIVES}`)
+    } else {
+      navigate(`/folder/${folder.dir_id}`)
+    }
+  }, [navigate, folder])
```
rezk2ll:
```suggestion
  const navigateToParentFolder = useCallback(() => {
    navigate(
      folder.dir_id === SHARED_DRIVES_DIR_ID
        ? `/sharings?tab=${SHARING_TAB_DRIVES}`
        : `/folder/${folder.dir_id}`
    )
  }, [navigate, folder])
```

doubleface: Frankly, I prefer to have a named intermediate variable to explicitly show what `folder.dir_id === SHARED_DRIVES_DIR_ID` means.  And with that said, if you'd rather I use a ternary operator, why not, but is it really worth making a change for that?

PR #3470, opened 2025-09-15
fix: Open correct shared drive view by user :bug:

lethemanh:
### Change:

Set the correct path to open shared drive by user's role.

#### User is owner

https://github.com/user-attachments/assets/ff9ae41a-7059-4449-bf02-b524059a5ab1

#### User is the recipient

https://github.com/user-attachments/assets/bbc73a90-86f5-460f-9b39-d61e5088de30

### Need:

[cozy/cozy-viewer#2818](https://github.com/cozy/cozy-libs/pull/2818)

### DIFF
--- a/src/modules/navigation/hooks/useSharedDriveLink.tsx
+++ b/src/modules/navigation/hooks/useSharedDriveLink.tsx
@@ -26,7 +26,18 @@ const useSharedDriveLink = (sharing: SharedDrive): UseFileLinkResult => {
   const cozyUrl = client?.getStackClient().uri as string
 
   const app = 'drive'
-  const path = `shareddrive/${sharing._id}/${sharing.rules[0].values[0]}`
+
+  if (!sharing.rules[0].values[0]) {
+    throw new Error('Missing folder id in shared drive')
+  }
+
+  /** Set shared drive path
+   * if user is owner of this shared drive, the path should be `folder/:folderId`
+   * otherwise the path should be `shareddrive/:driveId/:folderId`
+   */
+  const path = sharing.owner
+    ? `folder/${sharing.rules[0].values[0]}`
+    : `shareddrive/${sharing._id}/${sharing.rules[0].values[0]}`
 
   const currentURL = new URL(window.location.href)
   const currentPathname = currentURL.pathname
--- a/src/modules/views/Sharings/index.jsx
+++ b/src/modules/views/Sharings/index.jsx
@@ -68,7 +68,7 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
   const { pushModal, popModal } = useModalContext()
   const { isSelectionBarVisible, toggleSelectAllItems, isSelectAll } =
     useSelectionContext()
-  const { allLoaded, refresh } = useSharingContext()
+  const { allLoaded, refresh, isOwner } = useSharingContext()
   const { isNativeFileSharingAvailable, shareFilesNative } =
     useNativeFileSharing()
   const { sharedDrives } = useSharedDrives()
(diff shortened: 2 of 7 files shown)


src/modules/navigation/hooks/useSharedDriveLink.tsx
```diff
@@ -26,7 +26,14 @@ const useSharedDriveLink = (sharing: SharedDrive): UseFileLinkResult => {
...
   const app = 'drive'
-  const path = `shareddrive/${sharing._id}/${sharing.rules[0].values[0]}`
+
+  /** Set shared drive path
+   * if user is owner of this shared drive, the path should be `folder/:folderId`
+   * otherwise the path should be `shareddrive/:driveId/:folderId`
+   */
+  const path = sharing.owner
+    ? `folder/${sharing.rules[0].values[0]}`
+    : `shareddrive/${sharing._id}/${sharing.rules[0].values[0]}`
```
zatteo, line 40: Agreeing with the general philosophy, but I know Christophe also worked on it friday and ended up with a slightly different solution, let's wait him to compare both solutions

src/modules/views/Sharings/index.jsx
```diff
@@ -121,21 +121,25 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
...
-        id: SHARED_DRIVES_DIR_ID,
-        _type: 'io.cozy.files',
-        type: 'directory',
-        name: driveName,
-        path: `/Drives/${driveName}`,
-        dir_id: SHARED_DRIVES_DIR_ID,
-        driveId: sharing.id,
-        attributes: {
+      if (fileInSharingSection && isOwner(fileInSharingSection?.id)) {
+        return { ...fileInSharingSection }
```
rezk2ll: is it necessary to create a new object here?

lethemanh: Actually, it's unnecessary, I've removed it

src/modules/navigation/hooks/useSharedDriveLink.tsx
```diff
@@ -26,7 +26,14 @@ const useSharedDriveLink = (sharing: SharedDrive): UseFileLinkResult => {
...
 
   const app = 'drive'
-  const path = `shareddrive/${sharing._id}/${sharing.rules[0].values[0]}`
+
+  /** Set shared drive path
+   * if user is owner of this shared drive, the path should be `folder/:folderId`
+   * otherwise the path should be `shareddrive/:driveId/:folderId`
+   */
+  const path = sharing.owner
+    ? `folder/${sharing.rules[0].values[0]}`
```
rezk2ll, line 39: so we assume that `sharing.rules[0].values[0]` to always be defined in this case? What if values didn't exist

src/modules/views/Sharings/index.jsx
```diff
@@ -121,21 +121,25 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
...
-        _id: rootFolderId,
-        id: SHARED_DRIVES_DIR_ID,
-        _type: 'io.cozy.files',
-        type: 'directory',
-        name: driveName,
-        path: `/Drives/${driveName}`,
-        dir_id: SHARED_DRIVES_DIR_ID,
-        driveId: sharing.id,
-        attributes: {
+      if (fileInSharingSection && isOwner(fileInSharingSection?.id)) {
```
rezk2ll: The `transformedSharedDrives` could be cleaned a bit

```ts
const transformedSharedDrives = (sharedDrives || []).map(sharing => {
  const [rootFolderId, driveName] = [
    sharing.rules?.[0]?.values?.[0],
    sharing.rules?.[0]?.title
  ]

  const fileInSharingSection = result.data?.find(item =>
    item.relationships?.referenced_by?.data?.[0]?.id === sharing.id
  )

  if (isOwner(fileInSharingSection?.id)) return fileInSharingSection

  const directoryData = {
    type: 'directory',
    name: driveName,
    dir_id: SHARED_DRIVES_DIR_ID,
    driveId: sharing.id
  }

  return {
    ...fileInSharingSection,
    _id: rootFolderId,
    id: SHARED_DRIVES_DIR_ID,
    _type: 'io.cozy.files',
    path: `/Drives/${driveName}`,
    ...directoryData,
    attributes: directoryData
  }
})
```

src/modules/navigation/hooks/useSharedDriveLink.tsx
lethemanh: @zatteo do we have any case that `sharing.rules[0].values[0]` is not defined? 
@rezk2ll if it did not exist I think we should throw the error here.

zatteo: Yes it is true for shared drive folder. See https://github.com/cozy/cozy-stack/blob/master/docs/shared-drives.md#shared-drives

Agree to throw if it does not exists.

PR #3476, opened 2025-09-16
🐛 fix: select/disselect all label display on mobile

rezk2ll:
### Before

image

image

### After

image

image


src/locales/fr.json
```diff
@@ -69,7 +69,9 @@
       "slide": "Présentation"
     },
     "select_all": "Tout sélectionner",
+    "select_all_mobile": "Tout",
     "clear_selection": "Effacer la sélection",
+    "clear_selection_mobile": "Annuler",
```
zatteo, line 74: Not convinced with "Annuler" to deselect everything 🤔

zatteo: "Rien" or "Aucun" ? Idk

rezk2ll: we have "Retirer" but can be confusing people can consider it as delete.
"Ôter" i think it is old / unknown
"Décocher" makes sense for a single checkbox

zatteo: Agree... Maybe go for it and we ll see

PR #3477, opened 2025-09-16
feat(shareddrives): Add realtime on shared drives list

doubleface:
Since there is no realtime on sharings, we use realtime on shared drives folder list of folders. Whenever a change is done in the list of folders in this directory, a request to the list of shared drive is done.


zatteo (approved):
Cowboy solution as needed 🤠 It can also be triggered when adding or removing a Nextcloud file but it is very rare.

Does it work when : 
- adding
- deleting
for : 
- owner
- recipient
?

doubleface:
Yes I confirm it works when adding/deleting for owner/recipient

PR #3478, opened 2025-09-17
feat: Implement renaming files or folders in shared drive :sparkles:

lethemanh:
### Change:

Implement renaming files or folders in recipient's shared drives.

### DIFF
--- a/src/modules/drive/RenameInput.jsx
+++ b/src/modules/drive/RenameInput.jsx
@@ -14,7 +14,7 @@ import FilenameInput from '@/modules/filelist/FilenameInput'
 // If we set the _rev then CozyClient tries to update. Else
 // it tries to create
 const updateFileNameQuery = async (client, file, newName) => {
-  return client.save({
+  return client.collection('io.cozy.files', { driveId: file.driveId }).update({
     ...file,
     name: newName,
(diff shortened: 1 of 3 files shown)


src/modules/drive/RenameInput.jsx
```diff
@@ -14,7 +14,7 @@ import FilenameInput from '@/modules/filelist/FilenameInput'
 // If we set the _rev then CozyClient tries to update. Else
 // it tries to create
 const updateFileNameQuery = async (client, file, newName) => {
-  return client.save({
+  return client.collection('io.cozy.files', { driveId: file.driveId }).update({
```
zatteo, line 17: I wonder what the comment means or if by update from save to update we change the behavior 🤔

zatteo: Can you just double check that this method is not used for something else than renaming and there is no regression?

lethemanh: I saw this document about the API update https://github.com/cozy/cozy-stack/blob/master/docs/shared-drives.md#patch-sharingsdrivesidfile-id
It can update the metadata of a file or directory, rename it or to move it

rezk2ll (approved):

PR #3479, opened 2025-09-17
feat: Implement moving files or folders in shared drive :sparkles:

lethemanh:
### Change:

Implement moving files or folders in recipient's shared drives.

#### Moving in shared drive

https://github.com/user-attachments/assets/16d3b743-9917-4625-b43c-b263746899e5


src/components/FolderPicker/FolderPicker.spec.jsx
```diff
@@ -71,33 +84,36 @@ describe('FolderPicker', () => {
     )
   }
 
-  it('should be able to move inside another folder', async () => {
+  it('should render with the provided folder', async () => {
```
rezk2ll, line 122: Here you are removing the old test and testing something else entirely.
Tests are meant to ensure that we still have that behavior. Here, we no longer test "should be able to move inside another folder".

If a test fails, it means the code we introduced is not mocked properly or the behavior has changed.
it does not mean we should remove that test and replace it with a new one that tests something else

src/components/FolderPicker/FolderPicker.spec.jsx
```diff
@@ -71,33 +84,36 @@ describe('FolderPicker', () => {
...
 
-    const moveButton = screen.queryByRole('button', {
-      name: 'Move'
-    })
-    fireEvent.click(moveButton)
-    expect(onConfirmSpy).toHaveBeenCalledWith(rootCozyFolder)
-  })
+    const props = FolderPickerBody.mock.calls[0][0]
 
-  it('should display the folder creation input', async () => {
```
rezk2ll, line 92: the same thing here, we should not do this ever!

src/components/FolderPicker/FolderPickerBody.tsx
```diff
@@ -31,6 +37,37 @@ const FolderPickerBody: React.FC<FolderPickerBodyProps> = ({
     )
   }
 
+  // Display content of recipient's shared drive folder
```
rezk2ll, line 56:
```suggestion
```

src/components/FolderPicker/FolderPickerBody.tsx
```diff
@@ -31,6 +37,37 @@ const FolderPickerBody: React.FC<FolderPickerBodyProps> = ({
...
+        folder={folder}
+        isFolderCreationDisplayed={isFolderCreationDisplayed}
+        hideFolderCreation={hideFolderCreation}
+        entries={entries}
+        navigateTo={navigateTo}
+      />
+    )
+  }
+
+  // Display content of `Drives` folder
```
rezk2ll:
```suggestion
```

src/components/FolderPicker/helpers.ts
```diff
@@ -171,5 +175,7 @@ export const getParentFolder = async (
     }
   }
 
-  return await getCozyParentFolder(client, folder.dir_id)
+  const driveId =
+    folder.dir_id === 'io.cozy.files.shared-drives-dir' ? '' : folder.driveId
```
rezk2ll: we use the constant here, like SHARED_DRIVES_DIR_ID

src/hooks/useKeyboardShortcuts.tsx
```diff
@@ -8,7 +8,7 @@ import flag from 'cozy-flags'
 import { useAlert } from 'cozy-ui/transpiled/react/providers/Alert'
 import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
 
-import { isEditableTarget, normalizeKey } from './helpers'
+import { isEditableButNotCheckbox, normalizeKey } from './helpers'
```
rezk2ll: isEditableButNotCheckbox is a weird function name 🤨

src/modules/paste/index.js
```diff
@@ -9,10 +14,59 @@ import { joinPath } from '@/lib/path'
 import { hasOneOfEntriesShared } from '@/modules/move/helpers'
 import { computeNextcloudFolderQueryId } from '@/modules/nextcloud/helpers'
 
+const generateMoveOrCopySharedDriveApi = async (
```
rezk2ll: i can't tell what this function does; the name is not clear.
also if it just calls moveRelateToSharedDrive, why not call it directly instead of creating a wrapper function?

src/modules/paste/index.js
```diff
@@ -9,10 +14,59 @@ import { joinPath } from '@/lib/path'
...
+        ? destDirectory.cozyMetadata?.createdOn
+        : '',
+      sharing_id: destDirectory.driveId,
+      dir_id: destDirectory._id
+    },
+    isCopy
+  )
+}
+
+export const buildMoveApi = async (
```
rezk2ll: same thing, could be named differently and also missing JSdoc

src/modules/paste/index.js
lethemanh: actually, this function will reuse in the copy but it is in this PR https://github.com/cozy/cozy-drive/pull/3480/files#diff-f0b04775257f97eab0b86e5781befc4f5cc06bc0e07effa7139f368e0048443fR65
btw, I will change the function name

rezk2ll (approved):

PR #3492, opened 2025-09-23
Add manage access and remove/quit shared drive in menus

zatteo:
- Refactoring with extraction of shared drive helpers
- UI improvments of menu appearing in sidebar on hover
- Adding manage access and remove/quit shared drive in menu appearing when viewing a folder and in sharing section (see screenshot)

Screenshot 2025-09-23 at 06 28 14

### DIFF
--- a/src/modules/shareddrives/components/actions/deleteSharedDrive.js
+++ b/src/modules/shareddrives/components/actions/deleteSharedDrive.js
@@ -10,6 +10,7 @@ import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
 import { showModal } from '@/lib/react-cozy-helpers'
 import DeleteConfirm from '@/modules/drive/DeleteConfirm'
 import { getFolderIdFromSharing } from '@/modules/shareddrives/helpers'
+import { isFolderFromSharedDriveOwner } from '@/modules/shareddrives/helpers'
 
 export const deleteSharedDrive = ({
   sharedDrive,
@@ -25,9 +26,15 @@ export const deleteSharedDrive = ({
     name: 'deleteSharedDrive',
     label: label,
     icon,
-    displayCondition: () => sharedDrive.owner,
-    action: async () => {
-      const sharedDriveFolderId = getFolderIdFromSharing(sharedDrive)
+    // This action can be triggered with an io.cozy.sharing when called from sidebar
+    // or with an io.cozy.files when called from the sharing tab that's why we need to
+    // manage two kind of data
+    displayCondition: docs =>
+      sharedDrive ? sharedDrive.owner : isFolderFromSharedDriveOwner(docs[0]),
+    action: async docs => {
+      const sharedDriveFolderId = sharedDrive
+        ? getFolderIdFromSharing(sharedDrive)
+        : docs[0]._id
 
       if (!sharedDriveFolderId) {
         showAlert({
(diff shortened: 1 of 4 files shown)


src/modules/drive/Toolbar/components/ManageAccessItem.jsx
```diff
@@ -0,0 +1,30 @@
...
+import ListItemIcon from 'cozy-ui/transpiled/react/ListItemIcon'
+import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
+import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
+
+const ManageAccessItem = ({ onClick }) => {
+  const navigate = useNavigate()
+  const { t } = useI18n()
+
+  const handleClick = () => {
+    navigate('./share')
```
rezk2ll, line 16: Why do we need this navigation?

src/modules/shareddrives/components/actions/deleteSharedDrive.js
```diff
@@ -0,0 +1,84 @@
...
+        return
+      }
+
+      dispatch(
+        showModal(
+          <DeleteConfirm
+            files={[folder]}
+            afterConfirmation={() =>
+              showAlert({
+                message: t('alert.trash_file_success'),
```
rezk2ll, line 65: this will display `The selection has been moved to the Trash.`

can we have a dedicated message to shared drive?

src/modules/shareddrives/components/actions/manageAccess.js
```diff
@@ -6,6 +6,10 @@ import PeopleIcon from 'cozy-ui/transpiled/react/Icons/People'
 import ListItemIcon from 'cozy-ui/transpiled/react/ListItemIcon'
 import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
 
+import { isFolderFromSharedDriveOwner } from '@/modules/shareddrives/helpers'
+import { getFolderIdFromSharing } from '@/modules/shareddrives/helpers'
+
+// Bien documenter ce bordel et faire la même pour les autres
```
rezk2ll: 😳

zatteo: Removed...

src/modules/drive/Toolbar/components/ManageAccessItem.jsx
zatteo: Good point. This is a first iteration of this manage access button. For the moment we have no way to display the sharing modal in every page. We could implement it but it will take some time. So I leverage the fact that when you go to "/folder/:id/share" it goes to the folder and display share modal. 

If we have time we can improve this this week.

src/modules/shareddrives/components/actions/manageAccess.js
zatteo: More explanation. This PR where I tried to do some refactoring make me understand that we have a lot of ways to deal with shared drive, depending where we are in drive app and if we are an owner or a recipient.

Owner : 
- sometimes the real io.cozy.files of the shared drive
- sometimes the io.cozy.sharing of the shared drive

Recipient : 
- sometimes the io.cozy.sharing of the shared drive
- sometimes the remote io.cozy.file with driveId
- sometimes the shared drive sharing or the shared drive shortcut transformed in a fake io.cozy.file (in sharing section)

Extracting some helpers like I did here make it possible to have some common things between all this ways of dealing with shared drives, but I think we can do better.

rezk2ll: thank you

PR #3527, opened 2025-10-01
fix(shared drives): upload button and create menu was not disabled when in read-only 🐛

rezk2ll:
### changes

- Properly propagate writing permission to child components; previously, it was true by default
- Disable the new upload button on the emptyWrapper component
- Disable the create menu & upload button on the layout when the shared drive is read-only

### DIFF
--- a/src/modules/layout/Layout.jsx
+++ b/src/modules/layout/Layout.jsx
@@ -71,11 +71,12 @@ const LayoutContent = () => {
                 componentsProps={{ button: { className: 'u-w-100 u-bdrs-6' } }}
                 label={t('upload.label')}
                 displayedFolder={displayedFolder}
+                disabled={isFolderReadOnly}
               />
               <AddMenuProvider
                 canCreateFolder={true}
-                canUpload={true}
-                disabled={false}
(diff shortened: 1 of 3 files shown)


src/modules/layout/Layout.jsx
```diff
@@ -71,11 +71,12 @@ const LayoutContent = () => {
                 componentsProps={{ button: { className: 'u-w-100 u-bdrs-6' } }}
                 label={t('upload.label')}
                 displayedFolder={displayedFolder}
+                disabled={isFolderReadOnly}
               />
               <AddMenuProvider
                 canCreateFolder={true}
-                canUpload={true}
-                disabled={false}
+                canUpload={!isFolderReadOnly}
```
lethemanh, line 78: how about using only prop `disabled`, because the `UploadButton` is disabled as well if cannot upload to the displayed folder?

rezk2ll: because the `disabled` property controls the selectionBar visibility, and the canUpload is passed down to child components alongside the isDisabled prop.

The AddMenuProvider can be reworked in another PR.

zatteo (approved):

PR #3530, opened 2025-10-02
feat: Add right click on shared drive item in recipient :sparkles:

lethemanh:
### Change:

Add right click on shared drive item in recipient.

### Result:

CleanShot 2025-10-02 at 17 12 46@2x

### DIFF
--- a/src/modules/folder/components/FolderBody.jsx
+++ b/src/modules/folder/components/FolderBody.jsx
@@ -8,6 +8,7 @@ import styles from '@/styles/folder-view.styl'
 
 import { EmptyDrive } from '@/components/Error/Empty'
 import Oops from '@/components/Error/Oops'
+import RightClickFileMenu from '@/components/RightClick/RightClickFileMenu'
 import { useThumbnailSizeContext } from '@/lib/ThumbnailSizeContext'
 import { useViewSwitcherContext } from '@/lib/ViewSwitcherContext'
 import AddFolder from '@/modules/filelist/AddFolder'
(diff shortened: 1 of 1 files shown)


rezk2ll:
@lethemanh does it also work on an empty folder? or inside a folder in general.

For example, right click > create folder as a recipient

src/modules/folder/components/FolderBody.jsx
```diff
@@ -135,18 +136,26 @@ const FolderBody = ({
                 <React.Fragment key={queryIndex}>
                   {query.data.map(file => {
                     return (
-                      <File
+                      <RightClickFileMenu
```
JF-Cozy, line 139: what about mobile? I think FolderViewBody is used on mobile also, isn't it? It's strange that we "forget" this file when adding the right-click feature 🤔 Besides, why we need to modify this files, I though it was not used anymore (expect for mobile maybe, that's the question?)

lethemanh: This file is used in SharedDriveFolderBody so that why I add RightClickFileMenu here

lethemanh: But if this file is not being used any more, I think we should use the same file as Drive is using, what do you think?

JF-Cozy: Why we don't use the virtualized one for SharedDrive? It's intended or we just missed it?

lethemanh: I think we missed it, but if need to change now I will add virtualized as well

PR #3532, opened 2025-10-03
fix: Display correct view by sharing rule in shared drive from recipient :bug:

lethemanh:
### Change:

- Check correct write access in shared drive.
- Display sharing badges, share button in shared drive based on sharing rule.

### Need:

cozy/cozy-sharing#


src/components/RightClick/RightClickAddMenu.jsx
```diff
@@ -48,7 +48,10 @@ const RightClickAddMenu = ({ children, ...props }) => {
   const { hasWriteAccess } = useSharingContext()
 
   const isFolderReadOnly = displayedFolder
-    ? !hasWriteAccess(displayedFolder._id)
+    ? !hasWriteAccess(
+        displayedFolder.driveId ?? displayedFolder._id,
+        !!displayedFolder.driveId
+      )
```
zatteo: nit: Is it simplier to have an hasWriteAccess signature like this `hasWriteAccess(docId, driveId)`

I was also thinking this when reviewer the corresponding PR on cozy-libs but I am not sure 🤔

zatteo (approved):

src/components/RightClick/RightClickAddMenu.jsx
lethemanh: Oh, you're right. I should change the params of `hasWriteAccess`.

zatteo: More readable! I was not 100% sure it was better, but now I really think it is better.

src/modules/views/SharedDrive/SharedDriveFolderView.jsx
```diff
@@ -43,7 +45,12 @@ const SharedDriveFolderView = () => {
         <Content className={isMobile ? '' : 'u-pt-1'}>
           <FolderViewHeader>
             <SharedDriveBreadcrumb driveId={driveId} />
-            <Toolbar canUpload={false} canCreateFolder={false} />
+            <Toolbar
+              canUpload={false}
+              canCreateFolder={false}
```
rezk2ll, line 50: i know this is not your change, but why canUpload and canCreateFolder are false?

lethemanh: I don't think it's useful in this case but let @doubleface explain for more detail

doubleface: No real intent here. Office editing was not activated at the moment I did it but I don't really remember

PR #3542, opened 2025-10-08
feat: Open OO files in edit mode by default if has right access :sparkles:

lethemanh:
### Change:

Check OO file access before open to open correct editor mode.

#### Before:

https://github.com/user-attachments/assets/f978acc4-74aa-48b9-98a5-5c3cacbf536a

#### After:

https://github.com/user-attachments/assets/acd70eec-371b-40ba-b945-d37fc9d69031


src/modules/views/OnlyOffice/OnlyOfficeProvider.jsx
```diff
@@ -82,11 +84,12 @@ const OnlyOfficeProvider = ({
   useEffect(() => {
     if (
       searchParam.get('fromCreate') === 'true' ||
-      searchParam.get('fromEdit') === 'true'
+      searchParam.get('fromEdit') === 'true' ||
+      hasWriteAccess(fileId, driveId)
```
JF-Cozy: why keeping searchParam and "edit button"? Now we don't want to use the edit button anymore, If a user has a write access it doesn't make sens to keep it. And if a user doesn't have write access, it probably doesn't make sens either 🤔 I think here we can simplify things and rely only on hasWriteAccess, isn't it?

src/modules/views/OnlyOffice/Toolbar/index.jsx
```diff
@@ -128,7 +132,9 @@ const Toolbar = ({ sharingInfos }) => {
       {!isPublic && isEditorReady && (
         <>
           <Sharing file={file} />
-          <EditButton />
+          {!flag('drive.onlyoffice.editor-mode-by-access.enabled') ? (
+            <EditButton />
+          ) : null}
```
JF-Cozy: nit:
```
{!flag('drive.onlyoffice.editor-mode-by-access.enabled') && (
          <EditButton />
        )}}`
```
like you did above

src/modules/views/OnlyOffice/OnlyOfficeProvider.jsx
```diff
@@ -82,11 +85,13 @@ const OnlyOfficeProvider = ({
...
       searchParam.get('fromCreate') === 'true' ||
-      searchParam.get('fromEdit') === 'true'
+      searchParam.get('fromEdit') === 'true' ||
+      (flag('drive.onlyoffice.editor-mode-by-access.enabled') &&
+        hasWriteAccess(fileId, driveId))
     ) {
       setEditorMode('edit')
     }
-  }, [searchParam])
+  }, [searchParam, fileId, driveId, hasWriteAccess])
```
rezk2ll, line 99:
```suggestion
useEffect(() => {
  if (!hasWriteAccess(fileId, driveId)) return

  if (flag('drive.onlyoffice.editor-mode-by-access.enabled')) {
    setEditorMode('edit')
    return
  }

  if (
    searchParam.get('fromCreate') === 'true' ||
    searchParam.get('fromEdit') === 'true'
  ) {
    setEditorMode('edit')
  }
}, [searchParam, fileId, driveId, hasWriteAccess])
```

src/modules/views/OnlyOffice/OnlyOfficeProvider.jsx
```diff
@@ -82,11 +85,13 @@ const OnlyOfficeProvider = ({
   useEffect(() => {
     if (
       searchParam.get('fromCreate') === 'true' ||
-      searchParam.get('fromEdit') === 'true'
+      searchParam.get('fromEdit') === 'true' ||
+      (flag('drive.onlyoffice.editor-mode-by-access.enabled') &&
+        hasWriteAccess(fileId, driveId))
```
JF-Cozy: to avoid confusion and useless thinking, I thing we should add a comment to describe how to remove the flag when it will be necessary. It's not clear for me if `searchParam.get('fromEdit') === 'true'` will be relevant again

PR #3543, opened 2025-10-08
fix: Display again folder in favorite

zatteo:
See https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.files/#trash for more explanation about trash for files and folders.

### DIFF
--- a/src/queries/index.ts
+++ b/src/queries/index.ts
@@ -521,8 +521,8 @@ export const buildFavoritesQuery: QueryBuilder<buildFavoritesQueryParams> = ({
       })
       .partialIndex({
         'cozyMetadata.favorite': true,
-        trashed: false,
-        dir_id: { $nin: [TRASH_DIR_ID] }
+        path: { $or: [{ $exists: false }, { $regex: '^(?!/.cozy_trash)' }] },
+        trashed: { $or: [{ $exists: false }, { $eq: false }] }
       })
       .indexFields([sortAttribute])
       .sortBy([{ [sortAttribute]: sortOrder }]),


src/queries/index.ts
```diff
@@ -521,8 +521,8 @@ export const buildFavoritesQuery: QueryBuilder<buildFavoritesQueryParams> = ({
       })
       .partialIndex({
         'cozyMetadata.favorite': true,
-        trashed: false,
-        dir_id: { $nin: [TRASH_DIR_ID] }
+        path: { $or: [{ $exists: false }, { $regex: '^(?!/.cozy_trash)' }] },
+        trashed: { $or: [{ $exists: false }, { $eq: false }] }
```
JF-Cozy, line 525: are we sure about the trash doc? Do we have ask @paultranvan? Is it not a problem to use `or`? (is it creating correct indexes?).

Besides the doc talk about `_id`, no need for that here?

Finally, we have other queries in that file using `trashed: false`, do we have to modify them as well?

zatteo: No need for the _id because we can not set the trash as favorite.

About other queries, I don't think so, otherwise we would have encounter the problem before? Here I fix my fix from yesterday afternoon.

paultranvan: > No need for the _id because we can not set the trash as favorite.

True, but it does not cost much to add it

> Finally, we have other queries in that file using trashed: false, do we have to modify them as well?

Depends on the query objective / context, there is no generic answer. Basically, if you target only files (and not folders), evaluating `trashed: false` is enough

PR #3442, opened 2025-08-22
feat: Implement range selection :sparkles:

lethemanh:
### Need:

[cozy/cozy-ui#2835](https://github.com/cozy/cozy-ui/pull/2835)

### Evidence:

https://github.com/user-attachments/assets/b146af9b-a1c5-43a7-9737-e7e343c8a21b

PR #3545, opened 2025-10-08
Add double click possibility on files/folder

JF-Cozy:
Previously we could open a file/folder with simple click on it. Now we have to double click them. Clicking on them only once will select them. I didn't add a debounce for now, but we will see if it is necessary later...

No change on mobile.

Lots of refactor to split mobile / desktop code inside the hook in charge of the click features.

As the shift+click needs to be refactored and fixed, the compatibility with it is real basic for now. We decided to create the PR anyway, to facilitate the all refactoring. That's why we finally put the double click feature behind `drive.doubleclick.enabled` flag.


src/hooks/useOnLongPress/helpers.spec.jsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  afterEach(() => {
+    jest.clearAllMocks()
+    MockDate.reset()
+  })
+
+  // should create a real life test to replace toggle by final func
+  xit('should only toggle by default', () => {
```
rezk2ll, line 103:
```suggestion
  it('should only toggle by default', () => {
```

src/modules/views/Folder/virtualized/helpers.js
```diff
@@ -13,6 +13,7 @@ export const makeRows = ({ queryResults, IsAddingFolder, syncingFakeFile }) => {
     rows.push(syncingFakeFile)
   }
 
+  // TODO: we should not modify io.cozy.files for frontend purpose
```
rezk2ll, line 16: Where did it happen?

src/hooks/useOnLongPress/index.js
```diff
@@ -0,0 +1,55 @@
...
+  toggle
+}) => {
+  const timerId = useRef()
+  const isLongPress = useRef(false)
+  const [lastClickTime, setLastClickTime] = useState(0)
+  const { isDesktop } = useBreakpoints()
+  const {
+    setSelectedItems,
+    setFocusedIndex,
+    setLastSelectedIndex,
```
rezk2ll, line 23: `setLastSelectedIndex` is not exported by the selection provider

src/hooks/useOnLongPress/helpers.js
```diff
@@ -0,0 +1,155 @@
...
+    } else {
+      return openLink(event)
+    }
+  }
+  // --
+
+  const currentTime = Date.now()
+  const isDoubleClick = currentTime - lastClickTime < DOUBLECLICKDELAY
+
+  if (isDoubleClick) {
```
rezk2ll, line 37:
```suggestion
  if (isDoubleClick && flag('drive.doubleclick.enabled')) {
```

rezk2ll (commented):
- When disabling the flag `drive.doubleclick.enabled`, the shift select is expected to function as usual, but it does not in this case.

We should have the default behavior ( shift arrow select ) to work as usual if the flag is not enabled.

- shift select is not working

src/hooks/useOnLongPress/helpers.spec.jsx
JF-Cozy: As comments (here and in the feature) and commit description try to explain, the func triggered in that case should be unique and simple as possible. I think we must refactor/fix shift selection before. So for now, the test is ignored.

src/modules/views/Folder/virtualized/helpers.js
JF-Cozy: in the code below this comment. Introduced here https://github.com/cozy/cozy-drive/pull/3442/commits/f31ccd627b0dff6d6b15e62cb031471a180c1988#diff-910cec369da31c0662775504f1f60d32463e533e625db0f46976cb9ce20958ca

src/hooks/useOnLongPress/index.js
JF-Cozy: good catch I forgot to save/commit the modification (I rebased the branch to add it in the penultimate commit

src/hooks/useOnLongPress/helpers.js
JF-Cozy: No need there is `return` inside the flag condition, the code is not triggered. I preferred to isolate this temporary code instead of merging it with the one that will remain.

rezk2ll (approved):

PR #3547, opened 2025-10-09
fix: Adjust validations of copy cut by shortcuts :bug:

lethemanh:
### Change:

- Paste action is disabled in Recent and Sharing.
- Check correct source folder of the files when moving.


src/contexts/ClipboardProvider.tsx
```diff
@@ -80,7 +83,7 @@ const initialState: ClipboardState = {
   operation: null,
   timestamp: null,
   cutItemIds: new Set(),
-  sourceFolderId: null,
+  sourceFolderIds: new Set(),
```
JF-Cozy, line 86: why using a Set instead of a classic Array? We used to get array with `xxxxxYyyyyIds` name func. I don't say we must not, just that the API between Array and Set are not the same so...

src/hooks/useKeyboardShortcuts.tsx
```diff
@@ -76,6 +80,15 @@ export const useKeyboardShortcuts = ({
       return
     }
 
+    const parentFolderIds = selectedItems.map(item => item.dir_id)
+    if (parentFolderIds.includes(SHARED_DRIVES_DIR_ID)) {
+      showAlert({
+        message: t('alert.cannot_move_shared_drive'),
+        severity: 'info'
```
JF-Cozy: we don't use `info` much with alert, is it ok with Swann?

src/modules/views/Recent/index.jsx
```diff
@@ -85,7 +83,7 @@ export const RecentView = () => {
   const result = useQuery(query.definition, query.options)
 
   useKeyboardShortcuts({
-    canPaste: hasClipboardData,
+    canPaste: false,
```
JF-Cozy: maybe we should set canPast as false by default in `useKeyboardShortcuts` ? So no need to specify here (just remove it)

src/hooks/useKeyboardShortcuts.tsx
```diff
@@ -76,6 +80,15 @@ export const useKeyboardShortcuts = ({
       return
     }
 
+    const parentFolderIds = selectedItems.map(item => item.dir_id)
+    if (parentFolderIds.includes(SHARED_DRIVES_DIR_ID)) {
+      showAlert({
+        message: t('alert.cannot_move_shared_drive'),
```
rezk2ll: this is not a move operation, a copy one

src/contexts/ClipboardProvider.tsx
lethemanh: Actually, `sourceFolderIds` refers to the file’s `dir_id`.
When I copy or cut multiple files from the same folder, those files share the same `dir_id`, which causes duplicate IDs => that’s why I use a `Set` to remove duplicates.

As for why I store multiple IDs: in views like `Recent` or `Sharings`, users can select files from different folders, so keeping multiple folder IDs helps check whether the user is pasting into one of the original source folders.

PR #3554, opened 2025-10-13
refactor: Refactor shift selection :recycle:

lethemanh:
### Change:

- Moved all logics that're related to shift selection out of `SelectionProvider`.
- Implemented new hook and helper functions for shift selection.
- Implemented unit test for shift selection.
- Modified `useLongPress` to be available with shift selection.
- Modified folder view component to trigger shift selection.

### DIFF
--- a/src/hooks/useShiftSelection/helpers.ts
+++ b/src/hooks/useShiftSelection/helpers.ts
@@ -1,3 +1,5 @@
+import cloneDeep from 'lodash/cloneDeep'
+
 import { IOCozyFile } from 'cozy-client/types/types'
 
 import type { SelectedItems } from '@/modules/selection/types'
@@ -41,21 +43,6 @@ export interface HandleShiftClickParams {
   endIdx: number
   selectedItems: SelectedItems
   items: IOCozyFile[]
-}
-
-/**
- * Returns a new selectedItems object without the removed key
- * @param {SelectedItems} selectedItems The current selected items object
- * @param {string} removedKey The key/ID of the item to remove from selection
- *
- * @returns {SelectedItems} A new SelectedItems object without the removed key
- */
-const returnValidSelections = (
-  selectedItems: SelectedItems,
-  removedKey: string
-): SelectedItems => {
-  const { [removedKey]: _, ...validSelections } = selectedItems
-  return validSelections
 }
 
 /**
@@ -128,8 +115,6 @@ const toggleSelection = ({
   isMovingToSelect,
   isItemSelected
 }: ToggleSelectionParams): SelectedItems => {
-  let newSelected = { ...selectedItems }
-
   // Identify which item to modify (depends on selection direction)
   const targetItem = isMovingToSelect
     ? items[currentIdx]
@@ -137,13 +122,20 @@ const toggleSelection = ({
     ? items[lastInteractedIdx]
     : items[currentIdx]
 
-  if (isMovingToSelect) {
-    newSelected[targetItem._id] = targetItem
-  } else {
-    newSelected = returnValidSelections(newSelected, targetItem._id)
-  }
-
-  return newSelected
+  return Object.entries(selectedItems).reduce<SelectedItems>(
+    (acc, [key, value]) => {
+      if (isMovingToSelect) {
+        acc[key] = value
+        acc[targetItem._id] = targetItem
+      } else {
+        if (key !== targetItem._id) {
+          acc[key] = value
+        }
+      }
+      return acc
+    },
+    {}
+  )
 }
 
 /**
@@ -232,20 +224,24 @@ export const handleShiftClick = ({
   selectedItems,
   items
 }: HandleShiftClickParams): HandleShiftSelectionResponse => {
-  let newSelectedItems = { ...selectedItems }
   const endItem = items[endIdx]
-  const isMovingToSelect = !Object.hasOwn(newSelectedItems, endItem._id)
+  const isMovingToSelect = !Object.hasOwn(selectedItems, endItem._id)
   const start = Math.min(startIdx, endIdx)
   const end = Math.max(startIdx, endIdx)
 
-  for (let i = start; i <= end; i++) {
+  const newSelectedItems = Array.from(
+    { length: end - start + 1 },
+    (_, i) => start + i
+  ).reduce((acc, i) => {
     const item = items[i]
     if (isMovingToSelect) {
-      newSelectedItems[item._id] = item
+      acc[item._id] = item
     } else {
-      newSelectedItems = returnValidSelections(newSelectedItems, item._id)
+      const { [item._id]: _, ...rest } = acc
+      return rest
     }
-  }
+    return acc
+  }, cloneDeep(selectedItems))
 
   return {
     newSelectedItems,


src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  for (let i = start; i <= end; i++) {
+    const item = items[i]
+    newSelectedItems[item._id] = isMovingToSelect ? item : undefined
+  }
+
+  setSelectedItems(newSelectedItems as SelectedItems)
+}
+
+const clamp = (maxLength: number, index: number): number =>
```
rezk2ll, line 55: Please add JSDoc to your functions

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Find the next index (in given direction) where selection state flips.
+ * This defines the next "boundary" for select/deselect operations.
+ */
+const findNextBoundaryIndex = ({
+  items,
+  startIdx,
+  direction,
+  isMovingToSelect,
```
rezk2ll, line 77: What is the purpose if this isMovingToSelect?
Please add JSDoc to explain what the prop in the object does / means.

The code should be simple to understand

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,223 @@
...
+    ? items[currentIdx]
+    : isItemSelected(items[lastInteractedIdx]._id)
+    ? items[lastInteractedIdx]
+    : items[currentIdx]
+
+  if (isMovingToSelect) {
+    newSelected[targetItem._id] = targetItem
+  } else {
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+    delete newSelected[targetItem._id]
```
rezk2ll: Please don't use delete, look [here](https://javascript.plainenglish.io/avoid-the-delete-keyword-in-javascript-87ff2a47f26c)

src/hooks/useShiftSelection/helpers.ts
lethemanh: The JSDoc has been added

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  setSelectedItems(newSelectedItems)
+}
+
+/**
+ * Clamps an index value to be within valid array bounds.
+ *
+ * @param maxLength - The maximum length of the array
+ * @param index - The index to clamp
```
rezk2ll: I would prefer a human generated JSDoc .

param {type} name description

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+  )
+
+  const lastInteractedIdx = useMemo(() => {
+    return lastInteractedItem
+      ? items.findIndex(item => item._id === lastInteractedItem)
+      : 0
+  }, [lastInteractedItem, items])
+
+  const selectItemByIds: SelectedItems = useMemo(() => {
+    return selectedItems.reduce<SelectedItems>(
```
rezk2ll, line 72: Are you sure selectedItems is an array?

it is an object in the selection provider
```ts
 const [selectedItems, setSelectedItems] = useState({})
```

src/modules/selection/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  hideSelectionBar: () => void
+
+  /** Clear all the selected items */
+  clearSelection: () => void
+
+  /** Whether the SelectionBar is visible or not */
+  isSelectionBarVisible: boolean
+
+  /** List of selected items as an array */
+  selectedItems: IOCozyFile[]
```
rezk2ll, line 21: it's not an array it's an object

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setSelectedItems,
+      setLastInteractedItem
+    ]
+  )
+
+  const viewTypeRef = useRef<ViewType>()
+  viewTypeRef.current = useMemo(() => viewType, [viewType])
+
+  const itemsRef = useRef<IOCozyFile[]>()
+  itemsRef.current = useMemo(() => items, [items])
```
rezk2ll: This is unnecessary and double-storing the value

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  useEffect(() => {
+    setIsSelectAll(selectedItems.length === itemsRef.current?.length)
+  }, [setIsSelectAll, selectedItems])
```
rezk2ll: Please avoid useEffect

src/modules/views/Folder/virtualized/Grid.jsx
```diff
@@ -9,82 +9,93 @@ import GridWrapper from './GridWrapper'
...
-          <>
-            {file.type != 'tempDirectory' ? (
-              <RightClickFileMenu
-                key={file?._id}
-                doc={file}
-                actions={actions.filter(action => !action.selectAllItems)}
-              >
-                <GridFile
+    const onToggleSelect = (fileId, event) => {
+      event.stopPropagation()
```
rezk2ll: why do you need to do it here outside the onSelect?

src/modules/views/Folder/virtualized/helpers.js
```diff
@@ -13,11 +13,7 @@ export const makeRows = ({ queryResults, IsAddingFolder, syncingFakeFile }) => {
     rows.push(syncingFakeFile)
   }
 
-  // TODO: we should not modify io.cozy.files for frontend purpose
-  return rows.map((row, index) => ({
-    ...row,
-    index
-  }))
+  return rows
```
rezk2ll, line 16: does this change have other impacts on virtualized table? @JF-Cozy

src/modules/views/Folder/virtualized/Table.jsx
```diff
@@ -44,92 +39,80 @@ const components = {
...
-  )
-  const { isNew } = useNewItemHighlightContext()
-  const sortedRow = useMemo(() => {
-    const sortedData = stableSort(rows, getComparator(order, orderBy))
-    return secondarySort(sortedData)
-  }, [rows, order, orderBy])
-
-  useShiftArrowsSelection({ items: sortedRow, viewType: 'list' }, tableRef)
+    const handleRowSelect = (row, event) => {
+      event.stopPropagation()
```
rezk2ll: same comment

src/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -129,6 +132,13 @@ const FolderViewBody = ({
 
   const { syncingFakeFile } = useSyncingFakeFile({ isEmpty, queryResults })
 
+  const onToggleSelect = (fileId, e) => {
+    setLastInteractedItem(fileId)
+    if (e.shiftKey) {
```
rezk2ll: we shouldn't check the keyboard event key in each place we use the onShiftClick

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+      selectItemByIds,
+      setSelectedItems,
+      setLastInteractedItem
+    ]
+  )
+
+  const viewTypeRef = useRef<ViewType>()
+  viewTypeRef.current = useMemo(() => viewType, [viewType])
+
+  const itemsRef = useRef<IOCozyFile[]>()
```
rezk2ll:
```suggestion
  const itemsRef = useRef<IOCozyFile[]>([])
```

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,312 @@
...
+  endIdx: number
+  selectedItems: SelectedItems
+  items: IOCozyFile[]
+  setSelectedItems: (items: SelectedItems) => void
+  setLastInteractedItem: (id: string | null) => void
+}
+
+/**
+ * Returns a new SelectedItems object with the specified key removed.
+ * This is a safe way to remove items from selection without using delete operator.
```
rezk2ll: AI comment slipped in.

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+  viewType?: ViewType
+}
+
+interface UseShiftSelectionReturn {
+  lastInteractedIdx: number
+  setLastInteractedItem: (id: string | null) => void
+  onShiftClick: (clickedItemId: string) => void
+}
+
+const useShiftSelection = (
```
rezk2ll, line 49: JSDoc

src/hooks/useShiftSelection/index.tsx
lethemanh: I was surprised too but `SelectionProvider` return `selectedItems` like this https://github.com/cozy/cozy-drive/pull/3554/files#diff-da897bfcd1d89d8d569b1a068a1708413ae8097334b3c324f16e9843ff91c9e5L247

src/modules/selection/types.ts
lethemanh: same as above

src/modules/views/Folder/virtualized/helpers.js
lethemanh: Actually, I reverted my change to this version.
There was a comment from Julien to not modify `io.cozy.files` https://github.com/cozy/cozy-drive/pull/3554/files#diff-910cec369da31c0662775504f1f60d32463e533e625db0f46976cb9ce20958caL16

src/hooks/useShiftSelection/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * When shift key is held and an item is clicked, selects or deselects all items
+   * between the last interacted item and the clicked item (inclusive).
+   *
+   * @param {string} clickedItemId - ID of the item that was clicked
+   * @param {KeyboardEvent} event - The keyboard event (must have shiftKey = true)
+   */
+  const onShiftClick = useCallback(
+    (clickedItemId: string, event: KeyboardEvent) => {
+      event.stopPropagation()
```
rezk2ll: Wouldn't this make sense to move it after line 94:

```ts
if (!event.shiftKey) return

event.stopPropagation()
```

if it's not a shift key there is no reason to stop the event propagation

src/modules/views/Folder/virtualized/helpers.js
JF-Cozy: it depends how we use file.index in the table. But if we don't, no there is no impact "by default" with the basic virtualized table from cozy-ui

lethemanh: Before that, the `file.index` was used in shift selection only. But now, we don't depend on it anymore.

src/modules/selection/SelectionProvider.jsx
```diff
@@ -42,26 +34,15 @@ const SelectionProvider = ({ children }) => {
   const location = useLocation()
   const [selectedItems, setSelectedItems] = useState({})
   const [isSelectionBarOpen, setSelectionBarOpen] = useState(false)
-  const [focusedIndex, setFocusedIndex] = useState(0)
-  const [lastSelectedIndex, setLastSelectedIndex] = useState(null)
-  const [isKeyboardNavigating, setIsKeyboardNavigating] = useState(false)
-  const itemsListRef = useRef([])
+  const [isSelectAll, setIsSelectAll] = useState(false)
```
JF-Cozy, line 37: nit: this add is not part of the extraction refactor, it's a fix/feat isn't it? It should be in an other commit

src/modules/views/Folder/virtualized/helpers.js
JF-Cozy, line 16: nit: good thing to remove this, but it should be in an other commit I think

src/modules/views/Folder/virtualized/FolderViewBodyContent.jsx
JF-Cozy, line 1: I don't understand why we deal with sort here, since the goal of this commit is to extract logic of shift selection from selection provider.

Besides about the main goal, I understand we simply removed it and not just extracted it... 🤔

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @returns {HandleShiftSelectionResponse}
+ */
+export const handleShiftClick = ({
+  startIdx,
+  endIdx,
+  selectedItems,
+  items
+}: HandleShiftClickParams): HandleShiftSelectionResponse => {
+  let newSelectedItems = { ...selectedItems }
```
JF-Cozy: be careful here, doing that only create a shallow copy, not a deep copy. The difference is that simple values or not related, but complex values (object and array) are. It could lead to issues in the futur. To be sure, use `cloneDeep` from lodash.

nit: Besides I'm not a big fan of `let` usage, if it's possible, maybe use reduce instead (for objects you can `Object.entries(monObjet).reduce((acc, [key, value]) => {}, something)` )

src/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -115,6 +132,11 @@ const FolderViewBody = ({
 
   const { syncingFakeFile } = useSyncingFakeFile({ isEmpty, queryResults })
 
+  const onToggleSelect = (fileId, e) => {
+    setLastInteractedItem(fileId)
+    onShiftClick(fileId, e)
```
JF-Cozy, line 137: the goal here is to trigger shift selection. Is it possible to use only on func `onShiftClick`? Is `setLastInteractedItem` useful for something else?

src/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -67,6 +69,21 @@ const FolderViewBody = ({
   const folderViewRef = useRef()
   const IsAddingFolder = useSelector(isTypingNewFolderName)
 
+  const allFiles = useMemo(() => {
```
JF-Cozy, line 72: I think we should deal with query result differently. A query have loading status, and its life cycle should be dealt independently. The component should get only the result by prop.

src/hooks/useOnLongPress/helpers.js
```diff
@@ -43,6 +44,7 @@ export const handleClick = ({
     setSelectedItems({ [file._id]: file })
   }
 
+  onInteractWithFile?.(file?._id, event)
```
JF-Cozy: why `?.` here ?

src/modules/views/Folder/virtualized/Table.jsx
```diff
@@ -101,6 +101,7 @@ const Table = forwardRef(
                   currentFolderId={currentFolderId}
                   withFilePath={withFilePath}
                   actions={actions}
+                  onInteractWithFile={onSelect}
```
JF-Cozy: why using a new prop to pass something already in the main API? `onSelect` is a prop in `VirtualizedTable` why not using it?

src/hooks/useShiftSelection/helpers.ts
```diff
@@ -210,7 +202,7 @@ export const handleShiftArrow = ({
     direction,
     isMovingToSelect,
     isItemSelected,
-    isReturnCurrent: Object.keys(newSelectedItems).length < 1
+    isReturnCurrent: Object.keys(newSelectedItems).length <= 1
```
JF-Cozy, line 197: it's a fix, not a refactor

src/modules/views/Folder/virtualized/FolderViewBodyContent.jsx
lethemanh: I've moved it from this commit, and placed into the commit [refactor: Modified folder view component to trigger shift selection ♻️](https://github.com/cozy/cozy-drive/pull/3554/commits/555eaa7295d940453a112098dbe3bc2a1a6457ca)

src/modules/views/Folder/FolderViewBody.jsx
lethemanh: not every toggle is shift click, and `setLastInteractedItem` is used for shift arrow also. So I think using `onShiftClick` only is not enough

src/hooks/useShiftSelection/helpers.ts
lethemanh: this commit is a fix 
CleanShot 2025-10-15 at 19 50 19@2x

src/modules/selection/SelectionProvider.jsx
lethemanh: Before the implementation of shift selection, `isSelectAll` is a state. Btw, `isSelectAll` has involvement of `itemsListRef` (which is mainly used for shift selection), so after moving shift selection I think it's better to revert it.

src/modules/views/Folder/virtualized/helpers.js
lethemanh: I think it's a part of remove old shift selection flows, so placing it in this comment is reasonable.

src/modules/views/Folder/FolderViewBody.jsx
lethemanh: loading status is used in this comment and I don't want to affect the current logic, so I must compute it like the current.

src/modules/views/Folder/virtualized/Table.jsx
lethemanh: `onSelect` in `VirtualizedTable` only trigger when clicking on checkbox, so when clicking on the row to select file I have to pass the prop `onSelect` to trigger functions for shift selection. Btw, I've change the name of the prop to avoid confusion.

src/hooks/useOnLongPress/helpers.js
lethemanh: removed

src/hooks/useShiftSelection/helpers.ts
lethemanh: I've removed `let` and updated to use reduce

src/modules/views/Folder/virtualized/helpers.js
JF-Cozy: Yes good point, I was thinking more of a commit that would fix this object error separately, but anyway, that's not how it was implemented in the first place, so...

src/modules/views/Folder/FolderViewBody.jsx
JF-Cozy: ok. Maybe the main problem here is that is hard to understand what the API is and how use it. But jsdocs helps a lot :)

src/modules/views/Folder/FolderViewBody.jsx
JF-Cozy: I don't understand. You mean we already did that before?

lethemanh: I mean, we already used `queryResults` as prop before, and also used `fetchStatus` in this component, so I don't want to change that prop to prevent any effect on the current logic.

PR #3567, opened 2025-10-20
feat: Save user's sorting choice :sparkles:

lethemanh:
### Change:

When user changes the sorting, all the pages will apply the change.

### Related to:

https://github.com/cozy/cozy-stack/pull/4599

https://github.com/cozy/cozy-store/pull/956

https://github.com/cozy/cozy-doctypes/pull/283

https://github.com/cozy/cozy-client/pull/1643

https://github.com/cozy/cozy-ui/pull/2879

### Flag:

`drive.save-sort-choice.enabled`

### DIFF
--- a/src/hooks/useFolderSort/index.spec.jsx
+++ b/src/hooks/useFolderSort/index.spec.jsx
@@ -235,7 +235,7 @@ describe('useFolderSort', () => {
       })
 
       expect(mockClient.save).toHaveBeenCalledWith({
-        _type: DOCTYPE_DRIVE_SETTINGS,
+        ...existingSettings,
         attributes: {
           attribute: 'updated_at',
           order: 'desc'
--- a/src/hooks/useFolderSort/index.ts
+++ b/src/hooks/useFolderSort/index.ts
@@ -45,7 +45,7 @@ const useFolderSort = (
           Q(DOCTYPE_DRIVE_SETTINGS)
         )) as QueryResult
 
-        if (!data || !data.length) return
+        if (!data?.length) return
 
         setCurrentSort(data[0]?.attributes)
       } catch (error) {
@@ -84,16 +84,15 @@ const useFolderSort = (
           Q(DOCTYPE_DRIVE_SETTINGS)
         )) as QueryResult
 
-        const settingsToSave: DriveSettings =
-          data && data.length
-            ? {
-                ...data[0],
-                attributes: { attribute, order },
-              }
-            : {
-                _type: DOCTYPE_DRIVE_SETTINGS,
-                attributes: { attribute, order },
-              }
+        const settingsToSave: DriveSettings = data?.length
+          ? {
+              ...data[0],
+              attributes: { attribute, order }
+            }
+          : {
+              _type: DOCTYPE_DRIVE_SETTINGS,
+              attributes: { attribute, order }
+            }
 
         await client.save(settingsToSave)
         logger.info('Sort settings persisted', { attribute, order })


src/modules/navigation/duck/hooks.ts
```diff
@@ -1,28 +1,98 @@
-import { useCallback } from 'react'
+import { useCallback, useEffect } from 'react'
```
rezk2ll: First thing I see in the PR 😅

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,28 +1,98 @@
...
+    | undefined
+  ]
+}
+
 const useFolderSort = (folderId: string): [Sort, (props: Sort) => void] => {
+  const client = useClient()
+
   const defaultSort: Sort =
-    flag('drive.default-updated-at-sort.enabled') || folderId === TRASH_DIR_ID
+    flag('drive.default-updated-at-sort.enabled') ||
```
rezk2ll: please remove this flag. It's not useful anymore since we need to respect the user sorting choice .
This flag was added to force every one to have the updated at sort by default

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,28 +1,98 @@
...
+
+        const settings = result.data[0]?.attributes
+        const sortOrder = settings?.order ?? defaultSort.order
+        const sortAttribute = settings?.orderBy ?? defaultSort.attribute
+        dispatch(sortFolder(folderId, sortAttribute, sortOrder))
+      } catch (error) {
+        logger.error('Failed to load sorting settings:', error)
+        dispatch(sortFolder(folderId, defaultSort.attribute, defaultSort.order))
+      }
+    }
```
zatteo: Can you use directly `useQuery` hook?

lethemanh: I've tried but useQuery does not return data in this case

src/modules/navigation/duck/hooks.ts
lethemanh: it has been refactored

src/modules/navigation/duck/hooks.ts
lethemanh: Sorry, I've tried again, maybe I missed something in the previous. I will refactor by this way.

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,93 @@
...
   const dispatch = useDispatch()
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-  const currentSort = (useSelector(state => getSort(state)) ||
-    defaultSort) as Sort
-  const setOrder = useCallback(
-    ({ attribute, order }: Sort) => {
+
+  const driveSettingsResult = useQuery(Q(DOCTYPE_DRIVES_SETTINGS), {
+    as: DOCTYPE_DRIVES_SETTINGS,
+    enabled: !!client
```
zatteo: What is the purpose of the `enabled: !!client`? I never saw it in Cozy codebase (but maybe there is a good reason)

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,93 @@
...
+  const setSortsOrder = useCallback(
+    async ({ attribute, order }: Sort) => {
       dispatch(sortFolder(folderId, attribute, order))
+
+      if (!client) return
+
+      try {
+        const { data } = (await client.query(
+          Q(DOCTYPE_DRIVES_SETTINGS)
+        )) as QueryResult
```
zatteo: `driveSettingsResult` variable should be always up to date so no need to do a new query?

lethemanh: In the future, maybe we have a other settings, so I think we have to query the drive settings to get all the current attributes and override the sorting settings only.

src/modules/navigation/duck/hooks.ts
lethemanh: just want to ensure client was load before get drive settings

src/modules/navigation/duck/hooks.ts
zatteo: Normally when you use useQuery, data is always up to date _if you set up realtime_ for the doctype which is not the case here. So you are right.

src/modules/views/Folder/virtualized/FolderViewBody.jsx
```diff
@@ -22,7 +22,8 @@ const FolderViewBody = ({
   canUpload = true,
   canDrag,
   withFilePath = false,
-  sortOrder
+  sortOrder,
```
rezk2ll: i disagree with adding two new props to handle this:

We already use `orderProps`, which has

```ts
      orderProps={{
            order,
            orderBy,
            setOrder,
            setOrderBy
          }}

```

i saw you deleted the use of this orderProps, which breaks the virtualized table view also.

Please use the existing prop, and we should have the same props

src/modules/navigation/duck/hooks.ts
rezk2ll: not needed, please remove it

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,89 @@
...
+    const currentSortAttribute = settings?.attribute ?? defaultSort.attribute
+    dispatch(sortFolder(folderId, currentSortAttribute, currentSortOrder))
+    initRef.current = true
+  }
+
+  const setSortsOrder = useCallback(
+    async ({ attribute, order }: Sort) => {
       dispatch(sortFolder(folderId, attribute, order))
+
+      if (!client) return
```
rezk2ll:
```suggestion
      if (!client) {
        logger.warn('Cannot persist sort: client unavailable')
        return
      }
```

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,89 @@
...
+            order,
+            attribute
+          }
+        })
+      } catch (error) {
+        logger.error('Failed to save sorting preference:', error)
+      }
     },
-    [dispatch, folderId]
+    [client, dispatch, folderId]
```
rezk2ll: You have driveSettingsResult at the top, you can reuse it

```ts
     try {
        const existingSettings = driveSettingsResult.data[0]

        const settingsToSave: DriveSettings = existingSettings
          ? {
              ...existingSettings,
              attributes: { attribute, order }
            }
          : {
              _type: DOCTYPE_DRIVE_SETTINGS,
              attributes: { attribute, order }
            }

        await client.save(settingsToSave)
        logger.info('Sort settings persisted', { attribute, order })
      } catch (error) {
        logger.error('Failed to persist sort settings', { error, attribute, order })
      }
    },
    [client, dispatch, folderId, driveSettingsResult.data]
```

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,89 @@
...
+  }) as QueryResult
+
+  const initRef = useRef(false)
+  if (!initRef.current && driveSettingsResult.fetchStatus === 'loaded') {
+    const settings = driveSettingsResult.data[0]?.attributes
+    const currentSortOrder = settings?.order ?? defaultSort.order
+    const currentSortAttribute = settings?.attribute ?? defaultSort.attribute
+    dispatch(sortFolder(folderId, currentSortAttribute, currentSortOrder))
+    initRef.current = true
+  }
```
rezk2ll: please keep it simple 

```suggestion
const [settingsInit, setsettingsInit] = useState(false)

if (!settingsInit && driveSettingsResult.fetchStatus === 'loaded') {
  const settings = driveSettingsResult.data[0]?.attributes

  if (settings) {
    dispatch(sortFolder(folderId, settings.attribute, settings.order))
  }
  
  setsettingsInit(true)
}
```

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,89 @@
...
+          Q(DOCTYPE_DRIVE_SETTINGS)
+        )) as QueryResult
+
+        const existing =
+          data[0] && Object.keys(data[0]).length
+            ? data[0]
+            : ({
+                _type: DOCTYPE_DRIVE_SETTINGS,
+                attributes: {}
+              } as unknown as QueryResult['data'][0])
```
rezk2ll: we are lying to TS

src/modules/views/Public/PublicFolderView.spec.jsx
```diff
@@ -67,13 +67,25 @@ jest.mock('components/pushClient', () => ({
 useSharingContext.mockReturnValue({ byDocId: [] })
 
 describe('Public View', () => {
+  beforeAll(() => {
+    // Mock console methods to suppress redux-logger output
```
rezk2ll:
```suggestion
```

also it would be nice to have new tests for the useFolderSort hook and the new props we add

src/modules/navigation/duck/hooks.ts
```diff
@@ -1,33 +1,93 @@
...
+  data?: DriveSettings[]
+  fetchStatus?: string
+}
+
 const useFolderSort = (folderId: string): [Sort, (props: Sort) => void] => {
+  const client = useClient()
+
   const defaultSort: Sort =
-    flag('drive.default-updated-at-sort.enabled') || folderId === TRASH_DIR_ID
+    folderId === TRASH_DIR_ID || folderId === 'recent'
```
rezk2ll:
```suggestion
    folderId === TRASH_DIR_ID || folderId === RECENT_FOLDER_ID
```

PR #3572, opened 2025-10-20
fix(FilesRealTimeQueries): MapIterator.forEach not being compatible with outdated browsers 🐛

rezk2ll:
### The bug

- in sentry we started to notice an Exception being fired `TypeError: a.forEach is not a function`

image

image

### The cause

in the FilesRealTimeQueries we use a Map as a Buffer for the file operations and then we get the list of processed files by calling the `Map.keys()` which returns a `MapIterator`

```ts
const bufferCreatedFiles = new Map()

.... // later on

const fileIdsToProcess = bufferFiles.keys() // <-- returns MapIterator

// later on

fileIdsToProcess.forEach(fileId => bufferFiles.delete(fileId))
```

the MapIterator.forEach method was added in 2024 ( in ES2024 ) [mdn page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/forEach)

image

so for browsers with versions earlier than supported ones the `MapIterator.forEach` is undefined ( and the buffer is never cleared ).

#### checking affected browsers

image
image
image
image

all the browsers who had the issue are outdated

#### replicating the bug using an outdated browser ( firefox 102 )

https://github.com/user-attachments/assets/011d1fbb-7118-4024-8fc5-bba77b805ab7

### what's changed

- Replaced `forEach` with `for...of` loop (ES2015 compatible)
- Fixed Map size check: used `bufferFiles.size` instead of `bufferFiles.length`

### DIFF
--- a/src/components/FilesRealTimeQueries.jsx
+++ b/src/components/FilesRealTimeQueries.jsx
@@ -59,7 +59,7 @@ const processEvents = async (client, mutationType) => {
     mutationFn = Mutations.deleteDocument
     multipleMutationFn = Mutations.deleteDocuments
   }
-  if (bufferFiles.length === 0) return
+  if (bufferFiles.size === 0) return
 
   const fileIdsToProcess = bufferFiles.keys()
   const filesByFolder = groupFilesByFolder(bufferFiles)


zatteo:
> Don't we have babel for this kind of stuff? Should we update our targeted browsers?

Agree that this should not happen. I can take the point to see where we are with babel.

rezk2ll:
> Don't we have babel for this kind of stuff? Should we update our targeted browsers?

the target browsers we have: https://github.com/cozy/cozy-libs/blob/master/packages/browserslist-config-cozy/index.js

```
module.exports = [
  'last 2 Chrome major versions',
  'last 2 Firefox major versions',
  'last 2 FirefoxAndroid major versions',
  'last 2 Safari major versions',
  'last 3 iOS major versions',
  'last 2 Edge major versions',
  'Firefox ESR',
  'Android 4.4',
  '> 1% in FR',
  'not dead',
  'not ie <= 11'
]
```

rezk2ll:
also in our babel preset we have `useBuiltIns` here: https://babeljs.io/docs/babel-preset-env#usebuiltins

our preset: https://github.com/cozy/cozy-libs/blob/5ae1ac7e1cae4b6cab7958798ef174ec1751fcaa/packages/babel-preset-cozy-app/index.js#L19

this means that in RSbuild the `output.polyfill` is set to off 

it should be `usage`, we can add it here: https://github.com/cozy/cozy-drive/blob/master/rsbuild.config.mjs#L12
```js
const mergedConfig = mergeRsbuildConfig(config, {
  output: {
    polyfill: 'usage' 
})
```

JF-Cozy:
> > I think we can just ignore this issue

> Please don't take any decision like that without consensus.

It was not my intention.

Besides I don't think we should fix the code step by step by using old school code to have better compatibility. I think we should go deeper like you said and find a global solution. Unless it was a critical issue and we were on a rush, we should let the bug here to helps us finding a proper solution...

PR #3574, opened 2025-10-22
feat: Add folder color customization feature

doubleface:
* Add folder personalization menu action : 
image

* Folder personalization dialog : 
   - desktop : 
image

  - mobile : 
image

* Use personalized icon for folders

This feature will need modifications in cozy-stack to allow `decorations` attribute in `io.cozy.files`

### flag

`drive.folder-personalization.enabled`

### DIFF
--- a/src/modules/actions/components/personalizeFolder.tsx
+++ b/src/modules/actions/components/personalizeFolder.tsx
@@ -1,6 +1,7 @@
 import React, { forwardRef } from 'react'
 
 import { IOCozyFile } from 'cozy-client/types/types'
+import flag from 'cozy-flags'
 import { Action } from 'cozy-ui/transpiled/react/ActionsMenu/Actions'
 import ActionsMenuItem from 'cozy-ui/transpiled/react/ActionsMenu/ActionsMenuItem'
 import Icon from 'cozy-ui/transpiled/react/Icon'
@@ -22,6 +23,7 @@ const personalizeFolder = ({ t, navigate }: personalizeFolderProps): Action => {
     label,
     icon,
     displayCondition: docs =>
+      flag('drive.folder-personalization.enabled') &&
       docs.length === 1 &&
       docs[0].type === 'directory',
     action: (docs: IOCozyFile[]): void => {
@@ -29,7 +31,7 @@ const personalizeFolder = ({ t, navigate }: personalizeFolderProps): Action => {
         navigate(`/folder/${docs[0]._id}/personalize`)
       }
     },
-    Component: forwardRef(function personalizeFolder(props, ref) {
+    Component: forwardRef(function PersonalizeFolder(props, ref) {
       return (
         <ActionsMenuItem {...props} ref={ref}>
           <ListItemIcon>
--- a/src/modules/filelist/icons/FileIconMime.jsx
+++ b/src/modules/filelist/icons/FileIconMime.jsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 
 import { isDirectory } from 'cozy-client/dist/models/file'
+import flag from 'cozy-flags'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 
 import { isEncryptedFolder } from '@/lib/encryption'
@@ -15,7 +16,7 @@ const FileIconMime = ({ file, size = 32, isEncrypted = false }) => {
   if (
     isDir &&
     flag('drive.folder-personalization.enabled') &&
-    (file.metadata?.decorations?.color)
+    file.metadata?.decorations?.color
   ) {
     return (
       <CustomizedIcon
--- a/src/modules/navigation/AppRoute.jsx
+++ b/src/modules/navigation/AppRoute.jsx
@@ -84,7 +84,9 @@ const AppRoute = () => (
         element={<Navigate to={ROOT_DIR_ID} replace={true} />}
       />
       <Route path="folder/:folderId" element={<DriveFolderView />}>
-        <Route path="personalize" element={<FolderCustomizer />} />
+        {flag('drive.folder-personalization.enabled') && (
+          <Route path="personalize" element={<FolderCustomizer />} />
+        )}
         <Route
           path="file/:fileId"
           element={<OutletWrapper Component={FilesViewerDrive} />}


src/locales/en.json
```diff
@@ -15,7 +15,7 @@
     "btn-client": "Get Twake Drive for desktop",
     "btn-client-web": "Get Twake",
     "btn-client-mobile": "Take your personnal cloud with you: install %{name} on all your devices!",
-    "banner-txt-client": "Get %{name} for Desktop and synchronise your files safely to make them accessible at all times.",
+    "banner-txt-client": "Get %{name} for toto and synchronise your files safely to make them accessible at all times.",
```
rezk2ll: who is this toto?

src/modules/actions/components/personalizeFolder.tsx
```diff
@@ -0,0 +1,47 @@
...
+const personalizeFolder = ({ t, navigate }: personalizeFolderProps): Action => {
+  const icon = PaletteIcon
+  const label = t('actions.personalizeFolder.label')
+
+  return {
+    name: 'personalizeFolder',
+    label,
+    icon,
+    displayCondition: docs =>
+      !!flag('drive.folder-personalization.enabled') &&
```
rezk2ll:
```suggestion
      flag('drive.folder-personalization.enabled') &&
```
I think the conversion is not necessary here since the null can work here also as a falsy value

src/modules/filelist/icons/FileIconMime.jsx
```diff
@@ -2,23 +2,39 @@ import PropTypes from 'prop-types'
...
+  } else {
+    return (
+      <Icon
+        icon={getMimeTypeIcon(isDir, file.name, file.mime, {
+          isEncrypted: isDirEncrypted
+        })}
+        size={size}
+      />
+    )
+  }
```
rezk2ll:
```suggestion
  return isDir &&
    flag('drive.folder-personalization.enabled') &&
    file.decorations?.color ? (
    <CustomizedIcon selectedColor={file.decorations.color} size={size} />
  ) : (
    <Icon
      icon={getMimeTypeIcon(isDir, file.name, file.mime, {
        isEncrypted: isDirEncrypted
      })}
      size={size}
    />
  )
```

The isEncrypted was applied anywhere in the previous version

zatteo:
Code looks good.

Is this new "Personalize folder" menu item also available in other views like "Sharing" ? Because in this case be careful to routing.

Is it available for shared drive recipients? I think it is not possible.

src/locales/en.json
doubleface: oups

src/modules/actions/components/personalizeFolder.tsx
doubleface: right

src/modules/filelist/icons/FileIconMime.jsx
doubleface: Good catch! Done

src/modules/navigation/AppRoute.jsx
```diff
@@ -83,6 +84,7 @@ const AppRoute = () => (
         element={<Navigate to={ROOT_DIR_ID} replace={true} />}
       />
       <Route path="folder/:folderId" element={<DriveFolderView />}>
+        <Route path="personalize" element={<FolderCustomizer />} />
```
JF-Cozy: good point to use a route. FYI I think Bob didn't want to use route anymore for modal. Don't know what the final call about that... we can keep it here to be consistent, but maybe we can simplify these thing in the future... we have to ask him.

src/components/ColorPicker/ColorPicker.jsx
```diff
@@ -0,0 +1,90 @@
...
+]
+
+const NB_COLUMNS_MOBILE = 6
+const NB_COLUMNS_DESKTOP = 8
+const CELL_HEIGHT_MOBILE = 56
+const CELL_HEIGHT_DESKTOP = 40
+const CIRCLE_SIZE_MOBILE = 40
+const CIRCLE_SIZE_DESKTOP = 36
+const ICON_SIZE_MOBILE = 18
+const ICON_SIZE_DESKTOP = 12
```
JF-Cozy, line 46: We should extract all this in a helper to clean the code, and add maybe a "Default to first color (dark gray)" var

jest.config.js
JF-Cozy, line 1: some comment in the commit description would be great to explain the approach

src/modules/actions/components/personalizeFolder.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <Icon icon={icon} />
+          </ListItemIcon>
+          <ListItemText primary={label} />
+        </ActionsMenuItem>
+      )
+    })
+  }
+}
+
+export { personalizeFolder }
```
JF-Cozy, line 47: we have a makeAction func in cozy-ui to simplify and harmonize this process

src/modules/views/Folder/CustomizedIcon.jsx
```diff
@@ -0,0 +1,33 @@
...
+        height: size
+      }}
+    >
+      <Icon icon={FolderIcon} size={size || 32} />
+      {maskStyle && <div className={styles.iconOverlay} style={maskStyle} />}
+    </div>
+  )
+}
+
+CustomizedIcon.displayName = 'CustomizedIcon'
```
JF-Cozy: nit: no need I think if there is no wrapper around the main func

src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -0,0 +1,113 @@
...
+import { CustomizedIcon } from './CustomizedIcon'
+
+import { ColorPicker } from '@/components/ColorPicker/ColorPicker'
+import logger from '@/lib/logger'
+import { buildFileOrFolderByIdQuery } from '@/queries'
+
+export const FolderCustomizer = () => {
+  const navigate = useNavigate()
+  const { t } = useI18n()
+  const [selectedColor, setSelectedColor] = useState('#46a2ff') // Default to first color (dark gray)
```
JF-Cozy: should be a VAR and not hardcoded value...

src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -0,0 +1,113 @@
...
+  const { isMobile } = useBreakpoints()
+  const { showAlert } = useAlert()
+  const client = useClient()
+  const { folderId } = useParams()
+
+  const folderQuery = buildFileOrFolderByIdQuery(folderId)
```
JF-Cozy, line 23: If there is a query, we should extract it in a specific component to deal with it (and the loading process) and then pass the result value to a dumb component. It's simpler that way, and we probably can remove the useEffect after that.

src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -0,0 +1,113 @@
...
+  const folderResult = useQuery(folderQuery.definition, folderQuery.options)
+  const folder = folderResult?.data
+
+  useEffect(() => {
+    if (folder?.metadata?.decorations?.color) {
+      setSelectedColor(folder.metadata.decorations.color)
+    }
+  }, [folder?.metadata?.decorations?.color])
+
+  const handleColorSelect = color => {
```
JF-Cozy, line 47: no need to wrap a state setter with simple func

src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -0,0 +1,113 @@
...
+      showAlert({
+        message: t('FolderCustomizer.error'),
+        severity: 'error'
+      })
+    } finally {
+      navigate(`/folder/${folder.dir_id}`)
+    }
+  }
+
+  if (!folder) {
```
JF-Cozy, line 74: should be in the "query deal wrapper"

src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -0,0 +1,113 @@
...
+        <>
+          <Grid container justifyContent="center" className="u-mb-1">
+            <Grid item>
+              <CustomizedIcon selectedColor={selectedColor} size={52} />
+            </Grid>
+          </Grid>
+          <Grid
+            container
+            justifyContent="center"
+            className={`u-mt-1 u-mb-1${isMobile ? '' : '-half'}`}
```
JF-Cozy: you have `u-mb-1-s` to deal with responsive stuff... probably something like `u-mb-1-s u-mb-half-l`

src/styles/folder-customizer.styl
```diff
@@ -0,0 +1,29 @@
+.iconContainer
+    position relative
+    display inline-block
```
JF-Cozy: you can use u-classes for these two

src/styles/folder-customizer.styl
```diff
@@ -0,0 +1,29 @@
...
+    mix-blend-mode screen
+
+    // Mask properties (image URL will be set via inline style)
+    mask-size 105% // avoids black border around folder icon
+    mask-repeat no-repeat
+    mask-position center
+
+    -webkit-mask-size 105%
+    -webkit-mask-repeat no-repeat
+    -webkit-mask-position center
```
JF-Cozy: no need for webkit

src/styles/folder-customizer.styl
```diff
@@ -0,0 +1,29 @@
...
+
+    pointer-events none
```
JF-Cozy: try to use css only for stuff not in u-classes

src/modules/filelist/icons/FileIconMime.jsx
```diff
@@ -6,19 +6,29 @@ import Icon from 'cozy-ui/transpiled/react/Icon'
...
-      icon={getMimeTypeIcon(isDir, file.name, file.mime, {
-        isEncrypted: isDirEncrypted
-      })}
-      size={size}
-    />
-  )
+  if (isDir && file.metadata?.decorations?.color) {
+    return (
+      <CustomizedIcon
+        selectedColor={file.metadata?.decorations?.color}
```
JF-Cozy: no need for `?.` here

src/modules/navigation/AppRoute.jsx
```diff
@@ -84,7 +84,9 @@ const AppRoute = () => (
         element={<Navigate to={ROOT_DIR_ID} replace={true} />}
       />
       <Route path="folder/:folderId" element={<DriveFolderView />}>
-        <Route path="personalize" element={<FolderCustomizer />} />
+        {flag('drive.folder-personalization.enabled') ? (
+          <Route path="personalize" element={<FolderCustomizer />} />
+        ) : null}
```
JF-Cozy: nit: you can use `&&` instead of `? : null`

src/styles/folder-customizer.styl
doubleface: Did not find any equivalent class. But this rule was actually  not needed. Removed it

src/modules/navigation/AppRoute.jsx
doubleface: I get a "illustrations-desert.svg:3 No routes matched location "/folder/e507eb70566c70d7c54c9fb42e060052/personalize" when I use ??. Don't know why

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: Where is it documented ? Can't find anything about it, even in the code. And it does not work for u-mb-half-l, don't know why

src/components/ColorPicker/ColorPicker.jsx
doubleface: Sorry, I don't see the point of extracting constants from a small component like this one.

jest.config.js
doubleface: done

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: Let's use an index in the list of colors from the ColorPicker. This way there won't be hardcoded color value (the list of colors excepted)

src/modules/actions/components/personalizeFolder.tsx
doubleface: makeAction is used in https://github.com/cozy/cozy-drive/blob/014be161ca057b92619ef8947a459903238af098/src/modules/views/Drive/DriveFolderView.jsx#L184

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: Extracted a DumbFolderCustomizer and now FolderCustomizer only handles the loading of the folder.
Also add the handling of loading state

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: Removed this the DumbFolderCustomizer component

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: done

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: Ok https://docs.cozy.io/cozy-ui/styleguide/section-settings.html#kssref-settings-breakpoints

Now it works

src/modules/navigation/AppRoute.jsx
doubleface: done

rezk2ll (approved):

src/components/ColorPicker/ColorPicker.jsx
JF-Cozy: The point is to keep files as clean as possible. Starting adding stuff not directly relative to the logic is creating a path for the future that we put everything in one single file. At the end it's unreadable. Besides those var are not mandatory to understand what the code does, it's not a primary value.

src/modules/actions/components/personalizeFolder.tsx
JF-Cozy: no there is `makeAction` and `makeActions`. Here i'm talking about the first one ;)

src/styles/folder-customizer.styl
JF-Cozy: i'm not talking about pointer-events but all the others

doubleface: I also removed the others

src/modules/actions/components/personalizeFolder.tsx
doubleface: Ok, fixed in https://github.com/cozy/cozy-drive/pull/3592

src/components/ColorPicker/ColorPicker.jsx
doubleface: Ok, thanks for the explanation. This is done in https://github.com/cozy/cozy-drive/pull/3592

PR #3582, opened 2025-10-24
feat(favicon): dynamically change favicon for OnlyOffice documents :sparkles:

lethemanh:
### Change:

When opening OO files (doc, sheet, slide), the favicon will be changed by file type.

### What has been done?

1. Change favicon when opening doc, sheet, slide.
2. Update to default favicon when opening other file types, closing the file viewer.
3. Test cases for `useUpdateFavicon`.
4. Add feature flag.

#### In normal folder:

https://github.com/user-attachments/assets/0fad24a6-213f-4bde-9c5f-922dc5c37555

#### In shared drive:

https://github.com/user-attachments/assets/766e4589-da29-44c2-813b-3ba1ff679787

### Flag:

`drive.update-favicon.enabled`


src/components/useUpdateFavicon.tsx
```diff
@@ -0,0 +1,52 @@
...
+  ) {
+    updateFavicon(defaultFavicon)
+    return
+  }
+
+  const faviconByMimetype = {
+    text: '/favicons/icon-onlyoffice-text.ico',
+    sheet: '/favicons/icon-onlyoffice-sheet.ico',
+    slide: '/favicons/icon-onlyoffice-slide.ico'
+  }
```
zatteo: To move outside hooks (and var name capitalized ?)

src/components/useUpdateFavicon.tsx
```diff
@@ -0,0 +1,52 @@
...
+    links.forEach(link => {
+      link.rel = 'icon'
+      link.type = 'image/svg+xml'
+      link.href = url
+
+      document.head.appendChild(link)
+    })
+  }, [])
+
+  const defaultFavicon = '/assets/favicon.ico'
```
zatteo: To move outside hook (and var name capitalized ?)

src/components/useUpdateFavicon.tsx
```diff
@@ -0,0 +1,52 @@
...
+      document.querySelectorAll<HTMLLinkElement>("link[rel~='icon']")
+
+    links.forEach(link => {
+      link.rel = 'icon'
+      link.type = 'image/svg+xml'
+      link.href = url
+
+      document.head.appendChild(link)
+    })
+  }, [])
```
zatteo: We can move this function also outside hook no?

src/components/useUpdateFavicon.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const defaultFavicon = '/assets/favicon.ico'
+
+  if (
+    fetchStatus !== 'loaded' ||
+    !file ||
+    !flag('drive.update-favicon.enabled')
+  ) {
+    updateFavicon(defaultFavicon)
```
rezk2ll: we don't need to do anything here.

src/components/useUpdateFavicon.tsx
rezk2ll: - we don't need to call the appendChild in the forEach loop.
- what happens if no links found?
- what if the correct favicon is already applied? we shouldn't re-update the DOM

```js
// handle non existing case
if (!link) {
  link = document.createElement('link');
  link.rel = 'icon';
  link.type = 'image/svg+xml';
  document.head.appendChild(link);
}

// handle
if (link.href !== url) {
  // do the other required stuf, ref type etc
  link.href = url;
}
```

src/components/useUpdateFavicon.tsx
```diff
@@ -0,0 +1,52 @@
...
+    slide: '/favicons/icon-onlyoffice-slide.ico'
+  }
+
+  const type = getFileMimetype(faviconByMimetype)(
+    file.mime,
+    file.name
+  ) as string
+
+  const favicon = get(faviconByMimetype, type, defaultFavicon) as string
+  updateFavicon(favicon)
```
rezk2ll: all this shouldn't exist in the body of the hook can be moved ( that if it's useful ) somewhere else.

also this hook can be simplified link this:

```ts
import { useCallback } from 'react';

export function useFavicon() {
  return useCallback((url: string): void => {
    let link = document.querySelector<HTMLLinkElement>("link[rel~='icon']");

    if (!link) {
      link = document.createElement('link');
      link.rel = 'icon';
      link.type = 'image/svg+xml';
      document.head.appendChild(link);
    }

    if (link.href !== url) {
      link.href = url;
    }
  }, []);
}
```

src/components/useUpdateFavicon.spec.jsx
```diff
@@ -0,0 +1,175 @@
...
+      mockQuerySelectorAll.mockReturnValue([mockLink])
+
+      renderHook(() => useUpdateFavicon(undefined, 'loaded'))
+
+      expect(mockQuerySelectorAll).toHaveBeenCalledWith("link[rel~='icon']")
+      expect(mockAppendChild).toHaveBeenCalled()
+      expect(mockLink.href).toBe('/assets/favicon.ico')
+    })
+
+    it('should use default favicon when flag is disabled', () => {
```
rezk2ll: So we test if the flag is disabled when in the `when conditions are not met for custom favicon` section.
It should be the other way around.

Also, the flag is meant to be removed, so we should not consider it as a criterion in the tests.
( you can also put a TODO to remove etc, but it would be best to test as if it's on )

rezk2ll (changes requested):
thank you for the PR and providing tests, there are some points we should fix:

- we should not update the favicon when there is no need to. in the code we apply the default icon in the hook when it's not needed. if it's an OO file yes update the favicon, you can use cleanup when unmounting the component to revert back to the default one.
- some tests just for the 'testing ceremony" and not relevant enough. keep tests relevant to the changes
- for the PR title / commit message i would put "feat(favicon): dynamically change favicon for OnlyOffice documents"

another approach to achieve this is :
1 - we have a reference of the default favicon somewhere
2 - if we are on an onlyoffice file change the favicon.
3 - if we are not on an onlyoffice file and the current favicon is different than the defaut one we reset it to the old one.

src/components/useUpdateFavicon.tsx
lethemanh: the reason I use forEach loop here because we have 2 type of favicon size 16x16 and 32x32 so I have to update all of them

src/hooks/useUpdateFavicon/helpers.ts
```diff
@@ -0,0 +1,33 @@
...
+  const type = getFileMimetype(FAVICON_BY_MIMETYPE)(
+    file.mime,
+    file.name
+  ) as string
+
+  const faviconUrl = get(FAVICON_BY_MIMETYPE, type) as string
+
+  const links = document.querySelectorAll<HTMLLinkElement>("link[rel~='icon']")
+
+  if (!links.length || !faviconUrl) return
```
rezk2ll: If no favicon found we check if the default favicon is set, otherwise set the default one.

If no links are found we add a new link if the favicon is found

src/hooks/useUpdateFavicon/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  file: IOCozyFile | undefined,
+  fetchStatus: string
+): void => {
+  useEffect(() => {
+    const originalFavicon =
+      document.querySelector<HTMLLinkElement>("link[rel~='icon']")
+    const originalHref = originalFavicon?.href
+
+    return () => {
+      if (originalFavicon && originalHref) {
```
rezk2ll:
```suggestion
      if (originalHref) {
```

src/hooks/useUpdateFavicon/helpers.ts
```diff
@@ -0,0 +1,33 @@
+import get from 'lodash/get'
```
rezk2ll: Let's not use lodash, plus it's an object you can just object[key]

src/hooks/useUpdateFavicon/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { IOCozyFile } from 'cozy-client/types/types'
+import flag from 'cozy-flags'
+
+import { updateFavicon } from './helpers'
+
+const useUpdateFavicon = (
+  file: IOCozyFile | undefined,
+  fetchStatus: string
+): void => {
+  useEffect(() => {
```
rezk2ll, line 16: we should check the flag before executing anything in this hook

src/hooks/useUpdateFavicon/helpers.ts
lethemanh: > If no favicon found we check if the default favicon is set, otherwise set the default one.

I think we just need to check if `faviconUrl` exists, and don't care about it's default favicon or not. Example: in the error page, where it has a different favicon, so it should not be overridden in this case.

nit: I don't think we have any case that the `links` does not exist, but I've added the check btw.

rezk2ll:
When testing, the OnlyOffice icon does not reset when closing the document or switching to a new file in the viewer.
Something is missing in the favicon reset.

The cleanup function does not run, and we don't detect that the file has changed

https://github.com/user-attachments/assets/93716fd1-0dd3-494b-8ae8-e32b0b47440e

You can fix it by passing the originalLink to the updateFavicon helper

```ts
  if (!faviconUrl && originalHref) {
    const link = document.createElement('link')
    appendFaviconToHead(link, originalHref)
    return
  }
```

and to detect the changes in the useUpdateFavicon hook

in the initial hook, store the original favicon

```ts
  useEffect(() => {
    if (!flag('drive.update-favicon.enabled')) {
      return
    }

    // get the initially loaded favicon and store it using setOriginalHref

  }, [setOriginalHref])
```

```ts
  useEffect(() => {
    if (
      fetchStatus !== 'loaded' ||
      !file ||
      !flag('drive.update-favicon.enabled')
    ) {
      return
    }

  // get the file favicon url
  // get the currently set favicon URL
  // decide if changing this favicon is necessary
  // if the favicon URL is not found, you just pass the originalHref ( original URL) as a fallback
  
  // you can update updateFavicon to just accept a link and do the file type checking here

    updateFavicon(file, originalHref)
  }, [file, fetchStatus, originalHref])
```

You can get something like this

https://github.com/user-attachments/assets/4dec3429-637c-47c1-ba3e-429cc1a50820

rezk2ll:
@lethemanh let's :telephone_receiver:

PR #3583, opened 2025-10-24
fix: client queryS being called without an id ( query options )

rezk2ll:
### what's changed
- added the getDriveSettingQuery for io.cozy.drive.settings queries
- used getDriveSettingQuery in saving sort choice
- fixed switch context view using rogue query ( without an id )

### DIFF
--- a/jestHelpers/setup.js
+++ b/jestHelpers/setup.js
@@ -2,6 +2,27 @@ import React from 'react'
 
 global.cozy = {}
 
+global.fetch = jest.fn(() =>
+  Promise.resolve({
+    ok: true,
+    status: 200,
+    headers: {
+      get: jest.fn(name => {
+        if (name.toLowerCase() === 'content-type') {
+          return 'application/json'
+        }
+        return null
+      })
+    },
+    json: async () => ({
+      rows: [],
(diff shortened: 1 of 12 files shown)


zatteo:
I do not understand, when do we use io.cozy.files.settings and io.cozy.drive.settings?

zatteo:
> > I do not understand, when do we use io.cozy.files.settings and io.cozy.drive.settings?
> 
> in ViewSwitcherContextProvider we use ` io.cozy.files.settings` and in useFolderSort we use `io.cozy.drive.settings`.
> 
> i think it's best to mutualize them and use `io.cozy.drive.settings`, i think it's a great oppurtunity to do so since i started this refacto

Do we have two "Drive settings" doctype? I did not remember about io.cozy.files.settings

shepilov:
@rezk2ll is it still relevant or close? do we now override this "choice" in the settings update?

rezk2ll:
> @rezk2ll is it still relevant or close? do we now override this "choice" in the settings update?

Hi it's still relevant and it's more of a refactoring chore in the drive . And it's needed for the requests performance.

In general we should never have requests without id

PR #3589, opened 2025-10-28
feat: enable multiple files and folders drag&drop

rezk2ll:
### The issue
when selecting multiple files and try to drag&drop it only works on the dragged file but not on all the files

### changes
provided the selected items in the context when using the virtualized grid and table views that will be used in cozy-ui to perform the action on the selected files

### demo

https://github.com/user-attachments/assets/8651352d-74e5-4fed-9ac5-b5aadd4de88c

### DIFF
--- a/src/modules/views/Folder/virtualized/Table.jsx
+++ b/src/modules/views/Folder/virtualized/Table.jsx
@@ -85,7 +85,7 @@ const Table = forwardRef(
         style={{ outline: 'none' }}
       >
         <VirtuosoTableDnd
-          context={{ actions }}
+          context={{ actions, selectedItems, isSelectedItem }}
           components={components}
           rows={rows}
           columns={columns}


src/modules/views/Folder/virtualized/Grid.jsx
```diff
@@ -83,7 +85,7 @@ const Grid = forwardRef(
             </>
           )}
           endReached={fetchMore}
-          context={actions}
+          context={{ actions, selectedItems, isSelectedItem }}
```
lethemanh, line 88: Sorry, I'm confused about this change. The old one receive `context` as `actions` object, but the new one `actions` just a key of `context`

rezk2ll: we call the VirtualizedGridListDnd with that context, which calls VirtualizedGridList here https://github.com/cozy/cozy-ui/blob/master/react/GridList/Virtualized/Dnd/index.jsx#L50
after that it calls VirtuosoGrid here with the context https://github.com/cozy/cozy-ui/blob/master/react/GridList/Virtualized/index.jsx#L10
which is then used in the GridItem here https://github.com/cozy/cozy-ui/blob/72ad25b1a0694e9d2bbbbb2be7d97087e5d7dcc5/react/GridList/Virtualized/Dnd/GridItem.jsx#L34

in the GridItem we expect to have the selectedItems when performing the drag action.

image

zatteo (approved):

PR #3591, opened 2025-10-29
feat: Add icons to folder customizer

doubleface:
[Capture vidéo du 2025-10-29 16-39-36.webm](https://github.com/user-attachments/assets/cc7e4481-93d2-4dc7-b23f-afed38ce040b)

- Now, you can also add an icon to folders. 
- It is also possible to select the color of the icon. 
- The list of icons that have been previously assigned is stored in localStorage, up to a limit of 8.
- Option to delete the selection for the color and icon separately.

### DIFF
--- a/src/components/IconPicker/IconColorPicker.jsx
+++ b/src/components/IconPicker/IconColorPicker.jsx
@@ -10,23 +10,9 @@ import Paper from 'cozy-ui/transpiled/react/Paper'
 import Popper from 'cozy-ui/transpiled/react/Popper'
 
 import { getIcon } from './IconIndex'
+import { ICON_COLORS } from './constants'
 
 import styles from '@/styles/folder-customizer.styl'
-
-export const ICON_COLORS = [
-  '#2c2c2c',
-  '#9aa0a6',
-  '#e0e0e0',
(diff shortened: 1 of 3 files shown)


src/components/IconPicker/IconColorPicker.jsx
```diff
@@ -0,0 +1,80 @@
...
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import IconButton from 'cozy-ui/transpiled/react/IconButton'
+import Paper from 'cozy-ui/transpiled/react/Paper'
+import Popper from 'cozy-ui/transpiled/react/Popper'
+
+import { getIcon } from './IconIndex'
+
+import styles from '@/styles/folder-customizer.styl'
+
+export const ICON_COLORS = [
```
rezk2ll: Those colors are not shared with the previous folder colors? if yes we can reuse the previous ones

doubleface: No, only one color is shared by the two lists

doubleface: At least it reminded me to put the constants in a separated file for IconPicker too.

PR #3592, opened 2025-10-30
feat: Add folder personalization to shared folders

doubleface:
It is now possible to have folder personalization on shared folders (for the owner only). To make it possible, I had handle the routing differently, which also avoids the display of the content of the folder while we customize it.

Also added some small fixes

- **fix: Remove unused foldercustomizer.styl**
- **fix: Black border around folder icon in small size**

### DIFF
--- a/src/modules/views/Folder/FolderCustomizer.jsx
+++ b/src/modules/views/Folder/FolderCustomizer.jsx
@@ -92,7 +92,7 @@ const DumbFolderCustomizer = ({ folder, onClose }) => {
             className="u-mt-1 u-mb-1-t u-mb-1-half"
           >
             <Grid item>
-              <Typography variant="h6" noWrap>
+              <Typography variant="h6">
                 {t('FolderCustomizer.description')}
               </Typography>
             </Grid>


src/modules/views/Folder/FolderCustomizer.jsx
```diff
@@ -38,6 +38,14 @@ FolderCustomizer.displayName = 'FolderCustomizer'
 const DumbFolderCustomizer = ({ folder }) => {
   const navigate = useNavigate()
   const { t } = useI18n()
+
+  const handleClose = () => {
+    if (window.history.length > 1) {
+      navigate(-1)
```
JF-Cozy: we only use specified route, we don't rely on browser history. I don't have right now an example to illustrate why but I remember we had lot of trouble when doing that way.

doubleface: FYI we do it like this here too : https://github.com/cozy/cozy-drive/blob/20cbcaceebdb07b1a57db6dc4ec118f039ea2197/src/hooks/useRedirectLink.jsx#L95-L104

Let's see if I find another way

rezk2ll (approved):

src/modules/views/Folder/FolderCustomizer.jsx
doubleface: @JF-Cozy Is it better using location.state.from ? [`fe6ecdb` (#3592)](https://github.com/cozy/cozy-drive/pull/3592/commits/fe6ecdb8a10282fafe96c291683a2a81946b11bc)

doubleface: Anyway, we do not use router any more to display folder customizer

PR #3594, opened 2025-10-30
fix: sorting on mobile and legacy view when no choice persistance is disabled

rezk2ll:
### issue

When the `drive.save-sort-choice.enabled` flag was disabled, sorting would not work on mobile (or desktop with the legacy view) due to a state management issue between parent and child components.

### changes:
- useFolderSort: Always update local state (setCurrentSort) before
   checking flag, so UI updates immediately regardless of persistence
- DriveFolderView & TrashFolderView: Pass orderProps (sortOrder,
   setOrder, isSettingsLoaded) to legacy FolderViewBody
- FolderViewBody: Use orderProps from parent when provided, fallback
   to an internal hook for backward compatibility.
- Synchronize the mobile and desktop sort ( to not cause saving empty sorting order in the state )

### demo

https://github.com/user-attachments/assets/f3100955-3a6d-4e50-a11d-888a58474171

### DIFF
--- a/src/hooks/useFolderSort/index.ts
+++ b/src/hooks/useFolderSort/index.ts
@@ -65,6 +65,8 @@ const useFolderSort = (
 
   const setSortOrder = useCallback(
     async ({ attribute, order }: Sort) => {
+      setCurrentSort({ attribute, order })
+
       if (!flag('drive.save-sort-choice.enabled')) {
         logger.warn(
           'Cannot persist sort: flag drive.save-sort-choice.enabled is not enabled'
--- a/src/modules/views/Folder/FolderViewBody.jsx
+++ b/src/modules/views/Folder/FolderViewBody.jsx
@@ -61,7 +61,8 @@ const FolderViewBody = ({
   canUpload = true,
   withFilePath = false,
   refreshFolderContent = null,
-  extraColumns
+  extraColumns,
+  orderProps
 }) => {
   const { isDesktop } = useBreakpoints()
   const navigate = useNavigate()
@@ -110,8 +111,13 @@ const FolderViewBody = ({
 
   const { isBigThumbnail } = useThumbnailSizeContext()
   const { sharingsValue } = useContext(AcceptingSharingContext)
-  const [sortOrder, setSortOrder, isSettingsLoaded] =
+  const [internalSortOrder, internalSetSortOrder, internalIsSettingsLoaded] =
     useFolderSort(currentFolderId)
+  const sortOrder = orderProps?.sortOrder ?? internalSortOrder
+  const setSortOrder = orderProps?.setOrder ?? internalSetSortOrder
+  const isSettingsLoaded =
+    orderProps?.isSettingsLoaded ?? internalIsSettingsLoaded
+
   const vaultClient = useVaultClient()
   const changeSortOrder = useCallback(
     (_, attribute, order) => setSortOrder({ attribute, order }),
--- a/src/modules/folder/components/FolderBody.jsx
+++ b/src/modules/folder/components/FolderBody.jsx
@@ -77,8 +77,7 @@ const FolderBody = ({
   const { viewType, switchView } = useViewSwitcherContext()
 
   const changeSortOrder = useCallback(
-    (folderId_legacy, attribute, order) =>
-      setSortOrder({ sortAttribute: attribute, sortOrder: order }),
+    (folderId_legacy, attribute, order) => setSortOrder({ attribute, order }),
     [setSortOrder]
   )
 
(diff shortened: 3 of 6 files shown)


src/hooks/useFolderSort/index.ts
```diff
@@ -65,6 +65,8 @@ const useFolderSort = (
 
   const setSortOrder = useCallback(
     async ({ attribute, order }: Sort) => {
+      setCurrentSort({ attribute, order })
```
lethemanh, line 68: This change will apply for desktop view also, there for it will trigger the API call every time we change the sorting choice.
Which is not our expected behavior in desktop view.

src/modules/views/Folder/FolderViewBody.jsx
```diff
@@ -110,8 +111,13 @@ const FolderViewBody = ({
 
   const { isBigThumbnail } = useThumbnailSizeContext()
   const { sharingsValue } = useContext(AcceptingSharingContext)
-  const [sortOrder, setSortOrder, isSettingsLoaded] =
+  const [internalSortOrder, internalSetSortOrder, internalIsSettingsLoaded] =
```
lethemanh, line 114: I don't think we need `useFolderSort` here anymore, we can use the `orderProps` directly.

src/hooks/useFolderSort/index.ts
rezk2ll: > This change will apply for desktop view also, there for it will trigger the API call every time we change the sorting choice. Which is not our expected behavior in desktop view.

That's how we save the sorting choice, by calling the API.
In the previous implementation, we did not update the state and relied on the virtualized folder view to do the sorting.
In the legacy view, the sorting will not be triggered => won't work on mobile.

By using isMobile breakpoints like you did in your PR, you only trigger the change for the legacy view on mobile and not on desktop.


https://github.com/user-attachments/assets/2cd85d0e-e0a1-4beb-bc92-6aa8da0bb80d

src/modules/views/Folder/FolderViewBody.jsx
rezk2ll: That will work on a virtualized folder, but will not work on the legacy view, which relies on the internal state to sort

src/hooks/useFolderSort/index.ts
lethemanh: yeah, but for that the API will be triggered even in virtualized view

rezk2ll: We need the API call to be made when we change the sort. This is the idea of the feature.
we change the sort order => we make an API call to save it

src/modules/views/Folder/FolderViewBody.jsx
lethemanh: you mean the `orderProps.sortOrder` does not update? Because, I've tried in my PR and the sorting in legacy view worked as expected.

rezk2ll: it will work on the normal folder view.

but it will not work on the Public view: src/modules/views/Public/PublicFolderView.jsx

we do not pass the props and we can't persist the data for the public view

```tsx
  <FolderViewBody
    actions={actions}
    queryResults={[filesResult]}
    canSort={false}
    currentFolderId={currentFolderId}
    refreshFolderContent={refreshFolderContent}
    canUpload={hasWritePermissions}
    extraColumns={extraColumns}
    isPublic={true}
  />
```

It will not work also in the Recent view: src/modules/views/Recent/index.jsx

We do not pass the props here, and also we should not persist the sorting choice nor fetch it in this view

```tsx
  <FolderViewBody
    actions={actions}
    queryResults={[result]}
    canSort={false}
    withFilePath={true}
    extraColumns={extraColumns}
  />
```

when you test locally, you should not only test the obvious main view, test with virtualization enabled and disabled, test with mobile and desktop view, test the other views not just the main view.

lethemanh: ah, sorry my implementation was slidely different from yours so the prop update as well, btw it's ok.

src/hooks/useFolderSort/index.ts
lethemanh: I mean the API to fetch the list of files and folders will be triggered every time the sorting choice changes. That's why we did not `setCurrentSort` in `setSortOrder` in the previous.

rezk2ll: No, that is not true, we don't refetch when we update the state, we don't have a watcher on the currentSort state, nor does it trigger any events


https://github.com/user-attachments/assets/a8a074c3-943a-4991-a0c5-8174818bd115

lethemanh (approved):

PR #3595, opened 2025-10-30
fix(grid view): prevent overflow of long parent folder paths on mobile

rezk2ll:
### the issue

In the Recent view grid layout on mobile, very long parent folder paths were not being truncated properly, causing layout issues.

mobile-grid

### what's changed
- Added max-width and overflow constraints to .fil-file-description in grid view

### demo
image

### DIFF
--- a/src/styles/filelist.styl
+++ b/src/styles/filelist.styl
@@ -211,6 +211,10 @@ column-width-thumbnail-bigger = 7rem
         max-width 6rem
         margin 0 auto
 
+    .fil-file-description
+        max-width 6rem
+        overflow hidden
+
 .fil-file-thumbnail
     position relative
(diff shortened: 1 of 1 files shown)


src/styles/filelist.styl
```diff
@@ -211,6 +211,10 @@ column-width-thumbnail-bigger = 7rem
         max-width 6rem
         margin 0 auto
 
+    .fil-file-description
+        max-width 6rem
+        overflow hidden
+
```
Crash--, line 217: I don't know, but we had in the past, a component for that: https://docs.cozy.io/cozy-ui/react/#/Legacy?id=filepath 

But i don't know why this component is in Legacy now

PR #3596, opened 2025-10-30
fix: Resolve sorting in mobile view :bug:

lethemanh:
### Problem:

Since mobile view does not use virtualized folder view, we don't have frontend sort. The sort in mobile only works when fetching API again.

### Solution:

To keep sorting behavior in mobile view, we need to update `currentSort` so the query could trigger every time changing the sorting choice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,5 +131,6 @@
     "redux-mock-store": "1.5.4",
     "redux-thunk": "2.4.2",
     "whatwg-fetch": "3.0.0"
-  }
+  },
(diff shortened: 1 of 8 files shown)


rezk2ll:
i already created a PR to fix it here: 
https://github.com/cozy/cozy-drive/pull/3594

lethemanh:
> i already created a PR to fix it here: #3594

Oh I don't see it.
Btw, I saw you changed the `currentSort` in all the views, but I think it will cause fetching API again in desktop view also. My change only changed `currentSort` in mobile view. So that, it can keep the old behavior.

PR #3480, opened 2025-09-17
feat: Implement duplicating files in shared drive :sparkles:

lethemanh:
### Change:

Implement duplicating files in recipient's shared drives.

### Note:

~This PR needs the updates of folder picker in [cozy/cozy-drive#3479](https://github.com/cozy/cozy-drive/pull/3479), I will set it as Draft and open again when the PR [cozy/cozy-drive#3479](https://github.com/cozy/cozy-drive/pull/3479) is merged.~

PR #3598, opened 2025-10-31
fix: Display correct right click menu in file viewer :bug:

lethemanh:
### Change:

In file viewer, we need to display `RightClickFileMenu` instead of `RightClickAddMenu`.

#### Normal file viewer:

CleanShot 2025-10-31 at 15 46 43@2x

#### Shared drive file viewer:

CleanShot 2025-10-31 at 15 40 52@2x

### DIFF
--- a/src/components/RightClick/RightClickFileMenu.jsx
+++ b/src/components/RightClick/RightClickFileMenu.jsx
@@ -4,12 +4,22 @@ import ActionsMenu from 'cozy-ui/transpiled/react/ActionsMenu'
 import { useBreakpoints } from 'cozy-ui/transpiled/react/providers/Breakpoints'
 
 import { useRightClick } from '@/components/RightClick/RightClickProvider'
+import { getContextMenuActions } from '@/modules/actions/helpers'
 import { useSelectionContext } from '@/modules/selection/SelectionProvider'
 
-const RightClickFileMenu = ({ doc, actions, disabled, children, ...props }) => {
+const RightClickFileMenu = ({
+  doc,
+  actions,
+  disabled,
+  children,
+  prefixMenuId,
+  ...props
+}) => {
   const { position, isOpen, onOpen, onClose } = useRightClick()
   const { isDesktop } = useBreakpoints()
   const { selectedItems, isItemSelected } = useSelectionContext()
+
+  const contextMenuActions = getContextMenuActions(actions)
 
   if (!children) return null
   if (disabled || !isDesktop)
@@ -28,16 +38,18 @@ const RightClickFileMenu = ({ doc, actions, disabled, children, ...props }) => {
           ? React.cloneElement(child, {
               ...props,
               onContextMenu: ev => {
-                onOpen(ev, `FileMenu-${doc._id}`)
+                onOpen(ev, `${prefixMenuId ?? 'FileMenu'}-${doc._id}}`)
+                ev.preventDefault()
+                ev.stopPropagation()
               }
             })
           : null
       )}
-      {isOpen(`FileMenu-${doc._id}`) && (
+      {isOpen(`${prefixMenuId ?? 'FileMenu'}-${doc._id}}`) && (
         <ActionsMenu
           open
           docs={isItemSelected(doc._id) ? selectedItems : [doc]}
-          actions={actions}
+          actions={contextMenuActions}
           anchorReference="anchorPosition"
           anchorPosition={{ top: position.mouseY, left: position.mouseX }}
           autoClose
--- a/src/hooks/useMoreMenuActions.jsx
+++ b/src/hooks/useMoreMenuActions.jsx
@@ -50,7 +50,6 @@ export const useMoreMenuActions = file => {
   const showPrintAction = isPDFDoc && isPrintAvailable
   const isCozySharing = window.location.pathname === '/preview'
   const isSharedDrive = window.location.href.includes('/shareddrive/')
-  const isDisplayInSharedDrive = !(isSharedDrive && !canWriteToCurrentFolder)
 
   const actions = makeActions(
     [
@@ -61,10 +60,10 @@ export const useMoreMenuActions = file => {
       download,
       showPrintAction && print,
       hr,
-      moveTo,
-      isDisplayInSharedDrive && duplicateTo,
-      isDisplayInSharedDrive && addToFavorites,
-      isDisplayInSharedDrive && removeFromFavorites,
+      !isSharedDrive && moveTo, // TO DO: Remove condtion when moving is available in shared drive
+      !isSharedDrive && duplicateTo, // TO DO: Remove condtion when duplicating is available in shared drive
+      addToFavorites,
+      removeFromFavorites,
       hr,
       versions,
       hr,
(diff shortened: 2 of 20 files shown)


src/modules/views/Folder/virtualized/Table.jsx
```diff
@@ -15,9 +15,14 @@ import { useNewItemHighlightContext } from '@/modules/upload/NewItemHighlightPro
 const TableRow = forwardRef(({ item, context, children, ...props }, ref) => {
   const { isItemCut } = useClipboardContext()
   const isCut = isItemCut(item._id)
+  const { actions } = context
 
   return (
-    <RightClickFileMenu doc={item} actions={context.actions} {...props}>
+    <RightClickFileMenu
+      doc={item}
+      actions={actions.filter(action => !action.selectAllItems)}
```
rezk2ll: This is not the right way to do it; check how the `versions` action does it ( it does not display when having multiple files ) you can do the other way around.

src/hooks/useMoreMenuActions.jsx
```diff
@@ -60,10 +59,10 @@ export const useMoreMenuActions = file => {
       download,
       showPrintAction && print,
       hr,
-      moveTo,
-      isDisplayInSharedDrive && duplicateTo,
-      isDisplayInSharedDrive && addToFavorites,
-      isDisplayInSharedDrive && removeFromFavorites,
+      !isSharedDrive && moveTo,
```
rezk2ll: why? in your other PRs https://github.com/cozy/cozy-drive/pull/3479 and https://github.com/cozy/cozy-drive/pull/3480 you are implementing the move to and duplicate to for shared drives yet you disable it here?

src/components/RightClick/RightClickFileMenu.jsx
```diff
@@ -28,12 +35,14 @@ const RightClickFileMenu = ({ doc, actions, disabled, children, ...props }) => {
...
-                onOpen(ev, `FileMenu-${doc._id}`)
+                onOpen(ev, `${prefixMenuId ?? 'FileMenu'}-${doc._id}}`)
+                ev.preventDefault()
+                ev.stopPropagation()
               }
             })
           : null
       )}
-      {isOpen(`FileMenu-${doc._id}`) && (
+      {isOpen(`${prefixMenuId ?? 'FileMenu'}-${doc._id}}`) && (
```
rezk2ll, line 48: why this needed?

src/hooks/useMoreMenuActions.jsx
lethemanh: Yeah, because those features haven't merged yet, so I cannot display the move and duplicate at the moment, it will be reopened once those PRs are merged. The `addToFavorites` and `removeFromFavorites` menus do not exist in shared drive so it should be disabled in this case anyway.

src/components/RightClick/RightClickFileMenu.jsx
lethemanh: The viewer is displayed above the list view => there will be 2 context menus opened with same `doc._id` on right-click in file viewer (one in the list, one is the current file viewer).
Changing the id will prevent it.

CleanShot 2025-10-31 at 16 51 15@2x

src/hooks/useMoreMenuActions.jsx
lethemanh: I've added TO DO comments remark to remove the condition when the features are available for shared drive.

src/modules/actions/helpers.js
```diff
@@ -25,3 +25,12 @@ export const navigateToModalWithMultipleFile = ({
...
 }
+
+export const driveActionsToContextMenuActions = (driveActions = []) => {
+  return driveActions.filter(driveAction => {
+    const action = Object.values(driveAction)[0]
+    return (
+      action.displayInContextMenu === undefined || action.displayInContextMenu
+    )
+  })
+}
```
rezk2ll: i liked the displayInContextMenu solution to handle displaying the action in context menu and top bar.

however this function needs a little refactoring,
- the name is off, it suggest it does some converting or mutation, instead it's just a filter. you can rename it to `getContextMenuActions`
- L33 check is weird, you can just check the explicit value: `!== false`
- missing JSDoc
- add unit tests for this helper

you can make it more clean like this:
```ts
/**
 * Returns the context menu visible actions
 * 
 * @param {Object[]} actions - the list of actions
 * @returns {Object[]} - the list of actions to be displayed
 */
export const getContextMenuActions = (actions = []) =>
  actions.filter(action =>
    Object.values(action)[0]?.displayInContextMenu !== false
  )

```

src/modules/filelist/File.jsx
```diff
@@ -138,6 +139,8 @@ const File = ({
     canInteractWithFile &&= canInteractWith(attributes)
   }
 
+  const convertedActions = driveActionsToContextMenuActions(actions)
```
rezk2ll: the name convertedActions is off, you can rename it to `contextMenuActions`

src/modules/navigation/AppRoute.jsx
```diff
@@ -89,8 +89,11 @@ const AppRoute = () => (
         >
           <Route path="v/revision" element={<FileHistory />} />
           <Route path="v/share" element={<ShareFileView />} />
-          <Route path="v/move" element={<MoveFilesView />} />
-          <Route path="v/duplicate" element={<FolderDuplicateView />} />
+          <Route path="v/move" element={<MoveFilesView isOpenInViewer />} />
+          <Route
```
rezk2ll: the change in this file is needed? i see it's a lint change but it's not the topic of this PR

src/modules/views/Modal/MoveFilesView.jsx
```diff
@@ -27,7 +27,7 @@ const MoveFilesView = () => {
 
   if (hasQueryBeenLoaded(fileResult) && fileResult.data && displayedFolder) {
     const onClose = () => {
-      navigate('..', { replace: true })
+      navigate(isOpenInViewer ? '../..' : '..', { replace: true })
```
rezk2ll, line 38: this is weird and not clear.

why ../.. and why ..

please explain in the code what is going on

src/components/RightClick/RightClickFileMenu.jsx
```diff
@@ -4,13 +4,23 @@ import ActionsMenu from 'cozy-ui/transpiled/react/ActionsMenu'
...
+  disabled,
+  children,
+  prefixMenuId,
+  ...props
+}) => {
   const { position, isOpen, onOpen, onClose } = useRightClick()
   const { isDesktop } = useBreakpoints()
   const { selectedItems, isItemSelected } = useSelectionContext()
 
+  const convertedActions = driveActionsToContextMenuActions(actions)
```
rezk2ll: Please rename it to `contextMenuActions`

src/modules/navigation/AppRoute.jsx
lethemanh: Yes, it's needed. In file viewer after moving success it will move back to folder view instead of file viewer because the file will not exist in the current folder anymore.

src/modules/views/Modal/MoveFilesView.jsx
lethemanh: In case moving file from file viewer, after moving success the file will not exist in the current folder so we should navigate to current folder view by using `../../`.
Btw, I've add a new function named `onMovingSuccess` to handle for this case, because the other `onClose` success does not need this check.

PR #3599, opened 2025-11-03
feat: Apply folder virtualized view in public view :sparkles:

lethemanh:
### Change:

For now, public view will display virutalized table in desktop and displays legacy table in mobile.

### Result:

https://github.com/user-attachments/assets/e941f459-20a9-4239-8829-2597bc51b8f1

### DIFF
--- a/src/modules/views/Public/PublicFolderView.jsx
+++ b/src/modules/views/Public/PublicFolderView.jsx
@@ -27,6 +27,7 @@ import FolderViewBody from '../Folder/FolderViewBody'
 import FolderViewBreadcrumb from '../Folder/FolderViewBreadcrumb'
 import FolderViewHeader from '../Folder/FolderViewHeader'
 import OldFolderViewBreadcrumb from '../Folder/OldFolderViewBreadcrumb'
+import FolderViewBodyVz from '../Folder/virtualized/FolderViewBody'
 
 import useHead from '@/components/useHead'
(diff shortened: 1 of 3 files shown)


src/modules/views/Public/PublicFolderView.jsx
```diff
@@ -221,16 +233,28 @@ const PublicFolderView = () => {
...
-            actions={actions}
-            queryResults={[filesResult]}
-            canSort={false}
-            currentFolderId={currentFolderId}
-            refreshFolderContent={refreshFolderContent}
-            canUpload={hasWritePermissions}
-            extraColumns={extraColumns}
-            isPublic={true}
-          />
+          {!isMobile ? (
```
rezk2ll: Let's also check if virtualization is enabled. to be in sync with the other views. until the day we decide to get rid of legacy view

src/hooks/useKeyboardShortcuts.tsx
```diff
@@ -30,6 +30,7 @@ interface UseKeyboardShortcutsProps {
   items?: IOCozyFile[]
   sharingContext?: unknown
   allowCopy?: boolean
+  allowCut?: boolean
```
rezk2ll: Are these allowCut changes related to the new view?

It seems that something is missing from the useKeyboardShortcuts hooks that were added here.
i suggest you create a new PR with those changes and keep this PR only related to the public view

PR #3603, opened 2025-11-03
fix(OnlyOffice): Duplicate edit button in toolbar

rezk2ll:
### the issue
In the OnlyOffice toolbar, we have duplicate edit buttons

### cause
We had a condition that was trying to handle both public and private views with a single variable, which led to confusion. The button was then rendered in multiple places

### changes

Refactored the edit button visibility logic in the OnlyOffice toolbar to prevent duplicate buttons from appearing. Separated public and private view logic with.


src/modules/views/OnlyOffice/Toolbar/index.jsx
```diff
@@ -134,7 +128,7 @@ const Toolbar = ({ sharingInfos }) => {
       {!isPublic && isEditorReady && (
         <>
           <Sharing file={file} />
-          {showEditButton ? <EditButton /> : null}
+          {canEdit && <EditButton />}
```
zatteo, line 131: In public you do not have the two buttons with this logic?

rezk2ll: in public, we have the `!isPublic` at L128 to prevent it

zatteo: Of course did not see it (it's still tricky, but less)

PR #3607, opened 2025-11-06
fix: Various visual adjustments on customized folder icons

doubleface:
- **fix: Adjust customized folder icon size and position**
- **fix: Close incon color picker when scrolling**
- **fix: Adjust padding in icon color picker**
- **feat: Do not change icon color in the list of icons**

### DIFF
--- a/src/components/IconPicker/IconPicker.jsx
+++ b/src/components/IconPicker/IconPicker.jsx
@@ -15,7 +15,6 @@ import { IconColorPicker } from './IconColorPicker'
 import { getIcon, getIconList } from './IconIndex'
 import { NoneIcon } from './NoneIcon'
 import {
-  ICON_COLORS,
   NB_COLUMNS_MOBILE,
   NB_COLUMNS_DESKTOP,
   CELL_HEIGHT_MOBILE,


src/components/IconPicker/IconPicker.jsx
```diff
@@ -47,6 +47,25 @@ export const IconPicker = ({
...
+    const container = scrollContainerRef?.current
+    if (!anchorEl || !container) return
+
+    const handleScroll = () => {
+      setAnchorEl(null)
+    }
+
+    // Listen to scroll events on the provided container and window
+    container.addEventListener('scroll', handleScroll, true)
+    window.addEventListener('scroll', handleScroll, true)
```
rezk2ll: Why do we need to listen on both ? Isn't the container alone enough?

rezk2ll: @doubleface can we get rid of `window.addEventListener('scroll', handleScroll, true)` ?

doubleface: Yes, it remained from a previous attempt, sorry. Removed now

PR #3611, opened 2025-11-10
fix: Fetch correct data in sharings section when no file shared :bug:

lethemanh:
### Change:

When there is no sharing files,  the API will fetch all the files in drive. To prevent it, we need to disable fetch when there is no any sharing document IDs.

### DIFF
--- a/src/modules/views/Sharings/index.jsx
+++ b/src/modules/views/Sharings/index.jsx
@@ -77,7 +77,7 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
   const { allLoaded, refresh, isOwner } = sharingContext
   const { isNativeFileSharingAvailable, shareFilesNative } =
     useNativeFileSharing()
-  const { sharedDrives } = useSharedDrives()
+  const { sharedDrives, isLoaded: sharedDrivesLoaded } = useSharedDrives()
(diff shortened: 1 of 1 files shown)


src/modules/views/Sharings/index.jsx
```diff
@@ -112,6 +116,11 @@ export const SharingsView = ({ sharedDocumentIds = [] }) => {
         []
       return {
         ...result,
+        // If there are no shared documents, we consider the data is loaded by setting fetchStatus to 'loaded' and lastFetch to now.
+        fetchStatus:
+          sharedDocumentIds?.length > 0 ? result.fetchStatus : 'loaded',
+        lastFetch:
+          sharedDocumentIds?.length > 0 ? result.lastFetch : Date.now(),
```
rezk2ll, line 123: This feels like a hack. do we need to do this? i think isLoaded ( aka sharedDrivesLoaded ) is enough

lethemanh: `sharedDrivesLoaded` is not enough, it checks if shared drive list has been loaded or not.
The loading status of shared files/folders need to be checked by `fetchStatus` and `lastFetch`.

rezk2ll (approved):

PR #3612, opened 2025-11-12
feat: Hide /Settings folder

doubleface:
To avoid unexpected side effects when modifying files and folders inside it (wallpapers and shortcuts)


doubleface:
> Can we have this only when the personalization flag is enabled ?

Done

PR #3636, opened 2025-11-24
fix: Refresh folder after personalization

doubleface:


src/modules/actions/components/personalizeFolder.js
```diff
@@ -33,7 +34,12 @@ const personalizeFolder = ({
           <FolderCustomizerModal
             folderId={folderId}
             driveId={driveId}
-            onClose={popModal}
+            onClose={() => {
+              popModal()
+              if (onClose) {
+                onClose()
+              }
```
rezk2ll: nit: onClose?.()

rezk2ll (approved):

src/modules/actions/components/personalizeFolder.js
doubleface: Done

PR #3644, opened 2025-12-01
feat: Shell integration with external bridge container apps

ecnivtwelve:
## Context
Shell containers (such as [LaSuite Docs](https://github.com/cozy/cozy-lasuite-docs)) can call **controller apps** (that runs inside them) to serve as a pilot to the embedded app.

> [!NOTE]
> LaSuite Docs needs https://github.com/cozy/cozy-lasuite-docs/pull/6 to be compatible with this PR

## Changes
Drive (running inside an iframe) can be used as a file manager, will be able to : 
- Open files into embed
- Sync opened path between embed and Drive
- Know if running in a shell via Shell context
  - UI changes when running in shell


src/hooks/useShell.tsx
```diff
@@ -0,0 +1,116 @@
...
+  children: React.ReactNode
+}): JSX.Element => {
+  const navigate = useNavigate()
+
+  const [runsInShell, setRunsInShell] = useState(false)
+  const [selectedFile, setSelectedFile] = useState<string | null>(null)
+
+  useEffect(() => {
+    if (window.top) {
+      window.top.postMessage('loaded', '*')
```
Crash--, line 32: who is the window.top here?

src/hooks/useShell.tsx
```diff
@@ -0,0 +1,116 @@
...
+      .coz-bar-container button[aria-label="Rechercher"] {
+        margin-right: -12px;
+      }
+    `
+
+    const style = document.createElement('style')
+    style.type = 'text/css'
+    style.appendChild(document.createTextNode(CSS))
+    document.head.appendChild(style)
+  }
```
Crash--, line 67: It should be done by a new entry / target.

src/hooks/useShell.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const openFileInParent = (file: File): void => {
+    if ('metadata' in file && window.top) {
+      const id = file.metadata.externalId || ''
+      window.top.postMessage('openFile:' + id, '*')
```
Crash--, line 72: Who is listening / definning this openFile ?

Crash--:
In my opinion we'll always block iframe on Drive. So maybe we should use something like intent to do it properly even if I'm not a big fan of that.

PR #3645, opened 2025-12-01
feat: Now fetching recents from dataProxy

doubleface:
This allow us to have recent files from shared drives
If dataProxy is not available, we fallback to the previous query


src/hooks/useRecentFiles.jsx
```diff
@@ -0,0 +1,105 @@
...
+import { buildRecentQuery } from '@/queries'
+
+const useDataProxyRecents = () => {
+  const [data, setData] = useState([])
+  const [fetchStatus, setFetchStatus] = useState('loading')
+  const [error, setError] = useState(null)
+  const [useFallback, setUseFallback] = useState(false)
+  const [fallbackData, setFallbackData] = useState(null)
+  const [fallbackStatus, setFallbackStatus] = useState('loading')
+  const [fallbackError, setFallbackError] = useState(null)
```
rezk2ll: too many independent state variables, i suggest you use a reducer and you can group everything into one state:
```ts
const initialState = {
  data: [],
  status: 'loading', // loading | loaded | error
  error: null,
  triedFallback: false
}
```
```ts
const reducer = (state, action) => {
 // ... code here
}
```

then
```ts
const [state, dispatch] = useReducer(reducer, initialState)
...

dispatch({ type: 'LOADING' })
```
an

src/hooks/useRecentFiles.jsx
```diff
@@ -0,0 +1,105 @@
...
+  const dataProxy = useDataProxy()
+  const client = useClient()
+
+  const recentQuery = useMemo(() => buildRecentQuery(), [])
+
+  useEffect(() => {
```
rezk2ll, line 18: We have 3 useEffects that are dependent on each other.

rezk2ll (changes requested):
in theory this works but i think we have stuff to improve:
- 3 useEffects and potential race condition(s)
- a lot of states to manage and juggle between
- the code should be easy to read and maintain

also for the logic why not:
1 - try the proxy
2 - if the proxy fails try the fallback

doubleface:
Right, this implementation was not the best. I chose to implement it now with only one useEffect and handled the logic of calling the fallback only when the dataproxy fails. I think it is much better now

src/hooks/useRecentFiles.jsx
doubleface: I finally prefer the solution to call the fallback only on error using async/await

rezk2ll (approved):

PR #3647, opened 2025-12-03
feat: Download all files with select all

doubleface:
When select all is checked, download all files in the folder, even
when all files are not displayed.

### DIFF
--- a/src/modules/actions/download.jsx
+++ b/src/modules/actions/download.jsx
@@ -26,7 +26,15 @@ const makeComponent = (label, icon) => {
   return Component
 }
 
-export const download = ({ client, t, vaultClient, showAlert, driveId }) => {
+export const download = ({
+  client,
+  t,
+  vaultClient,
+  showAlert,
+  driveId,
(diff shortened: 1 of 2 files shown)


src/modules/actions/download.jsx
```diff
@@ -45,9 +53,13 @@ export const download = ({ client, t, vaultClient, showAlert, driveId }) => {
       )
     },
     action: files => {
+      let selectedFiles = files
+      if (isSelectAll) {
+        selectedFiles = [displayedFolder]
+      }
```
zatteo, line 59: Nice 👍

zatteo (approved):

rezk2ll (approved):

rezk2ll:
image

what if we click on "3 dots" > download folder without having to select all?

doubleface:
> what if we click on "3 dots" > download folder without having to select all?

It works as before, it downloads the folder with all the elements, even those which are not displayed

PR #3650, opened 2025-12-05
feat: Display view switcher for recent files

doubleface:
And also allow to hide sorter when canSort is not true.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Mobile file list header layout corrected to preserve right-side alignment while properly showing sort controls.

* **New Features**
  * Centralized file-sorting applied across Folder and Recent views for consistent, stable sorting and selection behavior.

* **Tests**
  * Added comprehensive tests covering sorting logic, default and external sort orders, secondary sorting, and edge cases.

### DIFF
--- a/src/modules/filelist/FileListHeaderMobile.jsx
+++ b/src/modules/filelist/FileListHeaderMobile.jsx
@@ -36,24 +36,27 @@ const FileListHeaderMobile = ({
     [setIsShowingSortMenu]
   )
 
-  if (!canSort) return null
-
   return (
     <TableHead className={styles['fil-content-mobile-head']}>
       <TableRow className={styles['fil-content-row-head']}>
-        <TableHeader
-          onClick={showSortMenu}
-          className={cx(
-            styles['fil-content-mobile-header'],
-            styles['fil-content-header--capitalize'],
-            {
-              [styles['fil-content-header-sortasc']]: sort.order === 'asc',
-              [styles['fil-content-header-sortdesc']]: sort.order === 'desc'
-            }
-          )}
-        >
-          {t(`table.mobile.head_${sort.attribute}_${sort.order}`)}
-        </TableHeader>
+        {canSort ? (
+          <TableHeader
+            onClick={showSortMenu}
+            className={cx(
+              styles['fil-content-mobile-header'],
+              styles['fil-content-header--capitalize'],
+              {
+                [styles['fil-content-header-sortasc']]: sort.order === 'asc',
+                [styles['fil-content-header-sortdesc']]: sort.order === 'desc'
+              }
+            )}
+          >
+            {t(`table.mobile.head_${sort.attribute}_${sort.order}`)}
+          </TableHeader>
+        ) : (
+          <div className="u-flex-auto" /> // to keep the viewType switch to the right side
+        )}
+
         {isShowingSortMenu && (
           <MobileSortMenu
             t={t}
--- a/src/modules/views/Folder/hooks/useFileSorting.spec.js
+++ b/src/modules/views/Folder/hooks/useFileSorting.spec.js
@@ -1,0 +1,176 @@
+import { renderHook } from '@testing-library/react-hooks'
+
+import { useFileSorting } from './useFileSorting'
+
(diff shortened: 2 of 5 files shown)


src/modules/filelist/FileListHeaderMobile.jsx
```diff
@@ -36,13 +37,14 @@ const FileListHeaderMobile = ({
     [setIsShowingSortMenu]
   )
 
-  if (!canSort) return null
+  if (!canSort && !canSwitchViewType) return null
```
rezk2ll: i don't think we need to have this `canSwitchViewType` here, no matter what mobile or desktop we should be able to switch the view

src/modules/filelist/FileListHeaderMobile.jsx
```diff
@@ -63,6 +65,7 @@ const FileListHeaderMobile = ({
           />
         )}
         <TableHeader
+          style={{ visibility: canSwitchViewType ? 'visible' : 'hidden' }}
```
rezk2ll: it should be always displayed

rezk2ll:
i don't understand why this PR is needed

src/modules/filelist/FileListHeaderMobile.jsx
doubleface: Yes, but at first the only criteria to display the header was if canSort is thruthy. Here, for the recents view, I want to display the header but only with the "change view type" button. 
Yes, at the moment, "change view type" button will always be display, but canSwitchViewType gives me a new criteria to display the mobile header or not and allowing to hide it make the api more coherent I think.

doubleface: Or should we change the logic and add a "hideHeader" prop to FolderViewBody ?

doubleface:
> i don't understand why this PR is needed

This PR is needed to display the view switcher on the recents file view.

I am sorry because at first I also added the sort feature because it was the only way to display the toolbar.
After some thinking, I prefered to make it possible to display only the view switcher in the toolbar, without the sorting.

src/modules/views/Folder/hooks/useFileSorting.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('should apply both primary and secondary sort', () => {
+    const { result } = renderHook(() =>
+      useFileSorting('folder-1', mockQueryResults, mockOrderProps)
+    )
+
+    // Should be sorted by name in descending order, then by type
+    expect(result.current.sortedFiles).toHaveLength(3)
+  })
```
doubleface, line 144: Added some tests on sorting

src/modules/filelist/FileListHeaderMobile.jsx
doubleface: Let's always show the header with the view switch and hide the sorter when canSort is not true

src/modules/views/Folder/hooks/useFileSorting.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    const fileInfo = sortedFiles.map(file => ({
+      name: file.name,
+      type: file.type
+    }))
+    expect(fileInfo).toEqual([
+      { name: 'folder-c', type: 'directory' },
+      { name: 'file-b.txt', type: 'file' },
+      { name: 'file-a.txt', type: 'file' }
+    ])
+  })
```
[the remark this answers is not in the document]
doubleface: done

src/modules/views/Folder/hooks/useFileSorting.spec.js
```diff
@@ -0,0 +1,161 @@
...
+          }
+        ],
+        {}
+      )
+    )
+
+    // Secondary sort should put directories before files
+    expect(result.current.sortedFiles[0].type).toBe('directory')
+    expect(result.current.sortedFiles[1].type).toBe('file')
+  })
```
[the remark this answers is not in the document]
doubleface: done

rezk2ll (approved):

PR #3656, opened 2025-12-09
feat: Hide shared drive owner icon in desktop / grid mode

doubleface:
Since the owner is already displayed in another way in this case:

Before:
image

After:
image

https://www.notion.so/linagora/UI-UX-shared-drive-mobile-bug-2bd62718bad180b18251eb6268fcf262

## Summary by CodeRabbit

* **New Features**
  * File list component now implements responsive behavior, displaying sharing owner information exclusively on mobile devices.

### DIFF
--- a/src/modules/filelist/icons/FileThumbnail.tsx
+++ b/src/modules/filelist/icons/FileThumbnail.tsx
@@ -13,6 +13,7 @@ import LinkIcon from 'cozy-ui/transpiled/react/Icons/Link'
 import TrashDuotoneIcon from 'cozy-ui/transpiled/react/Icons/TrashDuotone'
 import InfosBadge from 'cozy-ui/transpiled/react/InfosBadge'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
+import useBreakpoints from 'cozy-ui/transpiled/react/providers/Breakpoints'
 
 import styles from '@/styles/filelist.styl'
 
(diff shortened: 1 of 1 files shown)


rezk2ll:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (1)
> > src/modules/filelist/icons/FileThumbnail.tsx (1)> `89-94`: **Behavior now hides shared‑drive owner avatar on all non‑mobile views—confirm this matches the spec.**
> > > With `isMobile && <SharingOwnerAvatar ...>`, the owner avatar for shared‑drive folders only appears on mobile, regardless of `viewType`. That means it is also hidden on desktop in list view, not just in grid. If the product intent is strictly “hide only on desktop/grid”, you may instead want a condition like `isMobile || viewType === 'list'` (or similar), otherwise this broader change looks fine.
> 
> 📜 Review details

@doubleface as the bot said, in the PR we hide it for the list view also not just grid.

also in the screens above you are testing with one recipient, what if you have 3 or 4 recipients how will it look like?

rezk2ll (approved):

doubleface:
> > **Actionable comments posted: 0**
> > 🧹 Nitpick comments (1)
> > > src/modules/filelist/icons/FileThumbnail.tsx (1)> `89-94`: **Behavior now hides shared‑drive owner avatar on all non‑mobile views—confirm this matches the spec.**
> > > > With `isMobile && <SharingOwnerAvatar ...>`, the owner avatar for shared‑drive folders only appears on mobile, regardless of `viewType`. That means it is also hidden on desktop in list view, not just in grid. If the product intent is strictly “hide only on desktop/grid”, you may instead want a condition like `isMobile || viewType === 'list'` (or similar), otherwise this broader change looks fine.
> > 
> > 
> > 📜 Review details
> 
> @doubleface as the bot said, in the PR we hide it for the list view also not just grid.
> 
> also in the screens above you are testing with one recipient, what if you have 3 or 4 recipients how will it look like?

Here it is with more than 3 recipients : 
image

PR #3671, opened 2025-12-16
feat: Display full breadcrumb in public view

doubleface:
When displaying a shared folder, the breadcrumb only displayed the shared folder and parent in breadcrumb. Now we display the full breadcrumb from the current folder to the root of the public folder.

[Capture vidéo du 2025-12-16 17-58-04.webm](https://github.com/user-attachments/assets/e1faa6ab-73c3-4d5b-913b-ccb40fae2662)

https://www.notion.so/linagora/I-don-t-see-my-all-breadcrumb-in-public-link-2b862718bad180c4ae30c95acdf3bcd5

## Summary by CodeRabbit

* **Bug Fixes**
  * Breadcrumbs for public/shared folder views now reliably show the correct hierarchy, display only after the path is resolved, and reset on failure (errors are logged).

* **Refactor**
  * Breadcrumb resolution moved to an asynchronous, client-side traversal for more robust parent-folder lookup.
  * Public folder view and routing now pass shared-document context through the view so breadcrumbs are built consistently.

### DIFF
--- a/src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
+++ b/src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
@@ -1,20 +1,50 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 
-import { useQuery } from 'cozy-client'
+import { useClient } from 'cozy-client'
 
+import logger from '@/lib/logger'
 import { MobileAwareBreadcrumb as Breadcrumb } from '@/modules/breadcrumb/components/MobileAwareBreadcrumb'
-import { buildFileOrFolderByIdQuery } from '@/queries'
 
-const FolderViewBreadcrumb = ({ currentFolderId, getBreadcrumbPath }) => {
+const FolderViewBreadcrumb = ({
+  displayedFolder,
+  sharedDocumentId,
+  getBreadcrumbPath
+}) => {
   const navigate = useNavigate()
-  const currentFolderQuery = buildFileOrFolderByIdQuery(currentFolderId)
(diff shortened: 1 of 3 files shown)


src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
```diff
@@ -1,20 +1,30 @@
...
+    if (!displayedFolder || !sharedDocumentId) return
+    const asyncGetPaths = async () => {
+      const paths = await getBreadcrumbPath({
+        client,
+        displayedFolder,
+        sharedDocumentId
+      })
+      setPath(paths)
+    }
+    asyncGetPaths()
```
rezk2ll, line 42: I think we might introduce a race condition here, if we have paths too quickly in the breadcramb.

if we change path and the getBreadcrumbPath of an old path is still executing we might set an outdated path

doubleface: I added a setPath(null) at the beginning of the useEffect and now only render the component when path is not null. I should avoid the kind of race condition you raised. Thank you

src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
```diff
@@ -1,20 +1,43 @@
...
+        logger.error(`Error while fetching breadcrumb path: ${error}`)
+        setPath(null)
+      }
+      const paths = await getBreadcrumbPath({
+        client,
+        displayedFolder,
+        sharedDocumentId
+      })
+      setPath(paths)
+    }
```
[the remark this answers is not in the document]
rezk2ll: @doubleface probably you pasted the same code by mistake here

rezk2ll (approved):

PR #3674, opened 2025-12-18
fix: Hide settings files from recents

doubleface:
We should hide those files in recents since we already hide theme in the other screens.

I first tried to filter out settings files in the dataproxy request, but files don't have paths in couchdb ou pouchdb databases. Looks like it will not be a big deal to have a little bit less thant 50 files in the results.

https://www.notion.so/linagora/In-recent-view-tab-i-see-my-files-shortcut-homepage-2c862718bad18095bb2bc32515872edf

## Summary by CodeRabbit

* **Bug Fixes**
  * Recent files list now excludes entries from the settings directory, yielding a cleaner, more relevant recent-items view.
* **Tests**
  * Added tests to ensure settings-directory entries are filtered from both primary and fallback recent-data sources and that items lacking a path are preserved.

### DIFF
--- a/src/hooks/useRecentFiles.jsx
+++ b/src/hooks/useRecentFiles.jsx
@@ -3,8 +3,13 @@ import { useEffect, useState, useMemo } from 'react'
 import { useClient } from 'cozy-client'
 import { useDataProxy } from 'cozy-dataproxy-lib'
 
+import { SETTINGS_DIR_PATH } from '@/constants/config'
 import logger from '@/lib/logger'
 import { buildRecentQuery } from '@/queries'
+
+const filterOutSettingsFiles = files => {
+  return files.filter(file => !file.path?.startsWith(SETTINGS_DIR_PATH))
+}
 
 const useDataProxyRecents = () => {
   const [data, setData] = useState([])
@@ -23,7 +28,7 @@ const useDataProxyRecents = () => {
       if (dataProxy.dataProxyServicesAvailable) {
         try {
           const data = await dataProxy.recents()
-          setData(data || [])
+          setData(filterOutSettingsFiles(data || []))
           setFetchStatus('loaded')
           return
         } catch (err) {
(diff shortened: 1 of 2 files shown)


Crash--:
> We should hide those files in recents since we already hide theme in the other screens.

No, no no no please. 

The first mistake was to start hiding those files. WE SHOULD NOT. Because, we can't manage all the clients making request to the files. Here you hide from recent, tomorrow, you'll hide from search because I'm pretty sure we didn't hide them for that. But on desktop the files are still displayed & so on. 

So, if we want to hide this settings, then you should not use files to store the information and use dedicated doctype for that. Please.

doubleface:
The problem is more with the implementation choice for home shortcuts. We chose to change this implementation to use a specific doctype for shortcuts. This way, shortcuts won't appear automatically in recents or even search.

Crash--:
We still use `io.cozy.files.shortcut` so we need to reopen this PR. 

But let's wait to fix issue with recents, because I want to use an index, not a JS filter.

src/hooks/useRecentFiles.jsx
```diff
@@ -3,9 +3,14 @@ import { useEffect, useState, useMemo } from 'react'
 import { useClient } from 'cozy-client'
 import { useDataProxy } from 'cozy-dataproxy-lib'
 
+import { SETTINGS_DIR_PATH } from '@/constants/config'
 import logger from '@/lib/logger'
 import { buildRecentQuery } from '@/queries'
 
+const filterOutSettingsFiles = files => {
```
rezk2ll, line 10: i don't think this is a good long-term solution, and we have to do it also for other drive clients ( Desktop etc, tomorrow the mobile app ). 

So, a better solution, as we discussed, is:
- to not use the Settings folder at all, to store the shortcuts/wallpapers
- Use a new doctype ( that is not io.cozy.files ) for shortcuts, and another for wallpapers or a shared doctype for user preferences, etc. I know this requires a migration too, but we can do it
- removes the hacks we did to hide the SETTINGS_DIR_PATH

Crash--: Of course it's what I want too. 

But it's also a lot of migrations tasks and so one. So before being able to do it, we need at least to hide it first.

PR #3681, opened 2026-01-07
fix: Delete key on Mac should be treated as delete

doubleface:
## Summary by CodeRabbit

* **Bug Fixes**
  * Backspace on Apple platforms is now treated the same as Delete, fixing inconsistent keyboard behavior.
  * Improves reliability of deletion and shortcut handling on Mac keyboards for more consistent user interactions.

### DIFF
--- a/src/hooks/helpers.js
+++ b/src/hooks/helpers.js
@@ -74,7 +74,7 @@ export const normalizeKey = (event, isApple) => {
 
   const key = event.key.toLowerCase()
 
-  if (key === 'delete' || key === 'del') {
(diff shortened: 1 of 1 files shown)


src/hooks/helpers.js
```diff
@@ -76,6 +76,9 @@ export const normalizeKey = (event, isApple) => {
 
   if (key === 'delete' || key === 'del') {
     keys.push('delete')
+  } else if (isApple && key === 'backspace') {
```
rezk2ll: i would just add `isApple && key === 'backspace'` to the previous if statement

rezk2ll (approved):

PR #3684, opened 2026-01-19
feat: Antivirus integration

doubleface:
Integrate antivirus feature in :
- files list
image
image
- file viewer
image
image
image
- selection bar
image
image
- sharings 
image

https://www.notion.so/linagora/Anti-Virus-for-Drive-21c62718bad180069c59c8a5a9874324

[Capture vidéo du 2026-01-22 14-25-28.webm](https://github.com/user-attachments/assets/9523ba42-1259-4fea-88d6-82783b89e798)

## Summary by CodeRabbit

* **New Features**
  * Infected files show a visible badge and an “infected” tooltip; selection bar highlights infection and shows an explanatory popover.

* **Behavior Changes**
  * Actions are filtered by policy: many actions are blocked for infected files while trashing remains allowed; context menus and selection actions respect these rules.

* **Localization**
  * Added infection-related UI messages in English and French.

* **Tests**
  * Added comprehensive unit tests for policy filtering logic.

* **Chores**
  * Updated several transitive dependencies.


src/components/RightClick/RightClickFileMenu.jsx
```diff
@@ -19,7 +20,11 @@ const RightClickFileMenu = ({
   const { isDesktop } = useBreakpoints()
   const { selectedItems, isItemSelected } = useSelectionContext()
 
-  const contextMenuActions = getContextMenuActions(actions)
+  const infected = isInfected(doc)
+  const contextMenuActions = getContextMenuActions(actions).filter(action => {
+    const isRemoveAction = Object.keys(action).includes('trash')
```
rezk2ll: this is a hack, we shouldn't do this

src/hooks/useMoreMenuActions.jsx
```diff
@@ -45,6 +46,7 @@ export const useMoreMenuActions = file => {
   const currentFolderId = useCurrentFolderId()
   const { isSharingShortcutCreated, addSharingLink, syncSharingLink } =
     useSharingInfos()
+  const infected = isInfected(file)
```
rezk2ll: we have `isInfected` scattered everywhere in the code, if we add new interfaces we have to include it there too

src/hooks/useMoreMenuActions.jsx
```diff
@@ -54,19 +56,19 @@ export const useMoreMenuActions = file => {
...
-      showPrintAction && print,
-      hr,
+      !infected && showPrintAction && print,
+      !infected && hr,
       moveTo,
       !isSharedDrive && duplicateTo, // TO DO: Remove condtion when duplicating is available in shared drive
       addToFavorites,
       removeFromFavorites,
-      hr,
+      !infected && hr,
```
rezk2ll: same here, we display `hr` if the file is not infected

src/locales/fr.json
```diff
@@ -232,7 +232,11 @@
     "qualify": "Qualifier",
     "history": "Versions",
     "more": "Afficher plus d'action",
-    "openWithinNextcloud": "Ouvrir dans Nextcloud"
+    "openWithinNextcloud": "Ouvrir dans Nextcloud",
+    "infectedPopover": {
```
rezk2ll: i would group antivirus translations in one location not scattered in popover and file.

```
"antivirus": {
  "popover.title": "...",
  "popover.description": "...",
  "infectedFile": "..."
```

src/modules/actions/components/addToFavorites.tsx
```diff
@@ -14,13 +14,15 @@ interface addToFavoritesProps {
...
+  isInfected?: boolean
 }
 
 const addToFavorites = ({
   t,
   client,
   isMobile,
-  showAlert
+  showAlert,
+  isInfected
```
rezk2ll: this will become very hard to maintain later, i guess in the PR we will have this prop injected in all action functions

src/modules/filelist/helpers.ts
```diff
@@ -55,3 +55,21 @@ export const getFileNameAndExtension = (
...
 }
+
+interface FileWithAntivirusScan {
+  antivirus_scan?: {
+    status?: string
+  }
+}
+
+export const isInfected = (
+  file: FileWithAntivirusScan | Record<string, unknown>
```
rezk2ll: This type is misleading. we shouldn't say maybe yes maybe no here.

```ts
interface AntivirusScan {
  status: 'clean' | 'infected' | 'pending'
}

interface FileWithSecurity {
  antivirus_scan?: AntivirusScan
}
```

or update the doctype

also you are already check the attribute existance in the function.

src/modules/filelist/helpers.ts
```diff
@@ -55,3 +55,21 @@ export const getFileNameAndExtension = (
...
+): boolean => {
+  return (
+    'antivirus_scan' in file &&
```
rezk2ll: return early if the attribute doesn't exist.

check Rabbit suggestion below

src/modules/filelist/icons/FileThumbnail.tsx
```diff
@@ -123,12 +134,7 @@ const FileThumbnail: React.FC<FileThumbnailProps> = ({
         >
           {isKonnectorFolder ? (
             <BadgeKonnector file={file}>
-              <FileIcon
-                file={file}
-                size={size}
-                isEncrypted={isEncrypted}
-                viewType={viewType}
-              />
+              {fileIcon}
```
rezk2ll, line 137: This change is not needed since Konnectors are not concerned

src/modules/selection/SelectionBar.tsx
```diff
@@ -26,18 +37,76 @@ const SelectionBar: React.FC<{
...
   if (isSelectionBarVisible && actions) {
-    const convertedActions = driveActionsToSelectionBarActions(actions)
+    let convertedActions = driveActionsToSelectionBarActions(actions)
+
+    let color = 'default'
+    let iconComponent = null
+    const hasInfectedItem = selectedItems.some(item => isInfected(item))
+    if (hasInfectedItem) {
+      convertedActions = convertedActions.filter(driveAction => {
+        const actionName = Object.keys(driveAction)[0]
```
rezk2ll: no we can't do this. we are filtering the action by key name and we hardcode the trash in the list of the allowed actions here. as i said in the comments. it should be done once in a policy

rezk2ll (commented):
this PR needs a lot of refactoring.

- we have isInfected scattered everywhere in the code, in hooks, in actions, in menus. instead we should have a policy in the action. like this action can be done with infected files or not. and when building the actions we should determine whether to display the action or not. like we have for canWrite etc

we do it once and it should be applied to all UI. we shouldn't have conditions using isInfected in the UI everywhere, this will be very hard to maintain later.

other developers will forget to include isInfected somewhere, sometimes we will for example enable moving infected files,  then we will have to change ui code etc.

src/modules/filelist/icons/FileThumbnail.tsx
doubleface: But it is just using fileIcon instead of rewriting the the same <FileIcon> component with the same attributes everywhere. There is no change regarding konnectors here

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,78 @@
...
+  // If no infected files, return all actions (no policy filtering needed)
+  if (!hasInfectedFile) {
+    return actions
+  }
+
+  return actions.filter(wrappedAction => {
+    const action = getActionFromWrapper(wrappedAction)
+    if (!action) return true
+
+    return isActionAllowedForInfectedFiles(action, hasInfectedFile)
```
rezk2ll: This is a bit redundant, no need to pass in the second param and call this function when hasInfectedFile is true

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param actions - Array of wrapped actions from makeActions
+ * @param files - Array of files to check policies against
+ * @returns Filtered array of actions that are allowed for the given files
+ *
+ * @example
+ * ```typescript
+ * const filteredActions = filterActionsByPolicy(actions, selectedFiles)
+ * ```
+ */
+export const filterActionsByPolicy = <T extends Record<string, DriveAction>>(
```
rezk2ll, line 130: This is a bit specific to antivirus; we need to think of filterActionsByPolicy to have a lot of other policies, and this filter will check them.

rezk2ll (commented):
much better, but we need to make sure filterActionsByPolicy is valid not only for infected files but also for all other sort of policies ( readable, writable, shared, etc ).

for example

```ts
export interface ActionPolicyContext {
  files: IOCozyFile[]
  hasInfectedFile: boolean
  hasMultipleFiles: boolean
  hasSharedFile: boolean
  hasReadOnlyFile: boolean
}
```

```ts
const buildPolicyContext = (files: IOCozyFile[]): ActionPolicyContext => ({
  files,
  hasInfectedFile: files.some(isInfected),
  hasMultipleFiles: files.length > 1,
  hasSharedFile: files.some(f => f.cozyMetadata?.shared),
  hasReadOnlyFile: files.some(f => f.cozyMetadata?.permissions === 'read')
})
```

and we declare the policy of this type

```ts
export interface ActionPolicy {
  name: string
  allows(action: DriveAction, ctx: ActionPolicyContext): boolean
}
```

and for example the infection policy like this

```ts
const infectionPolicy: ActionPolicy = {
  name: 'infection',
  allows: (action, ctx) =>
    !ctx.hasInfectedFile || action.allowInfectedFiles === true
}
```

we group the policies and export them

```ts
const ACTION_POLICIES: ActionPolicy[] = [
  infectionPolicy,
  readOnlyPolicy,
  ...etc
]
```

and we exapand the drive actions:

```ts
export interface DriveAction {
  label: string
  icon?: React.ComponentType
  action?: (files: IOCozyFile[]) => void

  // Policy flags
  allowInfectedFiles?: boolean
  allowReadOnly?: boolean
  allowMultiple?: boolean
...etc
}
```

and then in the `filterActionsByPolicy` we use them together

```ts
  return actions.filter(wrapped => {
    // do whatever here
    if (!action) return true

    return ACTION_POLICIES.every(policy =>
      policy.allows(action, ctx)
    )
  })
```

src/hooks/useMoreMenuActions.jsx
```diff
@@ -61,7 +62,7 @@ export const useMoreMenuActions = file => {
       showPrintAction && print,
       hr,
       moveTo,
-      !isSharedDrive && duplicateTo, // TO DO: Remove condtion when duplicating is available in shared drive
+      !isSharedDrive && duplicateTo, // TO DO: Remove condition when duplicating is available in shared drive
```
rezk2ll, line 65: unrelated

src/modules/actions/components/addToFavorites.tsx
```diff
@@ -31,6 +32,8 @@ const addToFavorites = ({
     name: 'addToFavourites',
     label,
     icon,
+    // Policy: addToFavorites is NOT allowed for infected files (default behavior)
```
rezk2ll: no need for this comment since `allowInfectedFiles: false` is self explanatory

src/modules/actions/components/duplicateTo.tsx
```diff
@@ -36,6 +36,8 @@ const duplicateTo = ({
     name: 'duplicateTo',
     label,
     icon,
+    // Policy: duplicateTo is NOT allowed for infected files (default behavior)
```
rezk2ll: same here

src/modules/actions/components/moveTo.jsx
```diff
@@ -19,6 +19,8 @@ const moveTo = ({ t, canMove, pathname, navigate, isMobile, search }) => {
     name: 'moveTo',
     label,
     icon,
+    // Policy: moveTo is NOT allowed for infected files (default behavior)
```
rezk2ll: same here

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,175 @@
...
+} from './types'
+
+import { isInfected } from '@/modules/filelist/helpers'
+
+/**
+ * Builds the policy context from the selected files.
+ * This computes all the information needed for policy checks once,
+ * so we don't have to recompute it for each policy.
+ *
+ * @param files - The files being acted upon
```
rezk2ll:
```suggestion
 * @param {IOCozyFile[]} files - The files being acted upon
```

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @returns The policy context with computed file information
```
rezk2ll:
```suggestion
 * @returns {ActionPolicyContext} The policy context with computed file information
```

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,175 @@
...
+const trashedPolicy: ActionPolicyDefinition = {
+  name: 'trashed',
+  allows: (action: DriveActionPolicyFlags, ctx: ActionPolicyContext): boolean =>
+    !ctx.allInTrash || action.allowTrashed === true
+}
+
+// =============================================================================
+// Policy Registry
+// =============================================================================
+
```
rezk2ll:
```suggestion
```

src/modules/actions/policies.ts
```diff
@@ -0,0 +1,175 @@
...
+export const ACTION_POLICIES: ActionPolicyDefinition[] = [
+  infectionPolicy,
+  multipleFilesPolicy,
+  foldersPolicy,
+  trashedPolicy
+]
+
+// =============================================================================
+// Action Filtering
+// =============================================================================
```
rezk2ll:
```suggestion
```

src/modules/actions/versions.jsx
```diff
@@ -33,6 +33,8 @@ export const versions = ({ t, navigate, pathname }) => {
     name: 'history',
     label,
     icon,
+    // Policy: versions is NOT allowed for infected files (default behavior)
```
rezk2ll:
```suggestion
```

src/modules/selection/SelectionBar.tsx
```diff
@@ -23,21 +33,78 @@ const driveActionsToSelectionBarActions = (
...
+    convertedActions = filterActionsByPolicy(convertedActions, selectedArray)
+    const hasInfectedItem = hasAnyInfectedFile(selectedArray)
+
+    let color = 'default'
+    let iconComponent = null
+
+    if (hasInfectedItem) {
+      color = 'error'
+      iconComponent = (): JSX.Element => (
+        <div onMouseEnter={handlePopoverOpen} onMouseLeave={handlePopoverClose}>
```
rezk2ll, line 65: can we have a video of the UI behaviour with mouse interaction? is it smooth? ( mouse enter and mouse leave )

doubleface: New video added in description of the PR

rezk2ll (approved):

PR #3692, opened 2026-02-02
feat: Dynamic selection

doubleface:
[Capture vidéo du 2026-02-02 15-05-03.webm](https://github.com/user-attachments/assets/bbfb16bb-4c02-4323-aef2-b3e678cac0b3)

- rectangular selection
- allow to click anywhere on the line to select it
- removed checkbox on Table and Grid view
- select a file while right clicking on it

https://www.notion.so/linagora/Dynamic-selections-2ea62718bad180fb895ff1364a073f62

## Summary by CodeRabbit

* **New Features**
  * Drag-to-select multiple files with a rectangular marquee in grid/list when dynamic selection is enabled
  * Right-click now auto-selects the target file before showing the context menu
  * Dynamic selection mode toggle to switch selection behaviors

* **Bug Fixes**
  * Fixed context menu identifier/formatting issue

* **Chores**
  * Updated UI dependencies (cozy-ui, cozy-ui-plus) and added react-selecto
  * Adjusted file-list spacing and selection styling


src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+   * Extracts file data from a DOM element using the data-file-id attribute.
+   *
+   * @param {Element} el - DOM element with data-file-id attribute
+   * @returns {Object|undefined} The file object matching the element's ID, or undefined if not found
+   */
+  const getFileFromElement = useCallback(
+    el => {
+      const fileId = el.getAttribute('data-file-id')
+      if (!fileId) return undefined
+      return items.find(item => item._id === fileId)
```
rezk2ll: This is O(n^2), and it will cause performance issues if I have a lot of files. Imagine for each file I need to perform a find operation in 200 files.

We can use a map for this.
at the top:

```ts
const itemsById = useMemo(() => {
  const map = {}
  for (const item of items) map[item._id] = item
  return map
}, [items])
```

and we use it like

```ts
const getFileFromElement = useCallback(
  el => itemsById[el.getAttribute('data-file-id')],
  [itemsById]
)
```

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+    <div
+      ref={containerRef}
+      className="u-h-100 rectangular-selection-container"
+      onClick={handleContainerClick}
+    >
+      {children}
+      <Selecto
+        className={styles['cozy-selecto-box']}
+        container={containerRef.current}
+        dragContainer={window}
```
rezk2ll: Why the whole window? We should only apply it to the concerned container ( the main table container )

i don't think it's a good idea to start dragging from the sidebar for example.

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+        selectableTargets={['[data-file-id]']}
+        selectByClick={false}
+        selectFromInside={false}
+        hitRate={0}
+        ratio={0}
+        toggleContinueSelect="ctrl"
+        continueSelect={false}
```
rezk2ll: Those two lines contradict each other. I believe we need to remove one of them ?

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+      const distance = Math.sqrt(dx * dx + dy * dy)
+
+      // Only consider it a real drag if the mouse moved more than 5 pixels
+      // This distinguishes between a click and a rectangular selection
+      if (distance > 5) {
+        isDraggingRef.current = true
+        // Reset the flag after the click event has been processed
+        setTimeout(() => {
+          isDraggingRef.current = false
+        }, 0)
```
rezk2ll: This will likely break, and some browsers will behave differently than we expect. 
Here, you rely on the browser event ordering and timing.

We can use a different approach without timing hacks, 

selecto supports a `handleDrag` prop that you can use, and we couple it with the  `isDraggingRef`.

```ts
const handleDragStart = useCallback(e => {
  dragStartPosRef.current = { x: e.clientX, y: e.clientY }
  isDraggingRef.current = false
}, [])
```

```ts
const handleDragEnd = useCallback(() => {
  dragStartPosRef.current = null
}, [])
```

and we add handleDrag

```ts

const handleDrag = useCallback(e => {
  const start = dragStartPosRef.current
  if (!start) return

  const dx = e.clientX - start.x
  const dy = e.clientY - start.y

  if (Math.hypot(dx, dy) > 5) {
    isDraggingRef.current = true
  }
}, [])
```

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+      const baseSelection =
+        e.inputEvent?.ctrlKey || e.inputEvent?.metaKey
+          ? selectedItems.reduce(
+              (acc, item) => ({
+                ...acc,
+                [item._id]: item
+              }),
+              {}
+            )
+          : {}
```
rezk2ll: no need for a reduced here, it can be simplified to:

```ts
const baseSelection =
  e.inputEvent?.ctrlKey || e.inputEvent?.metaKey
    ? { ...selectedItems }
    : {}
```

since selectedItems is already constructed like that.

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+
+      for (const el of e.selected) {
+        const file = getFileFromElement(el)
+        if (file) {
+          newSelection[file._id] = file
+        }
+      }
+
+      setSelectedItems(newSelection)
+      setIsSelectAll(Object.keys(newSelection).length === items.length)
```
rezk2ll, line 81: We will have a problem if newSelection is different from the `selectedItems`, maybe we can rely on `selectedItems` instead?

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+
+      // Check if clicked on a file or interactive element
+      const isOnFile = target.closest('[data-file-id]')
+      const isInteractive =
+        target.closest('button') ||
+        target.closest('a') ||
+        target.closest('input') ||
+        target.closest('select') ||
+        target.closest('textarea') ||
+        target.getAttribute('role') === 'button'
```
rezk2ll: a small improvement:

```ts
const isInteractive = target.closest(
  'button,a,input,select,textarea,[role="button"]'
)
```

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+    >
+      {children}
+      <Selecto
+        className={styles['cozy-selecto-box']}
+        container={containerRef.current}
+        dragContainer={window}
+        selectableTargets={['[data-file-id]']}
+        selectByClick={false}
+        selectFromInside={false}
+        hitRate={0}
```
rezk2ll: 10% or 20%, what do you think?

I'm not sure about 0% hit rate, honestly.

src/modules/selection/RectangularSelection.jsx
```diff
@@ -0,0 +1,221 @@
...
+        if (file) {
+          newSelection[file._id] = file
+        }
+      }
+
+      setSelectedItems(newSelection)
+      setIsSelectAll(Object.keys(newSelection).length === items.length)
+    },
+    [items, selectedItems, getFileFromElement, setSelectedItems, setIsSelectAll]
+  )
```
rezk2ll, line 90: good catch @coderabbitai , i already suggested

```ts
const baseSelection =
  e.inputEvent?.ctrlKey || e.inputEvent?.metaKey
    ? { ...selectedItems }
    : {}
```

wdyt?

src/modules/selection/RectangularSelection.jsx
doubleface: Let's go with 10%

src/modules/selection/RectangularSelection.jsx
doubleface: Ok, now the drag container is the folder view body

src/modules/selection/RectangularSelection.jsx
doubleface: It's  not a contradiction: `continueSelect` will continue the selection by default, without any special key and `toggleContinueSelect` allows to continue the selection with the defined special key. I will add comments to make this clearer

rezk2ll (approved):

PR #3694, opened 2026-02-04
fix: Set the container of dynamic selection to window

doubleface:
We want to be able to start the drag selection from anywhere on the page and also clear the selection from anywhere on the page.

[Capture vidéo du 2026-02-04 14-15-41.webm](https://github.com/user-attachments/assets/6011d560-398c-4129-b32b-6d1beadb4a79)

Also removed the logic to get the element corresponding to container

https://www.notion.so/linagora/Dynamic-selection-fixes-2fd62718bad180809010e3f9eae9d0c3

## Summary by CodeRabbit

* **Refactor**
  * Selection now waits until its container is ready using a ref-based readiness model for more reliable mounting and rendering.
  * Removed external drag-container input from the public API; selection now uses the global drag context.
  * Preserved scroll container and scroll element support to maintain auto-scroll behavior during selections.

### DIFF
--- a/src/modules/selection/RectangularSelection.jsx
+++ b/src/modules/selection/RectangularSelection.jsx
@@ -1,4 +1,4 @@
-import React, { useRef, useCallback, useState, useMemo } from 'react'
+import React, { useRef, useCallback, useMemo, useState, useEffect } from 'react'
 import Selecto from 'react-selecto'
 
 import styles from './RectangularSelection.styl'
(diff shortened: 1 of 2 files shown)


src/modules/selection/RectangularSelection.jsx
```diff
@@ -193,27 +197,18 @@ const RectangularSelection = ({
...
     >
       {children}
-      {containerElement && dragContainer && (
+      {isContainerReady && (
         <Selecto
           className={styles['cozy-selecto-box']}
-          container={containerElement}
-          dragContainer={dragContainer}
+          container={containerRef.current}
+          dragContainer={window}
```
rezk2ll, line 211: why do we still need window?

doubleface: because we want to be able to start and clear a selection from anywhere on the page

rezk2ll (approved):

PR #3695, opened 2026-02-04
feat: Treat not scanned files as infected files

doubleface:
With the `drive.not-scanned-file-action.enabled` flag

When this flag is enabled, the user cannot do any action on files which are not scanned, "remove" action excepted.

https://www.notion.so/linagora/Antivirus-CNB-2fd62718bad1801da516c5831272425b

## Summary by CodeRabbit

* **New Features**
  * Support for not-yet-scanned files: actions can be allowed to operate on files with pending/incomplete antivirus scans when explicitly enabled.
  * Introduced an opt-in option to permit specific actions (e.g., trash) on such files.

* **Policies**
  * New enforcement that blocks actions on not-scanned files unless the action is explicitly allowed.

* **Tests**
  * Added tests covering allow/block behavior across pending, allowed, and disallowed scenarios.

### DIFF
--- a/src/modules/actions/policies.spec.ts
+++ b/src/modules/actions/policies.spec.ts
@@ -8,15 +8,6 @@ import {
 } from './policies'
 import type { DriveAction, ActionPolicyContext } from './types'
 
-// Mock the isInfected helper
-jest.mock('@/modules/filelist/helpers', () => ({
-  isInfected: jest.fn(
-    (file: { antivirus_scan?: { status?: string } } | null | undefined) => {
-      return file?.antivirus_scan?.status === 'infected'
-    }
-  )
-}))
-
 // Mock cozy-client isDirectory
 jest.mock('cozy-client/dist/models/file', () => ({
   isDirectory: jest.fn((file: { type?: string }) => file.type === 'directory')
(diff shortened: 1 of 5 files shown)


src/modules/actions/policies.ts
zatteo, line 54: How often does this method run? And on how many files? Because we have a lot of loops that could be converted in only one.

src/modules/actions/policies.spec.ts
```diff
@@ -8,12 +8,18 @@ import {
...
   isInfected: jest.fn(
     (file: { antivirus_scan?: { status?: string } } | null | undefined) => {
       return file?.antivirus_scan?.status === 'infected'
     }
+  ),
+  isNotScanned: jest.fn(
+    (file: { antivirus_scan?: { status?: string } } | null | undefined) => {
+      const status = file?.antivirus_scan?.status
+      return status === 'pending' || status === 'skipped' || status === 'error'
+    }
```
zatteo: Here you are rewriting the method exactly as it is no?

src/modules/filelist/helpers.ts
```diff
@@ -68,3 +68,10 @@ export const isInfected = (
 ): boolean => {
   return file?.antivirus_scan?.status === 'infected'
 }
+
+export const isNotScanned = (
+  file?: (FileWithAntivirusScan & Partial<IOCozyFile>) | null
+): boolean => {
+  const status = file?.antivirus_scan?.status
+  return status === 'pending' || status === 'skipped' || status === 'error'
+}
```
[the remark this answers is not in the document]
doubleface: I did it on purpose. Files won't have antivirus attributes when antivirus feature is not activated. When antivirus is not activated at all, we don't want to block files actions

src/modules/actions/policies.spec.ts
doubleface: right, this mock was totally useless. I removed it

src/modules/actions/policies.ts
doubleface: It was not run so often, but it is a good catch. Now all these checks are gathered in on loop. Thanks

zatteo (approved):

PR #3713, opened 2026-02-24
feat: Add federated shared folders

doubleface:
This allows to share an existing folder using the shared drives API.
The feature is gated behind the drive.federated-shared-folder.enabled flag.

[Capture vidéo du 2026-02-18 17-17-48.webm](https://github.com/user-attachments/assets/2b321632-d754-4435-a738-800faaa8d0ad)

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent duplicate files by excluding items referenced by shared drives from standard listings.
  * Refined shortcut handling so drive-backed shortcuts render via regular file rendering.
  * Ensure shared-drive items link to the central sharings path where applicable.

* **New Features**
  * Support federated shared folders alongside shared drives; shared-drive route shows when either feature is enabled.

* **Tests**
  * Added test verifying exclusion of shared-drive-referenced files.

* **Chores**
  * Bumped cozy-sharing dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "cozy-pouch-link": "^60.19.0",
     "cozy-realtime": "^5.8.0",
     "cozy-search": "^0.14.1",
-    "cozy-sharing": "^28.3.0",
+    "cozy-sharing": "^28.4.0",
     "cozy-stack-client": "^60.19.0",
     "cozy-ui": "^135.8.0",
     "cozy-ui-plus": "^4.4.1",
--- a/src/constants/config.js
+++ b/src/constants/config.js
@@ -20,3 +20,4 @@ export const MAX_PAYLOAD_SIZE = MAX_PAYLOAD_SIZE_IN_GB * 1024 * 1024 * 1024
 export const SHARING_TAB_ALL = 0
 export const SHARING_TAB_DRIVES = 1
 export const DEFAULT_UPLOAD_PROGRESS_HIDE_DELAY = 5000
+export const SHARINGS_VIEW_ROUTE = '/sharings'
--- a/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.spec.jsx
+++ b/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.spec.jsx
@@ -63,7 +63,7 @@ describe('useTransformFolderListHasSharedDriveShortcuts', () => {
       expect(result.current.sharedDrives).toHaveLength(1)
       expect(result.current.sharedDrives[0]).toMatchObject({
         _id: 'folder-1',
-        id: SHARED_DRIVES_DIR_ID,
+        id: 'folder-1',
         _type: 'io.cozy.files',
         type: 'directory',
         name: 'Shared Drive 1',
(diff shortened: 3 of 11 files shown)


rezk2ll:
I see in the url that it's still a shared drive technically no?

doubleface:
> I see in the url that it's still a shared drive technically no?

@rezk2ll Yes, a federated shared folder is technically a shared drive which can be associated to a folder which is in the the `/Drive` path.

rezk2ll:
The PR is good overall

a few things we can fix:

- `FileThumbnail.tsx` is missingthe  federated flag check
  The shared drive icon condition on line 88 only checks drive.shared-drive.enabled, but everywhere else in this PR the pattern is flag('drive.shared-drive.enabled') || flag('drive.federated-shared-folder.enabled'). If a federated shared folder produces an item where isSharedDriveFolder(file) is true, it won't get the shared drive icon when only the federated flag is on.

-` Sharings/index.jsx`: SharingTab not shown for federated folders?
  Line 240 only renders <SharingTab> when isEnabledSharedDrive is true. Should it also show when isEnabledFederatedSharedFolder is true? The loading guard already waits on sharedDrivesLoaded in both cases, so the tab logic seems like an oversight.

- Worth considering:

  Only referenced_by.data[0] is checked (hook lines 60 & 107). Both the transform and the filter only look at the first referenced_by entry. If the shared drive reference isn't at index 0, items slip through. A .some() check against sharedDriveIds would be more robust.

- Hardcoded '/sharings' in two places helpers.ts:14 and `FileNamePath.jsx`:48, both hardcode '/sharings'. If the route ever changes, these will silently break. Also, getSharedDrivePath was removed from the import in `FileNamePath.jsx` is it now dead code in routeUtils? If so, clean it up.

- Nits:

  - jest.clearAllMocks() -> jest.resetAllMocks() is intentional? resetAllMocks also wipes mock implementations, not just call counts. Works here since mocks are re-set in beforeEach, but worth a heads-up.
  - The !file.driveId guard appears in 3 places (FileIcon.jsx:12, FileThumbnail.tsx:120,124). A small helper like isDriveBackedFile(file) would make the intent clearer and avoid repetition.
  - The Nextcloud shortcut filter (!isNextcloudShortcut(...)) runs identically on sharedDrives in both transformedSharedDrives and sharedDriveIds — could extract a shared filteredSharedDrives memo.

doubleface:
@rezk2ll 

>  Sharings/index.jsx: SharingTab not shown for federated folders?
Line 240 only renders when isEnabledSharedDrive is true. Should it also show when isEnabledFederatedSharedFolder is true? The loading guard already waits on sharedDrivesLoaded in both cases, so the tab logic seems like an oversight.

It's intentional. Federated shared folders are shared drives only technically, but for the the user, it is just normal shared folder. That is why we display federated shared folders in the normal sharing section

doubleface:
@rezk2ll 

> Also, getSharedDrivePath was removed from the import in FileNamePath.jsx is it now dead code in routeUtils? If so, clean it up.

No it's not dead code. It is still used in FilesViewerSharedDrive

doubleface:
@rezk2ll I took your remarks into account

rezk2ll (approved):

PR #3716, opened 2026-02-26
fix: Defer rename action dispatch to prevent focus loss

doubleface:
Rename input field started losing focus with dynamic selection

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the rename input from losing focus when starting a rename, improving typing and keyboard reliability.

* **Refactor**
  * Rewrote the filename input UI as a hook-based functional component, improving maintainability, state clarity, and focus/selection behavior.

* **New Features**
  * Added a confirmation dialog when a rename changes a file extension.

* **Tests**
  * Updated tests to await user interactions and state updates, increasing reliability of rename-related test cases.

### DIFF
--- a/src/modules/filelist/FilenameInput.jsx
+++ b/src/modules/filelist/FilenameInput.jsx
@@ -1,5 +1,5 @@
 import cx from 'classnames'
-import React, { Component } from 'react'
+import React, { useState, useRef, useEffect, useCallback } from 'react'
 import { translate } from 'twake-i18n'
 
 import { isDirectory } from 'cozy-client/dist/models/file'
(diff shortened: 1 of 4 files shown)


rezk2ll (commented):
I understand the problem but `setTimeout(fn, 0)` is a timing workaround, not a proper fix.

The `setTimeout` pushes the dispatch to the next macrotask so the input mounts after all the selection cleanup is done. It works today. But it's held together by the assumption that nothing else will touch focus between "now" and "next tick" , that's not something anyone can guarantee as the codebase evolves.

This is the only action in `src/modules/actions/` using `setTimeout`. Next time someone hits a similar race with `RectangularSelection` (and they will , any action that triggers a focus-sensitive UI has the same exposure), they'll either cargo-cult this pattern or rediscover the root cause from scratch.

### you can try another solution:

The real issue is `handleContainerClick` in `RectangularSelection.jsx`. It clears selection on clicks that aren't on a file or an interactive element , but it doesn't check whether the click happened *inside its own container*. Portaled elements (action menus, context menus) live outside the container DOM, so their clicks look like "background clicks" and trigger the cleanup.

Add this before the `setSelectedItems({})` call and you don't need the `setTimeout` in rename, or in any future action with the same problem.
```jsx
if (!containerRef.current?.contains(target)) return
```

### other stuff to take into account

`FilenameInput` uses `autoFocus` HTML attribute + a `setTimeout(..., 5)` in `handleFocus` for selection range. That's two timing hacks stacked. A `useEffect` with `ref.focus()` would be more predictable, though that's a separate concern

doubleface:
@rezk2ll 
> you can try another solution: ....

The loss of focus is actually not caused by RectangularSelection. It actually happens even if I remove RectangularSelection from FolderViewBodyContent
I could not find any other fix for it

> other stuff to take into account

Cannot do a useEffect since it it a class component. But I tried using componentDidMount and componentDidUpdate for it without any success

doubleface:
@rezk2ll I converted the FilenameInput component to functional component to be able to use useEffect and then remove the call to setTimeout in it and it works

src/modules/filelist/FilenameInput.jsx
```diff
@@ -16,172 +16,177 @@ const ESC_KEY = 27
...
-        working: false,
-        error: true
-      })
+      setWorking(false)
+      setError(true)
+      isSubmittingRef.current = false
     } finally {
-      this.isSubmitting = false
+      setWorking(false)
+      isSubmittingRef.current = false
```
rezk2ll: this is a redundant cleanup in catch + finally, because `finally` always runs after `catch`. The catch-block resets are dead code. Keep `setError(true)` in catch, move all cleanup to `finally`:

```jsx
catch (e) {
  setError(true)
} finally {
  setWorking(false)
  isSubmittingRef.current = false
}
```

src/modules/filelist/FilenameInput.jsx
```diff
@@ -16,172 +16,177 @@ const ESC_KEY = 27
...
-                variant="primary"
-                label={t('RenameModal.continue')}
-                onClick={this.save}
-              />
-            </>
-          }
-          actionsLayout="row"
-        />
-      </div>
+  useEffect(() => {
```
rezk2ll, line 139: this `useEffect` for selection doesn't re-apply on re-focus

The old class component applied selection on every focus via `handleFocus` + `setTimeout`. This is a behavioral regression, unlikely to matter in practice (the input is transient), but the ref+effect indirection adds complexity for no benefit over just doing it inline:

let's try:
```jsx
const handleFocus = () => {
  if (!initialName || !textInput.current) return
  const { filename } = CozyFile.splitFilename({ name: initialName, type: 'file' })
  textInput.current.setSelectionRange(
    0,
    isDirectory(file) ? initialName.length : filename.length
  )
}
```

src/modules/filelist/FilenameInput.jsx
```diff
@@ -16,172 +16,177 @@ const ESC_KEY = 27
...
+          <>
+            <Button
+              variant="secondary"
+              onClick={abort}
+              label={t('RenameModal.cancel')}
+            />
+            <Button
+              variant="primary"
+              label={t('RenameModal.continue')}
+              onClick={save}
```
rezk2ll, line 189: minor: All handlers (`abort`, `save`) are re-created on every render. The `Dialog` now re-renders on every keystroke in the input

rezk2ll (commented):
I liked the refactor @doubleface. Kudos for that.

still we have to get rid of the setTimeout, let's try to fix the `handleContainerClick` instead with the guard:

```jsx
if (!containerRef.current?.contains(target)) return
```

src/modules/filelist/FilenameInput.jsx
doubleface: I tried this at first (and retried just now to be sure) but when `handleFocus` is called, `textInput.current` is still null and then the range selection is not done at all

src/modules/filelist/FilenameInput.jsx
doubleface: Now handlers are wrapped with useCallback

doubleface:
> still we have to get rid of the setTimeout, let's try to fix the `handleContainerClick` instead with the guard:
> 
> ```js
> if (!containerRef.current?.contains(target)) return
> ```

I also already tried this without any success. Actually, the focus is lost even if I don't wrap the view content with RectangularSelection component. It looks like the loss of focus is not caused by this selection.

rezk2ll (approved):

PR #3759, opened 2026-03-25
fix: Make rectangular selection more sensitive

doubleface:
We found we tend to miss to many items in many regular selections


zatteo (approved):
Is there any negative impact for this value? Like increased CPU consumption?

PR #3760, opened 2026-03-25
fix: Handle FetchError Forbidden on public pages

Crash--:
## Summary

Fixes Sentry issue [DRIVE-ZG2](https://errors.cozycloud.cc/organizations/cozycloud/issues/306214/) — **117,320 occurrences/day**, affecting **12,717 users**.

- Pass `isPublic` prop to `SharingProvider` in `DriveProvider.jsx` so it can skip realtime subscriptions on public pages (which use a read-only sharecode token that gets `403 Forbidden`)
- Add tests for `DriveProvider` verifying the `isPublic` prop is correctly forwarded

## Root Cause

When users access shared documents via public links, the app authenticates with a read-only sharecode token. The `SharingProvider` attempts to subscribe to realtime events, but the token lacks permissions → `FetchError: {"error":"Forbidden"}`.

## Test Plan

- [ ] Verify `DriveProvider.spec.jsx` tests pass
- [ ] Verify public link access to shared docs no longer generates Forbidden errors in Sentry

🤖 Generated with [Sentry Watcher](https://github.com/linagora/sentry-watcher) + Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Sharing now respects the public context setting so permissions adjust correctly for public vs non-public cases.

* **Tests**
  * Added tests to validate sharing behavior across both public and non-public contexts.


zatteo:
Added tests looks useless?

Can't wait to see more about how this PR was done 👀

doubleface:
Yes, the test looks useless and should be removed but the fix looks quite relevant

rezk2ll:
After testing locally i still see some 403 messages when sharing a file by link:

```
917.4a111f03.js?6abedbce7e9d1761:25 cozy-realtime Stack returned an error {status: '403 Forbidden', code: 'forbidden', title: "The application can't subscribe to io.cozy.sharings", source: {…}}
```

```
GET instance.localhost:8080/data/io.cozy.files.settings/_normal_docs?include_docs=true&limit=100 403 (Forbidden)
```

```
POST http://instance.localhost:8080/files/_find 403 (Forbidden)
```

```
cozy-drive Failed to load settings: FetchError: {"error":"code=403, message=Forbidden"}
```

Crash--:
I think we can merge it since this PR already reduce the number of errors, right?

rezk2ll (approved):

PR #3762, opened 2026-03-25
fix: Maintain selection when scrolling past screen in virtualized mode

doubleface:
When using rectangular selection with autoscroll on a virtualized list, items that scrolled out of view were being lost from the selection.

Now we accumulate all items selected during the drag operation in a Set, ensuring that once an item is selected, it stays selected even if it scrolls out of the viewport and is removed from the DOM by virtuoso.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rectangular drag selection now reliably accumulates selections across the drag, preventing lost items during multi-item drag selections.
  * Holding Ctrl/Cmd when starting a drag correctly preserves and extends the current selection.
  * Drag start/end and container click behavior refined to avoid accidental clears while dragging or when using keyboard modifiers.

### DIFF
--- a/src/modules/selection/RectangularSelection.jsx
+++ b/src/modules/selection/RectangularSelection.jsx
@@ -41,6 +41,7 @@ const RectangularSelection = ({
   const isDraggingRef = useRef(false)
   const dragStartPosRef = useRef(null)
   const mutationObserverRef = useRef(null)
+  const selectedDuringDragRef = useRef(new Set())
 
   useEffect(() => {
     if (containerRef.current) {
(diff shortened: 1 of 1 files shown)


src/modules/selection/RectangularSelection.jsx
```diff
@@ -90,15 +91,22 @@ const RectangularSelection = ({
...
+      }
+
+      // Build newSelection from accumulated items
+      const newSelection = {}
+      let lastSelectedId = null
+      for (const fileId of selectedDuringDragRef.current) {
+        const file = itemsMap.get(fileId)
+        if (file) {
+          newSelection[fileId] = file
+          lastSelectedId = fileId
```
zatteo: nit: In an helper function? 🤷

zatteo (approved):

PR #3764, opened 2026-03-25
fix: Fix rename input click behavior (selection toggle + Firefox cursor)

doubleface:
- When the `drive.dynamic-selection.enabled` flag is disabled, clicking on the rename input was toggling the row's selection state instead of positioning the cursor
- On Firefox, the cursor was not repositioned at all when clicking inside the rename input

## Changes

**Skip FileLink wrapper during rename** (`FileOpener.jsx`)
When renaming, the `FileOpener` returns children directly instead of wrapping them in a `FileLink` (`<a>` tag). This prevents the browser from navigating to the file viewer. The `<a>` tag's activation behavior fires regardless of `stopPropagation` (per HTML spec), and using `preventDefault` to block it also breaks cursor positioning on Firefox.

**Prevent row selection toggle during rename** (`Table.jsx`)
`handleRowSelect` now returns early when a rename is active. This is needed because cozy-ui's `VirtualizedTable` Cell calls the row `onClick` with `(row, column)` — not `(row, event)` — so it's not possible to filter clicks by event target.

**Manually reposition cursor on Firefox** (`FilenameInput.jsx`, `getCaretPositionFromPoint.js`)
Firefox does not natively reposition the cursor when clicking inside the rename input in this DOM context. A `mouseup` handler uses `document.caretPositionFromPoint` (standard/Firefox) or `document.caretRangeFromPoint` (Chrome/Safari) to manually compute and apply the caret position.

Other approaches were tested and failed:
- `stopPropagation` (React synthetic or native) on the input wrapper — did not fix cursor positioning on Firefox
- `preventDefault` on click to block `<a>` navigation — blocks cursor repositioning on Firefox
- `user-select: text` on the input element — no effect, computed value already resolves to `text`
- Changing `display: table-cell` to `display: block` on the wrapper — no effect

https://www.notion.so/linagora/Difficulty-renaming-a-file-or-folder-31962718bad180239db4f7cd43ba952f?d=f4a62718bad18314870903fbfa68c86d

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text cursor positioning in filename inputs when clicking, including more consistent behavior across browsers.
  * Prevented row selection and interaction from interrupting an active file rename session.
  * Ensured file entries render appropriately while renaming to avoid unintended interactions.

### DIFF
--- a/src/modules/filelist/FileOpener.jsx
+++ b/src/modules/filelist/FileOpener.jsx
@@ -25,6 +25,10 @@ const FileOpener = ({
     onInteractWithFile
   })
 
+  if (isRenaming) {
+    return children
+  }
+
   return (
     <FileLink
       ref={rowRef}
--- a/src/modules/filelist/FilenameInput.jsx
+++ b/src/modules/filelist/FilenameInput.jsx
@@ -10,6 +10,7 @@ import { translate } from 'twake-i18n'
 import styles from '@/styles/filenameinput.styl'
 
 import { CozyFile } from '@/models'
+import { getCaretPositionFromPoint } from '@/modules/filelist/getCaretPositionFromPoint'
 
 const ENTER_KEY = 13
 const ESC_KEY = 27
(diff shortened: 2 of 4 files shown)


zatteo (approved):
In the first commit, there is a change related to the third commit

PR #3772, opened 2026-03-30
Remove some flags that are not enabled by default

zatteo:
* **Features now enabled by default without flag**
  * Keyboard shortcuts enabled unconditionally
  * New-item highlighting always active
  * Folder personalization shown by default
  * ~~Sort preferences saved automatically~~
  * Favicon updates applied when available
  * Health theme always listed
  * Click/selection behavior simplified for consistent item selection and interactions

### DIFF
--- a/jestHelpers/setup.js
+++ b/jestHelpers/setup.js
@@ -21,10 +21,7 @@ jest.mock('cozy-dataproxy-lib', () => ({
 
 // Mock cozy-flags with jest mock function that supports both flag checking and test mocking
 jest.mock('cozy-flags', () => {
-  const mockFn = jest.fn(flagName => {
-    if (flagName === 'drive.highlight-new-items.enabled') {
-      return true
-    }
+  const mockFn = jest.fn(() => {
     // Return false for all other flags to avoid issues
     return false
   })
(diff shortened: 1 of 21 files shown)


rezk2ll (approved):

JF-Cozy (approved):
don't forget to remove them from Notion after merging :)

doubleface (commented):
Looks like there are still some unit tests to fix

zatteo:
I will check that all these flags are already enabled in every environments before merging.

Crash--:
@zatteo Please, move forward with this PR. It's complicated to have Drive locally without the behavior we have in production because of the flags. 

So yes, let's have this PR merged in some way :)

zatteo:
> @zatteo Please, move forward with this PR. It's complicated to have Drive locally without the behavior we have in production because of the flags.
> 
> So yes, let's have this PR merged in some way :)

It's in progress. We found a production environment where we did not enabled these flags so we are trying to enable them smoothly before removing them totally in the code.

zatteo:
`drive.save-sort-choice.enabled` removal has been postponed because it is still not activated in a production environment. Otherwise all flags removed in this PR are now activated for a week or more in all environment. We can safely merge it and deploy it.

PR #3781, opened 2026-04-01
feat: Quick and dirty implementation of pagination for shared drives

doubleface:
This implementation works but I think this is code which should be in cozy-client.

## Summary by CodeRabbit

* **New Features**
  * Added paginated loading for shared drive folders with load-more capability, a fetch status indicator, and has-more signaling to improve browsing large folders.

* **Bug Fixes**
  * Safer pagination: concurrent loads are serialized and errors during pagination preserve already-loaded items while logging failures.

* **Tests**
  * Added comprehensive tests covering initial load, pagination, error cases, and concurrent fetch behavior.


src/modules/views/SharedDrive/SharedDriveFolderView.jsx
```diff
@@ -64,13 +64,20 @@ const SharedDriveFolderView = () => {
...
+  const { sharedDriveResult, hasMore, fetchMore } = useSharedDriveFolder({
     driveId,
     folderId
   })
 
   const queryResults = sharedDriveResult
-    ? [{ fetchStatus: 'loaded', data: sharedDriveResult.included }]
+    ? [
+        {
+          fetchStatus: 'loaded',
```
rezk2ll: can we derive the fetchStatus here from the query instead of hardcoding it?

src/modules/shareddrives/hooks/useSharedDriveFolderHelpers.js
```diff
@@ -0,0 +1,26 @@
+const PAGE_LIMIT = 100
```
rezk2ll: This can be a typescript helper

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
```diff
@@ -87,11 +101,43 @@ const useSharedDriveFolder = ({
       }
       debouncedFetch.cancel()
     }
-  }, [client, driveId, folderId, sharedDriveQuery])
+  }, [client, driveId, folderId, sharedDriveQuery, statById])
```
rezk2ll: do we need `sharedDriveQuery` ? it seems stale to me

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
```diff
@@ -87,11 +101,43 @@ const useSharedDriveFolder = ({
...
+
+  const fetchMore = useCallback(async (): Promise<void> => {
```
rezk2ll, line 131: This produces a new function on each page load. Can we check if no component memoizes it

rezk2ll (commented):
Overall, it's fine, just one thing that I think about:

What happens if you rename a file after loading 3 pages?

cause when a file is created, updated, or deleted, the real-time subscription triggers a debounced re-fetch. That re-fetch calls `statById(folderId)` without a cursor, so it loads only page 1 and replaces the entire state

so the scenario to test:
- User scrolls -> page 1 + page 2 + page 3 loaded (300 files visible)
- File is renamed (realtime event)
   -> debouncedFetch fires
   -> fetches page 1 only (100 files)
   -> setSharedDriveResult({ included }) // overwrites everything
- User now sees only 100 files

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
doubleface: No indeed. I'm removing it

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
doubleface: Now I stabilize fetchMore reference using a cursor reference

doubleface:
> Overall, it's fine, just one thing that I think about:
> 
> What happens if you rename a file after loading 3 pages?
> 
> cause when a file is created, updated, or deleted, the real-time subscription triggers a debounced re-fetch. That re-fetch calls `statById(folderId)` without a cursor, so it loads only page 1 and replaces the entire state
> 
> so the scenario to test:
> 
>     * User scrolls -> page 1 + page 2 + page 3 loaded (300 files visible)
> 
>     * File is renamed (realtime event)
>       -> debouncedFetch fires
>       -> fetches page 1 only (100 files)
>       -> setSharedDriveResult({ included }) // overwrites everything
> 
>     * User now sees only 100 files

Ok, now I refetch all loaded page on realtime event

rezk2ll (approved):

PR #3784, opened 2026-04-02
feat: block uploads exceeding 500 files with a warning dialog

rezk2ll:
## Summary

- Uploading large numbers of files via the web silently fails after several minutes due to browser limitations (stale file references, tab switching, sleep, network drops). A technical fix landed in beta but the web path remains fragile for bulk uploads.
- Before starting the upload, all files are now counted (recursively for dropped folders). If the count reaches the configured limit, a dialog warns the user and suggests downloading the desktop sync tool instead.
- The limit defaults to 500 and can be overridden via the `drive.max-upload-file-count` flag.
- The "Download on Desktop" button is hidden on public share links where the desktop app is not available.
- The download link is controlled by the existing `cozy.desktop-app-download-link` flag, falling back to a platform-aware default URL.

https://github.com/user-attachments/assets/8773998d-cf61-47c6-b612-924172b974bd

## Summary by CodeRabbit

* **New Features**
  * Enforced a maximum upload limit (default 500 files per upload); uploads that exceed the limit are blocked and show a localized upload-limit dialog with actions including “Download on Desktop” when available.
* **Tests**
  * Added coverage for the upload flow, entry extraction, recursive directory counting, and limit-handling behaviors.

### DIFF
--- a/src/modules/navigation/duck/actions.jsx
+++ b/src/modules/navigation/duck/actions.jsx
@@ -9,14 +9,20 @@ import {
   ROOT_DIR_ID,
   TRASH_DIR_ID,
   FILES_FETCH_LIMIT,
-  MAX_PAYLOAD_SIZE_IN_GB
+  MAX_PAYLOAD_SIZE_IN_GB,
+  MAX_UPLOAD_FILE_COUNT
 } from '@/constants/config'
 import { createEncryptedDir } from '@/lib/encryption'
 import { getEntriesTypeTranslated } from '@/lib/entries'
 import logger from '@/lib/logger'
 import { showModal } from '@/lib/react-cozy-helpers'
 import { getFolderContent, getFolderContentQueries } from '@/modules/selectors'
-import { addToUploadQueue } from '@/modules/upload'
+import {
+  addToUploadQueue,
+  extractFilesEntries,
+  exceedsFileLimit
+} from '@/modules/upload'
+import UploadLimitDialog from '@/modules/upload/UploadLimitDialog'
 
 export const SORT_FOLDER = 'SORT_FOLDER'
 export const OPERATION_REDIRECTED = 'navigation/OPERATION_REDIRECTED'
@@ -82,7 +88,7 @@ export const uploadFiles =
     driveId,
     addItems
   ) =>
-  dispatch => {
+  async dispatch => {
     let targetDirId = dirId
     let navigateAfterUpload = false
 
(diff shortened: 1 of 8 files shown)


doubleface:
nit: I think prettier fixes should be merged with the first commit

src/modules/upload/index.js
```diff
@@ -597,3 +597,36 @@ const extractFilesEntries = items => {
...
+  if (fileCount >= limit) return fileCount
+
+  const dirCounts = await Promise.all(
+    directories.map(e => countDirectoryFiles(e.entry))
+  )
+
+  let count = fileCount
+  for (const dirCount of dirCounts) {
+    count += dirCount
+    if (count >= limit) return count
```
zatteo: This make the name of the method wrong no? Even if I understand the speed gain.

rezk2ll: fair point, i changed the behaviour and name to `exceedsFileLimit`

src/modules/navigation/duck/actions.jsx
```diff
@@ -91,9 +97,21 @@ export const uploadFiles =
...
     }
 
+    const maxFileCount =
+      flag('drive.max-upload-file-count') ?? MAX_UPLOAD_FILE_COUNT
+
+    // Extract entries synchronously before browser clears dataTransfer
+    const entries = extractFilesEntries(files)
+    const totalFileCount = await countFilesFromEntries(entries, maxFileCount)
+
+    if (totalFileCount >= maxFileCount) {
```
doubleface: When you look at the error message, it should be ">". Here 500 files upload is forbidden

zatteo (approved):
Looks good apart my single comment 👍

src/modules/navigation/duck/actions.jsx
rezk2ll: correct, fixed

doubleface (approved):
Just one tiny comment

src/modules/upload/index.js
```diff
@@ -597,3 +597,36 @@ const extractFilesEntries = items => {
...
+    directories.map(e => countDirectoryFiles(e.entry))
+  )
+
+  let count = fileCount
+  for (const dirCount of dirCounts) {
+    count += dirCount
+    if (count >= limit) return true
+  }
+
+  return false
```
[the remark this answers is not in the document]
rezk2ll: @coderabbitai The parallel approach (Promise.all) is faster for the 99% case where uploads are under the limit. The FileSystem API calls are local/in-memory, so even over-limit scenarios resolve in milliseconds. Adding a remaining budget parameter would complicate countDirectoryFiles for a negligible gain in an edge case. Fixed the JSDoc to accurately describe the behavior.

src/modules/navigation/duck/actions.jsx
```diff
@@ -91,9 +97,20 @@ export const uploadFiles =
...
+    const maxFileCount =
+      flag('drive.max-upload-file-count') ?? MAX_UPLOAD_FILE_COUNT
+
+    // Extract entries synchronously before browser clears dataTransfer
+    const entries = extractFilesEntries(files)
+
+    if (await exceedsFileLimit(entries, maxFileCount)) {
+      dispatch(showModal(<UploadLimitDialog maxFileCount={maxFileCount} />))
+      return
+    }
```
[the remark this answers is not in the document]
rezk2ll: @coderabbitai If exceedsFileLimit throws (a directory reader rejection), showing the upload limit dialog would be wrong, the user didn't exceed any limit, something broke.

Silently swallowing the error and blocking the upload with a misleading "too many files" modal is worse than letting the error propagate naturally.

The existing upload error handling further down the chain already catches and reports failures to the user.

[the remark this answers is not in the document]
rezk2ll: no 😏

PR #3785, opened 2026-04-06
fix: use fileId for upload queue identity instead of file.name

Crash--:
## Summary

- Fix a bug where the upload queue reducer matches items by `file.name`, causing two files with the same name (e.g. `summer/photo.jpg` and `winter/photo.jpg`) to collide — an action targeting one incorrectly updates both
- Introduce a `fileId` field on queue items (defaults to `file.name` for backward compatibility) used as the identity key in the reducer
- This is a prerequisite for the folder upload flattening feature, where files from different sub-folders will appear individually in the queue

## Test plan

- [x] 3 new unit tests demonstrate the collision bug and pass after the fix
- [x] All 33 existing tests pass
- [ ] Manual test: upload two files with the same name to different folders and verify independent progress/status tracking

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Upload queue now assigns a stable per-file identifier so uploads, progress updates, overwrites, removals and error handling reliably target the correct file—even when files share the same name. Conflict-overwrite and quota/error flows now report progress and completion consistently.

* **Tests**
  * Expanded test coverage validating per-file identifier behavior across progress, success, error, overwrite and duplicate-name scenarios, plus backward compatibility.

### DIFF
--- a/src/modules/upload/index.js
+++ b/src/modules/upload/index.js
@@ -55,7 +55,7 @@ const CONFLICT_ERROR = 409
 
 const itemInitialState = item => ({
   ...item,
-  fileId: item.fileId ?? item.file.name,
+  fileId: item.fileId ?? item.file?.name ?? item.entry?.name,
   status: PENDING,
   progress: null
 })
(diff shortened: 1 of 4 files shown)


src/modules/upload/index.js
```diff
@@ -202,12 +204,12 @@ export const processNextFile =
       return dispatch(onQueueEmpty(queueCompletedCallback))
     }
 
-    const { file, entry, isDirectory } = item
+    const { file, fileId, entry, isDirectory } = item
```
Crash--: I think we should merge this fix as is. 

i'm working on improving things in an other PR.

Crash--:
@rezk2ll Can i close it too?

rezk2ll:
> @rezk2ll Can i close it too?

yes, it's fully integrated in https://github.com/linagora/twake-drive/pull/3813

PR #3789, opened 2026-04-08
fix: adapt upload limit dialog for public shared links

rezk2ll:
## Summary

- On public shared links, the upload limit dialog was suggesting to download the desktop sync app, which makes no sense since the user is not logged in and shared links are not accessible from the desktop app
- Show a simpler message asking to reduce the number of files, with just a "Close" button instead of "Cancel" + "Download on Desktop"

## Summary by CodeRabbit

* **New Features**
  * Improved upload-limit messaging advising users to reduce file count and retry.
  * Context-aware upload dialog actions: public uploads show a single "Close" action; non-public uploads show "Cancel" and "Download for desktop".
  * Added localized strings for the new message and the "Close" label in English and French.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -362,7 +362,9 @@
     "limit": {
       "title": "You cannot upload more than %{limit} files at a time.",
       "content": "Need to upload more? Consider downloading the synchronization tool to your computer",
+      "content_public": "Please reduce the number of files and try again.",
       "cancel": "Cancel",
+      "close": "Close",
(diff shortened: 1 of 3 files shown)


doubleface (approved):
It's a bit of a unfortunate to have a commit just just for prettier formatting. Let's have just one commit. LGTM for the rest

rezk2ll:
> It's kind of a shame to have a commit just just for prettier formatting. Let's have just one commit. LGTM for the rest

good remark, fixed

PR #3796, opened 2026-04-09
fix: Reset lastClicked in NavLink when route matches

### LINKED ISSUE #3793: Sidebar: wrong selected section
You can see that I'm on my root folder, but the "Partages" section is still active

The "Fichiers" section should be active 

To reproduce every time (I'm pretty sure the first time I didn't do that, but at least it's a working scenario): go to you "Fichiers" section, then click on "Partages" then click on your browser's back button. 

The sidebar should be connected to the router. We should not handle that by hand. 

Image

doubleface:
Keeps the optimistic highlight in the sidebar but also works when using the browser back button.

Also added unit tests for these cases

fixes https://github.com/linagora/twake-drive/issues/3793

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation link active-state logic to support an optimistic "pending click" state and reliably clear that pending state when navigation completes or when the route changes (including back/alternate navigation), ensuring accurate active indication.

* **Tests**
  * Added a comprehensive test suite covering click handling, optimistic active-state behavior, and reset logic across route changes.

### DIFF
--- a/src/modules/navigation/NavLink.jsx
+++ b/src/modules/navigation/NavLink.jsx
@@ -1,6 +1,6 @@
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import React, { useEffect } from 'react'
+import React, { useEffect, useRef } from 'react'
 import { useLocation } from 'react-router-dom'
 
 import { NavLink as UINavLink } from 'cozy-ui/transpiled/react/Nav'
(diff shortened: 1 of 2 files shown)


Crash--:
@coderabbitai Does it fix also this use case?  https://github.com/linagora/twake-drive/issues/3793#issuecomment-4211318636 

I can't see a UT for this one

doubleface:
> To fully fix this, the effect should **also** reset `lastClicked` whenever the route changes to a path that **does not** match the clicked destination:
> 
> ```js
> useEffect(() => {
>   if (!lastClicked) return
>   if (navLinkMatch(rx, lastClicked, location.pathname)) {
>     setLastClicked(null) // route arrived at destination
>   } else if (!navLinkMatch(rx, to, location.pathname)) {
>     setLastClicked(null) // route went somewhere else (e.g., back button)
>   }
> }, [location.pathname, rx, lastClicked, to, setLastClicked])
> ```
> 

Good catch on the missing case. However the proposed fix has a subtle flaw: `navLinkMatch` ignores the `to` argument entirely when `rx` is provided (it just runs `rx.test(pathname)`), so the `else if (!navLinkMatch(rx, to, location.pathname))` branch is always the logical complement of the first — meaning `lastClicked` would be cleared **immediately on click**, before any route change happens, defeating the whole optimistic-highlight purpose.

The fix I went with instead tracks the previous `pathname` in a `useRef`. On each effect run we compare the new `location.pathname` to the one from the last run/

This way:
- **Click fires** → `lastClicked` changes but `pathname` is still the same as `prevPathnameRef.current` → nothing happens, we wait for navigation
- **Route reaches destination** → `pathname !== prevPathname`, first condition matches → clear
- **Back button / unexpected navigation** → `pathname !== prevPathname`, second condition matches → clear

Also updated the existing test that was encoding the bug (`not.toHaveBeenCalledWith(null)`) to assert the correct behaviour. All 8 tests pass.

PR #3799, opened 2026-04-13
feat: Added migration section leading to settings application 

doubleface:
behind settings.migration.enabled flag

image

## Summary by CodeRabbit

* **New Features**
  * Added a new "Migration" navigation item in the app header; visible when the related feature flag is enabled. Translations added for English, French, Russian, and Vietnamese.

* **Chores**
  * Bumped cozy-ui dependency to ^138.5.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "cozy-search": "^0.25.3",
     "cozy-sharing": "^30.1.1",
     "cozy-stack-client": "^60.23.0",
-    "cozy-ui": "^138.3.0",
+    "cozy-ui": "^138.5.0",
     "cozy-ui-plus": "^7.0.0",
     "cozy-viewer": "^28.0.7",
     "date-fns": "2.30.0",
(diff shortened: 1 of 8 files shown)


JF-Cozy (approved):

rezk2ll:
### Code review

Found 1 issue:

1. `clickState[1](false)` passes a boolean to a setter that expects `string | undefined | null`. The `clickState` tuple from `NavContext` is `useState(null)` and tracks the last-clicked route path for optimistic active-state rendering in `NavLink`. Every other consumer resets it with `undefined` (see `FavoriteListItem`, `ExternalDriveListItem`). Passing `false` is accidentally harmless today because `NavLink` only uses truthy checks, but it corrupts the semantic contract of the shared state and would break any future strict-equality check on `lastClicked`. The correct value is `undefined`.

https://github.com/linagora/twake-drive/blob/c4226a8ce90b9f99343a5ee63186e24011327d8f/src/modules/navigation/ExternalNavItem.jsx#L29-L33

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3813, opened 2026-04-24
refactor: simplify upload module, drop encryption writes

rezk2ll:
## Problem

Re-dropping an existing folder containing files used to show "no internet connection". Investigation surfaced more: conflict handling lived in three places, `processNextFile` recursed into folders inline, and the encrypted-folder write path bolted a fragile FileReader fire-and-forget branch onto the same code (gated by a flag never enabled in production). The flow had become hard to follow and hard to extend safely.

## Solution

**Flatten folder uploads at enqueue time.** Walk the dropped tree once, create (or reuse on 409) every intermediate folder upfront, and turn each file into its own queue item with a relative path. `processNextFile` then only ever handles single files.

**Show a placeholder while the tree walk runs.** Folder drops dispatch one placeholder row per top-level folder immediately so the upload tray reflects the drop. Flatten runs in the background; a `RESOLVE_FOLDER_ITEMS` action swaps placeholders for real per-file rows when it finishes.

Conflict handling collapses into two parallel helpers: `uploadOrOverwriteFile` (file 409 → overwrite) and `createFolderOrGetExisting` (folder 409 → reuse). Error classification is one function (`classifyUploadError`); the prep-time alert routes through one handler (`reportPrepError`).

## Demo

https://github.com/user-attachments/assets/2e189d91-8f6c-4566-b297-37e80fd236bb

## What has been removed and why

- **The encrypted-folder write path** (`drive.enable-encryption` flag, `AddEncryptedFolderItem`, `encryptAndUploadNewFile`, `createEncryptedDir`, `getEncryptionKeyFromDirId`'s flag gate). The flag was never enabled in production, so the entire branch was dead code. The encrypted upload branch also had a synchronous `fr.readAsArrayBuffer` that returned `undefined` immediately, breaking conflict detection on encrypted uploads.
- **`uploadDirectory`** (the recursive in-process helper) — replaced by flatten-at-enqueue.
- **`vaultClient` threading** through every upload call site (`Dropzone`, `DropzoneDnD`, `UploadButton`, `UploadItem`, `AddFolder`, `FolderPickerAddFolderItem`, plus the `createFolder` and `uploadFiles` actions).

Read-side encryption helpers (`decryptFile`, `isEncryptedFolder`, `downloadEncryptedFile`, `getDecryptedFileURL`) stay, so existing encrypted content remains viewable and downloadable.

## The gain

- The "no internet" bug for folder re-drops is structurally gone — inner-file 409s no longer escape into a directory catch.
- Each file in a folder drop shows its own progress row with its relative path (`photos/2024/img.jpg`) instead of one line for the whole folder.
- The upload tray appears the instant after the drop, even on deep trees, thanks to the placeholder rows.
- `processNextFile` is 25 lines top-to-bottom, no nested try/catch.
- `uploadFile` is roughly half its previous size.

## What we lost

- New encrypted folders can no longer be created, and files added to existing encrypted folders are now plaintext. Deliberate given the flag was off in production, but a real behaviour change for anyone who had it on.
- Folder drops still issue N folder-create calls before files start uploading; deep trees sit in "resolving" for a few seconds. The placeholder masks the wait. Removing it entirely would mean lazy per-file folder resolution — left as a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Folder drops now emit resolving placeholders, expand into per-file upload items, and show a preparing state in the upload queue.

* **Bug Fixes**
  * Uploads track progress by stable IDs, classify oversized payloads consistently, and queue-empty events wait for folder resolution.

* **Refactor**
  * Removed client-side encryption/vault flows and related UI gating across uploads, downloads, viewer, picker, and folder tooling.

* **Tests**
  * Added coverage for folder flattening, queue lifecycle, and limit handling; removed encryption-specific tests.

* **Chores**
  * Removed encrypted-folder UI strings and menu entries.


Crash--:
I don't understand the flow anymore... We've added a lot of if/else case lately this is not sustainable.

rezk2ll:
> I don't understand the flow anymore... We've added a lot of if/else case lately this is not sustainable.

i'm thinking of refactoring the whole thing

zatteo:
What is the status of file encryption support after your second commit? Is there still some working code? Did you inform Zoé and @paultranvan ?

Crash--:
I don't want to keep this feature (folder encryption) for now. It adds a lot of complexity for nothing.

src/lib/encryption.js
```diff
@@ -24,60 +22,19 @@ export const isEncryptedFile = file => {
...
 }
 
+/**
+ * Fetch the encryption key for a previously-created encrypted directory.
+ * Returns null when no key document exists, which is the case for any
+ * folder created after the write-side feature was removed. Read-side
+ * callers (downloadFiles, FilesViewer) keep this around so existing
+ * encrypted content stays decryptable.
+ */
 export const getEncryptionKeyFromDirId = async (client, dirId) => {
```
Crash--: you can remove

src/modules/upload/Dropzone.jsx
```diff
@@ -55,19 +46,16 @@ export const Dropzone = ({
   const onDrop = async (files, _, evt) => {
     if (!canDrop(evt)) return
 
-    const filesToUpload = canHandleFolders(evt) ? evt.dataTransfer.items : files
+    // react-dropzone already unwraps folders into individual File objects
```
Crash--: is it always enabled?

src/lib/encryption.js
```diff
@@ -24,60 +22,19 @@ export const isEncryptedFile = file => {
...
-      name,
-      dirId: dirID,
-      onUploadProgress,
-      encrypted: true
-    })
-  return resp.data
+  return res?.data?.key ?? null
 }
 
 const getBinaryFile = async (client, fileId) => {
```
Crash--: i'm pretty sure you can remove this file? It should not be used anymore

src/modules/upload/index.js
```diff
@@ -48,16 +52,18 @@ export const status = {
...
   ERR_MAX_FILE_SIZE
 }
 
 const CONFLICT_ERROR = 409
+const PAYLOAD_TOO_LARGE = 413
 
 const itemInitialState = item => ({
   ...item,
-  status: PENDING,
+  status: item.status || PENDING,
```
Crash--, line 69: why this change?

Crash--:
> Conflict handling collapses into two parallel helpers: uploadOrOverwriteFile (file 409 → overwrite) and createFolderOrGetExisting (folder 409 → reuse). Error classification is one function (classifyUploadError); the prep-time alert routes through one handler (reportPrepError).

Why do we still have 2 paths?

rezk2ll:
> > Conflict handling collapses into two parallel helpers: uploadOrOverwriteFile (file 409 → overwrite) and createFolderOrGetExisting (folder 409 → reuse). Error classification is one function (classifyUploadError); the prep-time alert routes through one handler (reportPrepError).
> 
> Why do we still have 2 paths?

because we have two different cases to handle when dealing with conflict. For files, we overwrite the file content and for folders, we reuse the existing folder.

We can have a single function that handles both conflicts but i think it will obscure the behaviour:

like:

```ts
const createOrReuseOnConflict = async (create, reuse) => { 
  try { return await create() }                                            
  catch (err) {
    if (err.status !== CONFLICT_ERROR) throw err
    return reuse()
  }
}
```
and have both helpers become two-line wrappers.

But I think the duplication is needed here since the two resource types are genuinely different

src/modules/upload/index.js
rezk2ll: When dropping a folder with deeply nested folders, the folder creation takes place before the file upload and it takes time and we will notice the queue ui is not shown ( and probably the user will think something is broken ) so i introduced a RESOLVING status that displays the destination is folder is being resolved and when it's ready we fill the queue ui with the files being uploaded.

rezk2ll:
@coderabbitai full review

src/modules/upload/Dropzone.jsx
rezk2ll: The dropzone honours the `disabled` prop the parent passes, it is not hard-enabled

src/modules/upload/index.js
```diff
@@ -511,44 +481,277 @@ export const overwriteFile = async (
...
+
 export const removeFileToUploadQueue = file => async dispatch => {
-  dispatch({ type: RECEIVE_UPLOAD_SUCCESS, file, isUpdate: true })
+  dispatch({
+    type: RECEIVE_UPLOAD_SUCCESS,
+    fileId: file.name,
+    file,
+    isUpdate: true
+  })
+}
```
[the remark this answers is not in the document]
rezk2ll: Out of scope here; worth a separate cleanup PR

src/locales/en.json
```diff
@@ -295,8 +294,7 @@
     },
     "download_file": {
       "offline": "You should be connected to download this file",
-      "missing": "This file is missing",
-      "encryption_many": "You cannot download several encrypted files at once"
+      "missing": "This file is missing"
     },
```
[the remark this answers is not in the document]
rezk2ll: Out of scope. `it.json` has been missing this key since long before this PR; it's a pre-existing locale-parity gap, not a regression

src/locales/ko.json
```diff
@@ -109,4 +108,4 @@
       "chooseCustomIcon": "사용자 지정 아이콘 선택"
     }
   }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
rezk2ll: The `Shortcut` block we removed from `ko.json` was an empty `{}` (no translations). Other locales have a populated `Shortcut` block. Removing the empty stub doesn't change runtime behaviour because `t('Shortcut.foo')` would have fallen back to the default locale either way. Aligning the rest of the locales is a separate i18n housekeeping job

src/modules/actions/utils.js
```diff
@@ -27,49 +22,18 @@ const downloadFileError = error => {
...
-        return client
-          .collection(DOCTYPE_FILES, { driveId })
-          .download(file, null, filename)
-      }
+      return await client
+        .collection(DOCTYPE_FILES, { driveId })
+        .download(file, null, file.name)
     } catch (error) {
       showAlert({ message: t(downloadFileError(error)), severity: 'error' })
     }
```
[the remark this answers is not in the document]
rezk2ll: this is hypothetical. trashFiles right below uses the exact same pattern: showAlert(...) and t(...) without guards. The "trust the caller" convention is project-wide, not specific to downloadFiles.

All UI call sites pass both showAlert and t; the = {} default just guards against the entire options arg being omitted.

src/modules/drive/Toolbar/components/InsideRegularFolder.jsx
```diff
@@ -1,15 +1,11 @@
...
 /**
  * Displays its children only if we are in a normal folder (eg. not the root folder or a special view like sharings or recent)
  */
 const InsideRegularFolder = ({ children, displayedFolder, folderId }) => {
   const insideRegularFolder =
-    folderId &&
-    displayedFolder &&
-    displayedFolder.id !== ROOT_DIR_ID &&
-    !isEncryptedFolder(displayedFolder)
+    folderId && displayedFolder && displayedFolder.id !== ROOT_DIR_ID
```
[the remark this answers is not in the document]
rezk2ll: You are asking me to "restore protection" for encrypted folders, but encrypted folders no longer exist as a concept anywhere in the codebase. we removed the entire feature

src/modules/upload/UploadQueue.jsx
```diff
@@ -1,101 +1,264 @@
-import React, { useEffect } from 'react'
-import { connect } from 'react-redux'
+import cx from 'classnames'
+import React, { useCallback, useEffect } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
 
-import UIUploadQueue from 'cozy-ui-plus/dist/UploadQueue'
```
zatteo, line 4: @JF-Cozy UploadQueue is not used anymore in our projects. I think we should remove it to avoid keeping unused components. What do you think?

zatteo (approved):
Some remarks:
- Well done for the encryption removal. It is so good to simplify the code.
- Easy to say know but the encryption removal should have been a separate PR before this upload work.
- Now that you have deep dived in the upload module understanding, if you think a documentation schema or markdown has value do not hesitate!

src/modules/upload/UploadQueue.jsx
JF-Cozy: If it's not used anymore, yes we can... but why haven't we updated the UploadQueue component instead of creating a new one here? 🤔

zatteo: Maybe it's good to move components from cozy-ui-plus to drive if they are used only in drive

PR #3815, opened 2026-04-27
feat(navigation): Add flagged new sidebar components

doubleface:
Had to centralise the fetching of Nav/NavNext components to avoid to duplicate the flag and component choosing logic

image

## Summary by CodeRabbit

* **Refactor**
  * Navigation system refactored to support dynamic component resolution, enabling seamless switching between sidebar UI variants through a feature flag.

* **Chores**
  * Updated cozy-ui dependency to the latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "cozy-search": "^0.25.3",
     "cozy-sharing": "^30.3.1",
     "cozy-stack-client": "^60.23.0",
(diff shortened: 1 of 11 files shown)


zatteo:
@JF-Cozy this is the approach you don't like when changing an existing component? Because I see only two possible options for this kind of update:
- like this
- iterate with a custom build until the product + design + dev agree

PR #3828, opened 2026-05-05
Drive E2E tests (Playwright)

zatteo:
## Why

The Drive web app has no end-to-end coverage. Lint and unit tests catch obvious breakage, but nothing exercises the actual UI against a real cozy-stack, which is exactly where sharing, multi-user flows, and cross-instance behaviour can regress unnoticed.

## What's covered

Folder CRUD (create / rename / move / duplicate / delete + restore), upload (single, multiple with the queue, file viewer), navigation (Recent, Favorites, Search, Trash), and the Shared Drives feature (Alice creates a drive, invites Bob, Bob browses content Alice puts inside). A second sharing flow exercises sharing a regular folder with another instance via the same auto-accept path.

## How sharing works without an SMTP server

cozy-stack normally delivers sharing invitations by email. Disabling SMTP in the test stack would leave invitations stuck at status=pending forever, so the suite drives the alternate path that bypasses email entirely:

- Both instances share the same OrgID and OrgDomain (set on instance creation via the admin REST API, the CLI doesn't accept those flags).
- The default context has auto_accept_trusted, auto_accept_trusted_contacts, and trusted_domains: cozy.localhost.
- Each instance is pre-populated with contacts for every other org member, with the peer's cozy URL filled in. With the URL on the contact, the share modal stamps the recipient's instance on the share, and cozy-stack takes the direct stack-to-stack path instead of falling through to SMTP.
- cozy-stack listens on port 80 inside the container so the receiving side's instance lookup (which doesn't strip the port from the Host header) finds the registered domain.

The sender PUTs the share, the recipient's stack matches the sender's domain against trusted_domains via the suffix rule, and the auto-accept job fires.

## When it runs

Three triggers: nightly cron at 03:00 UTC against the default branch, manual workflow_dispatch from the Actions tab, and a PR label \`e2e\` to opt a single PR in. The job is never required for merge or deploy.

## Why it ships with a dep alignment

\`cozy-sharing@31.2.0\` declares \`cozy-client >=60.24.0\` as a peer, but the lockfile kept 60.23.1 with a nested \`cozy-stack-client@60.23.0\`. The older \`createSharedDrive\` threw synchronously on a missing field, and an internal catch swallowed it, leaving the share modal open and the test waiting 15s for nothing. Bumping both packages to ^60.24.0 fixes the contract. The fixture now also attaches per-page console errors and uncaught page errors to failed tests so the next silent-catch breakage names itself instead of looking like a timeout.

## Local development

\`yarn build && npx playwright install chromium && yarn e2e\`. Set \`E2E_SKIP_TEARDOWN=1\` to keep the stack running between iterations.

## Summary by CodeRabbit

* **Tests**
  * Full end-to-end suite added covering folder CRUD, uploads & viewer, navigation, favorites, trash, shared drives and sharing flows with authenticated cross-instance scenarios and reusable fixtures.
* **Chores**
  * CI workflow to run E2E, cache browsers, and upload reports on failure.
  * Local reproducible E2E environment via Docker composition plus global setup/teardown and persisted auth state.
  * .gitignore updated; npm scripts added to run, debug and view E2E reports.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,65 @@
+name: E2E Tests
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+  pull_request:
+    types: [labeled, synchronize, reopened]
+
+jobs:
+  e2e:
+    name: Drive E2E
+    if: >
+      github.event_name != 'pull_request' ||
+      contains(github.event.pull_request.labels.*.name, 'e2e')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build Drive app
+        run: yarn build
+
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+
+      - name: Install Playwright system dependencies
+        run: npx playwright install-deps chromium
+
+      - name: Install Playwright browsers
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+        run: npx playwright install chromium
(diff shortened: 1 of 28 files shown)


zatteo:
😍 I will try to check the PR and run these tests locally today!

e2e/helpers/stack-api.ts
zatteo, line 1: We can remove this file no?

zatteo: Done

rezk2ll (approved):

PR #3837, opened 2026-05-08
fix: skip permissions fetch on public route without redirectLink

rezk2ll:
## Summary

- `useRedirectLink` was calling `fetchOwnPermissions` for every public viewer, but the result is only used when a `redirectLink` URL param is present.
- Public share tokens cannot always read `/permissions/self`, so the unnecessary call produced thousands of 403 errors with no user-facing impact.
- Gating the fetch on `redirectLink` removes the noise and skips wasted work in the common case.

Fixes DRIVE-ZH7

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary permission checks during public-share redirects, improving performance and avoiding needless access queries when no redirect target is present.

* **Tests**
  * Added tests to ensure permission fetching is skipped when redirect parameters are absent and to prevent cross-test URL state leakage.

### DIFF
--- a/src/hooks/useRedirectLink.jsx
+++ b/src/hooks/useRedirectLink.jsx
@@ -28,7 +28,17 @@ const useRedirectLink = ({ isPublic = false } = {}) => {
   const client = useClient()
   const navigate = useNavigate()
 
-  const isFromPublicFolder = searchParams.get('fromPublicFolder') === 'true'
+  /**
+   * We read query params from two sources because their position in the url
+   * changes how they are exposed :
+   * - for /#hash?searchParam, you need useSearchParams
+   * - for /?searchParam#hash, you need location.search
+   */
+  const getParam = key => searchParams.get(key) || params.get(key)
+
+  const isFromPublicFolder = getParam('fromPublicFolder') === 'true'
+
+  const redirectLink = getParam('redirectLink')
 
   const [currentMemberInstance, setCurrentMemberInstance] = useState(undefined)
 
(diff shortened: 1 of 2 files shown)


src/hooks/useRedirectLink.jsx
```diff
@@ -28,7 +28,17 @@ const useRedirectLink = ({ isPublic = false } = {}) => {
...
   const navigate = useNavigate()
 
-  const isFromPublicFolder = searchParams.get('fromPublicFolder') === 'true'
+  /**
+   * We read query params from two sources because their position in the url
+   * changes how they are exposed :
+   * - for /#hash?searchParam, you need useSearchParams
+   * - for /?searchParam#hash, you need location.search
+   */
+  const getParam = key => searchParams.get(key) || params.get(key)
```
zatteo, line 37: Is it normal that we have these two possibilities? Maybe this need to be fixed somewhere else?

zatteo (approved):
(check comment before)

doubleface (approved):

src/hooks/useRedirectLink.jsx
rezk2ll: we have two places:
- /#/onlyoffice/123?redirectLink=… -> params live inside the hash. useSearchParams()
- /?redirectLink=…&fromPublicFolder=true#/… -> params live before the hash. location.search has them; useSearchParams() doesn't.

i agree it's a duplication we need to fix

PR #3855, opened 2026-05-25
fix: stop unhandled 403 rejections on public sharing pages

rezk2ll:
## Summary

- Public sharing pages run on a read-only sharecode token that cannot read instance-level resources such as the root directory or disk usage. The app issued those queries anyway, and because cozy-client's useQuery does not catch a failed query's promise, every 403 became an unhandled rejection hitting thousands of public visitors.
- useDisplayedFolder and useHead now skip the root directory query when on a public route.
- PushBanner is no longer mounted on public pages, where it rendered nothing but still fetched disk usage through useInstanceInfo.

Fixes DRIVE-111P
Fixes DRIVE-112T

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented 403 errors when viewing public shares by disabling problematic root queries.
  * Push banner behavior refined: it no longer appears in public shares and is hidden while instance info is still loading.

* **Tests**
  * Added tests covering public-share behavior and conditional query enabling/disabling across hooks and layout components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-drive/pull/3855?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/modules/layout/Main.jsx
+++ b/src/modules/layout/Main.jsx
@@ -10,9 +10,9 @@ import { NEXTCLOUD_MIGRATIONS_DOCTYPE } from '@/lib/doctypes'
 
 const Main = ({ children, isPublic = false }) => (
   <MainUI>
-    <PushBanner isPublic={isPublic} />
     {!isPublic && (
       <>
+        <PushBanner />
         <RealTimeQueries doctype={NEXTCLOUD_MIGRATIONS_DOCTYPE} />
         <MigrationProgressBanner />
       </>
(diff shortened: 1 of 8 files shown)


src/modules/layout/Main.jsx
```diff
@@ -10,9 +10,9 @@ import { NEXTCLOUD_MIGRATIONS_DOCTYPE } from '@/lib/doctypes'
 
 const Main = ({ children, isPublic = false }) => (
   <MainUI>
-    <PushBanner isPublic={isPublic} />
     {!isPublic && (
       <>
+        <PushBanner />
```
zatteo, line 15: Can you clean `isPublic`from PushBanner if it is not used somewhere else?

zatteo (approved):

PR #3857, opened 2026-05-26
refactor: remove carbon copy and electronic safe certification columns

rezk2ll:
## Summary

Nothing in the stack populates `metadata.carbonCopy` or `metadata.electronicSafe` anymore, so the two metadata columns and the entire `useExtraColumns` indirection they hung off were dead surface area weighing on every file-list view. This rip-out cuts the cells, headers, file-name badges, query builders, locale tooltips, and per-view `desktopExtraColumnsNames` blocks in one go so the remaining `FolderViewBody` / view files can be read without that detour.

## Summary by CodeRabbit

* **Removals**
  * Removed carbon copy and electronic safe certification icons and columns from file lists and headers.
  * Removed certification tooltip texts from multiple language translations.
  * Simplified file/folder views by removing certification metadata visualizations and related UI elements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-drive/pull/3857?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zatteo (approved):
Great code cleaning 🧹

Did you double check with product before removing this? Even we can generate less and less documents with carbonCopy/electronicSafe, there are still existing documents using them.

Crash--:
>  there are still existing documents using them.

We're not displaying these columns on Drive since few months I think. https://github.com/linagora/twake-drive/pull/3418

PR #3858, opened 2026-05-26
refactor: extract shared folder-view hooks

rezk2ll:
The Drive, Public, Sharings, and Recent views all repeated the same dozen hook calls and the same FAB display effect, which made the per-view scaffolding drown out each view's actual logic. Bundling them behind small focused hooks lets each view show its own logic instead, and keeps the shared contract in one place so adding or removing a common hook stays a one-file change.

Two real bugs in Sharings surfaced during the extraction and are fixed inline: Select All operated on the raw query result instead of the displayed filtered list, and the placeholder gate blocked on shared-drive loading even when both shared-drive flags were off.

## Summary by CodeRabbit

* **Refactor**
  * Unified folder views for a more consistent experience across Drive, Recent, Sharings and Public views
  * Standardized selection, keyboard shortcuts and action menus for predictable behavior
* **Improved Mobile UX**
  * More reliable FAB display and mobile-responsive layouts
  * Better mobile selection-bar handling
* **Performance & Reliability**
  * More efficient folder/file fetching and refresh behavior for public/shared folders

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-drive/pull/3858?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/modules/views/Public/usePublicDisplayFlags.js
+++ b/src/modules/views/Public/usePublicDisplayFlags.js
@@ -1,0 +1,33 @@
+import flag from 'cozy-flags'
+
+/**
+ * Collects the boolean flags the Public view needs to decide which optional
+ * pieces of UI to render: the new vs old breadcrumb, the sharing banner,
+ * and the "add to my cozy" FAB. Keeping them in one place keeps the view
+ * body itself focused on layout.
+ */
+export const usePublicDisplayFlags = ({
+  sharingInfos,
+  isOnSharedFolder,
+  isMobile
+}) => {
+  const showNewBreadcrumbFlag = flag(
+    'drive.breadcrumb.showCompleteBreadcrumbOnPublicPage'
+  )
+  const isOldBreadcrumb = showNewBreadcrumbFlag !== true
+
+  // The sharing shortcut has not been created (or has been but not yet synced)
+  const isShareNotAdded =
+    !sharingInfos.loading && !sharingInfos.isSharingShortcutCreated
+  // Cozy-to-Cozy preview lives on the `/preview` route; link sharing on `/public`
+  const isPreview = window.location.pathname === '/preview'
+
+  return {
+    isOldBreadcrumb,
+    isShareNotAdded,
+    isPreview,
+    isSharingBannerPluginDisplayed:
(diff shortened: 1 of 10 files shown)


src/modules/views/Public/usePublicDisplayFlags.js
```diff
@@ -0,0 +1,33 @@
...
+ * Collects the boolean flags the Public view needs to decide which optional
+ * pieces of UI to render: the new vs old breadcrumb, the sharing banner,
+ * and the "add to my cozy" FAB. Keeping them in one place keeps the view
+ * body itself focused on layout.
+ */
+export const usePublicDisplayFlags = ({
+  sharingInfos,
+  isOnSharedFolder,
+  isMobile
+}) => {
```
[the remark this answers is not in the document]
rezk2ll: the cozy stack serves the Drive bundle at distinct URL paths (/, /public, /preview) , separate rsbuild targets. The router's pathname from useLocation() is the hash route, not the URL path. Passing it would break the === '/preview' check

src/modules/views/Public/usePublicDisplayFlags.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return {
+    isOldBreadcrumb,
+    isShareNotAdded,
+    isPreview,
+    isSharingBannerPluginDisplayed:
+      isShareNotAdded || (isOnSharedFolder && !isPreview),
+    isAddToMyCozyFabDisplayed: isMobile && isPreview && isShareNotAdded
+  }
+}
```
zatteo, line 33: This is not a React hook actually, just a usual function

zatteo (approved):
It is true that we went too big on logic in these components, so happy to try something like this (btw 1 commit per 1 hook refactored would have helped the review)

src/modules/views/Public/usePublicDisplayFlags.js
rezk2ll: Well, it's because we use the flag here, otherwise i would've moved it to ts

zatteo: There is no issue to use a flag in a function

PR #3864, opened 2026-05-27
feat: file-root shared drive handling in sharings

shepilov:
- Add support for shared-drive roots that point to a file: render them in Sharings, add dedicated viewer routes, exclude them from folder pickers, and treat them as regular files on the owner instance.         
   - Extract file-root detection, path helpers, and navigation into a `rootFileNavigation` module; the `infos` and `shareSharedDrive` actions delegate to it and the share action is split into a folder-root and a  
 file-root variant.                                                                                                                                                                                                  
   - Centralise the read-only sharing flag in a `resolveShouldHideSharingActions` helper used by the viewer.                                                                                                         
   - Route file-root `.url` shortcuts to the OnlyOffice viewer (or the generic shared-drive-root-file viewer) using the target class exposed by the cozy-stack in `metadata.target.class`.                           
   - Reword the "Leave shared folder" English label to "Leave sharing" and add the missing key in Russian and Vietnamese locales.                                                                                                                                                                                                      

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Shared-drive “root file” routing and viewer flow; share links correctly resolve for root-file drives. Viewer can disable/hide sharing UI; menus accept a hide-if-shared-drive option. Share action respects ownership.

* **Bug Fixes**
  * Folder picker now displays only directory entries.

* **Tests**
  * Added comprehensive tests for routing, share-paths, root-file handling, file-class/mime logic, and related helpers.

* **Style**
  * CSS to hide viewer sharing controls when disabled.

* **Chores**
  * Updated ignore rules to exclude IDE metadata.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/linagora/twake-drive/pull/3864?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


src/modules/filelist/sharePath.js
```diff
@@ -0,0 +1,29 @@
...
+      driveId: file.driveId,
+      fileId,
+      scope: getSharedDriveRootFilePathScope(pathname)
+    })
+  }
+
+  // For a folder-root shared drive recipient, the drive root IS the folder
+  // itself: the route's `:folderId` segment and the share modal's `:fileId`
+  // segment reference the same doc. Use one resolved id for both.
+  const folderRootId = file._id ?? file.id
```
zatteo: Equivalent to `fileId` line 10?

src/modules/navigation/AppRoute.jsx
```diff
@@ -202,6 +224,7 @@ const AppRoute = () => (
           <Route path="file/:fileId/revision" element={<FileHistory />} />
           <Route path="file/:fileId/share" element={<ShareFileView />} />
           <Route path="file/:fileId/qualify" element={<QualifyFileView />} />
+          {sharedDriveRootFileRoute()}
```
zatteo: Should also be behind same `flag('drive.federated-shared-folder.enabled')` no?

src/modules/shareddrives/components/actions/shareSharedDrive.js
```diff
@@ -19,16 +24,34 @@ export const shareSharedDrive = ({ navigate, t }) => {
     label: label,
     icon,
     displayCondition: docs => {
-      return docs.length === 1 && isFromSharedDriveRecipient(docs[0])
+      const [doc] = docs
+
+      return (
+        docs.length === 1 &&
+        isFromSharedDriveRecipient(doc) &&
+        isOwner?.(doc._id ?? doc.id)
```
zatteo: Do we really need to always `doc._id ?? doc.id` ? Normally we can rely on _id.

src/modules/viewer/styles.styl
```diff
@@ -1,5 +1,25 @@
...
+    // className is applied so a rename here is caught.
+    [data-test-id='share-button']
+        display none
+
+    [data-test-id='share-by-me-button']
+        display none
+
+    [class*='coz-btn-shared']
+        display none
+
```
zatteo: This is too fragile. Is it to hide sharing button when it is a file root shared drive? We should either have a prop to pass to the Viewer that says "sharing disabled" or maybe even something inside cozy-sharing.

src/modules/shareddrives/rootFile.ts
```diff
@@ -0,0 +1,73 @@
...
+} => {
+  const name = resolveSharedDriveRootFileName({
+    fetchedName: file.name || file.attributes?.name,
+    ruleName: rootRule?.title,
+    fallbackName
+  })
+  const mime = file.mime || file.attributes?.mime || rootRule?.mime
+  const inferredMime = mime || getMimeFromName(name)
+  const fileClass =
+    getFileClassFromMimeOrName({ mime: inferredMime, name }) ||
```
zatteo: Why do we need to infer this mime?

zatteo (commented):
A first pass already, I will need to check a second time and test it locally also with the size of the PR. Also interested by @rezk2ll and @doubleface opinion.

Some context here: https://github.com/cozy/cozy-stack/pull/4763

zatteo:
Is it possible to reuse existing shared drive viewer route? Instead of creating them again?

I understand why we need to customize sharing section for example, but not sure why we need to add routes.

src/modules/shareddrives/components/actions/shareSharedDrive.js
shepilov: I thought some of the helper could return onliy ids, but it looks like no, I'll remove it

shepilov:
> Is it possible to reuse existing shared drive viewer route? Instead of creating them again?
> 
> I understand why we need to customize sharing section for example, but not sure why we need to add routes.

We have folderId in the middle of the route (`/shareddrive/:driveId/:folderId/file/:fileId`) and we query for the folder content inside, so we will pass a fileId as a folderId

What confuses you the most in a separate route?

I think it's possible but it regular files and folder also have different routes

src/modules/viewer/styles.styl
shepilov: should be in this PR https://github.com/linagora/cozy-libs/pull/3011

zatteo:
> We have folderId in the middle of the route (`/shareddrive/:driveId/:folderId/file/:fileId`) and we query for the folder content inside, so we will pass a fileId as a folderId
> 
> What confuses you the most in a separate route?

This is convincing. Agreeing with creating new routes 👍

src/modules/navigation/AppRoute.jsx
```diff
@@ -70,6 +72,49 @@ const OutletWrapper = ({ Component }) => (
...
+      <Route path="share" element={<ShareDisplayedFolderView />} />
+      <Route path="move" element={<MoveSharedDriveFilesView />} />
+      <Route path="duplicate" element={<DuplicateSharedDriveFilesView />} />
+    </Route>
+  </>
+)
+
+const hasSharedDriveRoutes = () =>
+  flag('drive.shared-drive.enabled') ||
+  flag('drive.federated-shared-folder.enabled')
```
Crash--: you should always read the flag from the rendering method, not here!

src/modules/shareddrives/components/actions/shareSharedDrive.js
```diff
@@ -19,16 +24,34 @@ export const shareSharedDrive = ({ navigate, t }) => {
...
+        )
+        return
+      }
 
       navigateToModal({
         navigate,
-        pathname: `/shareddrive/${driveId}/${folderId}`,
-        files: docs,
+        pathname: `/shareddrive/${driveId}/${fileOrFolderId}`,
+        files: [{ ...doc, id: doc.id ?? doc._id }],
```
Crash--: why do we need to split and override id? It should be already normalized.

src/modules/shareddrives/rootFile.ts
```diff
@@ -0,0 +1,69 @@
...
+const resolveSharedDriveRootFileName = ({
+  fetchedName,
+  ruleName,
+  fallbackName
+}: {
+  fetchedName?: string
+  ruleName?: string
+  fallbackName?: string
+}): string | undefined => {
+  // Shared-drive root-file fetches can lose the extension, while the sharing
```
Crash--: Why?!

src/modules/viewer/FilesViewer.jsx
```diff
@@ -180,31 +190,39 @@ const FilesViewer = ({ filesQuery, files, onClose, onChange, viewerProps }) => {
...
-                    pathname: '',
-                    files,
-                    path: 'share'
-                  })
-                }
-              />
-            )}
+            <MoreMenu
+              file={viewerFiles[viewerIndex]}
+              shouldHideSharingActions={shouldHideSharingActions}
```
Crash--, line 199: Why do we add a new props here. I'm pretty sure we have a method to generate the actions, why don't we change this method to not generate the sharingAction in this situation?

src/modules/viewer/MoreMenu.jsx
```diff
@@ -9,11 +9,13 @@ import { useBreakpoints } from 'cozy-ui/transpiled/react/providers/Breakpoints'
...
 
-const MoreMenu = ({ file }) => {
+const MoreMenu = ({ file, shouldHideSharingActions = false }) => {
   const [showMenu, setShowMenu] = useState(false)
   const { isDesktop } = useBreakpoints()
   const anchorRef = useRef()
-  const actions = useMoreMenuActions(file)
+  const actions = useMoreMenuActions(file, {
+    shouldHideIfSharedDriveRecipient: shouldHideSharingActions
+  })
```
Crash--, line 18: Confirmed. Please don't do that. Can't we make useMoreMenuActions aware by itself of the context?

src/modules/shareddrives/rootFile.ts
```diff
@@ -0,0 +1,69 @@
...
+  const name = resolveSharedDriveRootFileName({
+    fetchedName: file.name || file.attributes?.name,
+    ruleName: rootRule?.title,
+    fallbackName
+  })
+  const mime = file.mime || file.attributes?.mime || rootRule?.mime
+  const fileClass =
+    getFileClassFromMimeOrName({ mime, name }) ||
+    file.class ||
+    file.attributes?.class
```
Crash--: Why are we creating a new indirection between the class and this new fileClass?!

Crash--:
While I encourage people to work across the stack, even outside their primary area of expertise, I also believe that when we introduce a major feature, it should be developed with the support of someone whose expertise matches that domain.

This PR introduces a significant amount of complexity that, in my opinion, shouldn't be necessary. If it is necessary, then we probably have a deeper issue that needs to be addressed.

src/modules/shareddrives/components/actions/shareSharedDrive.js
shepilov: yes, it's an aritfact, I thought I removed all of them, codex thought that for some APIs we have only id and for some only _id

src/modules/shareddrives/rootFile.ts
shepilov: what exactly ? I think it was problem to open OO as in filename there was not extension, but we had it in rule

src/modules/viewer/MoreMenu.jsx
shepilov: Agreed, passing shouldHideSharingActions through MoreMenu leaks viewer-specific sharing policy into a UI-only component

src/modules/shareddrives/rootFile.ts
shepilov: because it was missing, and I don't know where to expect it, so it's just a local fallback

shepilov:
> While I encourage people to work across the stack, even outside their primary area of expertise, I also believe that when we introduce a major feature, it should be developed with the support of someone whose expertise matches that domain.
> 
> This PR introduces a significant amount of complexity that, in my opinion, shouldn't be necessary. If it is necessary, then we probably have a deeper issue that needs to be addressed.

totally agree(about support), but what do you propose?

What exact complexity?

zatteo:
Maybe @shepilov:
- we can try to discuss 15 min tomorrow morning with you, @doubleface and me about two or three key points of the PR (for example, I understand now why we need new routes which create some boilerplate but why not; but I have still some _why_ questions). I propose tomorrow morning so that we can read it again one time.
- do you think it is possible to separate at least all bug fixes and the "file root shared drive" feature ? Shared folder/drives are already quite complex on frontend and it is complicated to think about the scope of interaction of this PR.
- and then we can see about how to get this merged

What do you think?

src/modules/viewer/MoreMenu.jsx
doubleface: The change in useMoreActions is not related to goal of this PR (and we actually don't want this change that way after discussion in daily. Let's remove it

doubleface:
Now the correct file class is sent by the stack with https://github.com/cozy/cozy-stack/pull/4783

src/modules/shareddrives/components/actions/shareSharedDrive.js
```diff
@@ -8,18 +8,29 @@ import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
...
     displayCondition: docs => {
-      return docs.length === 1 && isFromSharedDriveRecipient(docs[0])
+      const [doc] = docs
+
+      return (
+        docs.length === 1 &&
+        isFromSharedDriveRecipient(doc) &&
+        !isFileRootSharedDrive(doc) &&
+        isOwner?.(doc._id)
+      )
```
zatteo, line 33: `isFromSharedDriveRecipient(doc)` and `isOwner?.(doc._id)` should never be true at same time no?

src/modules/views/SharedDrive/rootFileViewer.js
```diff
@@ -0,0 +1,45 @@
...
+      ...metadata,
+      type: 'file'
+    },
+    // Display-only breadcrumb fallback for cozy-viewer's toolbar
+    // (`showFilePath: true`). Mirrors the synthetic `/Drives/<name>` path
+    // used by `useTransformFolderListHasSharedDriveShortcuts`; it is not a
+    // canonical Cozy path and must not be used for navigation or audit URLs.
+    ...(!file.path && metadata.name ? { path: `/Drives/${metadata.name}` } : {})
+  }
+}
```
zatteo: Why do we need this function?

src/modules/shareddrives/rootFile.ts
```diff
@@ -0,0 +1,65 @@
...
+  })
+  const mime = file.mime || file.attributes?.mime || rootRule?.mime
+  const fileClass = file.class || file.attributes?.class
+
+  return {
+    ...(name ? { name } : {}),
+    ...(mime ? { mime } : {}),
+    ...(fileClass ? { class: fileClass } : {})
+  }
+}
```
zatteo: What is the purpose of this function?

src/modules/navigation/hooks/helpers.ts
```diff
@@ -45,6 +53,8 @@ export const computeFileType = (
     return 'nextcloud-trash'
   } else if (
     file.dir_id === SHARED_DRIVES_DIR_ID &&
+    !isFileRootSharedDrive(file) &&
+    !isFileRootSharedDriveShortcut(file) &&
     !isNextcloudShortcut(file)
   ) {
     return 'shared-drive'
```
zatteo, line 67: I wonder if this path is still used? Any thoughts?

src/modules/views/SharedDrive/FilesViewerSharedDriveRootFile.jsx
```diff
@@ -0,0 +1,107 @@
...
+          )
+        }
+        viewerProps={{
+          panel: {
+            sharing: {
+              // Federated/transformed sharing docs may carry a canonical `_id`
+              // that differs from the route `fileId` — `makeSharedDriveRootFileViewerFile`
+              // normalises both to the route id, so ask the underlying fetched
+              // doc for the id that the sharing context actually knows about.
+              disabled: !isOwner(fileResult.data?._id ?? fileId)
```
zatteo: An owner should never end in this "SharedDrive viewer" no?

src/modules/views/SharedDrive/FilesViewerSharedDriveRootFile.jsx
```diff
@@ -0,0 +1,107 @@
...
+  makeSharedDriveRootFileViewerFile
+} from '@/modules/views/SharedDrive/rootFileViewer'
+import { buildSharedDriveFileOrFolderByIdQuery } from '@/queries'
+
+const isViewerReady = ({ allLoaded, sharedDrivesLoaded, fileResult, file }) =>
+  [allLoaded, sharedDrivesLoaded, hasQueryBeenLoaded(fileResult), file].every(
+    Boolean
+  )
+
+const FilesViewerSharedDriveRootFile = () => {
```
zatteo, line 19: Why is the Viewer so much more complicated than the viewer for other shared drives? https://github.com/linagora/twake-drive/blob/master/src/modules/views/SharedDrive/FilesViewerSharedDrive.jsx#L34

src/modules/shareddrives/components/actions/shareSharedDrive.js
doubleface: The problem is the name of the isFromSharedDriveRecipient function. Fixed in ef95625dad5256fec78ffca1a526133a62b6b452

src/modules/views/SharedDrive/rootFileViewer.js
doubleface: removed

src/modules/shareddrives/rootFile.ts
doubleface: no purpose anymore. Removed

src/modules/navigation/hooks/helpers.ts
doubleface: It is still useful for legacy file root shortcuts created before the update to the stack which populates metadata.target.mime. But in this case it won't really work anyway. I will remove this.

doubleface: I'll adjust the detection so all file-root shortcuts are excluded from the 
generic shared-drive branch, while only stack-enriched shortcuts are routed to the dedicated file-root viewer. Legacy ones will fall back to the regular shortcut behavior.

src/modules/views/SharedDrive/FilesViewerSharedDriveRootFile.jsx
doubleface: Yes, you're right. This viewer is only for the recipient-side shared-drive root file case, so I removed the owner check and disabled the sharing panel unconditionally.

src/modules/views/SharedDrive/FilesViewerSharedDriveRootFile.jsx
doubleface: I hope you find it a lot simpler now

src/modules/views/SharedDrive/FilesViewerSharedDriveRootFile.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const files = useMemo(() => (file ? [file] : []), [file])
+
+  const filesQuery = useMemo(
+    () => ({
+      ...fileResult,
+      data: files,
+      hasMore: false
+    }),
+    [fileResult, files]
+  )
```
zatteo, line 46: I think the two memo are useless no?

zatteo (approved):
Approved with a last question 🎉

PR #3865, opened 2026-05-29
fix: reserve breadcrumb space with a loading skeleton

### LINKED ISSUE #3863: Breadcrumb "skeleton" ?
We've fixed the issue of displaying the skeleton during a public navigation here https://github.com/linagora/twake-drive/issues/3851 

But we still have issue with the breadcrumb. We do not display any skeleton, we do not "reserve" the space on the left side. It results that the icones to change the display view (list / card) 

https://github.com/user-attachments/assets/ec3291f1-3daf-41d3-9af6-b145fb597f6a

rezk2ll:
Fixes #3863

## Summary

- The breadcrumb rendered nothing while its path loaded, so it reserved no space and the list/grid view toggle jumped from left to right once it resolved. This was most visible on public share pages, where the breadcrumb only appears after the folder and its ancestors load.
- A skeleton now holds the breadcrumb's space while loading, so the toggle stays put.
- The skeleton only shows while the folder is genuinely loading: it clears once the folder is known, so an inaccessible or revoked share no longer leaves a skeleton spinning forever.

## Summary by CodeRabbit

* **New Features**
  * Hook now exposes an isLoading flag for folder data.
  * Breadcrumb skeleton component added and shown while folder data is loading.

* **Bug Fixes**
  * Breadcrumb fallback refined: show placeholder while loading, hide after settled failures to avoid permanent skeletons.
  * Breadcrumb rendering now coordinates explicitly with folder loading state.

* **Tests**
  * Tests expanded for loading, settled-failure, and fallback breadcrumb behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-drive/pull/3865?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/hooks/useDisplayedFolder.spec.jsx
+++ b/src/hooks/useDisplayedFolder.spec.jsx
@@ -63,6 +63,24 @@ describe('useDisplayedFolder', () => {
     )
   })
 
+  it('reports isLoading while the folder query has not settled', () => {
+    useQuery.mockReturnValue({ data: null, fetchStatus: 'loading' })
+    useCurrentFolderId.mockReturnValue('folder-id')
+
+    expect(useDisplayedFolder().isLoading).toBe(true)
+  })
+
+  it('reports not loading once the folder query has settled', () => {
+    useQuery.mockReturnValue({
+      data: null,
+      fetchStatus: 'failed',
+      lastError: { status: 403 }
+    })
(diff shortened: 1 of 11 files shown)


zatteo (approved):
Good UX improvment

PR #3867, opened 2026-06-01
fix: stop public breadcrumb infinite loop on shared subfolders

### LINKED ISSUE #3866: Sharing Public link
With the latest changes, I've an issue on Twake Drive on public link. It seems that we have a loop somewhere. 

https://github.com/user-attachments/assets/26f34166-207e-41a0-b2c9-c3771fc545ec

rezk2ll:
## Summary

Fixes #3866. Opening a subfolder of a public link froze the tab.

- The breadcrumb effect depended on the whole `displayedFolder` object. `getBreadcrumbPath` writes parent folders into the cozy-client store, so `useDisplayedFolder` returns a new reference, re-triggering the effect in a fetch loop that starves the event loop.
- Now it depends on the folder identity (`id`, `name`, `dir_id`) instead, matching `useBreadcrumbPath`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Breadcrumbs no longer trigger unnecessary path refetches when the same folder is re-displayed with a different internal object, improving performance and stability; loading indicator behavior updated to rely on stable folder identity.

* **Tests**
  * Added test coverage to confirm breadcrumb path is fetched once and not retriggered on object-reference changes.

### DIFF
--- a/src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
+++ b/src/modules/views/Folder/OldFolderViewBreadcrumb.jsx
@@ -17,19 +17,30 @@ const FolderViewBreadcrumb = ({ sharedDocumentId, getBreadcrumbPath }) => {
   const [path, setPath] = useState(null)
   const [hasError, setHasError] = useState(false)
 
+  // Depend on the folder identity, not the object: getBreadcrumbPath writes to
+  // the cozy-client store, which churns the displayedFolder reference and would
+  // re-run this effect forever.
+  const folderId = displayedFolder?.id
+  const folderName = displayedFolder?.name
+  const folderDirId = displayedFolder?.dir_id
+
   useEffect(() => {
     let isMounted = true
 
     // eslint-disable-next-line react-hooks/set-state-in-effect
     setPath(null)
     setHasError(false)
-    if (!displayedFolder || !sharedDocumentId) return
+    if (!folderId || !sharedDocumentId) return
 
     const asyncGetPaths = async () => {
       try {
         const paths = await getBreadcrumbPath({
           client,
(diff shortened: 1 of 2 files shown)


Crash--:
How do you explain this regression? Because I didn't noticed it before, so it's a recent regression

rezk2ll:
> How do you explain this regression? Because I didn't noticed it before, so it's a recent regression

It was introduced this morning by this PR: https://github.com/linagora/twake-drive/pull/3865 (it adds a skeleton loader to breadcrumbs).

The parent-walking getBreadcrumbPath (which fetches each ancestor folder and writes them to the cozy-client store) has been in the public view since December (d0a3415e). The effect has always depended on the whole displayedFolder object. So the ingredients were there.

But until this morning, OldFolderViewBreadcrumb received displayedFolder as a prop from the parent view. The async path-fetch didn't force the parent to re-render, so that prop stayed referentially stable and the effect ran once.

aee776f52 (the breadcrumb-skeleton commit, dated the same morning the issue was filed) changed the breadcrumb to read the folder from its own useDisplayedFolder() subscription instead of the prop. That coupled the store writes that getBreadcrumbPath makes directly back into this component: each ancestor fetch pushes a fresh displayedFolder object into the effect's dependency, which re-runs the effect which fetches again. Infinite loop.

Why nobody hit it before / it's easy to miss: it only triggers when there's a parent chain to walk, i.e. inside a subfolder of a public share. At the share root displayedFolder.id === sharedDocumentId, so getBreadcrumbPath returns immediately with no store write and no loop. That's exactly the path most quick tests take

Crash-- (approved):

PR #3869, opened 2026-06-03
feat: Add shared drive in nav

zatteo:
Behind flag('drive.shared-drive.enabled'). It is a considered as a first version to have access to shared drive. 

image

## Release Notes

* **New Features**
  * Desktop navigation now includes shared drive support. Organization-shared drives are displayed in the sidebar navigation menu, allowing direct access to shared organizational resources. The shared drives section appears above the external drives list and is available when the feature is enabled. Clicking on a shared drive navigates directly to it.


doubleface (approved):

zatteo:
For the record, realtime is not working. I think we will need to rely on io.cozy.sharing realtime.

PR #3871, opened 2026-06-03
fix: Improve shared drive management for useSharedDrives

zatteo:
useSharedDrives was relying on a io.cozy.files hack to refetch the query but this was working only for the owner. Now we use realtime on io.cozy.sharing that work for both.

Follows https://github.com/linagora/twake-drive/pull/3869

## Summary by CodeRabbit

* **Refactor**
  * Enhanced shared drives data synchronization with real-time updates to ensure immediate reflection of changes instead of periodic refresh cycles.


doubleface (approved):
A lot better! Thank you

PR #3872, opened 2026-06-03
feat: Handle 2 versions of OO

Crash--:
It seems that the upgrade to OnlyOffice 9.3 on the server requires this stuff. (from 8.0) 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OnlyOffice document editor configuration handling to prioritize explicit editor configuration when present and correctly determine edit mode with a safer fallback, ensuring more reliable and flexible editor behavior.


zatteo (approved):
A comment about why/what version which attribute supports?

PR #3874, opened 2026-06-04
feat: add upload conflict resolution modal

shepilov:
## Summary

Adds a batch upload conflict resolution flow for duplicate file uploads.

When one or several uploaded files already exist in the target folder, Drive now shows a confirmation modal once per upload batch and lets the user choose whether to replace existing files, keep both versions, or cancel the conflicted uploads.

## Changes

- Add upload conflict resolution helpers for replace, keep-both, cancel, and generated rename handling.
- Add an upload conflict modal and wire it into the upload flow.
- Keep uploading non-conflicting files while conflicted files wait for the user choice.
- Apply the selected conflict strategy to all duplicate files in the upload batch.
- Update the upload queue to display files waiting for conflict resolution.

## Summary by CodeRabbit

* **New Features**
  * Interactive upload conflict dialog (Replace / Keep both / Cancel) with per‑batch apply-all behavior.
  * Queue shows a "conflict" status, treats conflicted items as in‑progress, and defers completion until conflicts are resolved.
  * Automatic rename flow when keeping both to avoid name collisions.

* **Localization**
  * Added conflict-related translations for EN, DE, ES, FR, RU, and VI.

* **Tests**
  * Added UI and service tests covering conflict dialogs, retry/rename flows, and edge cases.


src/modules/upload/conflictResolution.js
rezk2ll, line 1: a JSDoc on all functions in this file would be nice

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+      if (!isConflictError(error)) throw error
+      warnOnManyRenameAttempts(FILE_TYPE, attempt)
+      name = generateUploadConflictName(name)
+    }
+  }
+
+  throw buildRenameLimitError(FILE_TYPE)
+}
+
+export const resolveFileConflict = async (
```
rezk2ll: 7 arguments are a lot, let's use an object and also let's get rid of this `noop` thing ( by using optional chaining on the calls ).

onNameResolved?.(name)

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+
+const CONFLICT_ERROR = 409
+const DIRECTORY_TYPE = 'directory'
+const FILE_TYPE = 'file'
+export const MAX_UPLOAD_CONFLICT_RENAME_ATTEMPTS = 1000
+const UPLOAD_CONFLICT_RENAME_LOG_STEP = 100
+const noop = () => {}
+const CONFLICT_STRATEGY_REPLACE = 'replace'
+const CONFLICT_STRATEGY_KEEP_BOTH = 'keep-both'
+const CONFLICT_STRATEGY_CANCEL = 'cancel'
```
rezk2ll: those are not needed, just

```js
export const uploadConflictStrategies = {
  REPLACE: 'replace',
  KEEP_BOTH: 'keep-both',
  CANCEL: 'cancel'
}
```

also they belong on a shared const / util, like upload/constants

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+  // Cancel skips this conflicted file without stopping the rest of the queue.
+  if (selectedConflictStrategy === uploadConflictStrategies.CANCEL) {
+    return { isCancel: true }
+  }
+
+  // No modal choice yet: store the row as conflict and continue uploading.
+  return { isConflict: true }
+}
+
+export const createFolderWithRenameOnFileCollision = async (
```
rezk2ll: this is the same algorithm as `uploadWithRenamedFile`, we could extract the loop

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+      dirID,
+      uploadOptions,
+      driveId,
+      onNameResolved
+    )
+  }
+
+  // Cancel skips this conflicted file without stopping the rest of the queue.
+  if (selectedConflictStrategy === uploadConflictStrategies.CANCEL) {
+    return { isCancel: true }
```
rezk2ll: feels like go code 😅, i know it's not mandatory but can we have this file in typescript? since it's new ( conflictResolution.ts )

we can have a predicatable conflict outcome type etc,

```ts
type ConflictOutcome =
  | { resule: 'uploaded'; data: FileDoc; isUpdate: boolean; finalName: string }
  | { result: 'conflict' }
  | { result: 'cancelled' }
```

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+    name: originalName,
+    type
+  })
+  // Continue an existing "name-1" conflict suffix when there is one.
+  const match = filename.match(/^(.*)-(\d+)$/)
+
+  if (!match) return `${filename}-1${extension}`
+
+  const baseName = match[1]
+  const suffix = parseInt(match[2], 10) + 1
```
rezk2ll: zero padding is dropped here, for example:

parseInt("09", 10) is 9, +1 becomes 10, so file-09.txt becomes file-10.txt.

another issue about this if the matched number passes the Number.MAX_SAFE_INTEGER ( 9007199254740991 ) the arithmetic is wrong:
like file-1000000000000000000000.txt becomes file-1e+21.txt

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+
+const filesCollection = (client, driveId) =>
+  client.collection(DOCTYPE_FILES, { driveId })
+
+export const generateUploadConflictName = (originalName, type = FILE_TYPE) => {
+  const { filename, extension } = splitFilename({
+    name: originalName,
+    type
+  })
+  // Continue an existing "name-1" conflict suffix when there is one.
```
rezk2ll: this is a problem since it makes the function re-parses its own output every iteration.

The root cause is conflating two things: "what is the base name" and "what is the next free index.", the regex is trying to find both and fails the first one.

i propose to change the behaviour like this:
file-2024.txt -> file-2024 (1).txt

this way you filter by "name (n)" and not "name-n"

src/modules/upload/index.js
```diff
@@ -350,65 +509,74 @@ const getExistingDirectory = async (client, dirID, name, driveId) => {
...
+      { onUploadProgress },
+      driveId
+    )
     return { data, isUpdate: false }
   } catch (err) {
     if (err.status !== CONFLICT_ERROR) throw err
-    const path = await resolveFullpath(client, dirID, file.name, driveId)
-    const data = await overwriteFile(client, file, path, options, driveId)
-    return { data, isUpdate: true }
+    // Only 409 errors enter the conflict service; other upload errors stay unchanged.
```
rezk2ll, line 549: so it's reactive by default, as we discussed, this can lead to a problem for people who drop large files, and the conflict can be detected mid-session.

and an alternative better approach is to not be reactive, but scan the first level of files/folders ( we should already have them in cache so we won't make a new request ) and we can show the popup there.

let's say i have 10 files and folders in the root directory and i drag&dropped a folder, i will just scan the first layer and if i have that same folder name i will show the popup and then the choice will be applied to all nested children and potential other conflicting files or folders ( yes it comes with an edge case of drag&dropping a lot of files flat at the same level, but we can live with )

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+    `Could not find an available upload ${type} name after ${MAX_UPLOAD_CONFLICT_RENAME_ATTEMPTS} attempts`
+  )
+  error.code = 'UPLOAD_CONFLICT_RENAME_LIMIT_REACHED'
+  return error
+}
+
+const filesCollection = (client, driveId) =>
+  client.collection(DOCTYPE_FILES, { driveId })
+
+export const generateUploadConflictName = (originalName, type = FILE_TYPE) => {
```
rezk2ll: We have a serious problem with this, especially the -<digits> suffix, it's causing wrong names being generated and an O(N^2) collision cascade:

let's imaging i drag&dropped 40 files, numbered sequentially: file-00.txt -> file-39.txt, then i reupload the same files again, and in the modal, I decided to keep both files:
- file-00.txt -> file-01.txt ( not file-00-1.txt ), and it walks to other existing items: file-09.txt -> file-10.txt

and here we enter a 409 loop, so testing locally, for the case of 40 files, we do 130 sequential 409s.

i know the original spec wants file-1.txt but we can't have the same suffix for conflict detection and renaming, maybe let's consider: file (1).ext / file (2).ext or file-copy.txt

src/modules/upload/UploadConflictDialog.jsx
```diff
@@ -0,0 +1,78 @@
...
+          />
+        </>
+      }
+    />
+  )
+}
+
+UploadConflictDialog.propTypes = {
+  onCancel: PropTypes.func.isRequired,
+  onClose: PropTypes.func.isRequired,
```
rezk2ll, line 74: is it required? since it was not added in the call ( showModal )

src/modules/upload/index.js
```diff
@@ -958,6 +1146,11 @@ export const onQueueEmpty = callback => (dispatch, getState) => {
     .map(item => item.uploadedItem)
     .filter(item => item && item._id)
 
+  // Do not show an empty success alert after canceling conflicts.
+  const hasOnlyCanceledItems =
+    queue.length > 0 && queue.every(i => i.status === CANCEL)
+  if (hasOnlyCanceledItems) return
```
rezk2ll, line 1153: but we should give feedback to the user if all files he drag&dropped are conflicting and he chose to cancel.

i propose to have a modal showing something like "no files has been uploaded or something )

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+export const uploadWithRenamedFile = async (
+  client,
+  file,
+  dirID,
+  options = {},
+  driveId,
+  onNameResolved = () => {}
```
rezk2ll: i agree let's use objects instead of 6 args

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+  const match = filename.match(/^(.*)-(\d+)$/)
+
+  if (!match) return `${filename}-1${extension}`
+
+  const baseName = match[1]
+  const suffix = parseInt(match[2], 10) + 1
+  return `${baseName}-${suffix}${extension}`
+}
+
+const statByName = async (client, dirID, name, driveId) => {
```
rezk2ll: we should catch HTTP errors here

rezk2ll (commented):
so in general:
- Are we sure about reactive-on-409?
- the rename suffix should be reworked
- and in tests we don't have the cases of statByName rejecting or failed HTTP call

src/modules/upload/conflictResolution.js
shepilov: file (1).ext - it's also the same suffix, no ?

I was thinkg that we need just to look into the curent folder and to have max(suffix) + 1, but I didn't want to do it in the first iteration of this PR as far as it's already big

src/modules/upload/UploadConflictDialog.jsx
shepilov: docs said that yes, but also not sure

src/modules/upload/conflictResolution.js
```diff
@@ -0,0 +1,261 @@
...
+
+      if (existing.type !== FILE_TYPE) throw error
+      warnOnManyRenameAttempts(DIRECTORY_TYPE, attempt)
+      // folder-vs-file keeps incrementing until a folder name is available.
+      nextName = generateUploadConflictName(nextName, DIRECTORY_TYPE)
+    }
+  }
+
+  throw buildRenameLimitError(DIRECTORY_TYPE)
+}
```
Crash--: FYI : We already have some kind of tested strategies here https://github.com/linagora/cozy-client/blob/364d348ff17d2e0683b7bfec9f03ab5a680f9c29/packages/cozy-client/src/models/file.js#L638 

I think we should reuse stuff from there.

src/modules/upload/index.js
```diff
@@ -76,13 +87,30 @@ const getStatus = (state, action) => {
...
+  return `${relativePath.slice(0, lastSlash + 1)}${name}`
+}
+
+// Browser folder uploads can prefix relative paths with ./; it is not a real Drive folder.
+const normalizeRelativeUploadPath = path => {
+  let cleanPath = path || ''
+  while (cleanPath.startsWith('/')) cleanPath = cleanPath.slice(1)
+  while (cleanPath.startsWith('./')) cleanPath = cleanPath.slice(2)
+  return cleanPath
+}
```
Crash--, line 115: Why do we need that now?

src/modules/upload/index.js
```diff
@@ -212,34 +359,55 @@ export const uploadProgress = (fileId, event, date) => ({
...
   pendingItem,
   client,
   dirID,
   driveId,
-  { dispatch, safeCallback }
+  { dispatch, safeCallback, getConflictStrategy, onUploadConflict }
 ) => {
-  const { file, fileId, folderId } = pendingItem
+  const { file, fileId, folderId, uploadBatchId } = pendingItem
+  const { relativePath } = pendingItem
```
Crash--, line 376: I don't get the point of this relativePath (yet)

src/modules/upload/index.js
```diff
@@ -212,34 +359,55 @@ export const uploadProgress = (fileId, event, date) => ({
...
-      driveId
+      {
+        onUploadProgress,
+        driveId,
+        getConflictStrategy,
+        onNameResolved
+      }
     )
+    if (result.isConflict) {
+      // Keep this row waiting for the modal while processNextFile moves on.
```
Crash--, line 395: Does it means that we continue the upload if there is a conflict (I mean the files that do not have conflict?)

src/modules/upload/index.js
shepilov: becaus browser returns such kind of uris

src/modules/upload/conflictResolution.js
shepilov: yes, and it's intended behavior 09 > 10, the only case then it can heppen that this file was created by user and we had all the numbers before

src/modules/upload/conflictResolution.js
shepilov: But it's a change in the suffix generation behaviour, it's just a change of the remplate, where do we put it

src/modules/upload/index.js
Crash--: But why this is needed now and not before?

src/modules/upload/index.js
rezk2ll: i forgot we have the queue that has the feedback so, this comment can be ignored

src/modules/upload/index.js
shepilov: is needed for `File.path` values coming from browser folder uploads / react-dropzone.
ther can be prefixed with / or ./, for example ./Photos/2026/img.jpg.

src/modules/upload/UploadConflictDialog.jsx
shepilov: prop is required by the component at runtime, but not by the caller(

src/modules/upload/conflictResolution.js
shepilov: I've looked into it, so it will do statByPath for every file upload, and not reacting on 409, also to use this function we will need to modify it with callbacks, onStrategy update, etc...
Right now it/s only used for OnlyOffice template, if we can refactor it, we need to move all the conflict resolution to client and merge is this function

@rezk2ll @Crash--  do you have long term vision about how cozy client should look like ? will it be relevant for other consumers?

but also if we want to make `cozy-client` better it would be great not to force client to right code like
```
onst queryString = next.split('?')[1]
  if (!queryString) return null
  return new URLSearchParams(queryString).get('page[cursor]')
```
for the next page

Crash--: What do you mean by "long term vision about how cozy client should look like"? If we're talking for this drag & drop conflict resolution, I don't think all the stuff should be handled in Cozy-client because it's only needed for Drive and it's specific to Drive. 

But at least, we should reuse the generateFileNameOnConflict (or thing like that). 

> Right now it/s only used for OnlyOffice template, if we can refactor it, we need to move all the conflict resolution to client and merge is this function

Now Excalidraw too, but it's also used by other Cozy applications (konnectors / apps). 


> onst queryString = next.split('?')[1]
>  if (!queryString) return null
> return new URLSearchParams(queryString).get('page[cursor]')

Where do you have that? Because yes, I don't think consumer should write this kind of code.

rezk2ll (approved):
🍾 works great now, let's rebase and fix the commits and merge.

for the conflict resolution issue we talked about i filed an issue in cozy-client https://github.com/linagora/cozy-client/issues/1686 so that everyone benefits from the fix

src/modules/upload/index.js
shepilov: Because before we didn't change the path in the widget when filename changed, we just displayed what we recieved from browser

src/modules/upload/index.js
shepilov: yep, we continue upload, and files which has conflicts waiting for decision

src/modules/upload/index.js
shepilov: mmm, we use it as a part of the display label for every upload item

PR #3884, opened 2026-06-08
feat(pdf): add an embedpdf editor with annotations and comments

Crash--:
## Summary

- Add a PDF editor route powered by EmbedPDF, opened with an Edit button shown in the PDF viewer when the user has write access.
- Load the file binary, let the user annotate and comment, and save back to the drive with autosave plus an explicit Save button.
- Commit pending annotations before saving so an annotation still being edited is not lost on reload.
- Open at a stable fit-page zoom and keep the viewport bounded so the document does not auto-zoom.
- Sign annotations with the current user: instance public_name in private, share token then URL username in public.
- Add a back button to the editor toolbar.
- Hide the Edit button and block the editor route for read-only access, in both private and public shares.
- Expose the editor from public shares too, behind the same write-access checks.
- Toggle the viewer Edit button through cozy-viewer's isPdfEditorEnabled prop, keeping the flag in the app.
- Gate the whole feature behind the drive.pdf-editor.enabled flag.
- Bump cozy-viewer to 28.1.0, which ships the Edit button support.

Built on the cozy-viewer Edit button (linagora/cozy-libs#3020, released in cozy-viewer 28.1.0).

## Also 

Started to create this "editors concept" on Drive to be able to "easily" integrate new editors. OO is not yet migrated to it, but at some point, we'll have to. 

The design is not great. But I needed to provide something quickly. I'm thinking about replacing our existing PDFViewer by the new one, but I need confirmation from the product team. 

## Summary by CodeRabbit

* **New Features**
  * In-app PDF editor/viewer: open, annotate, and save PDFs.

* **Improvements**
  * Unified editor chrome and title/header across editors.
  * Viewer and routing flows integrated with PDF editor; shared loading spinner used by editors.
  * Autosave, save-on-hide/unmount and manual save behavior standardized.

* **Bug Fixes**
  * Corrected root-folder parent-path handling.

* **Localization**
  * Added PDF save strings (English, French).

* **Chores**
  * Updated viewer dependency and dev build settings.


src/modules/navigation/AppRoute.jsx
```diff
@@ -111,6 +112,14 @@ const sharedDriveRoutes = () => (
   </>
 )
 
+// Group the flag-gated editor routes so the AppRoute switch stays small.
```
zatteo: Useless comment. Generally speaking lot of comments with AI in PRs recently. Don't know what to think about it.

src/modules/views/editor/helpers.js
```diff
@@ -0,0 +1,54 @@
...
+ * when the file belongs to one.
+ *
+ * @param {object} client - cozy-client
+ * @param {object} file - The io.cozy.files document being edited
+ * @param {Blob|string|ArrayBuffer} content - The content to persist
+ * @param {string} contentType - The MIME type to store
+ */
+export const updateFileBinary = (client, file, content, contentType) =>
+  client
+    .collection('io.cozy.files', file.driveId ? { driveId: file.driveId } : {})
```
zatteo, line 49: nit: `{ driveId: file?.driveId }` should be equivalent ?

src/modules/views/Excalidraw/useSceneSync.js
```diff
@@ -79,17 +72,11 @@ const useAutosave = (save, intervalMs) => {
...
-      if (document.visibilityState === 'hidden') flush()
-    }
-    document.addEventListener('visibilitychange', handleHide)
-    return () => {
-      clearInterval(intervalId)
-      document.removeEventListener('visibilitychange', handleHide)
-      flush()
-    }
+    return () => clearInterval(intervalId)
   }, [flush, intervalMs])
```
zatteo, line 80: nit: Maybe we can create a `useSaveRegurlaly` also? Especially if the same is also used for PDFs.

zatteo:
I got strange result when opening a file inside the editor, like replacing the current file, saving it, but then when I open it with the editor again it comes back? Do you reproduce and understand this?

zatteo:
Is it not possible to use it as a readonly viewer as well? I saw that we can disable some tools https://www.embedpdf.com/docs/vue/viewer/customizing-ui

Crash--:
> I got strange result when opening a file inside the editor, like replacing the current file, saving it, but then when I open it with the editor again it comes back? Do you reproduce and understand this?

I didn't have this issue let's chat about it. 

>  Is it not possible to use it as a readonly viewer as well? I saw that we can disable some tools https://www.embedpdf.com/docs/vue/viewer/customizing-ui

We can. And this is what we should do IMO. But it'll take more time to implement since it'll change the first user facing view of a PDF (it'll bring us password support and a lot of useful stuff too). I need to check with the Product team. This is why I decided to only use this new "viewer" when you edit the PDF.

Crash--:
@zatteo your issue should be fixed by https://github.com/linagora/twake-drive/pull/3884/changes/750744617eb459701871ed33fa8ef0e3dbe453a8 ;)

zatteo (approved):

src/modules/views/Pdf/usePdfSave.js
```diff
@@ -0,0 +1,80 @@
...
+  const save = useCallback(async () => {
+    const registry = registryRef.current
+    if (isReadOnly || !registry || !isDirty.current) return
+    isDirty.current = false
+    setIsSaving(true)
+    try {
+      const bytes = await exportPdf(registry)
+      if (bytes) {
+        const blob = new Blob([bytes], { type: PDF_MIME })
+        await updateFileBinary(client, file, blob, PDF_MIME)
```
rezk2ll: i think we need to wait for edits one by one, since here two requests can modify the file at different order than intended.

1 - request A ( slow one ) starts updating the binary but hasn't finished yet
2 - request B ( fast ? ) starts updating the binary
3 - request B finished
4 - request A finished and the binary is reset to the state of A ( and B change is discarded )

src/modules/views/editor/useSaveOnHideAndUnmount.js
```diff
@@ -0,0 +1,21 @@
...
+ * Flushes pending editor changes when the tab is hidden and on unmount, so an
+ * edit is not lost when the user switches tab or leaves. Pass a flush that
+ * no-ops in read-only mode.
+ *
+ * @param {() => void | Promise<void>} flush
+ */
+export const useSaveOnHideAndUnmount = flush => {
+  useEffect(() => {
+    const handleHide = () => {
+      if (document.visibilityState === 'hidden') flush()
```
rezk2ll, line 13: if we close the tab during the flush, will the document be saved, i mean does updateFileBinary has keepalive: true ?

src/modules/views/Pdf/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+      return <Loader />
+    }
+    if (!hasWriteAccess(fileId, driveId)) {
+      return <Navigate to="/" replace />
+    }
+  }
+
+  // In public the share code already determines read-only (passed in as
+  // isReadOnly). In private, the user always has write access at this point.
+  const readOnly = isPublic ? isReadOnly : !hasWriteAccess(fileId, driveId)
```
rezk2ll: isReadOnly is dead code / prop here no ? since getPublicPdfRoutes will redirect the user to / per the comment ( i quote routes.tsx )

```ts
      // The editor is for editing only: a read-only share must not reach it,
      // even by typing the URL. Send those visitors back to the public viewer.
      element={
        isReadOnly ? (
          <Navigate to="/" replace />
````

we can just remove the isReadOnly altogether

File: src/modules/views/Pdf/usePdfSave.js [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+  if (annotation.commit) await annotation.commit().toPromise()
+}
+
+// Pulls the current document bytes out of the EmbedPDF engine (annotations and
+// edits baked in), ready to persist.
+const exportPdf = async registry => {
+  const engine = registry.getEngine()
+  const document = getActiveDocument(registry)
+  if (!engine || !document) return null
+  await finalizePendingAnnotations(registry)
```
rezk2ll, line 31: I think it will trigger a call to `scheduleAutosave` that will save the file again after 3 seconds.

so save -> exportPdf -> finalizePendingAnnotations ->  the listener we have on the Onready

 annotation?.onAnnotationEvent?.(() => scheduleAutosave())

i quote:

```ts
 const onReady = useCallback(
    registry => {
      registryRef.current = registry
      if (isReadOnly) return
      const annotation = registry.getPlugin('annotation')?.provides()
      annotation?.onAnnotationEvent?.(() => scheduleAutosave())
    },
    [isReadOnly, scheduleAutosave]
```

so a save will trigger another save after 3 seconds because scheduleAutosave marks the document as dirty before the bounce .

i suggest maybe to diff check or suppress the autosave while there is a save on progress

src/modules/views/Pdf/Title.jsx
```diff
@@ -0,0 +1,45 @@
...
+    <EditorTitle
+      file={file}
+      flushRef={flushRef}
+      icon={PdfIcon}
+      dataTestId="pdf-title"
+      isPublic={isPublic}
+      isReadOnly={isReadOnly}
+    >
+      {!isReadOnly && (
+        <Button
```
rezk2ll, line 35: we should give feedback when the document is being saved or even better wait for the save to finish before triggering a new one

src/modules/views/Pdf/usePdfSave.js
```diff
@@ -0,0 +1,80 @@
...
+ * debounce saves on annotation changes. Also saves on tab hide and on unmount.
+ *
+ * @returns {{ save: () => Promise<void>, flush: () => Promise<void>, scheduleAutosave: () => void, isSaving: boolean }}
+ */
+export const usePdfSave = ({ client, file, isReadOnly, registryRef }) => {
+  const [isSaving, setIsSaving] = useState(false)
+  const isDirty = useRef(false)
+  const debounceRef = useRef(null)
+
+  const save = useCallback(async () => {
```
rezk2ll: i think save memoizes the whole file object and if the file name changes ( elsewhere ) the save will always the same old name registered during the init of the hook.

maybe we can depend on file.name

src/modules/navigation/AppRoute.jsx
Crash--: Removed in abb5b21ac.

src/modules/views/editor/helpers.js
Crash--: Equivalent, yes: `FileCollection` builds its prefix as `options.driveId ? sharedDriveApiPrefix(driveId) : '/files'`, so `{ driveId: undefined }` resolves to `/files` just like `{}`. I kept the ternary to match the read paths (`useSceneSync`, `usePdfBlobUrl`), which use the same form. Happy to switch all three to the short version in a follow-up if you prefer.

src/modules/views/Excalidraw/useSceneSync.js
Crash--: This overlaps with the Excalidraw `useAutosave` (debounce + flush on hide/unmount). Extracting a shared `useSaveRegularly` makes sense, but it reshapes both editors, so I'd rather do it as a separate follow-up than in this PR.

src/modules/views/Pdf/usePdfSave.js
Crash--: Done in 63d249e1c. Saves are chained on a promise queue now, so A and B can't interleave: B runs only after A resolves, and `runSave` rechecks `isDirty` so it's a no-op when nothing changed in between.

src/modules/views/editor/useSaveOnHideAndUnmount.js
Crash--: `updateFileBinary` calls cozy-client's `updateFile`, which doesn't set `keepalive`. It wouldn't help here anyway: the fetch spec caps a keepalive request body at 64 KB, far below a typical PDF. So a save on tab close is best-effort: `visibilitychange: hidden` fires before unload and usually lets the in-flight request finish, but a hard close can still drop it.

src/modules/views/Pdf/index.jsx
Crash--: Right, the public route already redirects read-only shares, so the view only mounts with write access and the prop was always false. Removed in f92148431: `isReadOnly` is gone from the public PDF view and a public mount is always editable.

src/modules/views/Pdf/usePdfSave.js
Crash--: Confirmed. `finalizePendingAnnotations` deselects and commits, which fires annotation events that rescheduled autosave in a loop. Fixed in 63d249e1c: `scheduleAutosave` ignores events while a save is running (`isSavingRef`), so a save's own finalize no longer schedules another save.

src/modules/views/Pdf/Title.jsx
Crash--: Both addressed. The Save button shows a busy state and is disabled while saving (4a36ce57c), and saves are serialized on a queue (63d249e1c) so a new trigger waits for the in-flight one instead of overlapping. Autosave stays silent by design; the explicit Save is what surfaces progress.

src/modules/views/Pdf/usePdfSave.js
Crash--: The save callback depends on the whole `file`, so the explicit Save (via `flushRef`) already used the current name. The gap was autosave: its listener is attached once at `onReady` and had captured the initial `scheduleAutosave`. Fixed in 63d249e1c by reading `scheduleAutosave` through a ref, so autosave persists under the fresh name after a rename too.

rezk2ll (approved):

PR #3885, opened 2026-06-09
fix(shareddrives): stop file list blinking on realtime refresh

rezk2ll:
## Problem

- In a shared drive, a member with write access who uploads files sees the list blink.
- The recipient view refetches the folder on every realtime file event.
- Each refetch cleared the data and showed a loading state first, so the list was replaced by the skeleton and then redrawn.
- Uploads fire these events in bursts, so the list blanks out repeatedly.

## Solution

Refresh in the background and keep the current list on screen, only swapping once the new data arrives. The initial load still shows the skeleton, and a failed refresh keeps the existing list (unless nothing has loaded yet).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved realtime synchronization to preserve currently displayed folder contents when background refreshes are in progress
  * Fixed issue where failed realtime updates would overwrite existing data; existing content now remains on-screen until successful refresh

### DIFF
--- a/src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
+++ b/src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
@@ -59,15 +59,25 @@ const useSharedDriveFolder = ({
   )
 
   useEffect(() => {
-    const fetchSharedDriveFolder = async (pagesToLoad = 1): Promise<void> => {
+    const fetchSharedDriveFolder = async (
+      pagesToLoad = 1,
+      { isRefresh = false }: { isRefresh?: boolean } = {}
+    ): Promise<void> => {
       fetchGeneration.current += 1
       const currentGeneration = fetchGeneration.current
 
-      setSharedDriveResult({ data: undefined, included: undefined })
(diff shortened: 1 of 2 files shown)


doubleface (approved):

zatteo (approved):
Awesome

src/modules/shareddrives/hooks/useSharedDriveFolder.tsx
```diff
@@ -59,15 +59,25 @@ const useSharedDriveFolder = ({
...
+      // arrives. Wiping to a loading state here would replace the list with the
+      // loading skeleton on every realtime event and make the view blink while
+      // files are being added.
+      if (!isRefresh) {
+        setSharedDriveResult({ data: undefined, included: undefined })
+        setFetchStatus('loading')
+        nextCursorRef.current = null
+        setNextCursor(null)
+        loadedPagesCount.current = 0
+      }
```
Crash--, line 80: it's a pity that Drive has to handle that. It should be done by Cozy client at least

PR #3886, opened 2026-06-09
fix(shareddrives): hide edit actions for read-only recipients

rezk2ll:
## Context
On the recipient side of a shared drive or shared folder, a viewer with read-only access still saw the Create and Upload actions in the sidebar, the right-click add menu, and the empty-folder placeholder. Write access was inferred from the local folder document, but shared-drive folders are proxied from the owner and have no local document, so the check defaulted to writable and offered edit actions to users who cannot write.

## Solution
Resolve write access from the current route (folder id and drive id) through a new useCurrentFolderWriteAccess hook, and gate the sidebar, the right-click add menu, and the empty-folder upload button on it. The folder is treated as writable until the sharing context has loaded, so owned folders keep their actions during the initial load. The empty-folder placeholder now shows a read-only message instead of an upload prompt.

## Summary by CodeRabbit

* **Bug Fixes**
  * Upload and add options are no longer displayed in read-only folders.
  * Right-click context menu now respects folder permissions and won't show add operations when the folder is read-only.

* **Tests**
  * Added test coverage for folder write-access permissions checks.

### DIFF
--- a/src/components/Error/Empty.jsx
+++ b/src/components/Error/Empty.jsx
@@ -60,10 +60,9 @@ const EmptyCanvas = ({
       text={
         <>
           {text}
-          {showUploadLayout && (
+          {showUploadLayout && canUpload !== false && (
             <span className="u-db u-mt-1">
               <UploadButton
-                disabled={!canUpload}
                 componentsProps={{
                   button: { variant: 'secondary' }
                 }}
--- a/src/components/RightClick/RightClickAddMenu.jsx
+++ b/src/components/RightClick/RightClickAddMenu.jsx
@@ -1,16 +1,16 @@
 import React, { useContext } from 'react'
 import { useLocation } from 'react-router-dom'
 
-import { useSharingContext } from 'cozy-sharing'
 import { useBreakpoints } from 'cozy-ui/transpiled/react/providers/Breakpoints'
 
 import { useRightClick } from '@/components/RightClick/RightClickProvider'
 import { useDisplayedFolder } from '@/hooks'
+import useCurrentFolderWriteAccess from '@/hooks/useCurrentFolderWriteAccess'
 import AddMenuProvider, {
   AddMenuContext
 } from '@/modules/drive/AddMenu/AddMenuProvider'
 
-const AddMenu = ({ children, ...props }) => {
+const AddMenu = ({ children, isFolderReadOnly, ...props }) => {
   const { isDesktop } = useBreakpoints()
   const { onOpen } = useRightClick()
   const { handleToggle, handleOfflineClick, isOffline } =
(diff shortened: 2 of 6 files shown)


src/components/Error/Empty.jsx
```diff
@@ -21,6 +21,7 @@ import CreateSharedDriveButton from '@/modules/views/SharedDrive/CreateSharedDri
 const EmptyCanvas = ({
   type,
   canUpload,
+  isReadOnly,
```
Crash--: canUpload & isReadOnly => What is the diff?

rezk2ll: canUpload: should the Upload button render?
isReadOnly: should the empty state say "This folder is shared with you in read-only mode" ( this is a new string for shared drives only )

we can't use one prop here

Crash--: are you sure about canUpload === should the upload button render? I though it was to know if we have right access or not

rezk2ll: Write access is usually the input to canUpload, but the prop itself means "is uploading allowed in this view"
we have different callers:
- SharingsFolderView: passes canUpload={hasWrite}
- SharedDriveFolderView: passes canUpload={canWriteToCurrentFolder}
- PublicFolderView: passes canUpload={hasWritePermissions}
- TrashFolderView passes canUpload={false} ( where it's different from others, since we have full rights on the trash but we say no don't render it ).
- FolderPickerContentLoader: hardcodes canUpload={false} for every empty folder in the Move dialog.
- SharedDriveFolderView: passes canUpload={false} in the toolbar  ( https://github.com/linagora/twake-drive/blob/master/src/modules/views/SharedDrive/SharedDriveFolderView.jsx#L182 )

Crash--: Ok but I still think we can’t live with 2 props. As you can see lot of places can upload is used as write access. 

You can not write in the trash view. You can not write during the loading phase and so on. 

I don’t know about the shared folder view

rezk2ll: ok fair on the new prop, i removed it.

now we have users who can't upload anything seeing "Select files on your computer or drag them here" ( but it does nothing ). i will address it in a new PR

src/components/Error/Empty.jsx
```diff
@@ -60,10 +60,9 @@ const EmptyCanvas = ({
       text={
         <>
           {text}
-          {showUploadLayout && (
+          {showUploadLayout && canUpload !== false && (
```
zatteo, line 63: Why not `showUploadLayout && canUpload` ?

zatteo (approved):

src/components/Error/Empty.jsx
rezk2ll: because canUpload by default is undefined in the case of showSharedDriveLayout, it's passing <EmptyCanvas type="sharing" />  ( without canUpload ), so if we do canUpload only !undefined will be false and we will make a regression elsewhere

PR #3889, opened 2026-06-10
feat: add realtime collaboration to the excalidraw editor

Crash--:
## Summary

- Add realtime Excalidraw collaboration over the cozy-stack realtime hub on the io.cozy.files channel: scene sync (last-write-wins reconciliation), heartbeat presence, a SCENE_INIT handshake for new arrivals, and live cursors.
- Reach collaboration in both private and public contexts, since both an owner and a public-share guest can subscribe and POST on the shared file. Read-only viewers receive without broadcasting.
- Name remote cursors from the share token or URL recipient, with a localized guest fallback for anonymous links.
- Extract a shared useEditorAuthor hook and move the PDF editor onto it.
- Gate the feature behind the drive.excalidraw.collab flag.

## Summary by CodeRabbit

* **New Features**
  * Real-time Excalidraw collaboration: simultaneous editing, presence, cursor sharing, scene sync/init, targeted init for newcomers, and public/read-only handling

* **Localization**
  * Guest label added for Excalidraw (English: "Guest", French: "Invité")

* **Tests**
  * Extensive test coverage for collaboration protocol, presence/heartbeat, routing, scene relay, sender/router hooks, and author resolution

* **Chores**
  * Added collaboration feature flag and unified editor author resolution across viewers


src/modules/views/Excalidraw/useCollabRouter.js
```diff
@@ -0,0 +1,108 @@
...
+  )
+
+  return useCallback(
+    doc => {
+      const message = unwrapMessage(doc)
+      // Drop auto-echo: the cozy hub relays our own POST back to us.
+      if (!message || message.senderId === sessionIdRef.current) return
+      // A leaving peer should not be resurrected by its own goodbye.
+      if (message.type !== MESSAGE_TYPES.PRESENCE_BYE) touchPeer(message)
+      routeByType(message, {
```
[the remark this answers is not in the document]
Crash--: Valid for the shared `io.cozy.files` `notified` channel, which is a blind relay, so a foreign or malformed payload could otherwise spawn a phantom collaborator. Guarded in 9fa6f085a.

I didn't take the inline version though:
- Extracted a pure `isCollabMessage(message)` helper in `collabProtocol.js` (requires a string `senderId` and a known `MESSAGE_TYPES`), with unit tests, instead of inlining the checks in the handler. Keeps the router's cyclomatic complexity flat and makes the guard testable on its own.
- Hoisted the known-types `Set` to module scope so it isn't rebuilt on every message.

The handler now early-returns on `!isCollabMessage(message)` before `touchPeer`.

src/modules/views/Excalidraw/useCollabSender.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+export const useCollabSender = ({
+  active,
+  isReadOnly,
+  realtime,
+  fileId,
+  sessionIdRef,
+  usernameRef
+}) => {
+  const canSend = active && !isReadOnly
```
rezk2ll, line 29: Here a read-only viewer never sends PRESENCE_HELLO or PRESENCE_PING so people with a "view only" link see a frozen picture ( tested locally ), they see whatever was saved at that moment, then nothing else.

one fix is to drive the editor's outbound scene/handshake off the subscription itself rather than off received presence, so a silent viewer still gets SCENE_INIT and SCENE_UPDATE

src/modules/views/Excalidraw/useSceneRelay.js
```diff
@@ -0,0 +1,146 @@
...
+
+  const collectNewFiles = useCallback(
+    () => pickNewFiles(apiRef.current, knownFileIdsRef.current),
+    [apiRef]
+  )
+
+  const broadcastScene = useCallback(
+    elements => {
+      if (isReadOnly) return
+      const version = getSceneVersion(elements)
```
rezk2ll, line 101: I think we have an issue here: two people drawing together is fine, but the moment anyone deletes a shape, the two browsers get stuck bouncing the same change back and forth, hundreds of messages a second, and it never stops until a tab closes.

getSceneVersion is elements.reduce((acc, el) => acc + el.version, 0), so a sum of per-element versions.

- applyRemoteScene watermarks from getSceneVersion(api.getSceneElements()), which excludes deleted elements.
- broadcastScene checks getSceneVersion(elements) where elements is the onChange argument, which includes deleted (tombstoned) elements.

Once the scene holds any deleted element, the onChange sum is permanently higher than the watermark taken from the non-deleted set, so version <= watermark is false and each peer rebroadcasts every update it just applied, forever.

Every real session deletes something eventually, so any two-editor room degrades into a self-sustaining flood against /realtime. 

( The unit tests hide this by mocking getSceneVersion as elements.length )

src/modules/views/Excalidraw/useCollabSender.js
```diff
@@ -0,0 +1,50 @@
...
+        senderId: sessionIdRef.current,
+        username: usernameRef.current,
+        type
+      }
+      if (payload !== undefined) message.payload = payload
+      if (targetId !== undefined) message.targetId = targetId
+      // RealtimePlugin.sendNotification returns undefined (not a promise), so
+      // guard the call itself instead of chaining .catch on its result.
+      try {
+        realtime.sendNotification(COLLAB_DOCTYPE, fileId, message)
```
rezk2ll, line 43: this is async no? i think this way we swallow the promise and never catch failed POSTs

src/modules/views/Excalidraw/useSceneRelay.js
```diff
@@ -0,0 +1,146 @@
...
+        MESSAGE_TYPES.SCENE_INIT,
+        { elements: api.getSceneElements(), files: api.getFiles() },
+        targetId
+      )
+    },
+    [apiRef, sendMessage]
+  )
+
+  const reset = useCallback(() => {
+    lastBroadcastVersionRef.current = -1
```
rezk2ll, line 143: if the effect calling this reruns the next onChange will not boradcasted to the other people listenting since we have the non-broadcast baseline via a watermark < 0 sentinel.

src/modules/views/Excalidraw/useSceneRelay.js
Crash--: Confirmed and fixed in ef0be0691. `getSceneVersion` sums per-element versions; `applyRemoteScene` watermarked from `getSceneElements()` (tombstones dropped) while `broadcastScene` reads the `onChange` array (tombstones kept), so after any delete the watermark stayed permanently below the onChange version and every applied update rebroadcast forever. The watermark now uses `getSceneElementsIncludingDeleted()` so both sides share the same basis. Added a regression test that applies a remote update with a deleted element and asserts the echo isn't rebroadcast (it fails against the old code).

src/modules/views/Excalidraw/useCollabSender.js
Crash--: Confirmed — a read-only viewer has no POST on /realtime, so it never announces itself and is invisible to presence; `broadcastScene` gated on detected peers, so its canvas froze at the saved scene. Fixed in e31c66473 by dropping the peer gate: editors now broadcast every genuine edit and any subscriber receives it, visible or not.

Two notes:
- Trade-off: a lone editor now POSTs updates nobody receives. Acceptable for correctness, but flagging it.
- SCENE_INIT still can't reach a silent viewer, but each SCENE_UPDATE carries the full element set, so the viewer catches up on the first edit after joining (and it already loads the saved scene on mount).

I left cursor broadcasting gated on detected peers for now (volatile and noisier when alone); happy to lift that too if you want viewers to see live cursors.

src/modules/views/Excalidraw/useSceneRelay.js
Crash--: Right. `reset()` ran from the subscription's teardown, so any rerun of that effect re-armed the `watermark < 0` sentinel and the next edit was taken for the loaded-scene baseline and dropped. Fixed in 3c2ebf287: `reset()` now runs from an effect keyed on the file id, so it re-baselines only on an actual document change, not on every re-subscribe.

src/modules/views/Excalidraw/useCollabSender.js
Crash--: It looks async but isn't from our side. `CozyRealtime.sendNotification` is `async`, but the `RealtimePlugin` wrapper we actually call doesn't return it:

```js
sendNotification(...args) {
  this.checkRealtimeInstance()
  this.realtime.sendNotification(...args) // no return
}
```

So our call resolves to `undefined` — there's no promise to `await` or `.catch`, and a failed POST's rejection is swallowed inside cozy-realtime, not here. The `try/catch` only guards the synchronous path (e.g. `checkRealtimeInstance` throwing).

We could bypass the plugin and POST through the stack client to surface failures, but that drops the realtime abstraction. Given collab messages are fire-and-forget (a dropped frame self-heals on the next SCENE_UPDATE, which carries the full scene, and on the periodic PING), I left it as-is. Happy to revisit if you'd rather surface POST errors.

src/modules/views/Excalidraw/useCollabSender.js
Crash--: Lifted the cursor gate too in f6e76f3e7, so read-only viewers now see live cursors. `CURSOR_THROTTLE_MS` (100 ms) still bounds the rate; same trade-off as the scene fix (a lone editor streams cursor moves that may have no receiver).

rezk2ll (approved):

PR #3898, opened 2026-06-11
fix(sharings): show only received shares in the sharings section

rezk2ll:
## Context

The Sharings section listed everything the user was part of, including the files, folders, and drives the user created and shared out to others. Those items are already reachable from My Drive and the shared-drive views, so the section duplicated them instead of acting as an inbox of what was shared with the user.

## Solution

Filter the Sharings list down to shares the user received (where the user is not the owner). Owned files, folders, and shared drives are dropped; received shared-drive links stay, since they are the only way a recipient reaches the drive. The folder picker, which also reuses the shared-drive transform, is untouched and still shows owned drives so the user can move and copy into them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ownership tracking for shared drives and items in the Sharings view; items are now correctly filtered based on ownership status.

* **Tests**
  * Added comprehensive test coverage for shared-drive ownership handling and sharing filter logic.

### DIFF
--- a/src/modules/views/Sharings/useFilteredSharings.jsx
+++ b/src/modules/views/Sharings/useFilteredSharings.jsx
@@ -1,6 +1,7 @@
 import { useMemo } from 'react'
 
 import flag from 'cozy-flags'
+import { useSharingContext } from 'cozy-sharing'
 
 import { SHARED_DRIVES_DIR_ID } from '@/constants/config'
 import { useTransformFolderListHasSharedDriveShortcuts } from '@/hooks/useTransformFolderListHasSharedDriveShortcuts'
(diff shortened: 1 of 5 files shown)


src/modules/views/Sharings/useFilteredSharings.jsx
```diff
@@ -11,18 +12,35 @@ const buildBaseShape = (result, hasIds) => ({
...
   withoutSharedDrives,
   transformedSharedDrives,
   nonSharedDriveList
 }) => {
   if (withoutSharedDrives) {
     return (
-      result.data?.filter(item => item.dir_id !== SHARED_DRIVES_DIR_ID) || []
+      result.data?.filter(
+        item =>
+          item.dir_id !== SHARED_DRIVES_DIR_ID && isReceivedFile(isOwner)(item)
```
shepilov, line 36: org_drive != true, not SHARED_DRIVES_DIR_ID

shepilov: I'm not sure about the same name in the model, but in sharing json should be like this

rezk2ll: org_drive is the narrower "created on an org instance", so it wouldn't reliably catch non-org shared drives.

(dir_id !== SHARED_DRIVES_DIR_ID) is pre-existing and only runs on the flags-off branch over io.cozy.files docs, which don't carry org_drive. and we want to also exclude the files shared by me via links

shepilov: There are no non-org shared drives, and the "SHARED_DRIVES_DIR_ID".

Shared drive = folder in org instance

shepilov: what does it means "flags-off branch" ?

shepilov: if there are no already existing items with org_drive = true then we don't need this check at all.

rezk2ll: > what does it means "flags-off branch" ?

ah i mean the shared drives enabled flag. in this PR we are hiding everything even shares created without shared drives

shepilov: then probably we shouldn't do it at all, we are changing UX too much for the existing users, then won't be able to filter files that were shared by them at all

before at least have this ability in this list, and as far we don't have filter by people in file list we won't have it at all

shepilov: @Bobpodvalnyi what do you think? It looks like we shouldn't do it for all the sharings

rezk2ll:
on draft until PO decide

shepilov:
#3975 should be addressed here by splitting into tabs

PR #3902, opened 2026-06-11
feat: Move Download button in public page more menu

doubleface:
https://app.notion.com/p/linagora/Public-page-Move-Download-folder-button-to-context-menu-37c62718bad18001a180c8636c309b15

## Summary by CodeRabbit

* **Refactor**
  * Download functionality has been reorganized and moved from a dedicated button to the actions menu, making it consistently accessible across all screen sizes.


src/modules/public/PublicToolbarByLink.jsx
```diff
@@ -37,7 +36,7 @@ const PublicToolbarByLink = ({
 
   const actions = makeActions(
     [
-      isMobile && download,
```
rezk2ll, line 40: why are we removing the mobile check

doubleface: Because the download action was already displayed on mobile but not on desktop. Now we want to display it in the menu everywhere

rezk2ll (approved):

PR #3908, opened 2026-06-11
chore(build): share assets across the three web targets

rezk2ll:
## Problem
The main, public and intents targets are three separate rspack compilations, so React, the cozy libraries and the editors are re-bundled into each target's own static tree. The registry tarball pays for every shared module up to three times: it sits at 18MB against the 20MB limit, and every new editor or heavy dependency multiplies its cost by the number of targets that use it. That pressure is what previously forced the PDF editor out of the public build.

## Solution
The three targets become entries of a single compilation. Shared modules are emitted once under build/static, exposed through a new public /static route so share pages can load them without auth, and each target keeps its own index.html (the / route now points at /main). The react-pdf alias moves to the webpack5 entry so the pdf.js worker lands under static/ instead of the build root, where only the private route could serve it. The tarball drops from 18MB to 10.8MB and stays flat when a dependency is added to more targets.

Since the editor chunks are now shipped once regardless, the PDF editor is restored on shared links. It is also loaded lazily, keeping it out of the initial payload of both the public page and the logged-in app (about 1MB less each).

## Worth knowing
All JS becomes publicly fetchable through /static, including chunks only used by the logged-in app. The code is AGPL and already published, so nothing sensitive changes hands.

## Summary by CodeRabbit

* **New Features**
  * PDF editing capability is now available in shared and public file views.

* **Performance**
  * PDF viewer components now load on-demand, improving initial application performance.

### DIFF
--- a/.bundlemonrc
+++ b/.bundlemonrc
@@ -8,10 +8,19 @@
       "path": "static/js/<chunkId>.<hash>.js"
     },
     {
+      "path": "static/js/async/<chunkId>.<hash>.js"
+    },
+    {
       "path": "static/js/cozy.<hash>.js"
     },
     {
       "path": "static/js/main.<hash>.js"
+    },
+    {
+      "path": "static/js/public.<hash>.js"
+    },
+    {
+      "path": "static/js/intents.<hash>.js"
     },
     {
       "path": "static/js/lib-react.<hash>.js"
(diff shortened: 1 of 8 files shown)


rezk2ll:
@Crash--  -69.39%

zatteo (approved):
Interesting!

1. Did you properly tested local dev (`yarn start` and `yarn watch`) with HMR, Excalidraw viewer & PDF viewer, public links, etc ?

2. I propose to try 1 or 2 weeks with Drive and then move to rsbuild-config-cozy-app what is common with other app.

rezk2ll:
> Interesting!
> 
> 1. Did you properly tested local dev (`yarn start` and `yarn watch`) with HMR, Excalidraw viewer & PDF viewer, public links, etc ?
> 2. I propose to try 1 or 2 weeks with Drive and then move to rsbuild-config-cozy-app what is common with other app.

yes HMR is intact and the viewers also work fine in public and normal views ( while sharing assets of the editos ).

i'm not merging this PR yet until the new release of drive.

and yes the idea is to test it on drive and then bring it to the other apps via the shared build config

rezk2ll:
E2E tests pass, locally also. i'll merge it and see how it goes @Crash-- @zatteo

PR #3920, opened 2026-06-15
fix(selection): keep shift selection item visible

doubleface:
Share DOM scroll helpers between rectangular selection and keyboard range selection. Use the actual scroll container to reveal the current item when extending a selection with Shift+Arrow.

[Capture vidéo du 2026-06-15 17-28-11.webm](https://github.com/user-attachments/assets/951dc7df-7d01-4d82-aa08-298fd26a1313)

https://app.notion.com/p/linagora/Add-scroll-when-use-Shift-arrow-to-select-37c62718bad18053823dec07e36bf911

## Summary by CodeRabbit

* **New Features**
  * Shift selection now automatically scrolls the newly selected item into view when using Shift+Click or Shift+Arrow keyboard navigation (based on the active scroll container).
  * Rectangular selection auto-scrolls smoothly as you drag near the edges.

* **Tests**
  * Expanded shift-selection test coverage to verify proper event wiring and that auto-scrolling occurs only when the target item is outside the visible container.

### DIFF
--- a/src/hooks/useShiftSelection/index.tsx
+++ b/src/hooks/useShiftSelection/index.tsx
@@ -20,6 +20,7 @@ import {
 
 import { isEditableTarget } from '@/hooks/helpers'
 import { useSelectionContext } from '@/modules/selection/SelectionProvider'
+import { scrollElementIntoViewInContainer } from '@/modules/selection/scrollHelpers'
 import { SelectedItems } from '@/modules/selection/types'
 
 type ViewType = 'list' | 'grid'
@@ -27,6 +28,7 @@ type ViewType = 'list' | 'grid'
 interface UseShiftSelectionParams {
   items: IOCozyFile[]
   viewType?: ViewType
+  scrollElement?: HTMLElement | null
 }
 
 interface UseShiftSelectionReturn {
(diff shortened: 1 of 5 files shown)


src/hooks/useShiftSelection/index.tsx
```diff
@@ -41,22 +43,48 @@ interface UseShiftSelectionReturn {
...
 
+  /**
+   * Scrolls the lastInteractedItem into view inside the scroll container.
+   * This mirrors rectangular selection: use the actual scrollable element
+   * and scroll only by the amount needed to reveal the current item.
+   */
+  const scrollToItem = useCallback(
+    (id: string) => {
+      const container = scrollElement || ref.current
+      const lookupRoot = ref.current || container
```
zatteo: This is equivalent to `const lookupRoot = ref.current || scrollElement || ref.current`, normal?

zatteo (approved):

src/hooks/useShiftSelection/index.tsx
doubleface: right, now I use only one variable

PR #3932, opened 2026-06-18
feat: dispatch document editors through a registry

Crash--:
## Summary

- Add an editor registry as the single source of truth for document dispatch: each entry holds its slug, kind, enablement, file and shortcut matchers, route builder, and add-menu create gating.
- Make `computeFileType`, `computeApp` and `computePath` read from the registry instead of per-editor branches.
- Fold Docs and Grist (bridge documents opened through `/bridge/<app>`) into the registry next to the in-app editors, tagged with a `kind` field so only in-app editors mount routes and are used as in-app redirect targets, while bridge documents keep their `/bridge` link.
- Generate the add-menu "create a document" entries from the registry: each editor declares `create.isAvailable`, so the registry owns which editors offer a create entry and when it shows; the toolbar keeps only the rendering.
- Route an Excalidraw file shared as a shared-drive root (seen by the recipient as a `.url` shortcut) to its editor, and redirect editor documents from the shared-drive root-file viewer so direct links and reloads reach the editor too.

Note on the repeated slug keys between `registry.ts` and the slug maps in `routes.tsx` (`ROUTE_FACTORIES`) and `CreateDocumentItems.jsx` (`CREATE_ITEM_RENDERERS`): `registry.ts` holds only data and pure dispatch logic with no JSX, because it is imported by `computeFileType`/`computePath` and runs on every file row of every listing. The slug maps keep the React components out of that hot path. Pulling `ExcalidrawView`, `OnlyOfficeView` or the create buttons into the registry would bundle them into every listing and make the pure dispatch model depend on the view layer. So the keys repeat on purpose: the registry is the model, and those maps are thin view-layer adapters keyed back to it.

## Summary by CodeRabbit

* **Refactor**
  * Centralized editor/bridge routing and document-type detection through a single registry, enabling consistent in-app redirects (including shared-drive root files) and cleaner, context-aware route generation.
  * Updated the document “Create” menu to be driven by the registry, showing only supported editor options based on device, upload capability, and feature-flag/permission rules.

* **Tests**
  * Added/expanded coverage for editor selection, shortcut target resolution, route generation, redirect behavior, and create-menu filtering across multiple document types and flag states.


src/modules/views/editor/registry.js
```diff
@@ -0,0 +1,96 @@
...
+ * Order matters: editors are matched top to bottom, so the more specific editor
+ * wins. Excalidraw precedes OnlyOffice because a `.excalidraw` file is a
+ * text-class document that OnlyOffice would otherwise claim.
+ *
+ * @type {EditorDescriptor[]}
+ */
+export const EDITORS = [
+  {
+    slug: 'excalidraw',
+    isEnabled: ({ isExcalidrawEnabled }) => Boolean(isExcalidrawEnabled),
```
zatteo: Should the flag be added directly here?

src/modules/views/editor/routes.jsx
```diff
@@ -0,0 +1,43 @@
...
+ */
+const EDITOR_ROUTE_GROUPS = [
+  { slug: 'onlyoffice', flag: null, getRoutes: getOnlyOfficeRoutes },
+  {
+    slug: 'excalidraw',
+    flag: 'drive.excalidraw.enabled',
+    getRoutes: getExcalidrawRoutes
+  },
+  { slug: 'pdf', flag: 'drive.pdf-editor.enabled', getRoutes: getPdfRoutes }
+]
```
zatteo: slug and flag to derive from EDITORS to centralize these values?

src/modules/views/editor/registry.js
zatteo: What is the difference between isEnabled and the flag?

zatteo:
Very very promising refactoring I can't wait to see a new editor added in 10 lines 🎉

src/modules/views/editor/registry.ts
```diff
@@ -0,0 +1,210 @@
...
+   * `/bridge/<app>` (Grist, Docs); they have no in-app route and must never be
+   * the target of an in-app redirect. */
+  kind: 'editor' | 'bridge'
+  /** The app that opens the document, as returned by `computeApp`. Absent
+   * (defaulting to `'drive'`) for in-app editors. */
+  app?: string
+  /** cozy-flag that gates mounting the editor's routes; `null` mounts them
+   * unconditionally (the editor gates access internally). Irrelevant for
+   * `'bridge'` documents (they mount no route). */
+  flag: string | null
```
zatteo, line 75: "flag" name is a little misleading: we end with `flag: null` for editor that "have a flag" literally because it is only to gate editor route (which I understand must be different that isEnabled)

zatteo (approved):

Crash--:
@shepilov I'm pretty confident with this change, can I merge it? Or do you want to try & understand what has been done?

PR #3933, opened 2026-06-18
fix(sharings): deduplicate transient shared-drive entries

doubleface:
When a folder is shared from Sharings, the view can briefly receive that same file twice: the real Drive entry and a transformed shared-drive entry. The transformed entry uses the shared-drives magic directory, which is rendered as /sharings in the file path column.

Deduplicate by document id after merging the query result with transformed shared-drive entries, and prefer the real Drive entry when both representations are present.

https://app.notion.com/p/linagora/Folders-doubling-in-sherings-38262718bad180a4ac3de913943d4707?source=copy_link

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved duplicate sharing entries that appeared during sharing updates. The system now automatically removes redundant entries and prioritizes the primary Drive representation when duplicates occur.

* **Tests**
  * Added comprehensive test coverage for sharing deduplication functionality.

### DIFF
--- a/src/modules/views/Sharings/useFilteredSharings.jsx
+++ b/src/modules/views/Sharings/useFilteredSharings.jsx
@@ -11,18 +11,65 @@ const buildBaseShape = (result, hasIds) => ({
   lastFetch: hasIds ? result.lastFetch : Date.now()
 })
 
+const getDocId = item => item._id ?? item.id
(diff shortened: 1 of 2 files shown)


zatteo:
Owner or recipient or both have the issue?

doubleface:
> Owner or recipient or both have the issue?

@zatteo Only the owner

rezk2ll (approved):

PR #3937, opened 2026-06-19
revert: share assets across the three web targets (#3908)

rezk2ll:
## Context
PR #3908 merged a single-compilation build that shares assets across the main, public and intents targets. We need to ship 1.102.1 without that change, so it is reverted here on the release branch.

## Solution
Reverts #3908 in full, restoring the three separate build environments, the per-target Excalidraw font copies, the `/static`-less manifest, the eager PDF view, and the public build without the PDF editor. The later version bump to 1.102.1 is preserved.

### DIFF
--- a/.bundlemonrc
+++ b/.bundlemonrc
@@ -8,19 +8,10 @@
       "path": "static/js/<chunkId>.<hash>.js"
     },
     {
-      "path": "static/js/async/<chunkId>.<hash>.js"
-    },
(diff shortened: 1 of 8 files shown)


doubleface (approved):

Crash--:
it's needed because of the flagship app not respecting the manifest.wepabb with an harcoded path to ./index.html

PR #3941, opened 2026-06-21
fix(pdf): serve embedpdf runtime assets from same origin

Crash--:
## Summary

- Copy the Pdfium wasm, the default stamp library and the glyph-fallback fonts into `build/static` and point the PDF editor at those same-origin paths, so they no longer load from jsDelivr (blocked by the CSP).
- Disable the EmbedPDF Google Fonts UI and signature webfonts, falling back to the system font stack.
- Ship only Latin Regular/Bold fallback fonts and drop the duplicate Pdfium wasm rspack emits, keeping the build tarball under the 20 MB registry limit.

## Summary by CodeRabbit

* **New Features**
  * Updated the PDF editor to self-host PDF runtime assets (including PDFium WASM and stamp manifests) from the app origin for smoother, more reliable loading.
  * Added configurable font fallback/snippet font behavior and improved multilingual font support (notably Arabic and Hebrew).
* **Tests**
  * Added automated Jest coverage to verify generated asset URLs and runtime configuration.
* **Chores**
  * Updated project dependencies to include required EmbedPDF runtime packages and assets.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,12 @@
     "worker-loader": "2.0.0"
   },
   "dependencies": {
+    "@embedpdf/default-stamps": "0.0.1",
+    "@embedpdf/fonts-arabic": "1.0.0",
+    "@embedpdf/fonts-hebrew": "1.0.0",
+    "@embedpdf/fonts-latin": "1.0.0",
+    "@embedpdf/models": "2.14.3",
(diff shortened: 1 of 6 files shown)


Crash--:
You can test with this branch, it works git://github.com/cozy/cozy-drive.git#build-qv-210626

rezk2ll (approved):

PR #3945, opened 2026-06-22
fix(sharings): keep the path label from clipping under a short name

rezk2ll:
## Problem
In the sharings list, the `/sharings` link shown under a shared drive's name gets clipped when the name is shorter than the link text (a one-letter share showed `/s` instead of `/sharings`). cozy-ui's `Filename` stacks the name and the path label in a column that shrinks to the name's width, so a path wider than a short name is cut off.

## Solution
Let that column fill the name cell when a path is shown, so the path label uses the available width instead of being capped to the name. Long paths still truncate at the cell edge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed file list display to prevent path information from being clipped when displaying filenames alongside paths. The filename cell now properly utilizes available space to ensure all information remains visible without truncation.

### DIFF
--- a/src/modules/filelist/virtualized/cells/FileName.jsx
+++ b/src/modules/filelist/virtualized/cells/FileName.jsx
@@ -78,7 +78,10 @@ const FileName = ({
   }
 
   return (
-    <span title={infected ? t('antivirus.infectedFile') : title}>
+    <span
+      className={hidePath ? undefined : styles['fil-file-name-with-path']}
+      title={infected ? t('antivirus.infectedFile') : title}
+    >
       <Filename
         icon={
           <FileThumbnailComponent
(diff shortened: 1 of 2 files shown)


src/modules/filelist/virtualized/cells/FileName.jsx
```diff
@@ -78,7 +78,10 @@ const FileName = ({
   }
 
   return (
-    <span title={infected ? t('antivirus.infectedFile') : title}>
+    <span
+      className={hidePath ? undefined : styles['fil-file-name-with-path']}
```
doubleface, line 82: you don't need stylus file for that: you have `.u-w-100` class

doubleface (approved):

PR #3947, opened 2026-06-22
feat(recents): show progress bar while shared-drive files load

rezk2ll:
## Context
In the recents view, shared-drive and federated-share files are fetched from the stack through the dataproxy and arrive a few seconds after the local files (longer for accounts with many drives). Until then the list looks settled, then silently grows with no indication anything was pending.

## Solution
Show a thin progress bar under the header while that background fetch is in flight, once a list is already on screen. The empty initial load keeps its existing skeleton.

## Summary by CodeRabbit

* **New Features**
  * Added visual loading indicator in the Recent view that displays when additional shared or federated items are being fetched in the background.

* **Tests**
  * Added test coverage for the loading indicator behavior across different loading states.


Crash--:
let's try it

doubleface (approved):

PR #3958, opened 2026-06-24
fix(sharing): Redirect after leaving shared document

doubleface:
Will need the next version of cozy-sharing with https://github.com/linagora/cozy-libs/pull/3055

Had to add @linagora/twake-icons
Also using node 24 in .nvmrc

https://app.notion.com/p/linagora/Recipient-When-remove-myself-from-folder-still-see-folder-content-but-can-t-interact-with-content-38862718bad180f0844fd6ff0d13e24d?source=copy_link

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the redirect after successfully revoking access for shared files and shared folders, including drive-based sharing.
  * Ensured the “close share” flow consistently returns to the expected view/tab.
* **Tests**
  * Added/expanded Jest coverage to verify post-revoke and close-navigation behavior in the share modals.
* **Chores**
  * Updated the supported Node.js version.
  * Refreshed dependencies (including adding icon support and updating sharing-related packages).


package.json
```diff
@@ -102,7 +103,7 @@
     "cozy-pouch-link": "^60.19.0",
     "cozy-realtime": "^5.8.0",
     "cozy-search": "^0.25.3",
-    "cozy-sharing": "^33.6.1",
+    "cozy-sharing": "34.0.0",
```
JF-Cozy: we can use `^` version for cozy/twake packages

src/modules/views/Modal/revokeSuccessRedirect.js
```diff
@@ -0,0 +1,11 @@
+import { SHARINGS_VIEW_ROUTE, SHARING_TAB_DRIVES } from '@/constants/config'
+
+function getRevokeSuccessRedirectPath({ document, driveId } = {}) {
```
JF-Cozy: as we don't fetch anything, `makeRevokeSuccessRedirectPath` is more appropriate

package.json
JF-Cozy, line 1: should be merged in the previous commit

src/modules/views/Modal/revokeSuccessRedirect.js
JF-Cozy, line 1: It seams to be more complicated for me to use a func in that case. It doesn't really mutualize code because the arg are different from one use to an other... so it should be easier to directly use the string in the navigate without using any func

JF-Cozy (approved):

package.json
doubleface: It was just an intermediate commit to handle issues on this version before going to the next version

PR #3960, opened 2026-06-25
feat(recent): make recent files react to live store updates

doubleface:
This only fixes real time update on recents view for local files at the moment

The proxy path now subscribes to the cozy-client store via useQuery so that trashed or renamed documents disappear from the recent view without navigating away: getDocumentFromState merges fresh store data onto stale proxy results on every re-render.

- Replace one-shot fetchQueryAndGetFromState with reactive useQuery
- Merge proxy data with store via getDocumentFromState + spread
- Filter trashed documents from both proxy and store sources
- Unify duplicated fallback return blocks
- Add e2e tests for live trash and rename on /#/recent

https://app.notion.com/p/linagora/recents-cant-delete-file-or-rename-files-38462718bad1806ca5afe2546e29b717

## Summary by CodeRabbit

* **New Features**
  * Recent files can now be managed directly from the Recent view without reloading the page (trash and rename supported).

* **Bug Fixes**
  * Recent entries now stay in sync with the latest document state, including removing trashed items and updating renamed items instantly.

* **Tests**
  * Added end-to-end coverage for trashing and renaming from the Recent view, plus improvements to Favorites cleanup and Search locator formatting.
  * Updated unit tests for the Recent-files hook to use query-based mocking and validate fallback/error behavior.

### DIFF
--- a/e2e/tests/navigation.spec.ts
+++ b/e2e/tests/navigation.spec.ts
@@ -30,6 +30,61 @@ test.describe('Navigation surfaces', () => {
     }
   })
 
+  test('Recent: a trashed file disappears without page reload', async ({
+    alicePage,
+    aliceDrive
+  }) => {
+    await alicePage.goto(ALICE_ROOT)
+    const sidebar = new SidebarPage(alicePage)
+    const file = path.join(path.dirname(FIXTURE), `recent-trash-${stamp()}.txt`)
+    await copyFile(FIXTURE, file)
+    try {
+      const fileName = path.basename(file)
+      await aliceDrive.uploadFiles(file)
+      await aliceDrive.row(fileName).waitVisible()
+
+      // The freshly uploaded file appears on /#/recent
+      await sidebar.goToRecent()
+      await alicePage.waitForURL(/\/recent/)
+      await aliceDrive.row(fileName).waitVisible()
+
+      // Trash it directly from the Recent view — sendToTrash's waitHidden
+      // validates that the row disappears live, without navigating away
+      await aliceDrive.row(fileName).sendToTrash()
+    } finally {
+      await safeUnlink(file)
+    }
+  })
+
+  test('Recent: a renamed file updates without page reload', async ({
+    alicePage,
+    aliceDrive
+  }) => {
+    await alicePage.goto(ALICE_ROOT)
+    const sidebar = new SidebarPage(alicePage)
+    const original = `recent-rename-${stamp()}.txt`
+    const renamed = `recent-renamed-${stamp()}.txt`
+    const file = path.join(path.dirname(FIXTURE), original)
+    await copyFile(FIXTURE, file)
+    try {
+      await aliceDrive.uploadFiles(file)
+      await aliceDrive.row(original).waitVisible()
+
+      // The file appears on /#/recent
+      await sidebar.goToRecent()
+      await alicePage.waitForURL(/\/recent/)
+      await aliceDrive.row(original).waitVisible()
+
+      // Rename directly from the Recent view — rename() self-validates
+      // by waiting for the new name to appear in the file list
+      await aliceDrive.row(original).rename(renamed)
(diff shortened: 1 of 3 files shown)


e2e/tests/navigation.spec.ts
```diff
@@ -30,6 +30,61 @@ test.describe('Navigation surfaces', () => {
...
+      // The freshly uploaded file appears on /#/recent
+      await sidebar.goToRecent()
+      await alicePage.waitForURL(/\/recent/)
+      await aliceDrive.row(fileName).waitVisible()
+
+      // Trash it directly from the Recent view — sendToTrash's waitHidden
+      // validates that the row disappears live, without navigating away
+      await aliceDrive.row(fileName).sendToTrash()
+    } finally {
+      await safeUnlink(file)
```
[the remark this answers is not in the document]
Crash--: @doubleface can you fix this?

src/hooks/useRecentFiles.jsx
```diff
@@ -1,59 +1,95 @@
...
+    }
+  }, [dataProxy])
+
+  if (!client) {
+    return {
+      data: [],
+      fetchStatus: 'error',
+      error: new Error('Client not available')
+    }
+  }
```
Crash--: how can we have undefined client here?!

src/hooks/useRecentFiles.jsx
```diff
@@ -1,59 +1,95 @@
...
+        } else if (!file.trashed) {
+          acc.push(file)
         }
-      } else {
-        setError(new Error('Client not available'))
-        setFetchStatus('error')
-      }
+        return acc
+      }, [])
+    }
```
Crash--, line 77: Why don't we put data coming from the data proxy to the store ? Why do we have several sources of truth? 

Why do we need that to be honest? If I'm renaming a document from the recent view, we should receive a realtime event, and the query should re-render by itself. 

I know that I'm missing something here, but I want to understand. Please help me

src/hooks/useRecentFiles.jsx
```diff
@@ -1,59 +1,95 @@
...
 import logger from '@/lib/logger'
 import { buildRecentQuery } from '@/queries'
 
 const useDataProxyRecents = () => {
-  const [data, setData] = useState([])
-  const [fetchStatus, setFetchStatus] = useState('loading')
-  const [error, setError] = useState(null)
+  const [proxyData, setProxyData] = useState(null)
+  const [proxyFetchStatus, setProxyFetchStatus] = useState('loading')
+  const [proxyError, setProxyError] = useState(null)
```
Crash--: Since these 3 states are connected together (so not independant), we should either use a reducer or only one object.

src/hooks/useRecentFiles.jsx
doubleface: Branching data from data proxy in the store is just not implemented and I have no idea what are the risks to do it.

src/hooks/useRecentFiles.jsx
doubleface: No reason, removed this check

src/hooks/useRecentFiles.jsx
Crash--: cc @paultranvan Be ready for a fun monday ;)

Crash--:
I still don't think this is the right solution. 

We should find a way to `useQuery()` on the shared drive Id and force these queries to dataproxylink. I'm trying to work on something, let me try.

Crash--:
And if I understand completly how it works, this fix will only work for my files, not the shared ones.

doubleface:
> And if I understand completly how it works, this fix will only work for my files, not the shared ones.

Yes, this is the intent of this PR at the moment. I updated the description since it was not clear

Crash--:
So it does not fix the issue. Let me polish what I'm working on. I'll try to create an issue & PRs for that.

doubleface:
Ok, then I'm closing this PR

PR #3982, opened 2026-06-29
fix(sharings): make sidebar counter consistent with sharings view

### LINKED ISSUE #3981: Counter not update
## Steps to reproduce

- I share a OO file
- I open sharing tab in the recipient side
- I see the notif in OO line
- I open the file
- I reopen sharing tab in recipient side

## Expected behavior

- Notification in OO line disappaer
- Counter in decrease of 1

## Observed behavior

- Notification in OO line is still displayed

Image

doubleface:
Problem:

Sometimes we did not have the same number displayed in the sidebar counter of sharings section and the list of new sharings in sharings section body.

Replace the independent CouchDB query for new sharing shortcuts in the sidebar badge with the same query used by the sharings body (buildSharingsQuery via useSharingContext). This ensures the counter only counts items that are actually present in the sharings view.

fixes https://github.com/linagora/twake-drive/issues/3981

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Sharings navigation badge so it more reliably reflects the number of new shared items.
  * Updated the Sharings navigation to refresh its results based on the currently shared documents, keeping the badge count consistent with what’s displayed.
  * The badge now safely defaults when results are unavailable, avoiding incorrect or missing counts.

### DIFF
--- a/src/modules/navigation/SharingsNavItem.jsx
+++ b/src/modules/navigation/SharingsNavItem.jsx
@@ -1,18 +1,30 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 
 import { useQuery } from 'cozy-client'
+import { isSharingShortcutNew } from 'cozy-client/dist/models/file'
+import { useSharingContext } from 'cozy-sharing'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import ShareIcon from 'cozy-ui/transpiled/react/Icons/ShareExternal'
 
 import { NavItem } from '@/modules/navigation/NavItem'
-import { buildNewSharingShortcutQuery } from '@/queries'
+import { buildSharingsQuery } from '@/queries'
 
 const SharingsNavItem = ({ clickState }) => {
-  const newSharingShortcutQuery = buildNewSharingShortcutQuery()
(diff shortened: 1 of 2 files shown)


src/modules/navigation/SharingsNavItem.jsx
```diff
@@ -1,27 +1,41 @@
...
+import React, { useMemo } from 'react'
 
 import { useQuery } from 'cozy-client'
+import { isSharingShortcutNew } from 'cozy-client/dist/models/file'
+import { useSharingContext } from 'cozy-sharing'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import ShareIcon from 'cozy-ui/transpiled/react/Icons/ShareExternal'
 
 import { NavItem } from '@/modules/navigation/NavItem'
-import { buildNewSharingShortcutQuery } from '@/queries'
```
zatteo, line 8: Is this query still used?

zatteo (approved):

src/modules/navigation/SharingsNavItem.jsx
doubleface: Good catch, it is not used anymore. Removed it

PR #3983, opened 2026-06-29
feat: dispatch shared-drive realtime events into the store

Crash--:
## Update: reactivity now comes from the data-proxy push, no per-drive websocket

Instead of opening one `CozyRealtime` socket per recipient drive in every Drive tab, the shared worker now pushes shared-drive document changes to Drive. `FilesRealTimeQueries` listens for those pushes on `window` and feeds them into the same buffering and dispatch pipeline, so the store updates the same way, without any Drive-side websocket.

- `FilesRealTimeQueries` no longer constructs `CozyRealtime`; it consumes `{ type: 'DATAPROXYMESSAGE', payload: { kind: 'realtime', event, doctype, driveId, doc } }` messages (origin filtered to the data-proxy) and reuses the existing buffers, `driveIdByFileId`, path resolution and dispatch.
- The worker side (relay on pouch `onSync` + forwarder) is in cozy/cozy-web-data-proxy#63, driven by the data-proxy periodic sync.
- The own-files realtime path is unchanged.
- Latency follows the worker periodic sync; a later change adds realtime inside the worker to bring it back to near-real-time through the same relay.

Original description below.

---

## Summary

Goal: make the Recent view and the shared-drive folder listing update live (trash, rename, move, create) for files inside shared drives, the way they already do for a user's own files.

Approach: shared-drive files are replicated locally by the data-proxy worker under a per-drive doctype. We let any view query them as ordinary reactive `io.cozy.files` documents through cozy-client's store, by adding a `driveId` query option that reads the drive's database while keeping the documents typed `io.cozy.files`. Remote changes flow into the same store slice through one realtime socket per recipient drive.

The work is split across three repositories:
- cozy-client: a `forceLink` query option plus driveId-aware shared-drive reads in cozy-pouch-link (linagora/cozy-client#1691).
- cozy-web-data-proxy: the worker forwards the `driveId` option through to its pouch link (cozy/cozy-web-data-proxy#62).
- this repository: subscribe one realtime socket per recipient shared drive and dispatch the events into the `io.cozy.files` store.

What this PR contains:
- `FilesRealTimeQueries` opens one `CozyRealtime` per recipient shared drive (`owner` is false) and dispatches created, updated and deleted events into the store as `io.cozy.files`, reusing the existing buffering and dispatch pipeline.
- Shared-drive file paths are resolved through the drive-scoped collection so the dispatched documents carry a correct path.
- The own-files realtime path is unchanged; shared-drive support is additive.

Still in progress before this is ready for review:
- Recipient-drive detection filters `owner === false`; confirm the sharings API always returns a boolean `owner`, or widen the filter so no recipient drive is silently skipped.
- Code Health: the dispatch method and its tests carry added complexity and test duplication to revisit.
- The end-to-end check needs the patched libraries linked into the worker build.
- Follow-up PRs will move the Recent view and the shared-drive folder view onto this capability.

### DIFF
--- a/src/components/FilesRealTimeQueries.jsx
+++ b/src/components/FilesRealTimeQueries.jsx
@@ -19,16 +19,26 @@ const bufferDeletedFiles = new Map()
 // Populated by drive-socket handlers; absent (undefined) for own-instance files.
 const driveIdByFileId = new Map()
 
-const getParentFolder = async (client, dirId) => {
+const getParentFolder = async (client, dirId, driveId) => {
   let parentDir = client.getDocumentFromState('io.cozy.files', dirId)
   if (!parentDir) {
-    // Parent is not in the store: query it
-    const parentQuery = buildFileOrFolderByIdQuery(dirId)
-    const parentResult = await client.fetchQueryAndGetFromState({
-      definition: parentQuery.definition(),
-      options: parentQuery.options
-    })
-    parentDir = parentResult.data
+    if (driveId) {
+      // Drive file: resolve the parent via the drive-scoped collection so the
+      // request is authenticated against the correct shared drive.
+      // statById returns { data: folderDoc, included: children, links }.
+      const result = await client
+        .collection('io.cozy.files', { driveId })
+        .statById(dirId)
+      parentDir = result.data
+    } else {
+      // Own-instance file: fall back to the standard store query path.
+      const parentQuery = buildFileOrFolderByIdQuery(dirId)
+      const parentResult = await client.fetchQueryAndGetFromState({
+        definition: parentQuery.definition(),
+        options: parentQuery.options
+      })
+      parentDir = parentResult.data
+    }
   }
   return parentDir
 }
@@ -79,9 +89,17 @@ const processEvents = async (client, mutationType) => {
   }
 
   for (const folderId in filesByFolder) {
+    const filesInFolder = filesByFolder[folderId]
+    // Derive the originating drive ID from the first file in the group.
+    // All files sharing the same dir_id in a given processing cycle come from
+    // the same drive (or from own-instance files which have no entry).
+    const driveId =
+      filesInFolder.length > 0
+        ? driveIdByFileId.get(filesInFolder[0]._id)
+        : undefined
     const files = []
-    const folder = await getParentFolder(client, folderId)
-    for (const file of filesByFolder[folderId]) {
+    const folder = await getParentFolder(client, folderId, driveId)
+    for (const file of filesInFolder) {
       const fileWithPath = ensureFilePath(file, folder)
       files.push(fileWithPath)
     }
@@ -101,10 +119,11 @@ const processEvents = async (client, mutationType) => {
       )
     )
   }
-  // Remove processed files from buffer
+  // Remove processed files from buffer and clear drive-id tracking.
   // Do not clear all at once in case pending events arrived during the processing
   for (const fileId of fileIdsToProcess) {
     bufferFiles.delete(fileId)
+    driveIdByFileId.delete(fileId)
   }
 }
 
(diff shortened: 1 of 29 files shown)


Crash--:
## Why the `RecentScopeQuery` component pattern

The reactive recents need one query for the user's own files plus one per recipient shared drive, and that count changes at runtime (drives are added or removed via the `io.cozy.sharings` realtime). The rules of hooks forbid calling `useQuery` in a loop or conditionally, so a variable number of queries cannot be expressed as direct hook calls.

The headless-component-per-scope pattern is the idiomatic React way to vary the number of live reactive queries: each `RecentScopeQuery` renders `null`, runs exactly one `useQuery`, and lifts its result to `useRecentFiles` via `onResult`. Mounting and unmounting a component is the legal way to change how many queries are active. From the outside `useRecentFiles` stays a hook returning `{ data, fetchStatus, error, scopeQueries }`; the components are an internal detail the view renders invisibly.

Alternatives considered:
- An imperative loop (`client.query` in an effect) would lose reactivity, which is exactly the `dataProxy.recents()` approach this PR replaces.
- cozy-client's existing `useQueries` loops `useQuery` internally (with an eslint-disable), so it is only safe for a static set of queries, not a dynamic count.

A cleaner long-term fix, a dynamic-safe `useQueries` in cozy-client, is tracked in linagora/cozy-client#1692. Once available, `useRecentFiles` can drop this component pattern for a single `useQueries(specs)` call.

src/queries/index.ts
```diff
@@ -539,6 +568,35 @@ export const buildSharedDriveFolderQuery: QueryBuilder<
...
+> = ({ driveId, folderId, sortAttribute, sortOrder }) => ({
+  definition: () =>
+    Q('io.cozy.files')
+      .where({
+        dir_id: folderId,
+        driveId,
+        [sortAttribute]: { $gt: null }
+      })
+      .indexFields(['dir_id', 'driveId', sortAttribute])
+      .sortBy([{ dir_id: sortOrder }, { [sortAttribute]: sortOrder }])
```
rezk2ll: skipping `driveId` triggers the "sort order should be the same than the indexed fields" warning and an ad-hoc index. we should add `{ driveId: sortOrder }` to the `sortBy`

src/components/FilesRealTimeQueries.jsx
```diff
@@ -34,6 +61,32 @@ export const ensureFileHasPath = async (doc, client) => {
...
+  filesInFolder,
+  mutations
+) => {
+  // Derive the originating drive ID from the first file in the group.
+  const driveId =
+    filesInFolder.length > 0
+      ? driveIdByFileId.get(filesInFolder[0]._id)
+      : undefined
+  const folder = await getParentFolder(client, folderId, driveId)
+  const files = filesInFolder.map(file => ensureFilePath(file, folder))
```
rezk2ll: we need to reapply `driveId` (already tracked in `driveIdByFileId`) onto each dispatched file.

Realtime docs arrive straight from the stack without the synthetic `driveId` (that field is added only by the data-proxy on replication). The dispatch reapplies `path` but not `driveId`, so cozy-client re-evaluates the shared-drive folder query (whose selector filters on `driveId`) against a doc that lacks it and drops the file from the live results. A refresh re-runs the pouch find and it reappears.

you can notice this when you rename a file in recipient side, it vanishes and reappear only on refresh. for owner it works fine

rezk2ll (commented):
other stuff i found, on recipient side:

- The live view (store) can be correct while a refresh (which reads the recipient drive pouch) shows a just-added file missing or a just-deleted file still present. Root cause is on the worker and pouch-link side ( the other two PRs ), but flagging it here.

- in targets/browser/setupAppContext.js: The link that serves every shared-drive query is only added when `flag('dataproxy.queries.enabled')` is truthy, but cozy-flags is not populated this early in `setupApp`, so on a cold load the link is missing and every drive query fails with `No link could handle operation` (the folder shows "Something went wrong")

src/components/FilesRealTimeQueries.jsx
Crash--: OK for this one I think. Yes we alredy have the driveId we should reuse.

src/queries/index.ts
Crash--: Something to add to my skill ^^

PR #4004, opened 2026-07-01
Fix nav link for shared drive

zatteo:
In nav, shared drive (team drive) were not highlighted. Now they are.

https://github.com/user-attachments/assets/fb077308-5dc3-4d33-89c1-dff560b06c72

## Summary by CodeRabbit

* **New Features**
  * Improved navigation highlighting so the current section stays active more accurately across sharing and shared drive pages.
  * Shared drive items now follow the updated active-link behavior for clearer menu state.

* **Bug Fixes**
  * Fixed cases where navigation could appear active for the wrong shared drive.
  * Cleared stale link state when moving between pages, improving consistency in the sidebar.

### DIFF
--- a/src/modules/navigation/NavItem.jsx
+++ b/src/modules/navigation/NavItem.jsx
@@ -27,6 +27,7 @@ const NavItem = ({
   icon,
   label,
   rx,
+  isActive,
   clickState,
   badgeContent,
   secondary,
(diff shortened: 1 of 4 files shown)


JF-Cozy (approved):
not relative directly to this PR but the entire process seams to be complicated, fragile and hard to maintain (without talking about setLastClicked 😱 ) we probably should deep dive into it and see, working with cozy-ui too, if we can simplify it and how 🤔

zatteo:
> not relative directly to this PR but the entire process seams to be complicated, fragile and hard to maintain (without talking about setLastClicked 😱 ) we probably should deep dive into it and see, working with cozy-ui too, if we can simplify it and how 🤔

Could be interesting indeed. And talking with product before to ensure we are doing what is expected.

PR #4024, opened 2026-07-08
docs: Propose thumbnail in File Picker intent

zatteo:
## Context

When using the File Picker to share by link, we want to display nicely links in the email. So we want to display these beautiful cards for the sender on tmail but also for the recipient on whatever email client he uses. So these beautiful cards must be inline HTML. The question is about the illustration of these cards.

Screenshot 2026-07-08 at 09 37 54

## Plan

It makes sense that the File Picker return a thumbnail to keep consistency between all apps calling the File Picker. Also, even if today we only talk about illustration, we may return real thumbnail in the future and only the File Picker can return a real thumbnail easily.

According to tmail team, it is not possible to use base64 images as they won't be rendered by some clients. So we propose here the following plan:

1. On a public CDN on our side, we upload some illustrations like _https://cdn.twake.app/files/pdf.jpg_, _https://cdn.twake.app/files/image.jpg_, _https://cdn.twake.app/files/default.jpg_
⚠️ this CDN does not exist yet
2. File Picker map the type of the picked file to the list of illustrations available in the CDN or to the default one

## Summary by CodeRabbit

* **Documentation**
  * Updated File Picker response details to include an optional thumbnail link for selected files.
  * Expanded the example payload to show thumbnail data alongside existing link fields.
  * Added guidance explaining that thumbnails are public, intended as previews, and do not expire.

### DIFF
--- a/docs/file-picker-intent.md
+++ b/docs/file-picker-intent.md
@@ -191,6 +191,9 @@ interface FilePickerEntry {
   mimeType: string | null
   sharingLink?: string
   downloadLink?: string
+  thumbnail?: {
+    link: string
+  }
 }
 ```
 
@@ -206,13 +209,20 @@ Example:
       "name": "invoice.pdf",
       "size": 123456,
       "mimeType": "application/pdf",
-      "downloadLink": "https://alice.example/files/download/..."
+      "downloadLink": "https://alice.example/files/download/...",
+      "thumbnail": {
+        "link": "https://cdn.example.com/files/pdf.jpg"
+      }
     }
   ]
 }
 ```
 
 For folders, `size` is `0` and `mimeType` is `null`.
+
+### Thumbnails
+
+The File Picker may provide a thumbnail (an illustration or a preview) that may be used by the caller. The thumbnail link is public and has an unlimited lifetime.
 
 ## Error handling
 


docs/file-picker-intent.md
```diff
@@ -206,14 +209,21 @@ Example:
...
     }
   ]
 }
 ```
 
 For folders, `size` is `0` and `mimeType` is `null`.
 
+### Thumbnails
+
+The File Picker may provide a thumbnail (an illustration or a preview) that may be used by the caller. The thumbnail link is public and has an unlimited lifetime.
```
zatteo, line 225: I keep the specification very simple for now. If one day we want to return thumbnail with expiration date, we can add an _expireAt_ field.

docs/file-picker-intent.md
```diff
@@ -191,6 +191,9 @@ interface FilePickerEntry {
   mimeType: string | null
   sharingLink?: string
   downloadLink?: string
+  thumbnail?: {
+    link: string
+  }
```
zatteo, line 196: Do we need more? format of the image? size of the image? I think no for now. We should just enforce 64x64 and jpg for every images.

hoangdat (approved):

tddang-linagora (approved):

doubleface:
@zatteo I suppose we will need to put it behind a flag to avoid to have some special setup to do in local dev environment ? Looks good to me

doubleface (approved):

zatteo:
> @zatteo I suppose we will need to put it behind a flag to avoid to have some special setup to do in local dev environment ? Looks good to me

Good idea 👍

zatteo:
I am waiting work from ops team to deploy a CDN so that we can map something like
```
{
   image: 'https://cdn.twake.app/image.jpg',
   pdf: 'https://cdn.twake.app/pdf.jpg',
   default: 'https://cdn.twake.app/default.jpg',
}
```

PR #4044, opened 2026-07-10
feat(file-picker): improve file selection flow (step 2)

### LINKED ISSUE #3971: Basic file picker no ricents no search no grid view / Step 2
Basic file picker no ricents no search no grid view 
@doubleface

doubleface:
- Render intent picker files in a virtualized table aligned with the Drive layout.
- Add keyboard, range, and multiple-file selection support.
- Reuse the selection provider and clear selections when navigating, confirming, or closing.
- Update picker translations and unit and end-to-end coverage.

fixes #3971 

[Capture vidéo du 2026-07-10 18-13-21.webm](https://github.com/user-attachments/assets/135136c1-a0de-455a-8107-7749339189f0)

## Summary by CodeRabbit

* **New Features**
  * File Picker now supports multi-item selection, returning per-entry public sharing links and per-entry temporary download links (including mixed folder/file selection rules).
  * Updated File Picker UI with a virtualized table, streamlined breadcrumb, selection counter, and clear-selection action.
* **Bug Fixes**
  * Shift+Arrow navigation now keeps the current selection when attempting to move outside list bounds.
* **Style**
  * Shortened and standardized File Picker link/attachment button labels across supported languages.
* **Tests**
  * Expanded end-to-end and hook coverage for multi-file selection, confirmation outcomes, and keyboard/boundary behaviors.


src/modules/selection/SelectionProvider.jsx
```diff
@@ -27,24 +27,33 @@ import { useNewItemHighlightContext } from '@/modules/upload/NewItemHighlightPro
...
 
+const SelectionRouteCleaner = ({ onLocationChange }) => {
+  const location = useLocation()
+
+  useEffect(() => {
+    onLocationChange()
+  }, [location, onLocationChange])
+
+  return null
+}
```
zatteo, line 38: Can't you write a hook for this? Why we need a component?

src/modules/services/components/FilePicker/FilePickerTable.jsx
```diff
@@ -0,0 +1,130 @@
...
+    id: 'size',
+    label: t('table.head_size'),
+    width: 100,
+    sortable: false
+  }
+]
+
+const FilePickerTableRow = forwardRef(
+  ({ item, context, className, ...props }, ref) => {
+    const row = item || context.data?.[props['data-item-index']]
```
zatteo: Why do we need `context.data?.[props['data-item-index']]`?

src/modules/services/components/FilePicker/FilePickerTableCell.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <time
+      dateTime={updatedAt}
+      title={f(updatedAt, t('LastUpdate.titleFormat'))}
+    >
+      {formattedUpdatedAt}
+    </time>
+  )
+}
```
zatteo: Is it possible to reuse some classic virtualized cell like /src/modules/filelist/virtualized/cells/LastUpdate.jsx instead of creating a new one for the File Picker?

src/modules/services/components/FilePicker/FilePickerFooter.jsx
```diff
@@ -25,74 +33,76 @@ const FilePickerFooter = ({
...
+    }
+
+    const tooltipTitle =
+      state.reasonKey === 'FilePicker.constraints.disabledReasons.fileTooLarge'
+        ? t(state.reasonKey, {
+            maxFileSize: filesize(actionConfig?.maxFileSize ?? 0, {
+              base: 10
+            })
+          })
+        : t(state.reasonKey)
```
zatteo: nit: To move in an helper?

zatteo:
Good to have avoid Sprite 👍

src/modules/selection/SelectionProvider.jsx
doubleface: The component is intentional here because SelectionProvider is also rendered outside a Router when clearOnLocationChange is disabled. Calling useLocation() directly from the provider
would still throw in that case. Keeping it in a conditionally rendered child lets us use the router only when needed while respecting the Rules of Hooks.

src/modules/services/components/FilePicker/FilePickerTable.jsx
doubleface: Right, the fallback is not needed here. I'll remove it

src/modules/services/components/FilePicker/FilePickerTableCell.jsx
doubleface: Right, LastUpdate can be reused here, and also Size. The existing FileName cell is more tightly coupled to the regular file list behavior (FileOpener, rename state and selection context), so I would keep the File Picker-specific name cell. I will update the date and size cells to reuse the existing presentation components.

src/modules/services/components/FilePicker/FilePickerFooter.jsx
doubleface: right, done

src/modules/services/components/FilePicker/FilePickerTableCell.jsx
zatteo: Great! Totally fine to create new component if needed, but if they are 99% or 100% same lets reuse them.

zatteo (approved):
You can clean the custom cell for File Picker that are unused

Issue #4040, opened 2026-07-10
Sharings. Sort out shares. Tabs. ST4: Wire tabs into SharingsView


shepilov:
Part of #3975 (Sharings tabs). Depends on #4037, #4038 and #4039.

## Goal
Wire tab state + per-tab data + tabs UI together in the Sharings view.

## Implementation details
- `src/modules/views/Sharings/index.jsx`:
  - `const [tab, setTab] = useSharingsTab()` (#4037)
  - pass the active tab to the evolved `useFilteredSharings` (#4038)
  - render `<SharingsTabs tab={tab} onChange={setTab} showDrives={…} />` per the placement rules in #4039 (desktop: header row; mobile: below the header)
- Verify derived behaviors follow the active tab automatically (they already consume `filteredResult.data`):
  - Select All — `buildSharingsActionsOptions` in `src/modules/views/Sharings/helpers.js` builds `selectAll` from the rendered list
  - `useKeyboardShortcuts` items
  - selection bar / bulk actions
- **Selection on tab switch**: `SelectionProvider` (`src/modules/selection/SelectionProvider.jsx`) already clears the selection on every `location` change, and a tab switch is a navigation because tab state is URL-based. No new code expected — add a regression test so this stays true (a future in-memory tab switch would silently break it).
- Empty state: reuse the existing empty component for each tab; per-tab copy is pending design (#3963) — don't block on it. The Team drives tab shows an empty state when flags are on but no org drive exists.
- Loading conditions unchanged (`allLoaded`, `sharedDrivesLoaded`, `hasQueryBeenLoaded`). Optional nicety: skip the `sharedDrivesLoaded` gate on the two non-drive tabs.

## Acceptance criteria
- [ ] Switching tabs swaps the list without a refetch
- [ ] Select All selects only visible (active-tab) items
- [ ] **Switching tabs clears the current selection** — bulk actions can never operate on items hidden in another tab (regression test)
- [ ] Shared-drive actions (share drive / leave drive) still work from the Team drives tab
- [ ] Existing `index.spec.jsx` stays green

Issue #4039, opened 2026-07-10
Sharings. Sort out shares. Tabs. ST3: SharingsTabs segmented control UI


shepilov:
Part of #3975 (Sharings tabs). Depends on #4037; can run in parallel with #4038.

## Goal
Segmented pill control (**With me / By me / Team drives**) per the mockups in #3975 / #3963 (Figma in #3963) — visible on both desktop **and mobile**.

## Implementation details
- New `SharingsTabs` component in `src/modules/views/Sharings/`, fully controlled: `({ tab, onChange, showDrives })`.
- **Placement (important):** `Topbar` (`src/modules/layout/Topbar.jsx`) defaults to `hideOnMobile: true` and `FolderViewHeader` doesn't override it — tabs rendered only inside the header are **invisible on mobile**. The removed implementation rendered `SharingTab` below the header for exactly this reason (`git show 154128053:src/modules/views/Sharings/index.jsx`).
  - Desktop: segmented pills in the header row next to `MobileAwareBreadcrumb` (per mockup)
  - Mobile: full-width `Tabs` (`narrowed`) rendered **below the header, outside `Topbar`**
- Desktop control: cozy-ui `Tabs` supports a `segmented` prop; the removed implementation used `ToggleButtonGroup variant="rounded"` (reference: `git show 154128053:src/modules/views/Sharings/SharingTab.jsx`). Pick whichever matches the Figma pills closest.
- **Team drives tab visibility**: shown whenever `drive.shared-drive.enabled` or `drive.federated-shared-folder.enabled` is on — even while the drives list is empty or still loading (`useSharedDrives` is async; a data-driven tab would cause layout shift). The empty tab shows an empty state (#4040). Hidden only when both flags are off.
- i18n: replace the orphaned keys `toolbar.sharings_tab_all` / `toolbar.sharings_tab_drives` (`src/locales/en.json`) with `sharings_tab_with_me`, `sharings_tab_by_me`, `sharings_tab_drives`; update `en.json` + `fr.json` (other locales via Transifex).
- Mobile design is not specified in Figma yet (open request on #3963); interim default is the full-width tabs described above.

## Acceptance criteria
- [ ] Desktop matches mockup: pills inline with the title, active pill highlighted
- [ ] **Tabs are visible and usable on mobile viewports**
- [ ] Keyboard accessible: focus states, arrow/tab navigation between tabs
[...]

Issue #3975, opened 2026-06-29
Sharings. Sort out shares. Tabs


Bobpodvalnyi:
<img width="1830" height="1090" alt="Image" src="https://github.com/user-attachments/assets/452031da-a780-4ecb-881a-feb2ed96af68" />

### Tabs
The interface is divided into three logical tabs, allowing users to quickly switch between different access types:

**With me:** Files and folders shared with you by other users.

**By me:** Files and folders that you have shared with others.

**Team drives:** Shared workspaces intended for collaboration within a B2B context (organizational drives).

when i reload the page we keep tabs and filters applied in the same state 

---

## Implementation plan

All data needed for the three tabs is already client-side — the Sharings view runs a single query over all shared doc ids (`src/modules/views/Sharings/index.jsx`), so this is pure client-side classification + UI. **No backend changes.**

### State persistence
The URL query param is the **single source of truth**: `/sharings?tab=with-me|by-me|drives`. Reload keeps the URL (satisfies the requirement), links are shareable. User tab switches **push** history entries, so browser back/forward moves between visited tabs; rewrites of invalid or unavailable values (canonicalization, incl. the legacy `?tab=1`) use **replace** so the back button is never trapped. No sessionStorage — in-section navigations preserve the query string instead (#4041). Future filter chips (Type / Contact / Modification date / Sources — separate slice, not yet in the codebase) plug into the same query-string mechanism and inherit reload persistence for free.

### Tab semantics
- **With me** — non-drive shares where `useSharingContext().isOwner(docId) === false`, pending invitation shortcuts (`isSharingShortcut`; checked first because `isOwner` defaults to `true` for docs missing from `byDocId`), and received federated drive roots (sharing doc `owner === false`)
- **By me** — non-drive shares where `isOwner === true` (includes public-link-only owned shares) and owned federated drive roots
- **Team drives** — organizational drives only (`org_drive === true`, same predicate as the sidebar `SharedDriveList`); tab visible whenever a shared-drive flag is on (empty state if none yet), hidden when `drive.shared-drive.enabled` and `drive.federated-shared-folder.enabled` are both off
- **Every document appears in exactly one tab.** Non-org drive roots are classified via the sharing doc's `owner` attribute, **not** `sharingContext.isOwner` — a recipient's drive root is often absent from `byDocId`, where `isOwner` defaults to `true`.

### Cross-cutting behaviors
[...]

PR #4047, opened 2026-07-15
feat: classify sharings into with-me/by-me/drives tabs

### LINKED ISSUE #4038: Sharings. Sort out shares. Tabs. ST2: Split sharings into With me / By me / Team drives
Part of #3975 (Sharings tabs). Depends on #4037 (tab constants).

## Goal
Split the single sharings query result into the three tab collections (With me / By me / Team drives), keeping the query-result shape consumed by `FolderViewBody`. Pure client-side classification — no backend changes, no new queries.

## Classification rules
- **Team drives**: organizational drives only — drive-type sharings with `org_drive === true` (same predicate the sidebar `SharedDriveList.tsx` uses), rendered as the transformed entries from `useTransformFolderListHasSharedDriveShortcuts` (`src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx`).
- **Non-org drive roots** (federated shared folders) are classified by ownership like regular shares — but using the **sharing doc's `owner` attribute** (`SharedDrive.owner`, available on `useSharedDrives` results), NOT `sharingContext.isOwner(rootId)`:
  ⚠️ a recipient's drive root is often absent from `byDocId`, where `isOwner` defaults to `true` — using it would silently put received federated drives under "By me".
  `owner === true` → By me; otherwise → With me.
- **With me**: items from `nonSharedDriveList` that are sharing shortcuts (`isSharingShortcut` from `cozy-client/dist/models/file`) **or** where `useSharingContext().isOwner(docId) === false`. The shortcut check comes first (same `isOwner` default-true pitfall).
- **By me**: remaining `nonSharedDriveList` items (`isOwner === true`). This naturally includes **public-link-only owned shares**: their `byDocId` entries carry permissions but no sharings, and `isOwner` then resolves to `true`.

## Implementation details
- Evolve `useFilteredSharings` (`src/modules/views/Sharings/useFilteredSharings.jsx`) to accept the active tab and return `{ filteredResult, sharedDrivesLoaded }` with the `{ ...result, data, count }` shape.
- The transformed drive entries carry `driveId`; join them with the `useSharedDrives` docs to read `org_drive` / `owner` (or extend the transform hook to pass both through).
- Export the classifier itself (doc + sharing context → tab) so the viewer can reuse it for tab-aware navigation (#4041).
- Preserve existing behavior: `deduplicateSharingShortcuts`, trash/Nextcloud exclusions, and the flags-off path (both `drive.shared-drive.enabled` and `drive.federated-shared-folder.enabled` off → **all** drive entries dropped, hook must not block on `sharedDrivesLoaded`).

## Acceptance criteria
- [ ] **Every document appears in exactly one tab**, including transient duplicate representations (dedup + exhaustive, mutually exclusive classification)
- [ ] Fixtures: owned share → By me; received share → With me; new invitation shortcut → With me; **link-only owned share → By me**; org drive → Team drives; owned federated drive → By me; received federated drive → With me
- [ ] Both flags off → no drive entries in any tab, no load-blocking
- [ ] Existing dedup regression tests still pass

shepilov:
## What

Second subtask of #3975 (Sharings tabs). Adds the client-side classification layer that decides which tab — **Shared with me**, **Shared by me**, or **Team drives** — each entry of the Sharings view belongs to. No visible change yet: the view call-site is untouched; the tab control (#4039) and view wiring (#4040) build on this.

## How

- `useTransformFolderListHasSharedDriveShortcuts` stamps `orgDrive`/`driveOwner` (from the sharing doc's `org_drive`/`owner` attributes) onto transformed drive entries, on both the owner-file and the recipient-synthetic branch — the sharing attributes aren't available downstream otherwise, and this avoids refetching sharings in the view.

Part of #3975
Closes #4038

## Summary by CodeRabbit

* **New Features**
  * Added tab-based filtering and classification for sharings, including shortcut and shared-drive categories.
  * Shared-drive entries now expose organization and ownership flags to improve accurate tab placement.

* **Bug Fixes**
  * Improved deduplication so the selected (“regular”) row retains drive classification metadata.
  * Corrected tab classification for pending invitations, link-only shares, and shared-drive entries (including owner scenarios).

* **Tests**
  * Expanded unit tests for shared-drive metadata transformation, shortcut deduplication, and per-tab filtering behavior.

### DIFF
--- a/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
+++ b/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
@@ -82,9 +82,6 @@ const useTransformFolderListHasSharedDriveShortcuts = (
           ...(sharing.rules[0]?.mime ? { mime: sharing.rules[0].mime } : {})
         }
 
-        // org_drive/owner only exist on the sharing doc; stamp them on the
-        // entry so the Sharings view can classify drives into tabs without
-        // refetching the sharings.
         const sharedDriveData = {
           type: isFileDriveRoot ? ('file' as const) : ('directory' as const),
           name: driveName,
--- a/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
+++ b/src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
@@ -17,10 +17,14 @@ interface SharedDrive {
   id: string
   drive_root_type?: DriveRootType
   rules: SharingRule[]
+  org_drive?: boolean
+  owner?: boolean
 }
 
 interface TransformedSharedDrive extends SharedDriveFile {
   driveId: string
+  orgDrive?: boolean
+  driveOwner?: boolean
 }
 
 interface UseTransformFolderListReturn {
@@ -78,11 +82,16 @@ const useTransformFolderListHasSharedDriveShortcuts = (
           ...(sharing.rules[0]?.mime ? { mime: sharing.rules[0].mime } : {})
         }
 
+        // org_drive/owner only exist on the sharing doc; stamp them on the
+        // entry so the Sharings view can classify drives into tabs without
+        // refetching the sharings.
         const sharedDriveData = {
           type: isFileDriveRoot ? ('file' as const) : ('directory' as const),
           name: driveName,
           dir_id: SHARED_DRIVES_DIR_ID,
           driveId: sharing.id,
+          orgDrive: Boolean(sharing.org_drive),
+          driveOwner: Boolean(sharing.owner),
           ...(isFileDriveRoot
             ? {
                 ...fileMetadata,
@@ -105,6 +114,8 @@ const useTransformFolderListHasSharedDriveShortcuts = (
             {
               ...fileInSharingSection,
               driveId: sharing.id,
+              orgDrive: Boolean(sharing.org_drive),
+              driveOwner: Boolean(sharing.owner),
               ...(isFileDriveRoot
                 ? {
                     ...fileMetadata,
--- a/src/modules/views/Sharings/useFilteredSharings.jsx
+++ b/src/modules/views/Sharings/useFilteredSharings.jsx
@@ -1,8 +1,15 @@
 import { useMemo } from 'react'
 
+import { isSharingShortcut } from 'cozy-client/dist/models/file'
 import flag from 'cozy-flags'
+import { useSharingContext } from 'cozy-sharing'
 
-import { SHARED_DRIVES_DIR_ID } from '@/constants/config'
+import {
+  SHARED_DRIVES_DIR_ID,
+  SHARING_TAB_BY_ME,
+  SHARING_TAB_DRIVES,
+  SHARING_TAB_WITH_ME
+} from '@/constants/config'
 import { useTransformFolderListHasSharedDriveShortcuts } from '@/hooks/useTransformFolderListHasSharedDriveShortcuts'
 
 const buildBaseShape = (result, hasIds) => ({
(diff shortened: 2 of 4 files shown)


src/modules/views/Sharings/useFilteredSharings.jsx
```diff
@@ -95,21 +141,30 @@ export const useFilteredSharings = ({ result, sharedDocumentIds }) => {
...
+    // exactly one tab.
+    const combined = computeData({
       result,
       withoutSharedDrives,
       transformedSharedDrives,
       nonSharedDriveList
     })
+    const data = tab
+      ? combined.filter(entry => getSharingsTabForEntry(entry, isOwner) === tab)
+      : combined
```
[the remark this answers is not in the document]
shepilov: Keeping this as is — it was assessed while preparing the PR and is a deliberate decision:

- List membership on the flags-off path is pre-existing behavior (the `dir_id` filter predates this PR); this PR only adds classification on top of the same list.
- For the shape that occurs in practice — an owner-side federated drive root (`owner: true`) — **By me is the same tab it gets with the flags on** (`driveOwner` → By me), so there is no user-visible misclassification.
- Removing regular drive roots when the flags are off requires the sharing docs: `referenced_by` alone cannot distinguish a drive sharing from a classic sharing. That means coupling this path to the async `useSharedDrives()` fetch, and doing it "without blocking loading" trades the current stable outcome for a flicker (row rendered until the fetch lands, then removed). The flags-off path deliberately ignores that fetch and its load state.
- The behavior is pinned by a dedicated test (owner-side drive root → By me with flags off) so it stays an explicit decision rather than an accident.

If org-drive roots (`org_drive: true`) ever surface owner-side with a regular `dir_id` in practice, that would be worth a dedicated follow-up issue.

Crash--:
Not blocking for this PR — a note for later.

This classification layer works on the full list loaded upstream by `cozy-sharing`'s `fetchAllSharings`, which pulls **every** sharing of the doctype in a single call with no pagination (client side and server side — the stack's `GetSharingsByDocType` runs a CouchDB view with no limit). That's fine today, but at some point we'll need to rework `fetchAllSharings`: with a large number of sharings, loading everything up front will get noticeably slow.

When we get there, moving to dedicated Mango queries (paginated, and ideally scoped per tab: shared-with-me / shared-by-me / drives) could make sense, rather than fetching the whole set and partitioning it client side.

src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
```diff
@@ -78,11 +82,16 @@ const useTransformFolderListHasSharedDriveShortcuts = (
           ...(sharing.rules[0]?.mime ? { mime: sharing.rules[0].mime } : {})
         }
 
+        // org_drive/owner only exist on the sharing doc; stamp them on the
+        // entry so the Sharings view can classify drives into tabs without
+        // refetching the sharings.
```
doubleface: nit: I don't think this comment is needed

doubleface (approved):

src/hooks/useTransformFolderListHasSharedDriveShortcuts/index.tsx
shepilov:
```suggestion
```

shepilov:
Tracked as a dedicated follow-up: #4048

shepilov:
> Not blocking for this PR — a note for later.
> 
> This classification layer works on the full list loaded upstream by `cozy-sharing`'s `fetchAllSharings`, which pulls **every** sharing of the doctype in a single call with no pagination (client side and server side — the stack's `GetSharingsByDocType` runs a CouchDB view with no limit). That's fine today, but at some point we'll need to rework `fetchAllSharings`: with a large number of sharings, loading everything up front will get noticeably slow.
> 
> When we get there, moving to dedicated Mango queries (paginated, and ideally scoped per tab: shared-with-me / shared-by-me / drives) could make sense, rather than fetching the whole set and partitioning it client side.

https://github.com/linagora/twake-drive/issues/4048

PR #4051, opened 2026-07-15
feat: add sharings tabs with per-tab filtering

shepilov:
## What

Third and fourth subtasks of #3975 (Sharings tabs): the visible **With me / By me / Team drives** control, wired to per-tab filtering so the feature is fully functional. The Sharings view now shows a segmented pill control next to the page title on desktop and full-width tabs below the header on mobile; switching tabs filters the list client-side (classification landed in #4047) and persists in the URL (#4043).

**Design check against Figma** (`Cozy new UI`, node 33267-55389): token-level match on track color, typography and labels. 
Known deviations left as is for now: control height (Figma 32px vs cozy-ui theme's 40px `minHeight`), inactive label color (`text/disabled` vs MUI default) and the active-pill shadow — all theme-level in cozy-ui, worth a design-system alignment rather than local overrides. The filters row in the mockup is a separate feature outside #3975.

Part of #3975
Closes #4039
Closes #4040

## Summary by CodeRabbit

* **New Features**
  * Updated Sharings navigation to use separate **“With me”** and **“By me”** tabs (including new grouping label support).
  * Added/labelled the **“Team drives”** tab when drives are available.
  * Refreshed the Sharings header with responsive tab placement and tab-driven file filtering.
* **Bug Fixes**
  * Tab switching via URL parameters now updates query state without disrupting other routing, and correctly clears selection/hides the selection bar.
* **Localization**
  * Updated Sharings tab labels across supported languages (including “Team drives” wording).
* **Tests**
  * Expanded Sharings tab and filtering coverage (desktop/mobile) and updated related e2e helpers/assertions.

Issue #4037, opened 2026-07-10
Sharings. Sort out shares. Tabs. ST1: Tab state via URL param with reload persistence


shepilov:
Part of #3975 (Sharings tabs). No dependencies — foundation subtask.

## Goal
Tab selection lives in the URL (`/sharings?tab=with-me|by-me|drives`) as the **single source of truth**, so it survives page reload. No sessionStorage — in-section navigations preserve the query string instead (see #4041).

## Implementation details
- `src/constants/config.js` — replace the numeric tab constants:
  - `SHARING_TAB_WITH_ME = 'with-me'` (default tab)
  - `SHARING_TAB_BY_ME = 'by-me'`
  - `SHARING_TAB_DRIVES = 'drives'` (currently `1`)
  - drop `SHARING_TAB_ALL` (referenced nowhere outside config)
- New hook `useSharingsTab()` in `src/modules/views/Sharings/` returning `[tab, setTab]`:
  - reads `tab` via `useSearchParams` and validates it against the three slugs
  - **availability**: `drives` is only a valid target while `drive.shared-drive.enabled` or `drive.federated-shared-folder.enabled` is on; a deep link to an unavailable tab resolves to the default
  - any unrecognized (incl. legacy `?tab=1`), missing, or unavailable value resolves to `with-me`
  - **canonicalization**: when the resolved tab differs from the param, the hook rewrites the URL with `{ replace: true }` so reloads/copied links are canonical and the back button is never trapped
  - `setTab` (user action) **pushes** a history entry, so browser back/forward moves between previously visited tabs
  - `setTab` copies the current search params and only sets `tab` — future filter params (type/contact/date/source) are preserved; invalid or unavailable values are ignored with a logged warning
- The existing `?tab=` emitters keep working through the updated constant (they import it):
  - `SharedDrivesRedirect` — `src/modules/navigation/AppRoute.jsx`
  - `makeRevokeSuccessRedirectPath` — `src/modules/views/Modal/revokeSuccessRedirect.js`
  - specs building URLs from the constant: `ShareFileView.spec.jsx`, `ShareDisplayedFolderView.spec.jsx`

## Product note
Without a storage fallback, entering the section via the static sidebar link (`/sharings`) always lands on the default tab. If product wants "restore the last visited tab" on re-entry, that is a separate decision (it would reintroduce a storage fallback) — tracked as an open question on #3975.

## Acceptance criteria
[...]

PR #4052, opened 2026-07-15
feat: sharings tab across navigation and in the viewer

### LINKED ISSUE #4041: Sharings. Sort out shares. Tabs. ST5: Preserve tab across in-section navigation
Part of #3975 (Sharings tabs). Depends on #4040. **Load-bearing**: with URL-only tab state (no storage fallback, see #4037), preserving the query string is what keeps the tab stable across navigation.

## Goal
The active tab survives all in-section navigation (open file/folder, close viewer, breadcrumb back) and page reloads at any point — and the viewer never leaks into another tab.

## Implementation details
**Preserve `location.search` through in-section navigations:**
- Viewer close/change: `src/modules/views/Sharings/FilesViewerSharings.jsx` navigates to `basePath` (`/sharings`) and `${basePath}/file/:id` without the query string — append the current search
- Row click opening files (`/sharings/file/:fileId`) and folders (`/sharings/:folderId`)
- Breadcrumb link back to `/sharings`

**Tab-aware viewer:**
- `FilesViewerSharings` currently loads **every** shared file, so viewer next/previous can cross from one tab's files into another's. Filter the viewer file list with the classifier exported by #4038 so navigation stays within the active tab.

## Acceptance criteria
- [ ] Open a file from "By me", close the viewer → still on "By me" (URL still carries `tab=by-me`)
- [ ] **Viewer next/previous only cycles through files of the active tab**
- [ ] Enter a shared folder, go back via breadcrumb → same tab
- [ ] Reload on any tab, with or without an open viewer → same tab

shepilov:
> Rebased onto master — #4051 (the tabs UI + filtering this builds on) is merged.

## What

Fifth subtask of #3975 (Sharings tabs): the active tab survives all in-section navigation, and the viewer stops leaking across tabs. With URL-only tab state (#4037, no storage fallback), any navigation that drops the query string resets the tab — and `FilesViewerSharings` paged through **every** shared file, so next/previous could cross from one tab's files into another's.

Part of #3975
Closes #4041

## Summary by CodeRabbit

- **New Features**
  - Preserved the selected Sharings tab during navigation across breadcrumbs, file links, and the sharings viewer.
  - Applied tab-based filtering to file lists at the Sharings root (and avoided applying it to nested shared content).

- **Bug Fixes**
  - Prevented `?tab=...` from carrying into non-sharings routes.
  - Ensured closing/changing files in the viewer retains the active tab query parameter.

- **Tests**
  - Expanded unit tests for tab extraction/preservation, breadcrumb navigation behavior, and viewer filtering/navigation transitions.

### DIFF
--- a/src/modules/navigation/hooks/helpers.spec.js
+++ b/src/modules/navigation/hooks/helpers.spec.js
@@ -1,4 +1,9 @@
-import { computeFileType, computeApp, computePath } from './helpers'
+import {
+  computeFileType,
+  computeApp,
+  computePath,
+  getSharingsTabSearch
+} from './helpers'
 
 import { TRASH_DIR_ID, SHARED_DRIVES_DIR_ID } from '@/constants/config'
 import { DRIVE_ROOT_TYPE } from '@/modules/shareddrives/types'
@@ -743,5 +748,24 @@ describe('computePath', () => {
         pathname: '/sharings'
       })
     ).toBe('folder-1')
+  })
+})
+
+describe('getSharingsTabSearch', () => {
+  it('returns the tab param when navigating within the sharings section', () => {
+    expect(getSharingsTabSearch('/sharings', '?tab=by-me')).toBe('?tab=by-me')
+    expect(
+      getSharingsTabSearch('/sharings/folder-1', '?foo=bar&tab=drives')
+    ).toBe('?tab=drives')
+  })
+
+  it('returns an empty search outside the sharings section', () => {
+    expect(getSharingsTabSearch('/folder/dir-1', '?tab=by-me')).toBe('')
+    expect(getSharingsTabSearch('/favorites', '?tab=by-me')).toBe('')
+  })
+
+  it('returns an empty search when no tab is set', () => {
+    expect(getSharingsTabSearch('/sharings', '')).toBe('')
+    expect(getSharingsTabSearch('/sharings', '?foo=bar')).toBe('')
   })
 })
--- a/src/modules/navigation/hooks/helpers.ts
+++ b/src/modules/navigation/hooks/helpers.ts
@@ -139,6 +139,24 @@ export const computeFileType = (
   }
 }
 
+/**
+ * Search string preserving the active Sharings tab across in-section
+ * navigations.
+ *
+ * The `?tab=` query param is the only storage of the active tab (see
+ * useSharingsTab), so any navigation staying under /sharings must carry it
+ * over. Outside the section it returns an empty search, so other views
+ * never inherit the param.
+ */
+export const getSharingsTabSearch = (
+  pathname: string,
+  search: string
+): string => {
+  if (!pathname.startsWith('/sharings')) return ''
+  const tab = new URLSearchParams(search).get('tab')
+  return tab ? `?${new URLSearchParams({ tab }).toString()}` : ''
+}
+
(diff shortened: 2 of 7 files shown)


doubleface:
I tested this branch and when I'm in sharings section, in "by me", if I open a file in the viewer and come back, I'm not in the sharings section anymore

shepilov:
> I tested this branch and when I'm in sharings section, in "by me", if I open a file in the viewer and come back, I'm not in the sharings section anymore

Because you are jumping out of /sharings, it's working only in models and "with me" sharing URLs

When you "click on file" in "by me", you are jumping to my drive, kine of question behavior for me; I think we should stay in "by me" @Bobpodvalnyi what do you think?

Crash--:
Having to manipulate query strings by hand is a big code smell.

In the React / React Router world, query strings are meant to **qualify a resource** — a sort order, a search text, a filter — not to carry navigation state. Here we clearly have real navigation: switching tabs pushes an entry onto the browser history, back/forward move between tabs, etc. That's a "place", not a modifier on a resource.

Because the tab lives in `?tab=` as the single source of truth, and `navigate()` deliberately drops the current search on every navigation, we're forced to re-attach it by hand everywhere (`getSharingsTabSearch`, the `pathname.startsWith('/sharings')` guard, `new URLSearchParams(...)` spread across several components). We're fighting the router instead of using it.

Why not model the tabs as a real route instead? e.g. `/sharings/with-me`, `/sharings/by-me`, `/sharings/drives`, with folders nested underneath (`/sharings/with-me/folder-1`). Then relative navigation, breadcrumb "up", and the viewer's next/previous all preserve the tab for free — no query-string juggling, no section guard — while keeping the URL bookmarkable, shareable and back/forward-friendly.

shepilov:
> Having to manipulate query strings by hand is a big code smell.
> 
> In the React / React Router world, query strings are meant to **qualify a resource** — a sort order, a search text, a filter — not to carry navigation state. Here we clearly have real navigation: switching tabs pushes an entry onto the browser history, back/forward move between tabs, etc. That's a "place", not a modifier on a resource.
> 

mm, i thought the opposite, in the "react world" search changes equals to navigation.

> Because the tab lives in `?tab=` as the single source of truth, and `navigate()` deliberately drops the current search on every navigation, we're forced to re-attach it by hand everywhere (`getSharingsTabSearch`, the `pathname.startsWith('/sharings')` guard, `new URLSearchParams(...)` spread across several components). We're fighting the router instead of using it.
> 
> Why not model the tabs as a real route instead? e.g. `/sharings/with-me`, `/sharings/by-me`, `/sharings/drives`, with folders nested underneath (`/sharings/with-me/folder-1`). Then relative navigation, breadcrumb "up", and the viewer's next/previous all preserve the tab for free — no query-string juggling, no section guard — while keeping the URL bookmarkable, shareable and back/forward-friendly.

I agree that tabs as a part of routes as more clean approach, but it was too many changes: we need explicit folder/file segments, legacy redirects, modals.
So would've do it as a separate PR.

And what do you think about just saving tab in the state ?

Crash--:
> And what do you think about just saving tab in the state ?

Source of truth should be the URL. State should be derived from the URL.

doubleface (approved):
Then let's merge this one for the next release and create a task to integrate these tabs in the router

PR #4053, opened 2026-07-16
refactor(icons): remove sprite usage

doubleface:
Stop injecting the hidden SVG sprite in application layouts and replace its two string-based icon references with individual components. This avoids parsing and keeping the full sprite in the DOM while allowing the icons to be bundled independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved icon rendering for infected-file warnings and selection help information.
  * Standardized icon selection across delete confirmation and qualify-file views for more consistent UI.

* **UI Improvements**
  * Removed an unnecessary icon/sprite element from application, public, embedded, and placeholder layouts to keep the interface cleaner.


shepilov:
It looks like we have indirect consumers of this sprite:

- It's a delete confirmation model: 
      src/modules/drive/DeleteConfirm.jsx:21 generates trash, restore, album, people, and link strings.
- and classify a document
      - src/modules/views/Modal/QualifyFileView.jsx:14 receives numerous string icon names from Cozy Client’s 

I think document classification is not used right now at all, but not sure about conrfim delete

doubleface:
@shepilov good catch, now I fixed these files but I had to map string icon names to icon components

shepilov (approved):

PR #4055, opened 2026-07-16
perf(file-picker): reduce startup payload

### LINKED ISSUE #4057: performance file picker
- animation (Theo) 
- decrise loading time 
- prefetch files

doubleface:
## Summary

- Remove broad Cozy package force splitting from the unified web build.
- Prefetch the root folder while the intent service initializes.
- Load only the fallback and current intent locales.
- Stop loading viewer-only CSS from the File Picker intent while preserving sharing styles.
- Remove Drive's encrypted-file integration, including Vault providers, encryption query plumbing, and obsolete test mocks.
- Add reproducible build and startup measurements with independent reports.

Measurements use Chromium at 1280x720, 4x CPU throttling, 40 ms latency, 1.5 Mbps download, 750 Kbps upload, and a disabled browser cache. Each variant uses the same French instance and deterministic files, with one warm-up followed by five recorded runs.

Bundle size is the initial File Picker JavaScript gzip size. Interaction time is the median from the real picker button click until the first file row is actionable.

| Variant | Initial JS gzip | Change | First actionable row | Change |
| --- | ---: | ---: | ---: | ---: |
| Neutral baseline | 1,666.3 KiB | | 11.523 s | |
| Remove force splitting | 1,332.8 KiB | -333.5 KiB (-20.0%) | 10.395 s | -1.128 s (-9.8%) |
| Prefetch root folder | 1,666.4 KiB | +0.1 KiB (+0.0%) | 10.240 s | -1.283 s (-11.1%) |
| Split intent locales | 1,608.5 KiB | -57.8 KiB (-3.5%) | 11.430 s | -0.093 s (-0.8%) |
| Combined, measured directly | 1,275.0 KiB | -391.3 KiB (-23.5%) | 10.675 s | -0.848 s (-7.4%) |
| Remove encrypted-file integration | 1,022.3 KiB | -253.0 KiB (-19.8%) | 16.156 s | inconclusive due to outliers |

Removing the viewer stylesheet reduced the intent's initial CSS from 29,160 B to 27,807 B gzip (-4.6%) without changing its request count. The matched Chromium median changed from 8,141.5 ms to 8,142.4 ms, within measurement noise, and all 13 File Picker end-to-end scenarios passed.

Removing Drive's encrypted-file integration reduced the intent's initial JavaScript from 1,305,624 B to 1,046,846 B gzip (-258,778 B, -19.8%) and reduced the intent asset count from 11 to 10. Runtime JavaScript transferred before interaction and root-folder request counts were unchanged. The startup timing comparison is inconclusive because three of five candidate runs were large outliers. The raw result is in [the encrypted-file removal report](https://github.com/linagora/twake-drive/blob/feat/file-picker-intent-performance/docs/performance/file-picker-encrypted-files-removal-campaign.md).

The cozy-keys-lib package remains a dependency because cozy-harvest-lib imports it at runtime. Removing it from the application dependency graph makes the production build fail.

Timing values are synthetic medians, not typical user timings. The five-run distributions overlap substantially for prefetch, locale splitting, and the combined result, so those timing differences are non-conclusive. The force-splitting timing is directional and is not claimed as a typical user gain. See the [full decision report](https://github.com/linagora/twake-drive/blob/feat/file-picker-intent-performance/docs/performance/file-picker-optimization-decision.md) for raw values and build guardrails.


Crash--:
Nice finding! 

I'm pretty sure we will find other stupid stuff like that. 

Do we need to update bundlemon with this PR? Since we now have dedicated cozy-* bundle for different targets.

src/targets/intents/index.jsx
```diff
@@ -40,12 +41,10 @@ document.addEventListener('DOMContentLoaded', () => {
 
   registerClientPlugins(client)
 
+  const dictRequire = await loadIntentLocales(data.locale)
```
Crash--, line 43: this is something we can do also for other targets, right?

doubleface: I suppose so. But it looks easier in this limitted scope for now

JF-Cozy:
@doubleface we add a lot of stuff to analyze the bundle's weight, did you try and compare with the analyzer we already have in the rsbuild config?

src/targets/intents/index.jsx
Crash--: Why that? This kind of fix should be backported to public / private bundle as well. Should be easy to do. If we don't do it right now, we'll never do it...

scripts/measureBuildAssets.mjs
```diff
@@ -0,0 +1,343 @@
...
+  return filePaths
+}
+
+const measureFiles = filePaths => {
+  return filePaths.reduce(
+    (metrics, filePath) => {
+      const content = fs.readFileSync(filePath)
+      metrics.files += 1
+      metrics.bytes.uncompressed += content.byteLength
+      metrics.bytes.gzip += gzipSync(content, { level: 9 }).byteLength
```
Crash--, line 133: In order to be ultra precise, we should use the same encoding as the stack aka brotli (quality 6)

Crash--:
> @doubleface we add a lot of stuff to analyze the bundle's weight, did you try and compare with the analyzer we already have in the rsbuild config?

@JF-Cozy we're not talking about the same thing. Of course we use few tools from rsbuild to check the bundle size (it's like that we find out for cozy-keys-lib etc), but here we also compress the final results to see the real gain at the end for the end user. 

And we run also some perf tests. 

But I think it can be a good idea @doubleface to move your script to compress to our rsbuild config like that, we can have it for all our apps. It can be great

JF-Cozy:
Ok thanks for the explanation. 

> here we also compress the final results

I thought we already did that with webpack, so also with rsbuild 🤔 maybe a plugin or a param can do the same thing? Sounds strange to have to do it manually

> move your script to compress to our rsbuild config

Agreed, that was my second question...

PR #4061, opened 2026-07-17
feat: hide the team drives tab while it has no content

shepilov:
## What

UX improvement on the Sharings tabs (#3975): the **Team drives** tab is now shown only while it actually has content. Previously it was purely flag-gated, so instances with the shared-drive flags on but no org drive showed a tab opening onto an empty state.

> Deliberate design change: this supersedes #4039's "visible (with empty content) when flags are on and no org drive exists yet" acceptance criterion.

## How

- `useFilteredSharings` exposes `hasDrives`, computed inside its existing memo with the same classifier (`getSharingsTabForEntry`) over the same deduplicated pre-tab-filter list the drives tab renders — so "empty" can never drift from what the tab would show, and there is no duplicate `useSharedDrives` fetch.

Part of #3975

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tab visibility for shared/organization drives based on actual drive availability and filtering results.
  * Kept the drives tab accessible via direct deep links (e.g., `?tab=drives`), even when the current list is empty.

* **Tests**
  * Expanded automated coverage for team drives and drives-tab visibility across scenarios: drives available, unavailable, and disabled via flags.
  * Added assertions for correct behavior when deep-linking to the drives tab.

### DIFF
--- a/src/modules/views/Sharings/index.jsx
+++ b/src/modules/views/Sharings/index.jsx
@@ -22,6 +22,7 @@ import { useFolderViewBase } from '../Folder/hooks/useFolderViewBase'
 import FolderViewBodyVz from '../Folder/virtualized/FolderViewBody'
 
 import useHead from '@/components/useHead'
+import { SHARING_TAB_DRIVES } from '@/constants/config'
 import { useFolderSort } from '@/hooks'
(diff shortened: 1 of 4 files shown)


doubleface (approved):
Just fix ai comment

PR #4063, opened 2026-07-17
perf(file-picker): remove broad cozy force splitting

doubleface:
This will significantly reduce bundle size for file picker intent.
I did not apply it to global rsbuild config because I am not sure of the consequences on other apps.

related to #4057 

## Summary by CodeRabbit

* **Performance**
  * Improved web build chunking to produce more consistent dependency bundles.
  * Updated asset splitting behavior to support more efficient delivery and loading of web application resources.

### DIFF
--- a/rsbuild.config.mjs
+++ b/rsbuild.config.mjs
@@ -31,6 +31,18 @@ const templates = {
 
 config.environments = {
   web: {
+    performance: {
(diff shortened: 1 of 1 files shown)


zatteo (approved):
Let's try!

PR #4066, opened 2026-07-17
perf(file-picker): Prefetch the file picker’s root folder while the intent handshake is pending.

doubleface:
Why

 Reduce the time needed to display the file picker after receiving an intent.

 How

 - Build and start the root-folder query before completing the intent handshake.
 - Wait for the prefetch before rendering the picker.
 - Log prefetch failures and continue rendering normally.

 Testing

 Added tests covering concurrent prefetching and graceful handling of prefetch failures.

related to #4057 

## Summary

* **Bug Fixes**
  * Improved picker startup by preloading the root content folder before rendering the picker for compatible `PICK` intents.
  * Picker now waits for the preload attempt when applicable, and still renders if the preload fails.
  * Faster service readiness by fetching intent details and creating the service concurrently.

* **Tests**
  * Added/updated Jest + React Testing Library tests for preload success, preload failure resilience, and ignoring non-`PICK` intents.

### DIFF
--- a/src/modules/services/components/IntentHandler.jsx
+++ b/src/modules/services/components/IntentHandler.jsx
@@ -9,6 +9,18 @@ import Picker from './Picker'
 
 import { ROOT_DIR_ID } from '@/constants/config'
 
+async function initPicker(client) {
+  const rootFolderQuery = buildContentFolderQuery(ROOT_DIR_ID)
+
+  try {
+    await client.query(rootFolderQuery.definition(), rootFolderQuery.options)
+  } catch (error) {
+    logger.warn('File Picker root prefetch failed', error)
+  }
+
+  return Picker
+}
+
 const IntentHandler = ({ intentId }) => {
   const client = useClient()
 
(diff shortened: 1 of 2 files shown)


src/modules/services/components/IntentHandler.jsx
```diff
@@ -31,6 +42,7 @@ const IntentHandler = ({ intentId }) => {
           intent.attributes.action === 'PICK' &&
           intent.attributes.type === 'io.cozy.files'
         ) {
+          await rootFolderPrefetch
```
zatteo: Maybe to move in a separe initPicker method or something like this?

src/modules/services/components/IntentHandler.jsx
```diff
@@ -23,6 +26,14 @@ const IntentHandler = ({ intentId }) => {
       let service
       let intent
       try {
+        const rootFolderQuery = buildContentFolderQuery(ROOT_DIR_ID)
+        const rootFolderPrefetch = client
+          .query(rootFolderQuery.definition(), rootFolderQuery.options)
+          .catch(error => {
+            logger.warn('File Picker root prefetch failed', error)
+            return null
+          })
```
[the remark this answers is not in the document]
zatteo: Agree 👍

zatteo (approved):

src/modules/services/components/IntentHandler.jsx
```diff
@@ -19,20 +34,21 @@ const IntentHandler = ({ intentId }) => {
...
-        service = await intents.createService(intentId, window)
-        intent = service.getIntent()
-
-        if (
-          intent.attributes.action === 'PICK' &&
-          intent.attributes.type === 'io.cozy.files'
-        ) {
-          component = Picker
-        }
+        const intentPromise = intents.request.get(intentId)
```
zatteo, line 41: Does it change something to do
`const intentPromise = intents.request.get(intentId)`
instead of
`intent = service.getIntent()`

?

doubleface: Yes. service.getIntent() is only available once createService() has resolved, after the handshake with the parent window. Fetching the intent directly lets us identify a file picker and start prefetching its root folder while that handshake is still pending.

The trade-off is an additional GET because createService() also fetches the intent internally.
I will add a comment to make it clearer

PR #4067, opened 2026-07-17
perf(file-picker): Split intent locale loading into a dedicated asynchronous loader

doubleface:
Why

 Prevent all locale bundles from being merged into the intent chunk while preserving English fallback behavior.

 How

 - Dynamically load English and the requested locale.
 - Fall back to English if the requested locale is unavailable.
 - Load locales before rendering the intent handler.
 - Add coverage for supported locales and fallback cases.

We could also apply this to other targets but this will be the object of other PRs

related to #4057 

## Summary by CodeRabbit

* **New Features**
  * Intent translations now load before the interface renders.
  * Added fallback to English when the selected language cannot be loaded.

* **Bug Fixes**
  * Improved reliability of localized intent experiences across supported languages.

* **Tests**
  * Added coverage for locale loading, fallback behavior, and supported translations.

### DIFF
--- a/src/targets/intents/loadIntentLocales.js
+++ b/src/targets/intents/loadIntentLocales.js
@@ -1,0 +1,26 @@
+const FALLBACK_LOCALE = 'en'
+
+async function loadLocale(locale) {
+  // Keep intent locale modules separate from the synchronous locale context
+  // used by the full app, otherwise chunk optimization merges every locale.
+  const localeModule = await import(`@/locales/${locale}.json?intent`)
+  return localeModule.default ?? localeModule
+}
+
+export async function loadIntentLocales(locale, loadDictionary = loadLocale) {
+  const fallbackDictionaryPromise = loadDictionary(FALLBACK_LOCALE)
+  if (locale === FALLBACK_LOCALE) {
+    const fallbackDictionary = await fallbackDictionaryPromise
+    return () => fallbackDictionary
+  }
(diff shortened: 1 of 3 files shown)


src/targets/intents/loadIntentLocales.js
```diff
@@ -0,0 +1,26 @@
+const FALLBACK_LOCALE = 'en'
+
+async function loadLocale(locale) {
+  // Keep intent locale modules separate from the synchronous locale context
+  // used by the full app, otherwise chunk optimization merges every locale.
+  const localeModule = await import(`@/locales/${locale}.json?intent`)
```
zatteo, line 6: What does `?intent` do?

zatteo (approved):
Looks like a good catch too! Should helpers be implement in twake-i18n?

src/targets/intents/loadIntentLocales.js
doubleface: ?intent is an arbitrary Webpack resource query. It does not change the JSON content, but makes Webpack treat these imports as a separate module context from the locale imports used by
 the full app. Without it, chunk optimization merges both contexts and pulls every locale into the intent chunk.

doubleface:
> Looks like a good catch too! Should helpers be implement in twake-i18n?

Not sure of the constraints on larger scope than file picker but we should study this yes

PR #4068, opened 2026-07-20
feat: Send readyToUse message when File Picker is ready

zatteo:
Requires https://github.com/linagora/cozy-libs/pull/3085

## Summary by CodeRabbit

* **New Features**
  * Added a `readyToUse` lifecycle signal to the File Picker.
  * The signal fires once after the picker UI renders and the root folder finishes loading, including when the initial root load errors.
  * Routed the signal through the picker stack so it’s triggered at the correct time.
  * Updated documentation to reflect the new lifecycle behavior.
* **Chores**
  * Updated dependency versions for improved UI and integration compatibility.


src/modules/services/components/FilePicker/FilePickerBody.jsx
```diff
@@ -44,11 +45,20 @@ const FilePickerBody = ({
...
   } = useQuery(contentFolderQuery.definition, contentFolderQuery.options)
   const items = contentFolder || []
 
+  const onReadyToUseRef = useRef(false)
+
+  useEffect(() => {
+    if (isLoading || onReadyToUseRef.current) return
+    onReadyToUseRef.current = true
+    onReadyToUse?.()
+  }, [isLoading, onReadyToUse])
```
zatteo, line 57: @doubleface your thoughts on the best place to send this message?

doubleface: Looks like it is the right place to send this message

doubleface: But I would only consider moving the “send once” guard to IntentHandler, since the contract is once per intent rather than once per FilePickerBody mount.

zatteo: Agree 👍

doubleface (approved):

PR #4072, opened 2026-07-20
fix(sharings): default to the with-me tab

shepilov:
## Summary

- Route sharing navigation and revoke redirects to the With me tab.
- Hide the Team drives tab when no organizational drives are available.
- Canonicalize unavailable Team drives deep links without adding browser history.

## Summary by CodeRabbit

- **Bug Fixes**
  - Sharing and revoke-success redirects now consistently return users to the canonical **“With me”** sharing tab.
  - Deep-linked **“Team drives”** URLs are automatically rewritten to **“With me”** when team drives aren’t available.
  - Canonicalization uses history replacement to avoid stale or stuck tab states.

- **Navigation**
  - Sharing links and routes now use a shared default destination that includes the correct `tab` query parameter.
  - Tab switching supports optional history replacement to preserve canonical behavior.

### DIFF
--- a/src/modules/views/Modal/revokeSuccessRedirect.js
+++ b/src/modules/views/Modal/revokeSuccessRedirect.js
@@ -1,5 +1,0 @@
-import { DEFAULT_SHARINGS_VIEW_ROUTE } from '@/constants/config'
-
-const makeRevokeSuccessRedirectPath = () => DEFAULT_SHARINGS_VIEW_ROUTE
-
-export { makeRevokeSuccessRedirectPath }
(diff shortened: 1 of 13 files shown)


src/modules/views/Modal/revokeSuccessRedirect.js
```diff
@@ -1,11 +1,5 @@
...
+import { DEFAULT_SHARINGS_VIEW_ROUTE } from '@/constants/config'
 
-function makeRevokeSuccessRedirectPath({ document, driveId } = {}) {
-  if (driveId || document?.driveId) {
-    return `${SHARINGS_VIEW_ROUTE}?tab=${SHARING_TAB_DRIVES}`
-  }
-
-  return SHARINGS_VIEW_ROUTE
-}
+const makeRevokeSuccessRedirectPath = () => DEFAULT_SHARINGS_VIEW_ROUTE
```
doubleface: nit: Since this now leads to a constant, we should remove remove this and import DEFAULT_SHARINGS_VIEW_ROUTE directly in other files

doubleface (approved):

doubleface (approved):
If the e2e tests pass

PR #4080, opened 2026-07-22
fix(file-picker): render content in caller dialog

doubleface:
## What

Remove the FilePicker's internal dialog and render its content directly inside
the dialog provided by the calling application.

## Why

`IntentDialogOpener` already owns the modal containing the intent iframe. The
additional `FixedDialog` created nested dialogs with competing sizing and
closing behavior.

## How

- Replace `FixedDialog` with a full-size flex layout for the header, content,
  and footer.
- Preserve the previous dialog spacing and dividers.
- Remove the internal `onClose` and `service.cancel()` flow.
- Keep `LinkAccessModal` as a secondary modal over the picker.
- Update tests for the new component structure and ownership model.

## Summary by CodeRabbit

* **UI Improvements**
  * Updated the file picker to use a simpler layout with distinct header, body, and footer regions.
  * Adjusted component structure and rendering to match the new layout.
* **Bug Fixes**
  * Removed outdated close controls and related styling from the file picker.
  * Closing the picker no longer triggers service cancellation.
  * Preserved existing file selection and link-generation/termination behavior.
* **Tests**
  * Updated tests and mocks to use the new file picker test IDs.
  * Added assertions for header, body, and footer rendering.

### DIFF
--- a/src/modules/services/components/FilePicker/index.jsx
+++ b/src/modules/services/components/FilePicker/index.jsx
@@ -1,7 +1,8 @@
 import PropTypes from 'prop-types'
 import React, { useState, memo, useMemo } from 'react'
 
-import { FixedDialog } from 'cozy-ui/transpiled/react/CozyDialogs'
+import Box from 'cozy-ui/transpiled/react/Box'
+import Divider from 'cozy-ui/transpiled/react/Divider'
 import { useAlert } from 'cozy-ui/transpiled/react/providers/Alert'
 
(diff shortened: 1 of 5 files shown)


zatteo:
I think we still need a close button in the header? Is it here?

image

doubleface:
> I think we still need a close button in the header? Is it here?

We still have got the one from the intent. This screen capture is with my branch
image

zatteo (approved):

PR #4085, opened 2026-07-22
feat(file-picker): add reference selection mode

doubleface:
Allow PICK intents to return plain file references without generating sharing or download links. Keep the action opt-in, support folder-only constraints and multiple selection, and document the payload for callers.


src/modules/services/components/FilePicker/constraints.js
```diff
@@ -57,6 +59,13 @@ export const getActionDisabledState = (actionConfig, selectedItems) => {
     }
 
     if (selectedItem && fileModel.isFile(selectedItem)) {
+      if (actionConfig.onlyFolder === true) {
+        return {
+          disabled: true,
+          reasonKey: 'FilePicker.constraints.disabledReasons.fileNotAllowed'
+        }
+      }
```
Crash--, line 67: Why can't we use the "allowedMimeTypes" parameters? 

I don't understand the needs of `allowFolder` & `onlyFolder` .

docs/file-picker-intent.md
```diff
@@ -36,6 +36,10 @@ It is not a top-level `actions` field.
       "label": "Attach file",
       "maxFileSize": 52428800,
       "allowedMimeTypes": ["image/*", "application/pdf"]
+    },
+    "reference": {
+      "label": "Select",
+      "onlyFolder": true
     }
```
Crash--, line 44: what do you call reference? Only the `io.cozy.files` document? 

I don't like the naming, but I've nothing yet to suggest.

docs/file-picker-intent.md
```diff
@@ -62,6 +66,8 @@ interface FilePickerConfig {
    * Omit to use defaults. Set to null to hide the action.
    */
   downloadLink?: ActionConfig | null
+  /** Plain reference action. Hidden unless explicitly configured. */
```
Crash--: can you use the same comment as the other. `* Omit to use defaults. Set to null to hide the action.`

src/modules/services/components/FilePicker/FilePickerFooter.jsx
```diff
@@ -98,13 +103,23 @@ const FilePickerFooter = ({
...
+        <span className="u-ml-1">
+          {renderAction(
+            downloadLinkLabel,
+            downloadLinkState,
+            downloadLinkAction,
+            () => onConfirm(filePickerLinkModes.TEMPORARY_DOWNLOAD_LINK),
+            'temporary-download-link-btn',
+            Attachment
+          )}
+        </span>
```
Crash--: why are we duplicating everything. Please create a "registry" with all the actions and map on them.

src/modules/services/components/FilePicker/index.jsx
```diff
@@ -133,6 +137,8 @@ const FilePicker = ({
             downloadLinkState={downloadLinkState}
             publicLinkAction={publicLinkAction}
             downloadLinkAction={downloadLinkAction}
+            referenceState={referenceState}
+            referenceAction={referenceAction}
```
Crash--: same here, we should not duplicate everything. We should use a registry and map on it.

docs/file-picker-intent.md
```diff
@@ -199,10 +214,12 @@ interface FilePickerEntry {
   mimeType: string | null
   sharingLink?: string
   downloadLink?: string
+  type?: string
+  doctype?: 'io.cozy.files'
```
Crash--: why it's not the full FileDoctype Type from cozy-client?

src/modules/services/components/FilePicker/payload.js
```diff
@@ -8,22 +8,24 @@ const { file: fileModel } = models
...
 ) => ({
   id: file._id,
   name: file.name,
   size: fileModel.isFile(file) ? parseInt(file.size, 10) || 0 : 0,
   mimeType: fileModel.isFile(file) ? file.mime : null,
   ...(sharingLink ? { sharingLink } : {}),
-  ...(downloadLink ? { downloadLink } : {})
+  ...(downloadLink ? { downloadLink } : {}),
+  ...(reference ? { type: file.type, doctype: 'io.cozy.files' } : {})
 })
```
Crash--, line 32: if we ask a "reference" we should return the reference. So the complete `io.cozy.files`. 

I'm wondering if we should not do the same for the other actions. 

It should be caller responsibility to ask only specific fields.

zatteo: What we really want to achieve with reference is to **not create a sharing or download link** right @doubleface ?

zatteo: > I'm wondering if we should not do the same for the other actions.

Good question, should we always return the full io.cozy.files and in some case "more" like a download link or a sharing link?

Crash--: I think yes. And if the caller do not need to handle the full io.cozy.files, then it can specified the desired attributes (because for exemple, a note is a big io.cozy.files...). No?

src/modules/services/components/FilePicker/constraints.js
doubleface: Mime type only apply to files and not folders then I think we will need a specific option anyway. Or should we create our own custom mime type for folders ? I don't think so

doubleface: This is the adaptation from another branch from @paultranvan who did not have all the context. That is why this PR is still in draft. There is still some work to make it more compliant with the spec. I actually think onlyFolder or equivalent should not be only in reference section.

docs/file-picker-intent.md
doubleface: What we call a reference is "doctype + _id". Since the file picker will on return files or folder, I suppose we could remove the doctype part.

Maybe we could replace the `reference` with `identifier` ?

src/modules/services/components/FilePicker/constraints.js
Crash--: We can have : `image/**` why not `folder/**` or `io.cozy.files:folder` or things like that? 

Bu we should not add more attributes.

docs/file-picker-intent.md
Crash--: I don't like `identifier` specially because I've that in mind: https://github.com/linagora/twake-drive/pull/4085#discussion_r3636453165

src/modules/services/components/FilePicker/constraints.js
Crash--: Instead of adding onlyFolder (and keeping allowFolder), I'd rather collapse the whole "what can be selected" constraint into a single field. file and folder become symmetric tokens, and mime globs narrow within files. Two options:

Option A - flat accept list
```
"accept": ["folder"]                     // folders only
"accept": ["file"]                       // any file
"accept": ["image/*", "application/pdf"] // images + PDF only
"accept": ["folder", "image/*"]          // folders or images
```
Rule: a token without / matches the doctype (file/folder); a token with / is a mime glob (implies a file).

Option B - structured match list
```
"match": [{ "type": "folder" }, { "mime": "image/*" }]
```
More verbose, but self-documenting and extensible to other fields later (class, tags, size).

In both cases it's an OR across entries (item allowed if it matches any), and this replaces allowFolder + onlyFolder + allowedMimeTypes with one field. Absent = ["file", "folder"] (permissive).


@zatteo If we were only for stack, I would go for Option B. It's way more flexible & extensible. But with openburo in mind, maybe we should take Option A? What is your opinion? And in this case, it could still be acceptedMimeTypes

zatteo: It is a good point that:
- we don't have any mimetype for folders in cozy-stack
- but the client of the file picker don't care

`class` or `tags` or `favorite` are internal drive metadata. So indeed, to stay close to what openburo could look like,  we should go to option A that does not expose any internal and stay simple.

For `class` or `tags` or `favorite`, it is maybe more a filter in the file picker itself that could help, like a "favorite" checkbox or tab in the file picker itself if we want to quickly see favorite. I don't see any usecase where we want to _force_ the user to select a favorite.

About size, we have the _max size for a file_ but also the _max size for all files_ that is a mail requirement. The _max size for all files_ would be complicated to insert in this _match_ so we still need size management outside of this.

So I would go to A right now. In anycase I like the `accept` name that is explicit & generic.

doubleface: `accept` option avoid the confusion with mime type and is really clear. And since we have to avoid to expose any internal, option A looks good.

> For `class` or `tags` or `favorite`, it is maybe more a filter in the file picker itself that could help, like a "favorite" checkbox or tab in the file picker itself if we want to quickly see favorite. I don't see any usecase where we want to _force_ the user to select a favorite.

Since we will have a tab for recents for example, we could have a tab for favorites also. And those filters look more like something to help the user to find the files he wants to pick.

If we agree, I could update the spec

src/modules/services/components/FilePicker/payload.js
doubleface: > What we really want to achieve with reference is to **not create a sharing or download link** right @doubleface ?

yes

docs/file-picker-intent.md
doubleface: Good point. Then, if we return the full document, we could replace `reference` simply with `documents` (since we may return multiple documents) ?

docs/file-picker-intent.md
doubleface: In this case, and if we want to return the full document, we could call it `documents` ?

shepilov: if it's identifier with type, it can be doc as everywhere in stack, json, data, obj if we want to return the whole json doc

src/modules/services/components/FilePicker/constraints.js
shepilov: [application/x-directory](https://mimeapplication.net/x-directory) if it help somehow

src/modules/services/components/FilePicker/payload.js
Crash--: I'm wondering if we should not use this "reference" mode if there is any action set by the caller. Maybe it should be the default one.

zatteo: > I'm wondering if we should not use this "reference" mode if there is any action set by the caller. Maybe it should be the default one.

Agree!

doubleface: I agree. The default action should be the one with no side effect

zatteo:
I like it a lot with this new specification. Even if sometimes the io.cozy.files document can return a lot of data, e.g. with a cozy-notes, I consider this as a feature for now. We will see in the future if sometimes we send too much data but there is no reason otherwise to not send the entire io.cozy.files document.

docs/file-picker-intent.md
```diff
@@ -2,27 +2,53 @@
...
+- In raw intent attributes, place it in `attributes.data`.
 
-It is not a top-level `actions` field.
+A minimal request can omit `multiple` and every action:
+
+```json
+{
+  "action": "PICK",
+  "type": "io.cozy.files",
+  "permissions": ["GET"],
```
Crash--: do we really need to set this permissions?

docs/file-picker-intent.md
```diff
@@ -31,11 +57,14 @@ It is not a top-level `actions` field.
   "permissions": ["GET"],
   "data": {
     "multiple": false,
-    "sharingLink": { "label": "Share as link" },
+    "doc": {
+      "label": "Select",
+      "accept": ["folder"]
```
Crash--: so no x-directory?

docs/file-picker-intent.md
```diff
@@ -46,177 +75,238 @@ It is not a top-level `actions` field.
...
-When no config is provided, Drive uses:
+If none of `doc`, `sharingLink`, or `downloadLink` is present, Drive enables
+only `doc` with its defaults:
 
 ```js
 {
   multiple: true,
-  sharingLink: { allowFolder: true },
-  downloadLink: { allowFolder: false }
+  doc: { accept: ['file', 'folder'] },
```
Crash--: file is really handled?

docs/file-picker-intent.md
```diff
@@ -46,177 +75,238 @@ It is not a top-level `actions` field.
...
-  availableSize?: number
+  availableSize?: number;
 }
 ```
 
-## Defaults
+## Action resolution and defaults
 
-When no config is provided, Drive uses:
+If none of `doc`, `sharingLink`, or `downloadLink` is present, Drive enables
```
Crash--: if their is any action present, Drive enables... 

Because we'll add other actions in the futur

docs/file-picker-intent.md
zatteo: I am skeptical about application/x-directory as it is non standard and we still need a custom value for "all files" so it does not solve entirely our problem. 

Or maybe creating our own ones?
- application/vnd.twake.file => means all files
- application/vnd.twake.folder => means all folders 

It fit in mimetype mental model and allow custom needs.

docs/file-picker-intent.md
doubleface: Actually no (after verification). Let's remove this from the examples

docs/file-picker-intent.md
doubleface: No, I juste updated this documentation to reflect the changes in the notion spec. This is still to be implemented. But before, I want to make sure we agree on that

PR #4090, opened 2026-07-23
feat(sharings): move tabs into routes

shepilov:
## Summary

- Move Sharings tab state from query parameters into nested routes.
- Redirect pre-tab Sharings URLs to the default With me route.
- Preserve active tabs across folders, viewers, actions, shared drives, and editor return flows.
- Remove query-tab propagation and update route coverage.

## Summary by CodeRabbit

* **New Features**
  * Updated Sharings navigation to use dedicated tab-scoped routes for “With me,” “By me,” and shared drives (URL path segments instead of tab query params).
  * Added automatic legacy URL redirection to canonical Sharings routes.
* **Bug Fixes**
  * Improved Share/Revoke/Close navigation to return to the correct active tab and location, including shared-drive and nested views.
  * Preserved existing search/filter query parameters during Sharings navigation.
  * Refined empty-state rendering and active tab highlighting on nested Sharings pages.
* **Tests**
  * Updated and added unit and e2e coverage for the new routing and redirect behavior.


src/modules/views/Sharings/useSharingsTab.ts
```diff
@@ -39,39 +51,42 @@ export const areDrivesAvailable = (): boolean =>
...
- * unrecognized (including the legacy numeric `?tab=1`) or unavailable values
- * resolve to the default "With me" tab, and the URL is canonicalized with a
- * `replace` navigation so the back button is never trapped. User tab switches
- * push a history entry by default; callers can request `replace` for later
- * canonicalization once asynchronously loaded tab availability is known.
- */
-export const useSharingsTab = (): [SharingsTab, SetSharingsTab] => {
-  const [searchParams, setSearchParams] = useSearchParams()
-  const drivesAvailable = areDrivesAvailable()
+const getSearchWithoutLegacyTab = (search: string): string => {
```
Crash--: is it really needed? We never deploy to production the old code?

src/modules/navigation/AppRoute.jsx
```diff
@@ -81,6 +85,48 @@ const OutletWrapper = ({ Component }) => (
...
+    <Route path="folder/:folderId" element={<SharingsFolderView />}>
+      <Route path="file/:fileId" element={<SharingsFilesViewer />} />
+      <Route path="file/:fileId/revision" element={<FileHistory />} />
+      <Route path="file/:fileId/share" element={<ShareFileView />} />
+      <Route path="file/:fileId/qualify" element={<QualifyFileView />} />
+      <Route path="share" element={<ShareDisplayedFolderView />} />
+    </Route>
+    <Route path="move" element={<MoveFilesView />} />
+  </Route>
+)
```
Crash--, line 130: In an other PR we should mutualize all these routes.

src/modules/views/Sharings/useSharingsTab.ts
```diff
@@ -39,39 +51,42 @@ export const areDrivesAvailable = (): boolean =>
...
   useEffect(() => {
-    if (paramValue !== tab) {
-      setSearchParams(prev => withTab(prev, tab), { replace: true })
+    if (activeTab !== tab) {
+      navigate(
+        {
+          pathname: getSharingsTabRoute(activeTab),
+          search
+        },
+        { replace: true }
```
Crash--, line 76: why do we replace here?

src/components/EditorToolbar/Sharing.jsx
```diff
@@ -1,26 +1,34 @@
...
+  const { search } = useLocation()
 
   const toggleShareModal = useCallback(
     () => setShowShareModal(v => !v),
     [setShowShareModal]
   )
 
   const handleRevokeSuccess = () => {
-    navigate(DEFAULT_SHARINGS_VIEW_ROUTE, { replace: true })
+    const redirectLink = new URLSearchParams(search).get('redirectLink')
```
Crash--, line 25: why do we need this redirectLink? Where do you set it?

Crash--: ok it's for note...

src/modules/navigation/AppRoute.jsx
Crash--: it has been done a little bit in a next commit.

src/modules/navigation/AppRoute.jsx
```diff
@@ -81,6 +86,52 @@ const OutletWrapper = ({ Component }) => (
...
+        : null}
+    </Route>
+    <Route path="folder/:folderId" element={<SharingsFolderView />}>
+      <Route path="file/:fileId" element={<SharingsFilesViewer />} />
+      <Route path="file/:fileId/revision" element={<FileHistory />} />
+      <Route path="file/:fileId/share" element={<ShareFileView />} />
+      <Route path="file/:fileId/qualify" element={<QualifyFileView />} />
+      <Route path="share" element={<ShareDisplayedFolderView />} />
+    </Route>
+    {flag('drive.shared-drive.enabled') ||
```
doubleface: nit: We have these two flags 3 times. Let's create sharedDrivesEnabled boolean

src/modules/views/Sharings/useSharingsTab.ts
shepilov: For example, team drives are disabled, but the user opens `/sharings/drives` and without replace we will have it in history

doubleface:
You need to update the routes in z-share-modal-surfaces.spec.ts e2e tests

doubleface:
> You need to update the routes in z-share-modal-surfaces.spec.ts e2e tests

You did it in the interval, sorry

doubleface (approved):
I'm not so comfortable with this routing stuff. But it looks like it works

PR #4099, opened 2026-07-24
fix: use editor author for onlyoffice shared folder username

### LINKED ISSUE #3831: Shared folder Only Office integration
Alice creates a shared folder with Bob. Alice creates an OO file in this shared folder. 

If Bob open this OO file, OO should not ask him to set a username. It should be automatically set to the public_name of the user. 

Maybe it's related to https://github.com/linagora/twake-drive/pull/3823 but I don't know if the username is will set with the shared folder. 

Image

Crash--:
## Summary

- Resolve the OnlyOffice editor username through `useEditorAuthor`, the shared source already used by the PDF and Excalidraw editors, instead of the `public_name` from the `/office/:id/open` response. That value is empty for a file inside a shared folder, so the editor prompted the user to type a name.
- Wait for the author to resolve before building the editor config, since OnlyOffice reads it at mount.
- Remove the now unused `makeName` helper.

Fixes #3831

## Summary by CodeRabbit

* **Bug Fixes**
  * OnlyOffice now shows the correct editor author name, even when public profile details are missing.
  * Editor configuration now waits for author information to finish loading before opening.
  * Adjusted sharing banner display behavior to better match navigation context.
* **Tests**
  * Added/updated test coverage to ensure author-loading gating and correct username mapping in OnlyOffice.

### DIFF
--- a/src/modules/views/OnlyOffice/Editor.spec.jsx
+++ b/src/modules/views/OnlyOffice/Editor.spec.jsx
@@ -23,6 +23,10 @@ jest.mock('modules/views/OnlyOffice/helpers', () => ({
   isOfficeEnabled: jest.fn()
 }))
 
+jest.mock('@/modules/views/editor/useEditorAuthor', () => ({
+  useEditorAuthor: jest.fn(() => ({ author: 'Bob', isLoading: false }))
+}))
+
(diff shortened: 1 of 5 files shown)


Crash--:
@zatteo it's a gift. Bug fixed, IA generated code, test added and since I'm removing some lines ;)

zatteo (approved):

PR #4100, opened 2026-07-27
feat(file-picker): Support intent themes

doubleface:
Expose light and dark  through the File Picker intent config. cozy-ui-plus uses the caller's single options.theme value for the host dialog, close button and spinner, then forwards it to the service.

Drive must apply that forwarded value again inside the service document because React context and CSS do not cross the iframe boundary. This keeps the API to one option while theming both sides without changing the instance preference.

[Capture vidéo du 2026-07-28 11-43-16.webm](https://github.com/user-attachments/assets/dd72fcc5-69d1-42e2-b348-d31cbb667e7d)

## Summary by CodeRabbit

* **New Features**
  * Added File Picker theme selection with `auto`, `light`, and `dark`.
  * File Picker dialogs and embedded picker content now render with the chosen theme consistently.
* **Documentation**
  * Updated File Picker theming documentation, including `theme` options, defaults, and how `auto` behaves.
* **Bug Fixes**
  * Improved reliability of the “selected file missing” error UI validation by asserting the exact error text.
* **Tests**
  * Added end-to-end coverage for dark-theme rendering and iframe/dialog alignment.

### DIFF
--- a/src/modules/navigation/components/FilePickerButton.jsx
+++ b/src/modules/navigation/components/FilePickerButton.jsx
@@ -10,6 +10,7 @@ import IntentDialogOpener from 'cozy-ui-plus/dist/Intent/IntentDialogOpener'
 import { useI18n } from 'twake-i18n'
 
 import logger from '@/lib/logger'
+import { filePickerThemes } from '@/modules/services/components/FilePicker/constants'
 
 const PICKER_CONFIGS = [
   {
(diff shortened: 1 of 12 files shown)


src/modules/navigation/components/FilePickerButton.jsx
```diff
@@ -133,6 +138,20 @@ export const FilePickerButton = () => {
           />
         ))}
       </RadioGroup>
+      <RadioGroup
```
shepilov, line 141: why do we need this radio? we can't just set a theme ?

doubleface: he radio group is only for the File Picker demo page. It lets us manually test the three supported behaviors: auto, light, and dark. Calling applications do not need this UI; they simply pass the desired theme through the intent options, or omit it to use the default behavior.

zatteo (approved):

zatteo:
Great!

PR #4103, opened 2026-07-27
feat(filters): add standalone filter select

shepilov:
## Summary

- Add a reusable controlled filter select built on Cozy TextField.
- Use Cozy responsive select behavior for desktop menus and mobile bottom sheets.
- Support active, disabled, optional desktop leading content, and intrinsic auto-width while keeping 32px desktop geometry.

### DIFF
--- a/src/components/Filters/Filter.jsx
+++ b/src/components/Filters/Filter.jsx
@@ -1,0 +1,158 @@
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import React, { forwardRef } from 'react'
+
+import TextField from 'cozy-ui/transpiled/react/TextField'
+import { makeStyles } from 'cozy-ui/transpiled/react/styles'
+
+import styles from './styles.styl'
+
+const FILTER_HEIGHT = 32
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    height: FILTER_HEIGHT,
+    alignItems: 'center',
+    borderRadius: theme.shape.borderRadius,
+    cursor: 'pointer',
+    backgroundColor: theme.palette.background.contrast,
+    '&:hover, &$focused': {
+      backgroundColor: theme.palette.background.contrast
+    },
+    '&:before, &:after': {
+      display: 'none'
+    },
+    '&$disabled': {
+      cursor: 'default'
+    }
+  },
+  active: {
+    '& .MuiFilledInput-root': {
+      backgroundColor: theme.palette.primary.light,
+      '&:hover, &.Mui-focused': {
+        backgroundColor: theme.palette.primary.light
+      }
+    }
+  },
+  focused: {},
+  disabled: {},
+  selectIcon: {
+    top: '50%',
+    transform: 'translateY(-50%)'
+  },
+  selectIconOpen: {
+    transform: 'translateY(-50%) rotate(180deg)'
+  },
+  input: {
+    boxSizing: 'border-box',
+    height: FILTER_HEIGHT,
+    minWidth: 0,
+    ...theme.typography.body2,
+    lineHeight: `${FILTER_HEIGHT}px`,
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+    textAlign: 'left',
+    cursor: 'inherit',
+    '&:focus': {
+      backgroundColor: 'transparent'
+    },
+    '&&': {
+      padding: theme.spacing(0, 4, 0, 2)
+    }
+  },
+  value: {
+    display: 'flex',
+    alignItems: 'center',
+    minWidth: 0,
+    gap: theme.spacing(1)
+  }
+}))
+
+const Filter = forwardRef(function Filter(
+  {
+    active = false,
+    autoWidth = false,
+    children,
+    className,
+    disabled = false,
+    label,
+    onChange = null,
+    options,
+    startAdornment = null,
+    value = '',
+    ...textFieldProps
+  },
+  ref
+) {
+  const classes = useStyles()
+
+  return (
+    <TextField
+      {...textFieldProps}
+      ref={ref}
+      className={cx(
+        {
+          [classes.active]: active,
+          [styles.autoWidth]: autoWidth
+        },
+        className
+      )}
+      disabled={disabled}
+      fullWidth={!autoWidth}
+      hiddenLabel
+      options={options}
+      onChange={onChange}
+      select
+      size="small"
+      value={value}
+      variant="filled"
+      InputProps={{
+        classes: {
+          root: classes.root,
+          focused: classes.focused,
+          disabled: classes.disabled,
+          input: classes.input
+        }
+      }}
+      SelectProps={{
+        classes: {
+          icon: classes.selectIcon,
+          iconOpen: classes.selectIconOpen
+        },
+        displayEmpty: true,
+        renderValue: () => (
+          <span className={classes.value}>
+            {startAdornment}
+            <span className="u-ellipsis">{label}</span>
+          </span>
+        )
+      }}
+    >
+      {children}
+    </TextField>
+  )
+})
+
+Filter.displayName = 'Filter'
+
+Filter.propTypes = {
+  active: PropTypes.bool,
+  autoWidth: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  label: PropTypes.string.isRequired,
+  onChange: PropTypes.func,
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+        .isRequired
+    })
+  ).isRequired,
+  startAdornment: PropTypes.node,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+}
+
+export { Filter }
--- a/src/components/Filters/index.js
+++ b/src/components/Filters/index.js
@@ -1,0 +1,1 @@
+export { Filter } from './Filter'
--- a/src/components/Filters/styles.styl
+++ b/src/components/Filters/styles.styl
@@ -1,0 +1,3 @@
+.autoWidth
+    width fit-content
+    flex-shrink 1


zatteo:
Clicking on the icon of a filter should also trigger it?
Also it seems there is an issue with the active state.

https://github.com/user-attachments/assets/d2e3a844-5e46-486a-8129-7d117c013ef2

src/components/Filters/ClearFiltersButton.spec.jsx
zatteo, line 1: I am not sure these can of tests have any interest 🤷

zatteo:
cc @JF-Cozy for a cozy-ui check (to check if there is nothing closer to what you try to do in cozy-ui)

src/components/Filters/ClearFiltersButton.jsx
JF-Cozy, line 1: we can remove this component, it's just a button with text variant and a startIcon, no need to add one more abstraction layer for this

src/components/Filters/ClearFiltersButton.spec.jsx
JF-Cozy: agreed, we can remove it

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,162 @@
...
+        widthClasses.textField,
+        { [classes.autoWidth]: autoWidth },
+        className
+      )}
+      disabled={disabled}
+      hiddenLabel
+      variant="filled"
+      InputProps={{
+        inputComponent: 'button',
+        classes: {
```
JF-Cozy, line 111: I think we can simplify somehow without being that explicit by just spread `classes` from useStyles I suppose 🤔 and use conditional class in the first level and not inside an other one

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,162 @@
...
+    label,
+    maxWidth = null,
+    minWidth = null,
+    startAdornment = null,
+    ...inputProps
+  },
+  ref
+) {
+  const classes = useStyles()
+  const widthClasses = useWidthStyles({ maxWidth, minWidth })
```
JF-Cozy: `const classes = useStyles({ maxWidth, minWidth })` should be enough, I suppose, somehow (with root maybe)?. You can also use u-classes conditionally

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,162 @@
...
+const Filter = forwardRef(function Filter(
+  {
+    active = false,
+    autoWidth = false,
+    className,
+    disabled = false,
+    endAdornment = null,
+    label,
+    maxWidth = null,
+    minWidth = null,
```
JF-Cozy: it's strange to have maxWidth/minWidth in the API, we generally deal with that in className or even classes for Mui comp

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,162 @@
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import React, { forwardRef } from 'react'
+
+import InputAdornment from 'cozy-ui/transpiled/react/InputAdornment'
+import TextField from 'cozy-ui/transpiled/react/TextField'
+import { makeStyles } from 'cozy-ui/transpiled/react/styles'
+
+const useStyles = makeStyles(theme => ({
```
JF-Cozy, line 12: lot of style here, we sure all is necessary? We can remove borderRiadus and size for exemple and use TextField size small

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,162 @@
...
+    <TextField
+      ref={ref}
+      className={cx(
+        classes.textField,
+        widthClasses.textField,
+        { [classes.autoWidth]: autoWidth },
+        className
+      )}
+      disabled={disabled}
+      hiddenLabel
```
JF-Cozy, line 103: select=true should be here too, isn't it?

src/components/Filters/Filter.spec.jsx
JF-Cozy, line 1: we can remove this, no need to test React or Mui comp. The style override should be tested with screenshot, but we don't have this mechanic in Drive so let's forget it for now.

src/modules/views/FiltersDemo/FiltersDemo.jsx
```diff
@@ -0,0 +1,140 @@
...
+  },
+  clearFiltersButton: {
+    height: '32px',
+    minHeight: '32px',
+    padding: theme.spacing(0.5, 1),
+    borderRadius: theme.shape.borderRadius
+  }
+}))
+
+function FiltersDemo() {
```
JF-Cozy: please use const FiltersDemo = () => {} (everywhere) for consistency :)

src/modules/views/FiltersDemo/FileTypeMenuDemo.jsx
```diff
@@ -0,0 +1,134 @@
...
+        ariaLabelPrefix="Default example"
+        isTypeMenuOpen={isMenuOpen}
+        onClearFilter={handleClearType}
+        onTypeClick={handleToggleMenu}
+        selectedValues={selectedOption ? { type: selectedOption.label } : {}}
+        typeFilterRef={setAnchorElement}
+        typeMenuId={TYPE_MENU_ID}
+      />
+
+      {anchorElement && isMenuOpen ? (
```
JF-Cozy: no need for this condition, open should not be forced to true

src/modules/views/FiltersDemo/FileTypeMenuDemo.jsx
```diff
@@ -0,0 +1,134 @@
...
+        <ActionsMenu
+          ref={{ current: anchorElement }}
```
JF-Cozy: ref={anchorElement}

src/modules/views/FiltersDemo/FileTypeMenuDemo.jsx
```diff
@@ -0,0 +1,134 @@
...
+          actions={[]}
+          classes={{ paper: classes.menuPaper }}
+          docs={[]}
+          id={TYPE_MENU_ID}
+          onClose={handleCloseMenu}
+          open
```
JF-Cozy: `open={isMenuOpen}`

src/modules/views/FiltersDemo/FilterControlDemo.jsx
```diff
@@ -0,0 +1,125 @@
...
+        aria-controls={ariaControls}
+        aria-expanded={ariaExpanded}
+        aria-haspopup={ariaHaspopup}
+        aria-label={`${ariaLabelPrefix} ${filterLabel} filter`}
+        data-filter={filterName}
+        endAdornment={endAdornment}
+        label={label}
+        maxWidth={maxWidth}
+        minWidth={minWidth}
+        onClick={onClick ? handleClick : null}
```
JF-Cozy: `onClick={handleClick}`

src/modules/views/FiltersDemo/FilterControlDemo.jsx
```diff
@@ -0,0 +1,125 @@
...
+        startAdornment={startAdornment}
+        tabIndex={onClick ? 0 : -1}
```
JF-Cozy: is it necessary?

src/modules/views/FiltersDemo/FiltersDemo.spec.jsx
JF-Cozy, line 1: again here, no relevant test in my opinion

src/modules/views/FiltersDemo/FilterToolbarDemo.jsx
JF-Cozy, line 1: commit should be squashed to the previous one

JF-Cozy:
we can squash all the commit, and better separates things after that. Only one commit will not be easy to review... besides I don't understand why we created a demo?

shepilov:
> we can squash all the commit, and better separates things after that. Only one commit will not be easy to review... besides I don't understand why we created a demo?

because we don't have any story book or anything and I wanted to have a page where it's possible to see how it will look like, I can remove it after adding to sharing page

src/components/Filters/Filter.jsx
shepilov: you mean change the component into MUI `select`?

JF-Cozy:
> > I don't understand why we created a demo?
> 
> because we don't have any story book or anything and I wanted to have a page where it's possible to see how it will look like, I can remove it after adding to sharing page

I think we can drop everything about the demo, you can use it locally to test, but I think it's not necessary to commit it

src/components/Filters/Filter.jsx
JF-Cozy: You can check with Swann but I think we said that a TextField Select will do the trick yes, and it comes with a BottomSheet in mobile. And so you should be able to remove some css here, probably

src/components/Filters/Filter.spec.jsx
JF-Cozy: We still can remove everything here, isn't it?

src/components/Filters/index.js
JF-Cozy, line 1: no need for that, just put here everything from Filter file 👍

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,144 @@
...
+      paddingLeft: 0
+    },
+    '&&.MuiFilledInput-inputAdornedEnd': {
+      paddingRight: 0
+    }
+  })
+}))
+
+/**
+ * Presentational filter trigger built on a Cozy filled TextField.
```
JF-Cozy: I don't think these AI blabla is necessary... it's up to you 🤷‍♂️

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,144 @@
...
+import PropTypes from 'prop-types'
+import React, { forwardRef } from 'react'
+
+import InputAdornment from 'cozy-ui/transpiled/react/InputAdornment'
+import TextField from 'cozy-ui/transpiled/react/TextField'
+import { makeStyles } from 'cozy-ui/transpiled/react/styles'
+
+const FILTER_HEIGHT = 32
+
+const useStyles = makeStyles(theme => ({
```
JF-Cozy, line 12: maybe you can use stylus for autoWidth css class and keep useStyle to overrides Mui Input classes, and then rename surface/trigger into root/input and be able to point children withtout using MuiFilledStuff. To override Mui style, try to be as close as possible of the Mui code https://github.com/mui/material-ui/blob/v4.x/packages/material-ui/src/FilledInput/FilledInput.js

src/components/Filters/index.js
shepilov: I think there will be more then one filter

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+      fullWidth={!autoWidth}
+      hiddenLabel
+      options={options}
+      onChange={onChange}
+      select
+      size="small"
+      value={value}
+      variant="filled"
+      InputProps={{
+        classes: {
```
JF-Cozy, line 111: isolate classes should be better I think, so it means making useStyle for Input and useStyle for Select... but we can go this way for now too

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+      padding: theme.spacing(0, 4, 0, 2)
+    }
+  },
+  value: {
+    display: 'flex',
+    alignItems: 'center',
+    minWidth: 0,
+    gap: theme.spacing(1)
+  },
+  label: {
```
JF-Cozy: we can use u-ellipsis for that

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+  active: {
+    '& .MuiFilledInput-root': {
+      backgroundColor: theme.palette.primary.light,
+      '&:hover, &.Mui-focused': {
+        backgroundColor: theme.palette.primary.light
+      }
+    }
+  },
+  focused: {},
+  disabled: {},
```
JF-Cozy, line 38: why are these necessary? with empty object value

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+    },
+    '&:before, &:after': {
+      display: 'none'
+    },
+    '&$disabled': {
+      cursor: 'default'
+    }
+  },
+  active: {
+    '& .MuiFilledInput-root': {
```
JF-Cozy, line 30: I don't know if pointing .MuiFilledInput stuff will still work in production if all Mui classes are jss built 🤔 we will see 🤷‍♂️

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+const FILTER_HEIGHT = 32
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    boxSizing: 'border-box',
+    height: FILTER_HEIGHT,
+    alignItems: 'center',
+    borderRadius: theme.shape.borderRadius,
+    cursor: 'pointer',
+    backgroundColor: theme.palette.action.hover,
```
JF-Cozy: strange to have `action.hover` color for a default one. Be careful to avoid semantic color just according to there values, the semantic matters

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+import styles from './styles.styl'
+
+const FILTER_HEIGHT = 32
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    boxSizing: 'border-box',
+    height: FILTER_HEIGHT,
+    alignItems: 'center',
+    borderRadius: theme.shape.borderRadius,
```
JF-Cozy, line 16: I thought we were ok with the actual radius 🤔

src/components/Filters/Filter.jsx
```diff
@@ -0,0 +1,164 @@
...
+import TextField from 'cozy-ui/transpiled/react/TextField'
+import { makeStyles } from 'cozy-ui/transpiled/react/styles'
+
+import styles from './styles.styl'
+
+const FILTER_HEIGHT = 32
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    boxSizing: 'border-box',
```
JF-Cozy: is it really necessary?

src/components/Filters/index.js
JF-Cozy: so we can do it later when it will be necessary, but nevermind

JF-Cozy (approved):
I've lost track of what's going on at this point, so, I let you think about my comments if there are relevant or not. Feel free to modify before merging.

src/components/Filters/index.js
shepilov: It will be in the next pr with the FilyTypeFilter

src/components/Filters/Filter.jsx
shepilov: FilledInput only rounds the top corners by default

src/components/Filters/Filter.jsx
shepilov: JSS selectors &$focused and &$disabled don't work without them

PR #4105, opened 2026-07-28
feat(file-picker): pick files on double-click

### LINKED ISSUE #3988: File Picker/ Step 3
Double click on file - generate link (show link perrmission modal)
If file alroady has link generated do not show link permission modal 

**Figma** https://www.figma.com/design/XqLqMINlZw09BdEHI8yLb9/Teammail---1.1?node-id=9918-42568&t=6TUCJTlyYZ0oGDSg-0

doubleface:
Only reuse public links scoped to the picked file so a direct action cannot expose additional files. Open link access when no reusable link exists, while preserving download-only configurations.

fixes #3988 

## Summary by CodeRabbit

- **New Features**
  - Double-clicking a picker file now reuses an existing public sharing link when available, or opens link options when not found.
  - Supports sharing vs download behavior (including temporary download mode) and preserves previously generated sharing links on reopen.
- **Bug Fixes**
  - Double-click now safely handles invalid/unsupported files, prevents concurrent actions, and improves surfaced inline error messaging.
  - Link results are validated more precisely (sharing URL present; download link undefined when not applicable).
- **Tests**
  - Extended E2E and component test coverage for double-click flows, link reuse, multi-select behavior, and confirmation dialog result fields.

### DIFF
--- a/src/modules/services/components/FilePicker/index.jsx
+++ b/src/modules/services/components/FilePicker/index.jsx
@@ -11,6 +11,7 @@ import FilePickerHeader from './FilePickerHeader'
 import { LinkAccessModal } from './LinkAccessModal'
 import {
   defaultFilePickerConfig,
+  filePickerDoubleClickResults,
   filePickerErrorCodes,
   filePickerLinkModes,
   filePickerThemes
@@ -138,7 +139,7 @@ const FilePicker = ({
         }
 
         const result = await onFileDoubleClick(item)
-        if (result === 'open-modal') {
+        if (result === filePickerDoubleClickResults.OPEN_MODAL) {
           handleOpenLinkAccess()
         } else if (result) {
           setError(result)
--- a/src/modules/services/components/FilePicker/index.jsx
+++ b/src/modules/services/components/FilePicker/index.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types'
-import React, { useState, memo, useMemo } from 'react'
+import React, { useState, memo, useMemo, useRef, useCallback } from 'react'
 
 import Box from 'cozy-ui/transpiled/react/Box'
 import Divider from 'cozy-ui/transpiled/react/Divider'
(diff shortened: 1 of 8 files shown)


src/modules/services/components/Picker.jsx
```diff
@@ -113,9 +114,29 @@ const Picker = ({ service, intent, onReadyToUse }) => {
...
+      const result = await fetchExistingSharingLink(client, file, {
+        singleFileOnly: true
+      })
+      if (result.status === 'found') {
+        return handlePick([file], filePickerLinkModes.PUBLIC_LINK, [
+          { documentId: getFileId(file), url: result.url }
+        ])
+      }
+      if (result.status === 'not_found') {
+        return 'open-modal'
```
zatteo: I think we should use constant for all these keys

src/modules/services/components/FilePicker/index.jsx
```diff
@@ -96,6 +99,58 @@ const FilePicker = ({
...
+            [item],
+            filePickerLinkModes.TEMPORARY_DOWNLOAD_LINK
+          )
+          if (pickError) {
+            setError(pickError)
+          }
+          return
+        }
+
+        const result = await onFileDoubleClick(item)
```
zatteo: It is a little strange to have part of the `onFileDoubleClick` that does not call `onFileDoubleClick`

src/modules/services/components/FilePicker/sharing.js
```diff
@@ -20,65 +20,58 @@ const getShortcode = permission => {
...
+    return { status: 'not_found' }
+  }
 
-    if (!sharecode) {
-      return null
-    }
+  const sharecode = getShortcode(existingPermission)
+  if (!sharecode) {
+    return { status: 'no_sharecode' }
+  }
```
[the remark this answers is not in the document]
doubleface: We are not supposed to have multiple links on one file. It is just a guard

src/modules/services/components/Picker.jsx
doubleface: Good point. Done

src/modules/services/components/FilePicker/index.jsx
doubleface: Yes. I'll make onFileDoubleClick the single entry point by passing the selected link mode to it. The parent callback will delegate download mode to handlePick and keep the existing-link lookup for public links.

zatteo (approved):

PR #4108, opened 2026-07-28
feat(filters): add date filter

shepilov:
## Summary

- Add a controlled date filter with Today, Last 7 days, Last month, and This year presets.
- Match the Figma trigger, option rows, icon, and clear state.
- Add English, French, Russian, and Vietnamese translations.
- Cover preset selection and clearing behavior.


src/components/Filters/DateFilter.jsx
```diff
@@ -0,0 +1,95 @@
...
+      autoWidth
+      aria-label={t('filters.date.aria_label')}
+      className={classes.filter}
+      clearLabel={t('filters.date.clear')}
+      label={label}
+      onChange={handleChange}
+      onClear={onClear}
+      options={mobileOptions}
+      startAdornment={<Icon icon={Calendar} size={20} />}
+      value={value ?? ''}
```
zatteo, line 78: At component prop you have `value = null` and here `value ?? ''`, why not `value = ''` at component prop?

zatteo (approved):

src/components/Filters/DateFilter.jsx
shepilov: We need `''` for MUI, and it can be null when we pass a value that can't be found. It will be kind of ok, but with MUI warning in the logs

zatteo: If we can avoid MUI warning it's better 👍

PR #4111, opened 2026-07-29
feat(filters): add contact filter

shepilov:
## Summary

- Add a searchable contact filter for contacts and groups.
- Extend the shared filter to support non-select inputs and custom input adornments.
- Add localized labels and behavioral coverage for searching, result limits, loading, selection, and clearing.

## Summary by CodeRabbit

* **New Features**
  * Added a contact filter with a “me” option, clear/reset text, and avatar-based autosuggest suggestions.
  * Added contact filter translations in English, French, Russian, and Vietnamese.
* **Enhancements**
  * Improved the generic `Filter` component to support optional `InputProps` and configurable selectable behavior.
  * Updated contact filter behavior so a disabled filter won’t open the suggestions list.
* **Styling**
  * Simplified contact filter styling for the dropdown, search input, and suggestion states.
* **Tests**
  * Updated contact filter tests to render avatars from option data and added coverage for the disabled state.


src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  return (
+    <div ref={ref} className={className} id={id} onClick={onClick}>
+      {children}
+    </div>
+  )
+})
+
+ContactFilterMenuSurface.displayName = 'ContactFilterMenuSurface'
+
+function getSuggestionValue(option: ContactFilterOption): string {
```
JF-Cozy: we probably can remove this?

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  return option.label
+}
+
+function makeContactFilterMenuId(): string {
+  contactFilterMenuId += 1
+  return `contact-filter-menu-${contactFilterMenuId}`
+}
+
+function handleSuggestionsFetchRequested(): void {}
```
JF-Cozy: to be removed?

src/components/Filters/ContactFilter.tsx
JF-Cozy, line 1: maybe configure AI bot to create `const someFn = () => {}` for consistency :)

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  onClear: () => void
+  options: ContactFilterOption[]
+  value?: string | null
+}
+
+interface ContactFilterAvatarProps {
+  option: ContactFilterOption
+}
+
+interface ContactFilterMenuSurfaceProps {
```
JF-Cozy: move them in type file?

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  children: ReactNode
+  className?: string
+  id: string
+  onClick?: MouseEventHandler<HTMLDivElement>
+}
+
+let contactFilterMenuId = 0
+
+function ContactFilterAvatar({
```
JF-Cozy: we can extract comp in other files, and func in helpers.js file to keep this one as simple as possible

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+  const [anchorElement, setAnchorElement] = useState<HTMLElement | null>(null)
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [query, setQuery] = useState<string>('')
+  const [menuId] = useState<string>(makeContactFilterMenuId)
+  const selectedOption = options.find(option => option.value === value) ?? null
+  const label = selectedOption?.label ?? t('filters.contact.label')
+  const suggestions = loading ? [] : filterContactFilterOptions(options, query)
+
+  const handleOpen = (): void => {
+    if (!disabled) {
```
JF-Cozy: onClick should not be triggered is disabled is true, isn't it?

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+        </ListItemIcon>
+        <ListItemText
+          primary={primaryLabel}
+          secondary={option.secondaryLabel}
+        />
+      </ListItem>
+    )
+  }
+
+  const renderSuggestionsContainer = ({
```
JF-Cozy: renderSuggestion and renderSuggestionsContainer should be extracted in comp for better readability, in my opinion

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+    containerProps,
+    children
+  }: Autosuggest.RenderSuggestionsContainerParams): ReactElement => {
+    return (
+      <div {...containerProps}>
```
JF-Cozy:
```
if (loading) return {}

return {}
```
will be easier to maintain, in my opinion

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+                icon={Bottom}
+                rotate={isOpen ? 180 : 0}
+                size={16}
+              />
+            </InputAdornment>
+          )
+        }}
+      />
+
+      {isOpen && !disabled ? (
```
JF-Cozy: it's better to set open conditionally than this, to mount the comp in the DOM anyway

src/components/Filters/ContactFilter.tsx
shepilov: If we don't need keyboard navigation - yes

zatteo:
@shepilov why using a new intermediate type ContactFilterOption? As input, you have `io.cozy.contacts` and `io.cozy.contacts.group`?

We have helper and existing components that can help for this kind of UI like `getInitials` that you rewrite here for ContactFilterOption type.

src/components/Filters/ContactFilter.tsx
shepilov: we can rename it to noop or inline, but we need it nor to have exception from lib in console

src/components/Filters/ContactFilter.tsx
shepilov: but it's already configured for AI bots in project conventions to use `function` for top-level functions

src/components/Filters/ContactFilter.tsx
shepilov: does it make sense? no so many types? 2 interfaces, if props supposed to be near component

src/components/Filters/ContactFilter.tsx
shepilov: mmm, complicated, I'm not there is not gurantee that we propage this diabled to wrepper component and icons

shepilov:
> @shepilov why using a new intermediate type ContactFilterOption? As input, you have `io.cozy.contacts` and `io.cozy.contacts.group`?
> 
> We have helper and existing components that can help for this kind of UI like `getInitials` that you rewrite here for ContactFilterOption type.

@zatteo because I don't know what model we will have there, I haven't looked yet what we have inside the sharings. But it's supposed to be not all the contacts from the addressbook, but a set of members from the existing shares

zatteo:
> > @shepilov why using a new intermediate type ContactFilterOption? As input, you have `io.cozy.contacts` and `io.cozy.contacts.group`?
> > We have helper and existing components that can help for this kind of UI like `getInitials` that you rewrite here for ContactFilterOption type.
> 
> @zatteo because I don't know what model we will have there, I haven't looked yet what we have inside the sharings. But it's supposed to be not all the contacts from the addressbook, but a set of members from the existing shares

If you are more in a "build UI components" mode, let's go like this for now. But if at the end we get io.cozy.contacts and io.cozy.contacts.group OR sharing members, we may more components to mutualize/use from cozy-ui/cozy-sharing. For example there is already an "autocomplete on contact or group" in the sharing modal.

src/components/Filters/ContactFilter.tsx
JF-Cozy: ok so let's remove this :D Who decided that, and why?

src/components/Filters/ContactFilter.tsx
JF-Cozy: it should be, that's the purpose of "disable" prop on comp 🤷‍♂️

src/components/Filters/ContactFilter.tsx
JF-Cozy: I mean, it does nothing, and used only once, so we can directly write in the comp prop `getSuggestionValue={option => option.label}`

src/components/Filters/ContactFilter.styl
```diff
@@ -0,0 +1,78 @@
+.filter
+    min-width 9.3125rem
+    max-width 20rem
+
+.triggerRoot
+    padding 0 .5rem 0 1rem
```
doubleface: We should use cozy-ui utility classes whenever it is possible

src/components/Filters/contactFilterOptions.ts
```diff
@@ -0,0 +1,44 @@
...
+      )
+
+      return searchableValues.some(value =>
+        normalizeContactFilterText(value).includes(normalizedQuery)
+      )
+    })
+    .slice(0, CONTACT_FILTER_MAX_RESULTS)
+}
+
+function getContactFilterInitials(label: string): string {
```
JF-Cozy: no useful func from cozy-client/models/contact to avoid creating new one?

src/components/Filters/contactFilterOptions.ts
JF-Cozy, line 1: we can add some simple isolated test here

src/components/Filters/ContactFilterSuggestion.tsx
```diff
@@ -0,0 +1,49 @@
...
+  currentUserLabel,
+  option
+}: ContactFilterSuggestionProps): ReactElement {
+  const primaryLabel = option.isCurrentUser
+    ? `${option.label} (${currentUserLabel})`
+    : option.label
+
+  return (
+    <ListItem button>
+      <ListItemIcon>{getContactFilterAvatar(option)}</ListItemIcon>
```
JF-Cozy: using a comp with prop instead of a func is easier to follow and maintain, like `<ContactFileterAvatar option={option} />`

src/components/Filters/ContactFilterSuggestion.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  if (option.kind === 'group') {
+    return (
+      <Avatar size="s">
+        <Icon icon={Peoples} size={16} />
+      </Avatar>
+    )
+  }
+
+  return <Avatar size="s">{getContactFilterInitials(option.label)}</Avatar>
```
JF-Cozy: nit: makeContact instead of getContact https://github.com/cozy/cozy-guidelines#naming-of-functions

src/components/Filters/ContactFilter.styl
JF-Cozy, line 1: lot of custom css and divs, maybe we can avoid that using Mui/cozy-ui comp and u-classes. Maybe it's time to deep dive into cozy-ui doc or plug AI agent to it

src/components/Filters/ContactFilter.tsx
shepilov: twake-javascript-conventions https://github.com/linagora/twake-guidelines/blob/main/skills/twake-javascript-conventions/SKILL.md

shepilov:
> > > @shepilov why using a new intermediate type ContactFilterOption? As input, you have `io.cozy.contacts` and `io.cozy.contacts.group`?
> > > We have helper and existing components that can help for this kind of UI like `getInitials` that you rewrite here for ContactFilterOption type.
> > 
> > 
> > @zatteo because I don't know what model we will have there, I haven't looked yet what we have inside the sharings. But it's supposed to be not all the contacts from the addressbook, but a set of members from the existing shares
> 
> If you are more in a "build UI components" mode, let's go like this for now. But if at the end we get io.cozy.contacts and io.cozy.contacts.group OR sharing members, we may more components to mutualize/use from cozy-ui/cozy-sharing. For example there is already an "autocomplete on contact or group" in the sharing modal.

yes, at first I was going to use "autocomplete on contact or group" on a sharing modal, but it has different behavior then the existing one; we too many customizations in it, like the source of the option, and how the selected results should be displayed, multiselect ant etc, we can discuss how we can mutualize it, but for me it looked to many differences

src/components/Filters/contactFilterOptions.ts
shepilov: What will we test there if it's tested with the component?

src/components/Filters/ContactFilter.styl
shepilov: Probably we can modify this skill https://github.com/linagora/twake-guidelines/tree/main/skills/twake-react-conventions, do you have any ideas about the prompt, except for the link to the docs?

JF-Cozy: nope I'm using cozy-ui and mui by myself, a read all the docs... ask @doubleface maybe?

src/components/Filters/contactFilterOptions.ts
JF-Cozy: personally I prefere to make simple test near the code, and keep comp testing for behavior but not for testing the core. It's generally easier to understand and maintain to have multiple light and simple test files, that a big one

src/components/Filters/ContactFilter.styl
shepilov: the only idea I have is ask AI to prove all the choices, when it chose to make a custom component insted of the one from the cozy-ui

src/components/Filters/ContactFilterSuggestionsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+          role="status"
+        >
+          <Spinner size="small" />
+          <span className="u-ml-half">{loadingMessage}</span>
+        </div>
+      </div>
+    )
+  }
+
+  if (children) {
```
JF-Cozy: nit: if (!children) or even two separate comp would be more understandable I guess 🤔

src/components/Filters/ContactFilterSuggestionsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+const ContactFilterSuggestionsContainer = ({
+  children,
+  containerProps,
+  emptyMessage,
+  loading,
+  loadingMessage
+}: ContactFilterSuggestionsContainerProps): ReactElement => {
+  if (loading) {
+    return (
+      <div {...containerProps}>
```
JF-Cozy, line 24: nit: maybe it's relevant here (because it's a final rendering comp), but just for info, have a big prop spreaded is a dangerous way because it's very difficult to track what we pass to what comp

src/components/Filters/contactFilterOptions.ts
```diff
@@ -0,0 +1,44 @@
+import { remove as removeDiacritics } from 'diacritics'
+
+import type { ContactFilterOption } from './ContactFilter.types'
+
+const CONTACT_FILTER_MAX_RESULTS = 20
+
+function normalizeContactFilterText(value: string): string {
+  return removeDiacritics(value).trim().toLocaleLowerCase()
```
doubleface: Maybe we should unify with https://github.com/linagora/twake-drive/blob/master/src/modules/search/components/helpers.js#L9

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,232 @@
...
+  ): void => {
+    setQuery(newValue)
+  }
+
+  const handleSuggestionSelected = (
+    _event: FormEvent<HTMLElement>,
+    { suggestion }: Autosuggest.SuggestionSelectedEventData<ContactFilterOption>
+  ): void => {
+    handleSelect(suggestion)
+  }
```
JF-Cozy: no need to create an handler if it's just a wrapper that does nothing more, specially if there used only once.

we can remove handleOpen, handleInputChange and handleSuggestionSelected they just do `() => setIsOpen(true)`, `(_ev, {newValue}) => setQuery(newValue)` etc.

src/components/Filters/contactFilterOptions.ts
shepilov: Then I think we need to change the helper to make it work with diacritics

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,232 @@
...
+  loading = false,
+  onChange,
+  onClear,
+  options,
+  value = null
+}: ContactFilterProps): ReactElement => {
+  const { t } = useI18n()
+  const [anchorElement, setAnchorElement] = useState<HTMLElement | null>(null)
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [query, setQuery] = useState<string>('')
```
JF-Cozy: "query" is a really confusing word (used for queries and async) so inputValue or something like that should be more understandable I think

src/components/Filters/ContactFilter.tsx
```diff
@@ -0,0 +1,232 @@
...
+  ContactFilterMenuSurfaceProps
+>(({ children, className, id, onClick }, ref): ReactElement => {
+  return (
+    <div ref={ref} className={className} id={id} onClick={onClick}>
+      {children}
+    </div>
+  )
+})
+
+ContactFilterMenuSurface.displayName = 'ContactFilterMenuSurface'
```
JF-Cozy: ContactFilterMenuSurface and ContactFilterSearchInput could be in separate file to keep this one simple

src/components/Filters/contactFilterOptions.ts
doubleface: We should reuse https://github.com/linagora/twake-drive/blob/master/src/models/Contact.js#L5 but since we only have the label here, we should calculate the initials when we create the option (when we have the full contact)

src/components/Filters/ContactFilter.styl
JF-Cozy: you can tell it also that creating a custom component doesn't mean to create everything from scratch. We have to rely on cozy-ui comp and css to do it, as much as possible

JF-Cozy (approved):
approve to unlock but comments should be considered :) welcome to the frontend world, and well done so far 👍

src/components/Filters/ContactFilter.styl
shepilov: Replaced the custom search input with the Cozy UI SearchBar and removed the related custom search CSS. The remaining styles cover filter and menu sizing, list scrolling and highlighting, and status layout.

src/components/Filters/ContactFilterSuggestionsContainer.tsx
shepilov: Changed this to an explicit if (!children) empty-state branch.

src/components/Filters/ContactFilter.tsx
shepilov: Renamed the state and related variables from query to inputValue.

src/components/Filters/contactFilterOptions.ts
shepilov: Extracted a shared normalizeSearchText helper. The existing search and the contact filter now use the same normalization, including diacritic handling.

src/components/Filters/contactFilterOptions.ts
shepilov: The filter receives a ready-to-render avatar because it also supports groups and other recipient types. The caller that builds an option from a real contact should use the existing Contact helper there, where the full contact model is available.

src/components/Filters/ContactFilterSuggestionsContainer.tsx
shepilov: Kept this spread intentionally. containerProps comes from react-autosuggest and contains the listbox id, role, ref, and ARIA props. This is the final DOM boundary, so forwarding the full object avoids duplicating the library prop list.

src/components/Filters/ContactFilter.tsx
shepilov: Moved both helpers to ContactFilterMenuSurface.tsx and ContactFilterSearchInput.tsx.

src/components/Filters/contactFilterOptions.ts
shepilov: Added a small contactFilterOptions.spec.ts covering accent-insensitive matching, secondary-label matching, the 20-result limit, and finding an option beyond the initial limit.

src/components/Filters/ContactFilter.tsx
shepilov: Removed the three one-use wrapper handlers and inlined their callbacks.

Issue #4112, opened 2026-07-30
[Feature request] Add a UI to delete files versions


ClaudeDeTwake:
Every time a user updates a file, the server keeps the previous version. It keeps up to 20 versions of every file.
If a user works with large files, previous versions can occupy a lot of space.
For example:
```
files : 25.68 Go
versions : 23.67 Go
used : 49.35 Go
```

For now, there is no UI allowing to display the weight of the versions, and to delete them.
It would be nice to add such an UI, to allow users to delete versions.
The stack provides routes to [delete a single version](https://github.com/linagora/cozy-stack/blob/master/docs/files.md#delete-filesfile-idversion-id) or [every version of every file](https://github.com/linagora/cozy-stack/blob/master/docs/files.md#delete-filesversions).

PR #4119, opened 2026-07-31
feat: delete a file version from the history modal

Crash--:
## Summary

- Add a trash action on each older version in the file history modal, so users can reclaim the space versions take up. Closes part of #4112.
- Deleting a version goes through `DELETE /files/:file-id/:version-id`. The path is built from the file collection prefix, so shared drives use the right route without a second code path.
- A confirm dialog states the version date and size and warns that the deletion cannot be undone, since versions have no trash.
- The trash action is hidden when the user has no write access on the containing folder.
- Rebuild the version row on `ListItem` and `ListItemSecondaryAction` instead of cozy-ui's `HistoryRow`, which hardcodes a single download action and has no slot for a second one. This drops the dashed timeline connector: the rows now match the list components used elsewhere in the app, and both actions become keyboard reachable `IconButton`s.
- Move the history queries to `src/queries` with explicit aliases so the version list can be invalidated after a deletion.
- Set `min_delay_between_two_versions` to 1s in the e2e stack config. It defaults to 15 minutes, so no test could otherwise build a version history.

### Before

History modal before

One download icon per row, versions linked by a dashed connector.

### After

History modal after

The current version keeps download only. Older versions get download plus trash.

### Confirmation

Confirm dialog

### After deleting

After delete

The deleted revision disappears without a page reload.

## Summary by CodeRabbit

- **New Features**
  - Added file version history with clearer metadata and download actions.
  - Added permanent deletion of older versions with permission-aware controls and confirmation prompts.
  - Added success and error feedback for version deletion.
  - Enabled retention of up to 20 versions, with a one-second minimum delay between new versions.
- **Bug Fixes**
  - Improved file history loading and display for personal and shared-drive files.
- **Tests**
  - Added coverage for version creation, deletion, downloading, permissions, and error handling.


src/components/FileHistory/HistoryRow.jsx
```diff
@@ -0,0 +1,67 @@
...
+import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
+import { useI18n } from 'twake-i18n'
+
+/**
+ * One entry of a file's version list.
+ *
+ * cozy-ui's HistoryRow renders a single hardcoded download action and exposes
+ * no slot for a second one, hence this local row built on the standard list
+ * primitives.
+ */
```
zatteo: nit: Useless comment

src/locales/fr.json
```diff
@@ -541,7 +541,17 @@
...
+    "download": "Télécharger cette version",
+    "deleteVersion": {
+      "action": "Supprimer cette version",
+      "title": "Supprimer cette version ?",
+      "version": "Version du %{date} (%{size})",
+      "irreversible": "Cette version sera définitivement supprimée. Cette action est irréversible.",
+      "cancel": "Annuler",
+      "confirm": "Supprimer",
+      "error": "Cette version n'a pas pu être supprimée."
+    }
```
zatteo, line 560: nit: Missing ru an vn

src/components/FileHistory/HistoryModal.jsx
```diff
@@ -34,71 +36,81 @@ const HistoryModal = ({
...
+              )}
+              {capabilities.fetchStatus === 'loaded' &&
+                isFileVersioningEnabled && (
+                  <span>{t('History.description')}</span>
+                )}
+              {(capabilities.fetchStatus === 'failed' ||
+                (!isFileVersioningEnabled &&
+                  capabilities.fetchStatus !== 'loading')) && (
+                <span>{t('History.noFileVersionEnabled')}</span>
               )}
```
zatteo: A single span with an helper to compute the text?

edit: Already fixed 👍

src/components/FileHistory/HistoryModal.jsx
```diff
@@ -1,15 +1,18 @@
...
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 
-import { withClient, useCapabilities } from 'cozy-client'
+import { useClient, useCapabilities } from 'cozy-client'
+import { useSharingContext } from 'cozy-sharing'
 import { Dialog } from 'cozy-ui/transpiled/react/CozyDialogs'
-import HistoryRow from 'cozy-ui/transpiled/react/HistoryRow'
```
zatteo, line 8: HistoryRow can be removed from cozy-ui if we don't use it anymore

docs/delete-file-version/after-delete.png
zatteo, line 1: Why commiting this before/after screenshot?

src/lib/versions.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {object} params
+ * @param {import('cozy-client').default} params.client
+ * @param {string} [params.driveId] Id of the shared drive holding the file
+ * @param {object} params.version io.cozy.files.versions document
+ */
+export const destroyFileVersion = async ({ client, driveId, version }) => {
+  const { prefix } = client.collection('io.cozy.files', { driveId })
+
+  return client.getStackClient().fetchJSON('DELETE', `${prefix}/${version.id}`)
+}
```
zatteo, line 17: Could be relevant in cozy-client?

docs/delete-file-version/after-delete.png
Crash--: because I want to share with @Bobpodvalnyi and @shepilov how it compares. I'll remove that before the merge.

zatteo (approved):

src/components/FileHistory/HistoryRow.jsx
Crash--: Removed in 07f3895.

src/locales/fr.json
Crash--: Added in 4ec037a: `History.download` and `History.deleteVersion.*` are now in ru.json and vi.json alongside en and fr.

src/components/FileHistory/HistoryModal.jsx
Crash--: Agreed, but that belongs in a cozy-ui PR rather than this one, and it needs a sweep for other consumers before removing it. Tracking it separately.

src/lib/versions.js
Crash--: Maybe later if we need it. Keeping the helper local for now.

src/components/FileHistory/DeleteVersionConfirm.jsx
```diff
@@ -0,0 +1,93 @@
...
+    } catch (_e) {
+      showAlert({
+        message: t('History.deleteVersion.error'),
+        severity: 'error'
+      })
+    } finally {
+      onClose()
+      setBusy(false)
+    }
+  }, [client, file, version, onClose, showAlert, t])
```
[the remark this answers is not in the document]
Crash--: Fixed in 30fa25c. `onClose()` now runs only on the success path, `setBusy(false)` stays in `finally`. The error test asserts `expect(onClose).not.toHaveBeenCalled()`, and it does fail against the previous version.

e2e/tests/file-versions.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  })
+  for (let i = 1; i <= count; i++) {
+    await overwriteFile({
+      instance: ALICE.instance,
+      fileId,
+      content: `version ${i}`.repeat(i + 1)
+    })
+  }
+  return fileId
+}
```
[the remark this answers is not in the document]
Crash--: Fixed in 6928a90 via `waitForVersionWindow()` in `e2e/helpers/stack.ts`.

For the record, the mechanism is in cozy-stack `model/vfs/version.go`: `detectVersionsToClean` returns `CleanCandidateVersion` when `previousTime.Add(minDelay).After(candidateTime)`, so the candidate is dropped silently. `len(olds) == 0` short-circuits to `KeepCandidateVersion`, so the first overwrite is never at risk and the wait is only applied from the second one onward.

src/components/FileHistory/HistoryModal.jsx
```diff
@@ -34,71 +44,74 @@ const HistoryModal = ({
...
+                        CozyFile.generateFileNameForRevision(file, revision, f)
+                      )
+                    }
+                    onDelete={
+                      canDeleteVersions
+                        ? () => setVersionToDelete(revision)
+                        : undefined
+                    }
+                  />
+                ))}
```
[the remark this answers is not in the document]
Crash--: Not changing this one. `revision.id` is what master already does at `HistoryModal.jsx:83`, it is not introduced here, and cozy-client normalizes documents to carry both `id` and `_id`, so the download call resolves the same version either way.

docs/superpowers/specs/2026-07-30-delete-file-version-design.md
```diff
@@ -0,0 +1,149 @@
...
+refreshes the list, matching how `NextcloudDeleteConfirm` already invalidates
+its folder queries.
+
+### Permissions
+
+The history modal is reachable from shared folders and public links, where the
+user may hold read-only access. The trash icon is hidden unless
+`useSharingContext().hasWriteAccess(file.id, file.driveId)` is true, so the UI
+never offers an action the stack would reject.
+
```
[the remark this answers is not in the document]
Crash--: Obsolete. `docs/superpowers/specs/` is no longer part of this branch.

PR #4121, opened 2026-07-31
feat(harvest): remove Harvest integration

### LINKED ISSUE #4060: Remove cozy-harvest-lib from drive and all its dependencies
We don't want connectors visible in drive anymore and we to make twake-drive lighter to be faster

There is a first try in https://github.com/linagora/twake-drive/tree/chore/remove-konnector-support but we will need also to remove cozy-harvest-lib from cozy-viewer as a dependency to make it complete

doubleface:
Drive no longer hosts Konnector management flows. Remove the banner, nested routes, dedicated queries, and Konnector permission so the app can
drop cozy-harvest-lib and its transitive packages.

Also depends on https://github.com/linagora/cozy-libs/pull/3109 and an upgrade of cozy-viewer to be mergeable

## Summary by CodeRabbit

* **Removed Features**
  * Removed Konnector routes and related navigation.
  * Removed the Harvest banner and Konnector-related views from Drive folders.
  * Removed the Konnector permission.
* **Permissions**
  * Updated the accounts permission description to reflect AI Assistant usage.
* **Maintenance**
  * Removed legacy Harvest integration components and related configuration.
  * Updated supporting interface and document-viewing components for improved compatibility.


zatteo:
@doubleface any idea about why there is this increase? 

image

doubleface:
> @doubleface any idea about why there is this increase?
> image

If I do the test locally, the bundle looses 1.68Mb

Here is an AI generated explanation I don't actually fully understand:

  ### 🔍 What Actually Changed in the Bundles

  1. Removal of cozy-harvest-lib:
      • On master, the 3357.js chunk (which bundled the Harvest / Konnectors integration) weighed 1,824.5 KB (~1.8 MB).
      • This chunk has completely disappeared.
  2. cozy-viewer update (30.0.19 → 30.0.31) & chunk restructuring:
      • Two new, more granular chunks were created: 1636.js (279.6 KB) and 9256.js (258.7 KB), totaling ~538 KB.
      • The net result is a reduction of over 1.2 MB on these modules alone.
  3. Where did the apparent increase (+318%) come from?
      • Rspack outputs split chunks with numeric IDs (5098.js, 7449.js, 9838.js, etc.).
      • In .bundlemonrc:8, the rule static/js/<chunkId>.<hash>.js matches a single arbitrary file among all numbered chunks.
      • Because the dependency graph was restructured, BundleMon ended up comparing the new 9838.js chunk (8.72 KB gzipped) against a completely different residual small
      chunk from master (2.08 KB gzipped).
      • This is purely a false positive caused by the single-chunk tracking pattern, while the overall build is actually 1.68 MB lighter.


I think we should update .bundlemonrc to avoid this kind of problems -> https://github.com/linagora/twake-drive/pull/4125

zatteo:
I have also the feeling that bundlemon does not track *useful* stuff. Like it does not track for example the intent bundle we tried to optimize in the previous weeks. It will need some rework one day.

zatteo (approved):

PR #4124, opened 2026-08-24
feat(file-picker): enforce maxFileCount and availableSize constraints

### LINKED ISSUE #4123: Implement maxFileCount and availableSize constraints in file picker
These were left as todo in the code

doubleface:
getActionDisabledState only enforced per-file maxFileSize, so action buttons stayed enabled once the selection went over a client's count or total-size cap. Add two selection-level rules that disable the action and surface the matching reason key, and pipe those keys through the footer tooltip.

The ru.locale FilePicker keys that were still in English are translated as well.

fixes #4123 

## Summary by CodeRabbit

* **New Features**
  * File selection now enforces maximum file-count and total-size limits.
  * Folders count toward item limits but not total file-size limits.
  * Clear, localized messages explain when limits are exceeded.
  * Tooltips display relevant limit values.
  * Mobile selection indicators show a check icon and item count.
* **Bug Fixes**
  * Sharing-link configuration now preserves file-count and available-size limits.
* **Documentation**
  * Updated File Picker guidance to describe enforced selection limits.
* **Tests**
  * Added coverage for boundaries, multiple selections, undefined limits, and folder handling.

### DIFF
--- a/src/modules/services/components/FilePicker/constraints.js
+++ b/src/modules/services/components/FilePicker/constraints.js
@@ -8,8 +8,6 @@ export { matchMimeType }
 
 const { file: fileModel } = models
 
-// TODO: enforce maxFileCount and availableSize.
-
 const getFileMime = file => {
   const mime = file?.mime || mimeTypes.lookup(file?.name)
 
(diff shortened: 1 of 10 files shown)


src/modules/services/components/FilePicker/constraints.js
```diff
@@ -82,5 +84,31 @@ export const getActionDisabledState = (actionConfig, selectedItems) => {
...
+      reasonKey: 'FilePicker.constraints.disabledReasons.maxFileCountExceeded'
+    }
+  }
+
+  const availableSize = actionConfig.availableSize
+  if (typeof availableSize === 'number') {
+    let totalSize = 0
+    for (const selectedItem of items) {
+      if (selectedItem && fileModel.isFile(selectedItem)) {
+        totalSize += Number(selectedItem.size)
```
zatteo: nit: Required to cast to `Number`?

zatteo (approved):

src/modules/services/components/FilePicker/constraints.js
doubleface: Right. Since we already check that this is a file model, we don't need to double check. Removing it

PR #4128, opened 2026-08-25
fix(file-picker): normalize file sizes for available limits

doubleface:
Cozy file documents can expose size as a string. Normalize sizes before summing selected files so availableSize checks do not concatenate values and disable valid selections. Add a demo configuration to reproduce the multi-file case.

## Summary by CodeRabbit

* **New Features**
  * Added file attachments with a 50 MB per-file limit and 100 MB total limit.
  * Added an “Attach files” action; sharing links are unavailable.

* **Bug Fixes**
  * Improved file-size validation for numeric and text-based values.
  * Invalid, blank, negative, or non-numeric sizes are now rejected.
  * Ensured text-based file sizes are correctly included in total-size calculations.

### DIFF
--- a/src/modules/navigation/components/FilePickerButton.jsx
+++ b/src/modules/navigation/components/FilePickerButton.jsx
@@ -39,6 +39,10 @@ const PICKER_CONFIGS = [
     label: 'Max size 1 KB'
   },
   {
+    id: 'max-size-and-available-size',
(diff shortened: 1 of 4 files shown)


zatteo (approved):
> Cozy file documents can expose size as a string.

Where does it come from? It should not?

PR #4155, opened 2026-08-31
feat(file-picker): Sharings section

doubleface:
Add a Sharings section to the file picker for browsing shared files and
folders.

Track the selected section, folder, and shared drive independently, and reuse
the Sharings filtering and query pipeline when displaying shared content.

image

## Summary by CodeRabbit

- **New Features**
  - Browse shared items and shared drives directly from the file picker using Drive and Sharings tabs.
  - Navigate shared folders with updated breadcrumbs on desktop and mobile.
  - Attach files from nested shared-drive folders while preserving download links across mixed-drive selections.
  - Display tailored empty-state messages for shared items and regular folders.

- **Bug Fixes**
  - Disabled items can no longer be selected or activated accidentally.
  - Switching sections now resets selections correctly.
  - Breadcrumbs retain the current folder when loading a parent fails.
  - Loading and shared-drive errors are surfaced more reliably.


src/modules/services/components/FilePicker/constants.js
```diff
@@ -15,7 +15,14 @@ export const filePickerLinkModes = {
...
 
-export const filePickerThemes = ['light', 'dark']
+export const filePickerSections = {
+  DRIVE: 'drive',
+  SHARINGS: 'sharings'
+}
+
+export const FILE_PICKER_SHARINGS_ROOT_ID = 'file-picker-sharings-root'
+
+export const filePickerThemes = ['auto', 'light', 'dark']
```
zatteo: This is bringing `auto` again.

src/modules/services/components/FilePicker/FilePickerBody.jsx
```diff
@@ -126,25 +120,269 @@ const FilePickerBody = ({
...
+          ))}
+        </Box>
+      ) : items.length === 0 ? (
+        <Typography
+          className="u-ta-center u-pa-2"
+          color="textSecondary"
+          data-testid="file-picker-empty"
+        >
+          {t(emptyMessageKey)}
+        </Typography>
```
zatteo: To move out in a dedicated `EmptyMessage` component (so we can move outside this one `emptyMessageKey` too) ?

src/modules/services/components/FilePicker/FilePickerBody.jsx
```diff
@@ -126,38 +128,311 @@ const FilePickerBody = ({
...
+    () =>
+      filterReceivedShares && allLoaded
+        ? filterOutReceivedShares([result], isOwner)[0]
+        : result,
+    [allLoaded, filterReceivedShares, isOwner, result]
+  )
+  const fetchStatus =
+    isQueryLoading(filteredResult) || filteredResult.isFetching
+      ? 'loading'
+      : (filteredResult.fetchStatus ?? 'loaded')
```
zatteo: Why do we need to do all of this to get a status? `isQueryLoading` is not doing a good job itself?

src/modules/services/components/FilePicker/FilePickerBody.jsx
```diff
@@ -126,38 +128,311 @@ const FilePickerBody = ({
...
+
+  useEffect(() => {
+    if (fetchStatus !== 'loading') onLoaded?.()
+  }, [fetchStatus, onLoaded])
```
zatteo: It will call `onLoaded` even if it fail. Is it wanted?

src/modules/services/components/FilePicker/FilePickerHeader.jsx
```diff
@@ -1,29 +1,89 @@
...
+        onChange={handleSectionChange}
+        aria-label={t('Nav.item_file_picker')}
+        className={cx(
+          styles.filePickerNavigation,
+          'u-flex-self-start u-mt-half u-w-auto'
+        )}
+        narrowed
+        textColor="primary"
+        indicatorColor="primary"
+      >
```
zatteo: nit: Create a FilePickerHeaderTabs component?

src/modules/services/components/FilePicker/index.jsx
```diff
@@ -63,7 +70,40 @@ const FilePicker = ({
 
   const navigateTo = folder => {
     setError(null)
-    setFolderId(folder.id ?? folder._id)
+    setLocation(currentLocation => {
+      const folderId = folder.id ?? folder._id
```
zatteo, line 74: Why `folder.id ?? folder._id`?

src/modules/services/components/Picker.jsx
```diff
@@ -52,32 +56,38 @@ const Picker = ({ service, intent, onReadyToUse }) => {
...
-    )
     let queryResults
     try {
       queryResults = await Promise.all(
-        selectedFileIds.map(fileId =>
-          client.query(Q('io.cozy.files').getById(fileId), {
-            as: `picker-confirm-${fileId}`,
+        selectedFiles.map(async file => {
+          const fileId = typeof file === 'string' ? file : getFileId(file)
+          const driveId = typeof file === 'string' ? null : file.driveId
```
zatteo: nit: Can we make everything return a file here to avoid needing to manage string file or object file?

src/modules/services/components/FilePicker/FilePickerSharingsContent.jsx
```diff
@@ -0,0 +1,51 @@
...
+  })
+
+  return renderContent({
+    items: fetchStatus === 'loaded' ? (filteredResult.data ?? []) : [],
+    fetchStatus,
+    hasMore: false,
+    fetchMore: null,
+    breadcrumbPath: [rootBreadcrumbPath],
+    isItemDisabled: isSharingShortcutNew
+  })
```
zatteo, line 41: What is this `renderContent` pattern?

src/modules/services/components/FilePicker/FilePickerBody.jsx
```diff
@@ -126,38 +128,311 @@ const FilePickerBody = ({
...
+      folderSelectable,
+      itemTypesAccepted,
+      multiple,
+      navigateTo,
+      onFileDoubleClick,
+      section,
+      isSectionChanging,
+      onSectionReady
+    ]
+  )
```
zatteo, line 347: Is this `useCallback` really useful?

src/modules/services/components/FilePicker/FilePickerSharingsContent.jsx
```diff
@@ -0,0 +1,51 @@
...
+  getSharingsFetchStatus,
+  useSharingsQueryResult
+} from '@/modules/views/Sharings/useSharingsQueryResult'
+
+export const FilePickerSharingsContent = ({
+  renderContent,
+  rootBreadcrumbPath
+}) => {
+  const { allLoaded, byDocId } = useSharingContext()
+  const sharedDocumentIds = useMemo(() => Object.keys(byDocId ?? {}), [byDocId])
```
zatteo: This is already calculated by FilePickerBody it can be passed as props

src/modules/services/components/FilePicker/index.jsx
```diff
@@ -44,7 +46,12 @@ const FilePicker = ({
   onReadyToUse,
   onFileDoubleClick
 }) => {
-  const [folderId, setFolderId] = useState(ROOT_DIR_ID)
+  const [location, setLocation] = useState({
+    section: filePickerSections.DRIVE,
+    folderId: ROOT_DIR_ID,
+    driveId: null
+  })
+  const [isSectionChanging, setIsSectionChanging] = useState(false)
```
zatteo, line 54: What is achieved with this `isSectionChanging`?

src/modules/services/components/FilePicker/constants.js
doubleface: Good catch. I was doing this PR while you did yours

src/modules/services/components/FilePicker/FilePickerBody.jsx
doubleface: This is mainly to get the error status if any. The part with isFetching was actually not needed and I removed it.

src/modules/services/components/FilePicker/FilePickerBody.jsx
doubleface: It is wanted but the naming is bad. I renamed it to onReady

src/modules/services/components/FilePicker/index.jsx
doubleface: Actually, it can be a breadcrumb and breadcrumbs have only id attribute since even before the file picker

src/modules/services/components/FilePicker/FilePickerSharingsContent.jsx
doubleface: renderContent is a render-prop callback. The content components are responsible for fetching and preparing their data, while FilePickerContent owns the shared rendering and interaction logic. This lets local folders, shared-drive folders, and sharings reuse the same picker UI while providing their own data source.

I renamed renderContent to renderFilePickerContent to try to make it a little more explicit

src/modules/services/components/FilePicker/FilePickerBody.jsx
doubleface: No

src/modules/services/components/FilePicker/index.jsx
doubleface: isSectionChanging prevents stale or cached content from flashing when switching between Drive and Sharings. The body stays hidden until the new section's query is no longer loading, then onSectionReady clears the flag and displays the new content.

e2e/pages/FilePickerPage.ts
```diff
@@ -93,6 +95,25 @@ export class FilePickerPage {
...
+
+      await scroller.evaluate((element: HTMLElement) => {
+        element.scrollTop += element.clientHeight
+      })
+      await this.page.waitForTimeout(50)
+    }
+
+    await item.waitFor({ state: 'visible' })
+  }
+
```
zatteo, line 120: Maybe something cleaner to do with `expect.toPass` or `expect.poll` ? https://playwright.dev/docs/test-assertions#expecttopass

zatteo (approved):

PR #4167, opened 2026-09-02
feat(file-picker): Add recent files section

### LINKED ISSUE #3995: file picker / risentce - step 5

doubleface:
Let users select recent local and federated files directly from the picker, with localized loading states and confirmation handling.

image

## Summary by CodeRabbit

* **New Features**
  * Added a **Recents** tab to the file picker.
  * Recently used files can now be browsed and selected, including files from shared drives.
  * Recent files display their parent folder paths where available.
  * Added loading, empty, error, and incremental-loading states for recent files.
  * Folders remain selectable in the file picker when permitted.

* **Localization**
  * Added translations for recent-file messages in English, French, Russian, and Vietnamese.
  * Added translations for version download and deletion actions.


src/modules/services/components/FilePicker/FilePickerHeader.spec.jsx
```diff
@@ -0,0 +1,50 @@
...
+        activeSection={filePickerSections.DRIVE}
+        onSectionChange={jest.fn()}
+      />
+    )
+
+    expect(screen.getAllByRole('tab').map(tab => tab.textContent)).toEqual([
+      'Nav.item_drive',
+      'Nav.item_recent',
+      'Nav.item_sharings'
+    ])
```
zatteo: We have same assertions (that we have 3 tabs with the good names) in unit tests (here) and E2E tests (e2e/tests/file-picker-sharings.spec.ts) and idk what to think about it

src/modules/services/components/FilePicker/FilePickerRecentsContent.jsx
```diff
@@ -0,0 +1,55 @@
...
+      itemIds.has(item._id)
+    )
+
+    if (remainingSelectedItems.length === selectedItems.length) return
+
+    setSelectedItems(
+      Object.fromEntries(remainingSelectedItems.map(item => [item._id, item]))
+    )
+    setIsSelectAll(false)
+  }, [data, fetchStatus, selectedItems, setIsSelectAll, setSelectedItems])
```
zatteo, line 32: This does not need a useEffect?

zatteo (commented):
Curious if now we have 3 tabs and it looks stabilized, a ponytail pass/AI pass on the whole FilePicker find potentiel refactoring or dead code 🤷‍♀️

src/modules/services/components/FilePicker/FilePickerRecentsContent.jsx
doubleface: I think so since the are side effects

src/modules/services/components/FilePicker/FilePickerHeader.spec.jsx
doubleface: Good point. Since the E2E test already covers the tabs in the real picker flow, I removed the redundant FilePickerHeader unit test and kept the E2E coverage as the source of truth.

src/modules/services/components/FilePicker/FilePickerRecentsContent.jsx
zatteo: Allright, did not see that setSelectedItems was from a provider up in the tree

zatteo (approved):

PR #4171, opened 2026-09-04
feat(recents): Support dataproxy in file picker and add E2E tests for recents

doubleface:
- Wrap IntentProvider in DataProxyProvider with doctypes (files, contacts, apps)
- Update Picker.recents.spec.jsx to test integration with useDataProxy
- Install dataproxy app and configure flags in e2e/setup/global-setup.ts
- Add E2E test for federated shared folder files in File Picker Recents tab
- Add E2E test for federated shared folder files in Drive Recents view

## Summary by CodeRabbit

- **New Features**
  - Federated shared-folder files are now available in **Recents**.
  - Shared files can be selected through the file picker and used to generate temporary download links.
  - Files from shared drives can be opened from **Recents**, with downloads routed through the data proxy.
  - Recents refresh when switching between **Recents** and **My Drive**, ensuring newly available shared files are displayed.
  - Shared files can be viewed successfully after opening them from Recents.

### DIFF
--- a/e2e/setup/global-setup.ts
+++ b/e2e/setup/global-setup.ts
@@ -71,7 +71,9 @@ const FEATURE_FLAGS = {
   'drive.shared-drive.enabled': true,
   'drive.federated-shared-folder.enabled': true,
   'drive.federated-shared-modal.enabled': true,
-  'drive.file-picker-demo.enabled': true
+  'drive.file-picker-demo.enabled': true,
+  'cozy.search.enabled': true,
+  'dataproxy.force-trusted-device.enabled': true
 }
 
 async function waitForStack(url: string, timeoutMs = 60_000): Promise<void> {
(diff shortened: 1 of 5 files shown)


e2e/setup/global-setup.ts
```diff
@@ -178,6 +180,20 @@ async function setupUser(
     )
   }
 
+  if (isAppInstalled(user, 'dataproxy')) {
+    console.log(`[e2e] Reusing Dataproxy app for ${user.label}.`)
+  } else {
+    console.log(`[e2e] Installing Dataproxy app for ${user.label}...`)
```
shepilov: Why do we want to install it with UI query? Why not by default with Drive?

rezk2ll (approved):

e2e/setup/global-setup.ts
doubleface: Right, dataproxy should be installed by default with drive

PR #4179, opened 2026-09-08
test: Execute nested sharing test for 3 different flags

JF-Cozy:
We previously added some tests for nested sharing https://github.com/linagora/twake-drive/pull/4177 . But by default, the e2e are only executed in a specific flag context set in the config file.

Here we want to execute these tests by switching true/false for 3 different flags `drive.federated-shared-folder.enabled`, `cozy.hide-sharing-cozy-to-cozy`, `drive.shared-drive.enabled`

## Summary by CodeRabbit

- **Bug Fixes**
  - Shared-drive file changes made while disconnected or reconnecting are now detected after the connection is restored.
  - Improved reliability when loading nested items in the mobile file picker.

- **Tests**
  - Expanded end-to-end coverage for nested sharing across federation, sharing visibility, and shared-drive settings.
  - Added validation for recipient lists, modal behavior, unavailable contact sharing, and link-only sharing.
  - Improved test consistency with standardized feature flags and support for longer-running scenarios.

### DIFF
--- a/e2e/helpers/flags.ts
+++ b/e2e/helpers/flags.ts
@@ -1,5 +1,18 @@
 import { stackExec } from './config'
 
+/** Flags applied to every instance at provisioning time (global-setup).
+ * Specs overriding flags via setFlags should spread over this map so each
+ * call re-asserts a full, deterministic state. */
+export const DEFAULT_FLAGS: Record<string, boolean | string | number> = {
+  'cozy.hide-sharing-cozy-to-cozy': true,
+  'drive.shared-drive.enabled': true,
+  'drive.federated-shared-folder.enabled': true,
+  'drive.federated-shared-modal.enabled': true,
+  'drive.file-picker-demo.enabled': true,
+  'cozy.search.enabled': true,
+  'dataproxy.force-trusted-device.enabled': true
+}
+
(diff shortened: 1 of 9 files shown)


e2e/helpers/flags.ts
```diff
@@ -1,5 +1,16 @@
...
+/** Flags applied to every instance at provisioning time (global-setup).
+ * Specs overriding flags via setFlags should spread over this map so each
+ * call re-asserts a full, deterministic state. */
+export const DEFAULT_FLAGS: Record<string, boolean | string | number> = {
+  'cozy.hide-sharing-cozy-to-cozy': true,
+  'drive.shared-drive.enabled': true,
+  'drive.federated-shared-folder.enabled': true,
+  'drive.federated-shared-modal.enabled': true,
+  'drive.file-picker-demo.enabled': true
+}
```
zatteo, line 14: 'cozy.search.enabled': true,
'dataproxy.force-trusted-device.enabled': true

removed from default flags?

JF-Cozy: was a mistake, good catch 👍 fixed

zatteo (approved):

PR #4180, opened 2026-09-08
fix: open shared drawings and pdfs on the instance that owns them

Crash--:
## Summary

- Add `useEditorOpen`, which asks the stack `GET /editor/:fileId/open` (or `GET /sharings/drives/:driveId/editor/:fileId/open`) where a file must be opened, and sends the user to the owner's public page when the answer points at another instance.
- Hold the Excalidraw and PDF editors back until that answer says the file is local, so neither joins a collaboration channel nor autosaves on a replica.
- Move `shouldBeOpenedOnOtherInstance` from `views/OnlyOffice/helpers` to `views/editor/helpers`, shared by the three editors.
- Fall back to opening the local copy when the route errors, so a stack without it keeps working.
- Fix `setupRouter` in the public AppRouter spec: it pushed a pathname while `AppLike` mounts a `HashRouter`, so every case rendered the `/` route.

## Summary by CodeRabbit

* **New Features**
  * Shared PDF and Excalidraw files now open on their owning instance.
  * Loading indicators appear while the correct instance is being determined.
  * Shared-drive files open from their appropriate hosting location.
  * OnlyOffice documents support improved cross-instance redirection.

* **Bug Fixes**
  * Improved handling of shared-file access across instances.
  * Local files continue to open normally with existing editing and permission behavior.
  * Editors now wait for location resolution before loading, preventing premature editing or autosaving.


src/modules/views/Excalidraw/index.jsx
```diff
@@ -20,12 +23,16 @@ const Excalidraw = ({ isPublic = false, isReadOnly = false }) => {
...
+      {openStatus === 'local' ? (
+        <Editor
+          fileId={fileId}
+          driveId={driveId}
+          isPublic={isPublic}
+          isReadOnly={readOnly}
+        />
+      ) : (
+        <Loader />
+      )}
```
zatteo: What about a wrapper component that does the "use editor hook to display loader or editor"? To avoid doing potentially same hook + if + all the tests for all editors?

Something like 
```
const RedirectEditor = ({ fileId, driveId, slug, children }) => {
  const openStatus = useEditorOpen({ fileId, driveId, slug })

  return openStatus === 'local' ? children : <Loader /> 
}
```

But unsure it will work for all editors as it is different for Pdf already... It's especially for all those tests that I don't like 😂

zatteo (approved):

PR #4190, opened 2026-09-14
feat(move): add inline folder creation

### LINKED ISSUE #4173: Add destination folders validation and folder creation
Mark not possible destination folders as disabled :
- source folder
- descendant folders of the folder to move
- folders without writing permissions

A "create folder" button is visible when possible as in the figma

doubleface:
Add inline destination-folder creation to the MoveTo picker, including validation, accessibility feedback, and end-to-end coverage.

fixes #4173 

[Capture vidéo du 2026-09-14 11-48-55.webm](https://github.com/user-attachments/assets/5604e86d-b7e9-4a4b-893a-fdd0837c4388)

## Summary by CodeRabbit

- **New Features**
  - Move files and folders into newly created folders directly from the Move dialog.
  - View reasons why destinations are unavailable, including read-only folders, descendants, source folders, and loading permissions.
  - Create folders inline with validation, progress feedback, and duplicate-name error handling.
  - File pickers support additional content, filtered shared items, disabled items, and restricted navigation.
  - Unavailable folders remain visible with clear disabled states.

- **Accessibility**
  - Added accessible labels, error states, focus handling, and tooltips for disabled items.

- **Localization**
  - Updated folder-creation wording and added move-related status messages in supported languages.


zatteo:
There are two divider when creating a folder

Screenshot 2026-09-16 at 08 20 47

zatteo:
When there is an error during error creation, there are two error messages. I don't know if it's on purpose on accepted:
image

src/modules/move/MoveTo.jsx
```diff
@@ -12,24 +16,134 @@ import {
...
+
+  return null
+}
+
+function isMoveDestination(item, entries, hasWriteAccess, allLoaded = true) {
+  return (
+    isLocalMoveDestination(item) &&
+    !getMoveDestinationDisabledReason(item, entries, hasWriteAccess, allLoaded)
+  )
+}
```
zatteo: Lot of helpers to move in a dedicated files I think?

src/modules/move/MoveTo.jsx
```diff
@@ -12,24 +16,134 @@ import {
...
+
+function isLocalItem(item) {
   return (
     Boolean(item) &&
-    models.file.isDirectory(item) &&
+    item._type !== 'io.cozy.remote.nextcloud.files' &&
     !item.driveId &&
-    item.cozyMetadata?.createdByApp !== 'nextcloud' &&
-    (!entryIds || !entryIds.has(itemId))
+    item.cozyMetadata?.createdByApp !== 'nextcloud'
```
zatteo: Don't we have a `isNextcloud` or something else for this kind of checks?

src/modules/move/MoveTo.jsx
```diff
@@ -73,74 +174,169 @@ export function MoveTo({
...
+    if (isBusy) return 'Move.moveInProgress'
+    if (isCreatingFolder) return 'Move.folderCreationInProgress'
+    return getMoveDestinationDisabledReason(
+      item,
+      entries,
+      hasWriteAccess,
+      allLoaded
+    )
+  }
+  const isItemDisabled = item => Boolean(getItemDisabledReason(item))
```
zatteo: Is it really useful to pass this `isItemDisabled` props if descendant can just check if `getItemDisabledReason(item)` return a truthy value?

zatteo:
I don't know to which point it would have be possible to split but this is a big commit complicated to read.

zatteo:
Are we sure about the secondary colors here? It almost never used? And I think @swanndesign25 changed it in the design system.

image

src/modules/move/MoveTo.jsx
doubleface: Yes, added them to the helper file

src/modules/move/MoveTo.jsx
doubleface: yes

doubleface:
> const isItemDisabled = item => Boolean(getItemDisabledReason(item))

I suppose I should have separated folder creation and folder disabling in two different PRs but it was bigger than expected in the end

doubleface:
> Are we sure about the secondary colors here? It almost never used? And I think @swanndesign25 changed it in the design system.
> image

@zatteo I confirmed this ghost variant button with @swanndesign25

doubleface:
> I don't know to which point it would have be possible to split but this is a big commit complicated to read.

@zatteo 
I suppose I should have separated folder creation and folder disabling in two different PRs but it was bigger than expected in the end

zatteo (approved):

PR #4193, opened 2026-09-15
fix(move): preserve partial move results

### LINKED ISSUE #4174: Handle partial success
When only a part of the files/folders has been moved, handle it properly :
- display success and errors
- remove these files/folders from the list of files to move
- keep non moved files selected with submit button activated

doubleface:
Add inline destination-folder creation to the MoveTo picker, including validation, accessibility feedback, and end-to-end coverage.

fixes #4174 

## Summary by CodeRabbit

- **New Features**
  - Move operations support partial completion with successful and failed item counts.
  - Failed items can be retried without repeating successful moves.
  - Successful moves remain completed when an operation is cancelled.
  - Destinations are revalidated before confirmation and locked while moving.
  - Users can create folders during destination selection.
  - Offline users are prevented from starting move operations.

- **Bug Fixes**
  - Move controls remain disabled while operations are pending.
  - Success notifications identify items moved successfully.

- **Localization**
  - Updated move-status and destination-lock messages across supported languages.


src/modules/move/MoveModal.jsx
```diff
@@ -104,71 +113,125 @@ const MoveModal = ({
...
-    isMovingOutsideNextcloud,
-    isMovingInsideNextcloud,
-    folder
-  }) => {
+  const refreshNextcloudQueries = (folder, sourceEntry) => {
+    const isMovingInsideNextcloud =
+      folder?._type === 'io.cozy.remote.nextcloud.files'
+    const isMovingOutsideNextcloud =
+      !isMovingInsideNextcloud &&
+      sourceEntry?._type === 'io.cozy.remote.nextcloud.files'
```
zatteo: Why do we duplicate `isMovingInsideNextcloud` and `isMovingOutsideNextcloud`?

src/modules/move/MoveTo.spec.jsx
```diff
@@ -488,7 +493,32 @@ describe('MoveTo', () => {
...
-  it('does not select a folder and moves into the current folder after navigation', () => {
+  it('keeps the browser inside the content area below the file summary', () => {
+    setup()
+
+    expect(screen.getByTestId('move-to-browser')).toHaveClass(
+      'u-pos-relative',
+      'u-flex',
+      'u-flex-column'
+    )
+  })
```
zatteo: To delete

src/modules/move/MoveModal.jsx
```diff
@@ -47,10 +48,14 @@ const MoveModal = ({
     allLoaded
   } = useSharingContext()
   const { registerCancelable } = useCancelable()
-  const { showSuccess } = useMove({ entries })
+  const { showSuccess } = useMove()
   const { t } = useI18n()
   const { showAlert } = useAlert()
+  const isOffline = useBrowserOffline()
```
zatteo: What is the purpose of using `useBrowerOffline`? Move request will fail if it's offline and that's already enough no?

src/modules/move/MoveModal.jsx
```diff
@@ -104,71 +113,125 @@ const MoveModal = ({
...
+      const failedEntries = remainingEntries.filter(
+        (_entry, index) => results[index].status === 'rejected'
+      )
+      const trashedFiles = results.flatMap(result =>
+        result.status === 'fulfilled' && result.value?.deleted
+          ? [result.value.deleted]
+          : []
+      )
+      results.forEach(result => {
+        if (result.status === 'rejected') logger.warn(result.reason)
```
zatteo: nit: These array could be more efficient with one or two max for loops. Maybe not necessary.

e2e/pages/MoveToPage.ts
```diff
@@ -29,10 +32,70 @@ export class MoveToPage {
...
+        if (remainingFailures > 0) {
+          this.moveFailures.set(fileId, remainingFailures - 1)
+          await route.fulfill({
+            status: 500,
+            contentType: 'application/json',
+            body: JSON.stringify({ error: 'Injected move failure' })
+          })
+          return
+        }
+      }
```
zatteo: Are there no PlayWright helpers for this?

src/modules/move/MoveModal.jsx
doubleface: yes factorised it in helper

src/modules/move/MoveModal.jsx
doubleface: Right, let's have better performance by default

e2e/pages/MoveToPage.ts
doubleface: Playwright provides the `times` option on `page.route()`. I simplified the helper to use { times: 1 }

Good point!

src/modules/move/MoveModal.jsx
doubleface: The idea was to fail faster without sending any request. But you are right, this make quite a lot of code just for that. Removed

e2e/pages/MoveToPage.ts
zatteo: Awesome!

zatteo (approved):

Issue #4175, opened 2026-09-04
Add move to shared folders


doubleface:
Add sharings tab to the "move to" dialog and properly handle moving to here

PR #4208, opened 2026-09-17
feat(move): add received sharing destinations

doubleface:
## Summary

- Add received federated sharings as Move to destinations.
- Preserve shared-drive context when validating or creating destination folders.
- Cover writable and read-only destinations, cross-owner moves, and loading behavior.

fixes #4175


src/components/FilePicker/FilePickerBody.jsx
```diff
@@ -312,25 +313,33 @@ const SharedDriveFolderContent = ({
...
+  const items = useMemo(() => {
+    const sourceItems = (sharedDriveResult.included ?? []).map(item => ({
+      ...item,
+      driveId
+    }))
+    const sourceIds = new Set(sourceItems.map(item => item._id ?? item.id))
+    const addedItems = additionalItems.filter(
+      item =>
+        item?.dir_id === folderId &&
+        item.driveId === driveId &&
```
zatteo, line 325: Why one `item?` and one `item.`?

src/components/FilePicker/FilePickerBody.jsx
```diff
@@ -312,25 +313,33 @@ const SharedDriveFolderContent = ({
...
+        !sourceIds.has(item._id ?? item.id)
+    )
```
zatteo, line 327: Why do we have two source for items?

zatteo:
Thanks for the multiple commits

zatteo:
@doubleface why this PR is from a fork?

src/components/FilePicker/FilePickerBody.jsx
doubleface: Will remove unneeded optional chaining in new PR which will not come from a fork

src/components/FilePicker/FilePickerBody.jsx
doubleface: One source is data coming from server and the other contains data created locally, waiting from server update (optimistic data). I will clarify this with a better naming

PR #4210, opened 2026-09-18
test(e2e): enable virtualized file lists by default

doubleface:
## Summary

- Enable Drive file-list virtualization in the default E2E flags so the test environment stays closer to production.
- Adapt E2E page objects and sharing scenarios to virtualized rows and empty states.
- Skip unsupported synthetic file-drop coverage instead of bypassing the drop handler.

## Summary by CodeRabbit

- **Tests**
  - Expanded end-to-end coverage for virtualized and empty Drive folders, including browsing, file actions, folder creation, moving, and public-folder views.
  - Improved test reliability for delayed content, scrolling, sharing flows, role changes, restoration, and filtered shared-folder views.
  - Temporarily skipped drag-and-drop upload coverage while synthetic file drops remain unsupported.
  - Updated sharing and navigation checks to accommodate alternate list and dialog states.


e2e/pages/DrivePage.ts
```diff
@@ -137,6 +166,14 @@ export class DrivePage {
   /** Right-click the empty content area; the dropzone wrapper's onContextMenu
    * opens the AddMenu. Callers must be in an empty folder so the click lands
    * on blank space and not on a file row (which opens the file menu instead). */
+  private async scrollFileListToTop(): Promise<void> {
```
JF-Cozy, line 193: the jsdoc above is for openAddContextMenu, not for this one

e2e/pages/DrivePage.ts
```diff
@@ -53,17 +69,30 @@ export class DrivePage {
   }
 
   /** Locator for the file list cell whose filename contains the substring —
-   *  use for "the original and its (1) copy" style multi-row assertions. */
+   *  use for "the original and its (1) copy" style multi-row assertions.
+   *  The virtualized table exposes the filename through its title instead of
+   *  the legacy filename test id. */
```
JF-Cozy, line 98: don't know if we must know and document that...? 🤔

e2e/pages/FileRow.ts
```diff
@@ -34,23 +34,83 @@ export class FileRow {
 
   /** Locator for the row's filename cell — exposed so tests can assert
    * visibility / count without going through a wrapper method. Anchored
-   * so similar names ("Folder 1" vs "Folder 12") don't collide. */
+   * so similar names ("Folder 1" vs "Folder 12") don't collide.
+   *
+   * The legacy list exposes a dedicated test id while the virtualized table
+   * exposes the untruncated name through the title attribute. Keep both
+   * locators here so the page object supports either list implementation. */
```
JF-Cozy, line 43: no need

e2e/pages/FileRow.ts
```diff
@@ -34,23 +34,83 @@ export class FileRow {
...
 
   private get rowEl(): Locator {
-    // Each row is a plain <div> with no semantic role; locate the closest
-    // ancestor that contains the per-row "More" button — that's the row.
-    return this.cell.locator(
-      'xpath=ancestor::*[.//button[@aria-label="More"]][1]'
-    )
+    // Both list implementations render semantic table rows. Locating the
+    // nearest row also works when the virtualized filename is a span rather
+    // than the legacy filename cell.
```
JF-Cozy, line 54: no need

e2e/pages/FileRow.ts
```diff
@@ -34,23 +34,83 @@ export class FileRow {
...
+    return this.cell.locator('xpath=ancestor::tr[1]')
+  }
+
+  private get actionButton(): Locator {
+    // The virtualized action component currently spells aria-label as
+    // `arial-label`, so support it alongside the correctly labelled legacy
+    // action button.
```
JF-Cozy, line 61: no need

e2e/pages/FileRow.ts
```diff
@@ -34,23 +34,83 @@ export class FileRow {
...
+              await this.scrollFileListToTop()
+              startedFromCurrentPosition = true
+            }
+            return 0
+          }
+
+          // Virtualized rows can be replaced between count() and the action
+          // below while a query result is settling. Keep the whole lookup in
+          // the poll so a detached row is retried instead of escaping as a
+          // flaky Playwright error.
```
JF-Cozy, line 85: no need

e2e/tests/z-nested-sharing.spec.ts
JF-Cozy, line 1: > Inherited role downgrades can either open a confirmation dialog or apply
immediately, depending on the cozy-sharing version

Why and how?! Depending on the cozy-sharing version where? 🤔 We only have one single version of cozy-sharing in the app... I don't understand what we want

Beside we can remove the last paragraph "Validation: the five nested-sharing scenarios pass with federated,
hide-cozy-to-cozy and shared-drive enabled. Targeted ESLint passes."

e2e/tests/upload-and-viewer.spec.ts
JF-Cozy, line 1: we can remove all this : 

```
Keep the actual DataTransfer drop interaction and explicitly skip its
scenario until native file drops can be automated. Document the missing
coverage next to the test so upload-button success cannot hide it.

Validation: attempts using dragenter, dragover and drop reached the
upload queue but failed with Missing name argument. Targeted ESLint
passes. The scenario is intentionally not reported as a passing drop.
```

Could we fix it instead of ignoring a test and adding comments that only say "should be fixed"? I didn't understand what was the limitation to not fixing the test right now.

e2e/pages/FileRow.ts
```diff
@@ -13,6 +13,7 @@ interface ConfirmDialog {
 }
 
 const OPTIONAL_DIALOG_TIMEOUT = 2_000
+const DOUBLE_CLICK_DELAY = 400
```
JF-Cozy: this comes from the code?

e2e/pages/FileRow.ts
JF-Cozy, line 1: could be squashed

JF-Cozy (commented):
I don't think we test here both with and without the flag. Is it relevant to test both, do we want to maintain the "flag set to false" version?
If not, we really should remove it and remove the flag...

By the way, what about mobile version?

doubleface:
> I don't think we test here both with and without the flag. Is it relevant to test both, do we want to maintain the "flag set to false" version? If not, we really should remove it and remove the flag...
> 
> By the way, what about mobile version?

Mobile version still does not use virtualized grids. But yes, we should remove this flag as soon as possible

e2e/pages/FileRow.ts
doubleface: This is a duplicate from src/hooks/useOnLongPress/helper. Now I use the same constant

PR #4218, opened 2026-09-21
feat(move): add received sharing destinations

### LINKED ISSUE #4175: Add move to shared folders

doubleface:
## Summary

- Add received federated sharings as Move to destinations.
- Preserve shared-drive context when validating or creating destination folders.
- Cover writable and read-only destinations, cross-owner moves, and loading behavior.

fixes #4175

## Summary by CodeRabbit

- **New Features**
  - Move files and folders into received shared drives and shared folders.
  - Move items through editable public links.
  - Create destination folders directly within shared-drive locations.
  - Use the Move To dialog across Drive and Sharings destinations.

- **Bug Fixes**
  - Improved Move To behavior on mobile layouts.
  - Newly created folders now appear correctly in their destination drive.
  - Move failures can be retried while preserving visible items.
  - Cross-drive moves better preserve file content and source state.


src/modules/move/helpers.js
```diff
@@ -25,24 +25,42 @@ function isDirectoryItem(item) {
   return Boolean(item) && models.file.isDirectory(item)
 }
 
-function isLocalItem(item) {
+function isCozyItem(item) {
```
Crash--, line 28: why an item from a driveId is not a cozyItem?

doubleface: It's actually the opposite. I renamed this function because it now accepts both local files and files from a shared drive.

src/components/FilePicker/FilePickerBody.jsx
```diff
@@ -312,25 +313,33 @@ const SharedDriveFolderContent = ({
...
-      displayedTypes,
-      driveId,
-      isItemIncluded,
-      sharedDriveResult.included,
+  const items = useMemo(() => {
+    const sourceItems = (sharedDriveResult.included ?? []).map(item => ({
+      ...item,
+      driveId
+    }))
+    const sourceIds = new Set(sourceItems.map(item => item._id ?? item.id))
```
zatteo: Why `item._id ?? item.id`? (same below)

src/modules/move/helpers.js
```diff
@@ -25,24 +25,42 @@ function isDirectoryItem(item) {
...
 export function isLocalMoveDestination(item) {
   return isDirectoryItem(item) && isLocalItem(item)
 }
 
+export function isMoveDestinationFolder(item) {
+  return (
+    isDirectoryItem(item) &&
+    isCozyItem(item) &&
+    item.orgDrive !== true &&
+    item.driveOwner !== true
```
zatteo: `!item.orgDrive && !item.driveOwner` not possible?

zatteo (approved):

src/modules/move/helpers.js
```diff
@@ -25,24 +25,42 @@ function isDirectoryItem(item) {
...
   )
 }
 
+function isLocalItem(item) {
+  return isCozyItem(item) && !item.driveId
+}
+
 export function isLocalMoveDestination(item) {
   return isDirectoryItem(item) && isLocalItem(item)
 }
```
zatteo: Is still method still used?

src/modules/shareddrives/hooks/useQueryMultipleSharedDriveFolders.spec.tsx
zatteo, line 1: There is already a `useQueryMultipleSharedDriveFolders.spec.jsx`(jsx vs tsx file)

src/components/FilePicker/FilePickerBody.jsx
doubleface: fixed

src/modules/move/helpers.js
doubleface: good catch, removed it

PR #4219, opened 2026-09-21
fix(editor): bypass the http cache when loading excalidraw and pdf files

Crash--:
## Context

On an Excalidraw drawing created a while ago, edits were saved to the server but disappeared after a page refresh: the editor reopened the old content.

What we observed on a real instance, for a drawing created on June 26 and edited today:

| Read | Result |
|---|---|
| File metadata (`GET /files/:id`) | 980 bytes, updated today |
| `GET /files/download/:id` as done by the editor | 119 bytes, the initial empty scene |
| Same request with `cache: 'no-store'` | 980 bytes, the actual drawing |
| Drive "Download" action (`/files/downloads/<secret>/...`) | 980 bytes, the actual drawing |

So the save works; only the editor's read is stale.

### Cause

The stack serves `/files/download/:id` through `http.ServeContent` with an `ETag` and a `Last-Modified`, but no `Cache-Control`. The browser then applies heuristic freshness: it reuses the cached response without asking the server for about 10% of the time elapsed since `Last-Modified`.

- The drawing was last modified about 87 days before it was opened, so the empty scene stayed "fresh" for roughly 8 to 9 days.
- Saving sends `PUT /files/:id`, which does not invalidate the cached `/files/download/:id` entry.
- On refresh the editor gets the old binary back from the cache.

On a freshly created drawing, `Last-Modified` is only seconds old, so the cache expires almost immediately and a refresh reads the new content. That is why the bug only shows up on older files.

This can also lose data: once the stale scene is loaded, the next autosave overwrites the newer version on the server.

The PDF editor reads its binary the same way (`fetchFileContentById`), so it is exposed to the same issue.

Other file types are not affected: the Drive download action uses one-time secret links, and OnlyOffice or Notes do not read the binary from the browser through this route.

## Summary

- Add a `fetchFileBinary` helper that downloads a file binary with `cache: 'no-store'`, scoped to the shared drive when needed.
- Use it to load Excalidraw scenes and PDF documents.


zatteo (approved):
Should it have been an option to `fetchFileContentById`?

